Issue #1277, opened 2015-06-04
OutOfMemoryError while generating HTML/XML report


zapbot:
```
What steps will reproduce the problem?
1. Run session in ZAP including passive/active scans
2. Generate XML/HTML report
3. Check file is generated

What is the expected output? What do you see instead?

xml file and html file should be displayed with report content.
File is created, but with no content and has a file size of 0kb


What version of the product are you using? On what operating system?
2.3.1 on Windows 7 Pro


Please provide any additional information below.


```

Original issue reported on code.google.com by `daniel.billing.nvm` on 2014-07-24 13:40:30

thc202:
Closing, the core reports are being superseded by the Report Generation add-on:
https://www.zaproxy.org/docs/desktop/addons/report-generation/

thc202:
Referenced in #6483 in any case.

PR #1660, opened 2015-06-08
Add Request and Response Data to XML Report

andykais:
HttpMessage data is under the `<requestdata>` and `<responsedata>` tag

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/Alert.java
+++ b/src/org/parosproxy/paros/core/scanner/Alert.java
@@ -523,9 +523,31 @@ public class Alert implements Comparable<Object>  {
     public void setAlertId(int alertId) {
         this.alertId = alertId;
     }
-    
+
+    private String getHTML() {
+        //gets HttpMessage request and response data from each alert and removes illegal and special characters
+        StringBuilder httpMessage = new StringBuilder();
+        
+        String requestHeader = this.getMessage().getRequestHeader().toString();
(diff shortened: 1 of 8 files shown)


psiinon:
Unfortunately this will mean that the request and response data will always be added to all reports.
I'd expect most people _not_ to want this, and even if they did this wouldnt be a good way to do it as it would take up a lot of memory.
We already know ZAP runs out of memory with large numbers of issues even without including this data :/ #1277

AndrewKaiser:
Maybe this is a special case then, though I do believe there are a number of people who may appreciate this feature. What if I were to make it an option in the GUI, in the save dialog maybe?

psiinon:
I can definitely see the benefit of this.
However I believe Sam Hakim is already working on this: https://groups.google.com/d/msg/zaproxy-develop/Yxiv0VrCnTg/pIQD9808ZxIJ 
You could always contact him directly to see where he's up to and if you can help him.

AndrewKaiser:
Alright I'm sure Sam is working on a more sophisticated solution but in case you are interested, I pushed a change that makes adding html data an option in the save dialogue

thc202:
@andykais can this PR be closed with release of "Code Dx" add-on?

andykais:
Yes, that will be fine. We got what we were looking for with the plugin.
Thank you for your quick responses

kingthorin:
Closing per requester feedback.

PR #1707, opened 2015-07-17
Fixes #1684 - Enables to include and exclude technologies of a contex…

### LINKED ISSUE #1684: Unable to set technology via API
Currently, it is not possible to set the technology of the context. 
Topic from User Group: https://groups.google.com/forum/#!topic/zaproxy-users/6GBBWcJ9rqs

yannicsmeets:
Enables to include and exclude technologies of a context through the API

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ContextAPI.java
+++ b/src/org/zaproxy/zap/extension/api/ContextAPI.java
@@ -35,27 +35,38 @@ import org.zaproxy.zap.extension.api.ApiResponseElement;
 import org.zaproxy.zap.extension.api.ApiException.Type;
 import org.zaproxy.zap.extension.authorization.AuthorizationDetectionMethod;
 import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
 
 public class ContextAPI extends ApiImplementor {
 
-    private static Logger log = Logger.getLogger(ContextAPI.class);
+    private static final Logger log = Logger.getLogger(ContextAPI.class);
 
     private static final String PREFIX = "context";
+    private static final String TECH_NAME = "technologyName";
     private static final String EXCLUDE_FROM_CONTEXT_REGEX = "excludeFromContext";
     private static final String INCLUDE_IN_CONTEXT_REGEX = "includeInContext";
     private static final String ACTION_NEW_CONTEXT = "newContext";
     private static final String ACTION_SET_CONTEXT_IN_SCOPE = "setContextInScope";
     private static final String ACTION_EXPORT_CONTEXT = "exportContext";
     private static final String ACTION_IMPORT_CONTEXT = "importContext";
+    private static final String ACTION_INCLUDE_TECHS = "includeContextTechnologies";
+	private static final String ACTION_INCLUDE_ALL_TECHS = "includeAllContextTechnologies";
+	private static final String ACTION_EXCLUDE_TECHS = "excludeContextTechnologies";
+	private static final String ACTION_EXCLUDE_ALL_TECHS = "excludeAllContextTechnologies";
     private static final String VIEW_EXCLUDE_REGEXS = "excludeRegexs";
     private static final String VIEW_INCLUDE_REGEXS = "includeRegexs";
     private static final String VIEW_CONTEXT_LIST = "contextList";
     private static final String VIEW_CONTEXT = "context";
+    private static final String VIEW_ALL_TECHS = "technologyList";
+	private static final String VIEW_INCLUDED_TECHS = "includedTechnologyList";
+	private static final String VIEW_EXCLUDED_TECHS = "excludedTechnologyList";
     private static final String REGEX_PARAM = "regex";
     private static final String CONTEXT_NAME = "contextName";
     private static final String IN_SCOPE = "booleanInScope";
     private static final String CONTEXT_FILE_PARAM = "contextFile";
     private static final String CONTEXT_ID = "contextId";
+    private static final String PARAM_TECH_NAMES = "technologyNames";
 
     public ContextAPI() {
         List<String> contextNameAndRegexParam = new ArrayList<>(2);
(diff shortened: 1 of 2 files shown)


yannicsmeets:
What do you think of the way to search a tech based on a string in the method getTech()? It is not very efficient, yet maybe it is not that big of an issue..

src/org/zaproxy/zap/extension/technology/ExtensionTechnology.java
```diff
@@ -0,0 +1,54 @@
+/*
```
thc202: The new API can be added with the extension ExtensionAPI, seems excessive to use a new extension just for this.

src/org/zaproxy/zap/extension/technology/TechnologyAPI.java
```diff
@@ -0,0 +1,158 @@
+/*
```
thc202: The new API should be added to ApiGeneratorUtils so it gets generated for the (Java, Python, ...) API clients.
Good to have some docs (added to Messages.properties file, the keys can be checked by running, for example, the Java API generator).

src/org/zaproxy/zap/extension/technology/TechnologyAPI.java
```diff
@@ -0,0 +1,158 @@
...
+		String[] techNames;
+		switch(name) {
+		case ACTION_INCLUDE_TECHS:
+			techNames = getParam(params, PARAM_TECH_NAMES, "").split(",");
+			for(String techName : techNames) {
+				techSet.include(getTech(techName));
+			}
+			break;
+		case ACTION_INCLUDE_ALL_TECHS:
+			techSet = TechSet.AllTech;
```
thc202: This statement has not effect to the context. It should be used the method Context.setTechSet(TechSet).

src/org/zaproxy/zap/extension/technology/TechnologyAPI.java
```diff
@@ -0,0 +1,158 @@
...
+import org.zaproxy.zap.extension.api.ApiResponseList;
+import org.zaproxy.zap.extension.api.ApiView;
+import org.zaproxy.zap.extension.api.ApiException.Type;
+import org.zaproxy.zap.model.Tech;
+import org.zaproxy.zap.model.TechSet;
+import org.zaproxy.zap.utils.ApiUtils;
+
+public class TechnologyAPI extends ApiImplementor{
+
+	private static Logger log = Logger.getLogger(TechnologyAPI.class);
```
thc202: Consider adding the _final_ modifier.

thc202:
IMHO the new API actions/views should be added to ContextAPI, seems more natural to me.

Regarding the efficiency, I wouldn't worry too much about that (at least for now).

yannicsmeets:
I've considered that. However, since authentication, users, and forcedUser are also part of a context yet have their own api, I thought I would keep it consistent. But if it is better to add it to the ContextAPI, I will put it in there.

thc202:
True, there are some context "attributes" that are not in the ContextAPI.
I'd wait for other opinions before changing that ;)

yannicsmeets:
Of course I am not changing the other api's :-) but should I put the technology views and actions in the context api or in a separate one?

thc202:
I was referring to changing the new views/actions to ContextAPI.
If no one else says that it should be in the ContextAPI then TechnologyAPI is just fine :)

psiinon:
I agree that it should be in ContextAPI, the main reason being; thats closest to the functionality.
If the technology code was in its own package then I'd agree that the API should be there - in the same way that its been implemented for authentication etc, but as it hasnt been then I think the ContextAPI is the right place.

yannicsmeets:
Alright, i'll put it in there. One more question, I use the contextId to fetch the context. However, in the contextAPI the contextName is being used. Should I use that as well?

yannicsmeets:
I have put it all into the contextAPI, and refactored all the if/else statements into a switch statement. Made more sense to me since other api's do this as well.

src/org/zaproxy/zap/extension/api/ExtensionAPI.java
```diff
@@ -79,7 +79,6 @@ public void hook(ExtensionHook extensionHook) {
 
         API.getInstance().registerApiImplementor(new ContextAPI());
 
-
```
thc202: This indentation change should be removed.

src/org/zaproxy/zap/extension/api/ContextAPI.java
```diff
@@ -141,14 +171,46 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
-        } else {
-            throw new ApiException(ApiException.Type.BAD_ACTION);
+        case ACTION_INCLUDE_TECHS:
+        	context = ApiUtils.getContextByParamId(params, CONTEXT_ID);
+        	techSet = context.getTechSet();
+        	techNames = getParam(params, PARAM_TECH_NAMES, "").split(",");
+        	for (String techName : techNames) {
+        		techSet.include(getTech(techName));
+        	}
+        	context.setTechSet(techSet);
```
thc202: This statement can be removed, it's redundant, the changes are already done to the _TechSet_ of the _Context_.
The same comment applies to cases _ACTION_EXCLUDE_TECHS_ and _ACTION_EXCLUDE_ALL_TECHS_.

yannicsmeets: You mean the statement context.setTechSet(techSet); ?

thc202: Yes.

thc202:
The refactoring looks good to me although it would be better (for the history) if it's done in another commit.
The ExtensionAPI still has indentation changes ;)

yannicsmeets:
So for next time, having separate commits for each change is better? Or only a separate commit for the refactoring?
Ah I see. Apparently I removed the trailing new lines.

thc202:
It would be better if it was for this time too ;)
Depends a lot on the changes being done, but as general rule, if the changes are (logically) unrelated they should be in separate commits.
In this case the refactoring is unrelated (i.e. not required) for the addition of the new views/actions so it should be in another commit.
In any case the commit message should mention the refactoring as it's easier to review/digest the changes.

yannicsmeets:
Ah alright. Sorry, I don't have much experience with this, but know I know :-) Will fix it today/tomorrow.

thc202:
Sorry for not spotting this sooner, the context should be saved/persisted (_Context.save()_) after the tech changes, otherwise the changes are lost between restarts of ZAP (or load of other sessions) even if the session was previously saved.

yannicsmeets:
No problem, good spot! Should be fixed now.

thc202:
It's missing the save in the case _ACTION_INCLUDE_TECHS_.

thc202:
Thanks!

yannicsmeets:
Yannic Smeets is good.
Thank you very much for all the help :-)

thc202:
Done in zaproxy/zap-core-help@e16648d8adf45cc8fc76990499fc71409b917963

PR #1726, opened 2015-07-24
Update JRE to 1.7u79 (CPU) for 2.4.1

marumari:


kingthorin:
Was there no u80 for Mac?

I tried to go looking but Oracle's page now states:
> **Where can I get Java 7 downloads?**
> 
> July 2015: Updates for Java 7 are no longer available to the public. Oracle offers updates to Java 7 only for customers who have purchased Java support or have Oracle products that require Java 7.

thc202:
@kingthorin http://www.oracle.com/technetwork/java/javase/downloads/jre7-downloads-1880261.html

kingthorin:
OK so is there a reason we're not going to u80?

marumari:
Because that's what Oracle recommends?

marumari:
Closing this, as it's subsumed by #1728.

Issue #1853, opened 2015-08-27
Allow to active scan a Context through ZAP API


dap:
There appears to be no way to specify a list of technologies when invoking an active scan via the REST API. I think it would be beneficial if there was greater parity with the graphical UI in this respect, to allow for more specific scan targeting.

psiinon:
You can configure the technologies associated with a context via the 'context' component, and then specify that context when starting the active scan.
Hopefully that will meet your requirements?

dap:
I looked through the code base to determine where the technologies are pulled from the Context and passed to the ActiveScanController. It appears that this only happens when a user is specified. See https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java#L542

psiinon:
We need to fix that then - you should definitely be able to scan a context without specifying a user.

dap:
I put together a patch that merges contexts then sets them manually. It's referenced from this ticket and I'll send a pull request.

thc202:
I've to disagree with assessment of the issue, the problem is not that the technologies are not set, the problem is that it's not possible to specify a context (which has its technologies).
I think the API should be changed to specify a context not to set the technologies of the contexts, it will be more consistent with the functionalities provided in GUI.
Not saying that the API should not allow to specify the technologies used in the scan (it should be possible like it is in GUI with "Technology" tab in "Active Scan" dialogue).

dap:
The API already allows for the specification of a context -- multiple contexts, in fact -- via the combination of importContext and setContextInScope. Were the API to be modified in the way that you indicate, there would still need to be code changes to respect the context-selected technologies at scan time. This is the reason for the referenced patch merging TechSet's. A given target may appear in scope in multiple contexts.

And I concur that it would be reasonable and intuitive to have the API function similarly to the active scan GUI. In such case, I think that the API should respect the technology selections in both the context and active scan configuration.

thc202:
Sorry for not being clear, I was talking about the active scan API, not the ZAP API as whole.
The active scanner already uses the technologies of the specified context, no other changes would be required other than change active scan API to allow to specify the context.

Regarding the target appearing in multiple contexts, well it's possible but I wonder how would that work in practice (i.e. authentication, authorisation, structural parameters). As said in the review of the pull request, in other cases where a context is selected for a given target it's just used the first one.

dap:
[...]

PR #1855, opened 2015-08-27
Set technologies based on URL contexts

dap:
Retrieve contexts in which the specfied
URL is included, then merge their tech-
nology lists and pass them to the scan
controller.

Addresses zaproxy/zaproxy#1853

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
+++ b/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
@@ -60,6 +60,7 @@ import org.zaproxy.zap.model.GenericScanner2;
 import org.zaproxy.zap.model.SessionStructure;
 import org.zaproxy.zap.model.StructuralNode;
 import org.zaproxy.zap.model.Target;
+import org.zaproxy.zap.model.TechSet;
 import org.zaproxy.zap.users.User;
 import org.zaproxy.zap.utils.ApiUtils;
 import org.zaproxy.zap.utils.XMLStringUtil;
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -532,23 +533,41 @@ private int scanURL(URI url, User user, boolean scanChildren, boolean scanJustIn
...
+			if (policy != null) {
+				objs.add(policy);
+			}
+
+			/*
+			 * A side effect of setting the Context based on the specified user
+			 * is the limiting of the scan to technologies associated with the
+			 * Context. When user is not specified, technologies should be
+			 * limited based on Contexts containing the supplied URL.
+			 */
```
thc202: This is not a side effect, it is the intended effect. A user is associated to a context so when scanning as a user it's also scanning with the associated context thus constrained by its configurations (technologies, authentication, structural parameters, what's included/excluded...).

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -532,23 +533,41 @@ private int scanURL(URI url, User user, boolean scanChildren, boolean scanJustIn
...
+			TechSet mergedTechSet = new TechSet();
 			if (user != null) {
 				target.setContext(user.getContext());
 			}
-
-			Object [] objs = new Object[]{};
-			if (policy != null) {
-				objs = new Object[]{policy};
+			else {
+				List<Context> contexts = session.getContextsForUrl(url.getURI());
```
thc202: IIRC in no other place are the context's configurations being merged (it's picked the first context).
This also forces the use of technologies, whether the user wants it or not.

src/org/zaproxy/zap/model/TechSet.java
```diff
@@ -49,10 +49,14 @@ public TechSet (Tech[] include, Tech[] exclude) {
 	}
 	
 	public TechSet(TechSet techSet){
+		this.merge(techSet);
+	}
+	
+	public void merge(TechSet techSet){
```
thc202: The implementation doesn't look right, it's possible to have a technology included and excluded at the same time.

thc202:
Added a comment to the issue regarding the proposed changed.

kingthorin:
@day are you going to address the feedback?

kingthorin:
@dap do you intend to address the feedback? Should this pull be closed?

kingthorin:
@dap any news?

kingthorin:
Abandon?

thc202:
Closing, per previous comments. Will update #1853 to allow to active scan a context through the ZAP API, which matches the behaviour of other APIs (e.g. spiders).

PR #1870, opened 2015-09-03
Fix path to .ZAP_JVM.properties in zap.bat

bartleusink:
Hi!

I noticed that zap.bat wasn't picking up my .ZAP_JVM.properties file.
The added percent signs should fix the path.


psiinon:
Many thanks!

Issue #1875, opened 2015-09-07
Scanner temp history not cleaned on close


dbalaouras:
According to org.parosproxy.paros.model.HistoryReference, history entries with type TYPE_SCANNER_TEMPORARY (int:14) should be removed on cleanup: 

```
    /**
     * A HTTP message sent by the (active) scanner which is set as temporary (i.e. deleted when the session is closed).
     */
    public static final int TYPE_SCANNER_TEMPORARY = 14;
```

Though, org.parosproxy.paros.db.paros.ParosTableHistory only cleans up type HistoryReference.TYPE_TEMPORARY (int: 0) by default:

```
    static {
        temporaryHistoryTypes.add(Integer.valueOf(HistoryReference.TYPE_TEMPORARY));
    }
```

The end result is having a huge session data file, containing a large number of undeleted temp history records.

thc202:
Fixed with #1875.

PR #1876, opened 2015-09-07
Fixing issue #1875: added TYPE_SCANNER_TEMPORARY to ParosTableHistory and SqlTableHistory

dbalaouras:
Added HistoryReference.TYPE_SCANNER_TEMPORARY to ParosTableHistory.java and SqlTableHistory.java.

*Still wondering if ```private static Set<Integer> temporaryHistoryTypes``` and the methods that modify it should be part of the class HistoryReference instead of duplicating code in ParosTableHistory and SqlTableHistory.*

### DIFF
--- a/src/org/parosproxy/paros/db/paros/ParosTableHistory.java
+++ b/src/org/parosproxy/paros/db/paros/ParosTableHistory.java
@@ -92,13 +92,15 @@ public class ParosTableHistory extends ParosAbstractTable implements TableHistor
     /**
      * The {@code Set} of history types marked as temporary.
      * <p>
-     * By default the only temporary history type is {@code HistoryReference#TYPE_TEMPORARY}.
+     * By default the only temporary history types are {@code HistoryReference#TYPE_TEMPORARY} and
+     * {@code HistoryReference#TYPE_SCANNER_TEMPORARY}.
      * <p>
      * Iterations must be done in a {@code synchronized} block with {@code temporaryHistoryTypes}.
      * 
      * @since 2.4
      * @see #setHistoryTypeAsTemporary(int)
      * @see HistoryReference#TYPE_TEMPORARY
+     * @see HistoryReference#TYPE_SCANNER_TEMPORARY
      * @see Collections#synchronizedSet(Set)
      */
     private static Set<Integer> temporaryHistoryTypes = Collections.synchronizedSet(new HashSet<Integer>());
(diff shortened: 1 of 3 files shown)


src/org/parosproxy/paros/db/paros/ParosTableHistory.java
```diff
@@ -92,13 +92,15 @@
...
-     * By default the only temporary history type is {@code HistoryReference#TYPE_TEMPORARY}.
+     * By default the only temporary history types are {@code HistoryReference#TYPE_TEMPORARY} and
+     * {@code HistoryReference#TYPE_SCANNER_TEMPORARY}.
      * <p>
      * Iterations must be done in a {@code synchronized} block with {@code temporaryHistoryTypes}.
      * 
      * @since 2.4
      * @see #setHistoryTypeAsTemporary(int)
      * @see HistoryReference#TYPE_TEMPORARY
+     * @see HistoryReference#TYPE_SCANNER_TEMPORARY
```
kingthorin: This should probably be seperated and listed as "@since TODO add version"

src/org/zaproxy/zap/db/sql/SqlTableHistory.java
```diff
@@ -68,13 +68,15 @@
...
-     * By default the only temporary history type is {@code HistoryReference#TYPE_TEMPORARY}.
+     * By default the only temporary history types are {@code HistoryReference#TYPE_TEMPORARY} and
+     * {@code HistoryReference#TYPE_SCANNER_TEMPORARY}.
      * <p>
      * Iterations must be done in a {@code synchronized} block with {@code temporaryHistoryTypes}.
      * 
      * @since 2.4
      * @see #setHistoryTypeAsTemporary(int)
      * @see HistoryReference#TYPE_TEMPORARY
+     * @see HistoryReference#TYPE_SCANNER_TEMPORARY
```
kingthorin: This should probably be seperated and listed as "@since TODO add version"

kingthorin:
@dbalaouras what prompted submission of this change does it address some specific concern or deficiency?

kingthorin:
@dbalaouras what prompted submission of this change does it address some specific concern or deficiency?

src/org/parosproxy/paros/db/paros/ParosTableHistory.java
thc202: I think it's fine adding the new type (it was added in previous releases).

thc202:
This is to address issue #1875 (HTTP messages database leak),

The changes look OK to me (the commit message could be corrected and the temporary types moved to other place, but not blockers to me).

psiinon:
My worry is that this could increase the time it takes ZAP to exit, and it can already take quite a while with big scans :/ @thc202 mentioned we could disable the db log when deleting the requests - that a big change?

thc202:
Disabling DB recovery log didn't have great impact, although there are some changes that help reduce the time it takes to delete the temporary messages, #2511 and #2512.

PR #1877, opened 2015-09-08
fix path to .ZAP_JVM.properties on OSX

bartleusink:
Hi!

On OSX, I'm getting the following error on startup, which causes the jvm heap size to be set to 512 mB:

    /Users/bartl/Applications/OWASP ZAP.app/Contents/Java/zap.sh: line 63: [: ~/Library/Application: binary operator expected

This pull request puts quotes around the `JVMPROPS` variable, allowing for spaces in the path.

### DIFF
--- a/src/zap.sh
+++ b/src/zap.sh
@@ -54,15 +54,15 @@ else
 fi
 
 if [ "$OS" = "Darwin" ]; then
-  JVMPROPS="~/Library/Application Support/ZAP/.ZAP_JVM.properties"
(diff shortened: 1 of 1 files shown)


thc202:
Could you update the commit message to indicate the fix?
It's more useful, for the history, to have that in the commit message than the PR ;)

bartleusink:
I've added some text to the commit message indicating the fix.

thc202:
Thanks!

PR #1889, opened 2015-09-11
Update README.md

m0k1:
Competition is ended.


psiinon:
Good point, but we should really update that to say that the results will be coming soon, as per https://www.owasp.org/index.php/ZAP :)

Issue #1894, opened 2015-09-14
ZAP only scans "User-Agent", "Referer", and "Host" request headers


ElColmo:
When active scanning a request with the "HTTP Headers" input vector enabled, ZAP will only scan the following headers: "User-Agent", "Referer", and "Host".

This is due to the following code in org.parosproxy.paros.core.scanner.VariantHeader:

    public static final String[] injectableHeaders = {
        HttpRequestHeader.USER_AGENT,
        HttpRequestHeader.REFERER,
        HttpRequestHeader.HOST
    };

Currently, in order to scan request headers other than the headers named above, a "custom" input vector needs to be used.

@psiinon  suggests a black-list approach, rather than the white-list approach currently used.

I'd suggest we add the following headers to the black-list (of request headers not to scan).  I'm assuming all other request headers should be attacked by default:
- Content-Length
- Pragma
- Cache-Control

This also raises the question as to whether this list should be user specified,however.

psiinon:
It would be good to tie this in to the attack strength, but that could be problematic ;)
A user specified white list, with the defaults @ElColmo suggested would work for me.

ElColmo:
I have an initial change committed locally to fix this. I can't push this to the central repo though.  I'm getting " https://github.com/zaproxy/zaproxy: git-receive-pack not permitted" in Eclipse when I try to "Push To Upstream".  I have very little experience with Git, but I'm assuming this a permissions based error.

psiinon:
We should all be using Pull Requests.
I dont yet, but I really should do :/

ElColmo:
I'm happy to. I just don't know the specifics of how it works yet.  I'm off to do some reading on the subject!

I've already committed locally, so that might narrow my options without re-work.

kingthorin:
Starting place: https://github.com/zaproxy/zaproxy/wiki/Contributing-Changes

ElColmo:
Seems the process has gotten way more complicated to me.

Having said that, please see: https://github.com/zaproxy/zaproxy/pull/1899

kingthorin:
Re: your pull request.

Should set-cookie also be marked non-injectable here? Aren't cookie injections handled properly as their own variant? Wouldn't invalidating the set-cookie header generally result in wasted requests?

What about http standard authentication headers? www-authenticate

kingthorin:
Agreed as you said in the PR cookie not set cookie, and authz not authn

PR #1899, opened 2015-09-15
Add support for scanning headers other than "User-Agent", "Referer", and "Host".

ElColmo:
Added support for scanning headers other than "User-Agent", "Referer", and "Host".

As a side effect, allows ZAP to detect a number of issues that it previously could not detect on OWASP Benchmark.


kingthorin:
Should set-cookie also be marked non-injectable here? Aren't cookie injections handled properly as their own variant? Wouldn't invalidating the set-cookie header generally result in wasted requests?

What about http standard authentication headers? www-authenticate

ElColmo:
Set-Cookie, Set-Cookie2, www-authenticate are response headers, so are not relevant here.  The "Cookie" and "Authorization"  request headers should probably be blacklisted for the reasons you specify, however.

So far, i think we should also exclude the following from scanning:
- Cookie (since there is a variant for this)
- Authorization (since this would break authorisation)
- Proxy-Authorization (since this would break authorisation)
- X-Csrf-Token (since this would break authorisation)

Are there any other headers that jump out to anyone?

kingthorin:
Yes, agreed. (Need coffee.....)

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -33,14 +35,19 @@
  */
 public class VariantHeader implements Variant {
 
-    // I've found an XSS using this payload on the Host header
-    // "%s-->\">'>'\"<sfi%06uv%06u>"
-
-    public static final String[] injectableHeaders = {
```
thc202: Better deprecate? Just in case it's being used by 3rd party classes.

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -33,14 +35,19 @@
...
 public class VariantHeader implements Variant {
 
-    // I've found an XSS using this payload on the Host header
-    // "%s-->\">'>'\"<sfi%06uv%06u>"
-
-    public static final String[] injectableHeaders = {
-        HttpRequestHeader.USER_AGENT,
-        HttpRequestHeader.REFERER,
-        HttpRequestHeader.HOST
+    public static final String[] nonInjectableHeaders = {
```
thc202: Consider making it private (it's mutable). If we really need to expose the headers it's better with an unmodifiable `Collection`.

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -33,14 +35,19 @@
...
+        HttpRequestHeader.CONTENT_LENGTH,	//scanning this would likely break the entire request 
+        HttpRequestHeader.PRAGMA,			//unlikely to be picked up/used by the app itself.
+        HttpRequestHeader.CACHE_CONTROL,	//unlikely to be picked up/used by the app itself.          
+        HttpRequestHeader.COOKIE,			//The Cookie header has its own variant that controls whether it is scanned. Better not to scan it as a header.
+        HttpRequestHeader.AUTHORIZATION,	//scanning this would break authorisation
+        "PROXY-AUTHORIZATION",				//scanning this would break authorisation		//TODO: add to HttpRequestHeader?
```
thc202: Yes, better add now to one of the header classes (same for following headers). I'm OK with `HttpRequestHeader`.

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -76,11 +82,20 @@ public void setMessage(HttpMessage msg) {
             }
         }        
         
-        for (int idx = 0; idx < injectableHeaders.length; idx++) {
-            headerContent = msg.getRequestHeader().getHeader(injectableHeaders[idx]);
-            if (headerContent != null) {
-                params.add(new NameValuePair(NameValuePair.TYPE_HEADER, injectableHeaders[idx], headerContent, idx));
-            }
+        List<HttpHeaderField> httpHeaders = msg.getRequestHeader().getHeaders();
+        if (httpHeaders == null) return;
```
thc202: The returned list is never `null`.

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -76,11 +82,20 @@ public void setMessage(HttpMessage msg) {
...
-            headerContent = msg.getRequestHeader().getHeader(injectableHeaders[idx]);
-            if (headerContent != null) {
-                params.add(new NameValuePair(NameValuePair.TYPE_HEADER, injectableHeaders[idx], headerContent, idx));
-            }
+        List<HttpHeaderField> httpHeaders = msg.getRequestHeader().getHeaders();
+        if (httpHeaders == null) return;
+        int headerPos=0;
+        for (HttpHeaderField header : httpHeaders) {       	
+        	boolean ignoreHeader = false;
+        	for (int idx = 0; idx < nonInjectableHeaders.length; idx++) {
```
thc202: Might be clearer (and faster?) with a `HashSet`?
For example:
```Java
// NON_INJECTABLE_HEADERS were added in lower case
boolean ignoreHeader = NON_INJECTABLE_HEADERS.contains(header.getName().toLowerCase(Locale.ROOT));
```

thc202:
A couple more suggestions:
 - Connection;
 - Proxy-Connection;
 - If-Modified-Since (removed from active scanner requests);
 - If-None-Match (ditto);
 - X-Security-Proxy (intended for ZAP);
 - X-ZAP-RequestID (added by ZAP for logging/tracing purposes).

Could you fixup the commits?

ElColmo:
I've already committed and pushed the changes.  I didn't do a fixup because I had no idea what that even meant at the time.  Is it too late to do whatever tidyup you wanted?

src/org/parosproxy/paros/core/scanner/VariantHeader.java
```diff
@@ -32,15 +37,37 @@
...
 
-    // I've found an XSS using this payload on the Host header
-    // "%s-->\">'>'\"<sfi%06uv%06u>"
-
-    public static final String[] injectableHeaders = {
-        HttpRequestHeader.USER_AGENT,
-        HttpRequestHeader.REFERER,
-        HttpRequestHeader.HOST
+	//headers converted to lowercase to make comparison easier later.    
+    private static final String [] injectablesTempArray = {
```
thc202: Since the array is now just used to initialise the `HashSet`, it could be done in a static block to let it go way (not that it would make much difference).

thc202:
No, you can do the reworks whenever you want, just change your local branch and push, the PR will be updated with those changes.

ElColmo:
I *think* it's ready for merge.. Can you take a look?

https://github.com/zaproxy/zaproxy/pull/1899

thc202:
There are some commits that shouldn't be there.

ElColmo:
Sorry. I don't know what the issue is, or how to resolve it.

thc202:
Following the steps to fixup the commits, of this PR, through the command line:
1. Fetch from upstream (to rebase on latest changes);
2. `git rebase -i upstream/develop`, it should open a text editor (or, whatever program is configured) showing the existing commits in the branch;
3. Remove the two last commits (i.e. lines), `0041766` and `09ef03c`, we don't want those commits in the PR. Commit `a5c8458` could also be removed since it's just an indentation change;
4. Set remaining commits, except the first, to fixup;
5. Close the text editor. It should rebase successfully (without conflicts);
6. Review that everything is OK (just one commit with the correct changes);
7. Push.
The PR should be updated and ready for merge.

I recommend amend the commit before pushing to say that it fixes the issue (to automatically close the issue on PR merge):
1. `git commit --amend`
2. Replace `https://github.com/zaproxy/zaproxy/issues/1894` with `Fix #1894` (or any other variant that closes the issue). [1]

[1] https://help.github.com/articles/closing-issues-via-commit-messages/

thc202:
@ElColmo the previous steps are still up-to-date, would you have some spare time to give it a try? It shouldn't take more than 5 minutes (if the env is already set up).

thc202:
Rebased and merged in @e3a108cf87d37cef5b763ea6e635bbb0482f0030.

PR #1965, opened 2015-10-06
Added new active scanner TestUserAgent to alerts.xml

denniskniep:
Implemented a new active scanner. This scanner should be added to the alerts.xml. https://github.com/zaproxy/zap-extensions/pull/47

### DIFF
--- a/src/doc/alerts.xml
+++ b/src/doc/alerts.xml
@@ -86,7 +86,8 @@ This list the alerts
 10046	Insecure Component
 10047	HTTPS Content Available via HTTP
 10048	ShellShock
-10049  Image Location Scanner - passive scanner finding
+10049   Image Location Scanner - passive scanner finding
+10050   TestUserAgent
 
 10101	Insufficient Authentication
 10102	Insufficient Authorization


src/doc/alerts.xml
```diff
@@ -86,7 +86,8 @@ This list the alerts
 10046	Insecure Component
 10047	HTTPS Content Available via HTTP
 10048	ShellShock
-10049  Image Location Scanner - passive scanner finding
+10049   Image Location Scanner - passive scanner finding
```
thc202: Indentation changes are better if done in other commit than the actual changes (but I guess a little change like this is just fine). [1]

kingthorin: Not sure if it matters at all but over here you mentioned intent to use 10050: https://github.com/zaproxy/zap-extensions/pull/47#issuecomment-146001555

denniskniep: Yes, the id 10050 is in line 90. Additionally I removed a redundant space at line 89.

kingthorin:
Np, only saw the change the commented on.

src/doc/alerts.xml
kingthorin: Np, only saw the change the commented on.

PR #1968, opened 2015-10-08
Fixes #1967: View current session location via API

### LINKED ISSUE #1967: Query Session Location via API
Missing: http://localhost:8080/UI/core/view/sessionLocation/
to display the full path of of the current session.

The value is already used in the GUI:
https://github.com/zaproxy/zaproxy/blob/2.4/src/org/parosproxy/paros/view/SessionGeneralPanel.java#L139

Needed for moving the session into a FTP directory after automated scan.

yannicsmeets:
Allows users to retrieve the location of the current session file via the API

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -824,6 +824,7 @@ core.api.view.messages = Gets the HTTP messages sent by ZAP, request and respons
 core.api.view.numberOfMessages = Gets the number of messages, optionally filtering by URL
 core.api.view.version = Gets ZAP version
 core.api.view.excludedFromProxy = Gets the regular expressions, applied to URLs, to exclude from the Proxy
+core.api.view.sessionLocation = Gets the location of the current session file
 
 database.optionspanel.name = Database
 database.optionspanel.option.compact.label = Compact (on exit)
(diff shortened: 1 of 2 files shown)


psiinon:
This is actually a tricky one. One of the reasons we havnt supported this to date is because it leaks information - a remote site can use this to gain information about the pentesters computer.
I'm not sure we should support it without extra safeguards, eg such as requiring the API key.
This might be one to discuss on the dev group - I'll start a thread there in a bit..

Harinus:
There also exists the function: core/view/homeDirectory which exposes path information without the API key. Maybe you should consider reviewing this one as it is pretty similar to my requested feature in #1967. I don't know if it is intended behaviour but the homeDirectory is set when opening any saved session but not when opening a new session. Otherwise I could use this one for my needs. Requiring the API key would be fine for me though.

yannicsmeets:
Has anything been decided on this already?

Harinus:
Is there still no decision on these 5 lines of code?

kingthorin:
@yannicsmeets as Simon suggested please ensure this requires apikey

kingthorin:
@yannicsmeets is an update coming?

yannicsmeets:
I'm sorry no, I currently don't have any time

kingthorin:
Could you set this PR to allow modification by maintainers?
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

yannicsmeets:
I can't see the checkbox to allow edits. Probably because the source branch is missing (i don't know why that happened, seems to be deleted). Is there another way to do this?

thc202:
That happens when the fork is deleted, not sure if that can be fixed. The only options, I think, is to either manually merge the pull request or open a new one.

Issue #1975, opened 2015-10-09
RC4-SHA SSL cipher suite not supported


ElColmo:
When proxying a browser connection to an HTTPS URL, where the web server's (software terminated) HTTPS endpoint supports *only* the "TLSv1.0  128 bits  RC4-SHA" ciphersuite, ZAP returns the following error to the web browser: 

ZAP Error [javax.net.ssl.SSLHandshakeException]: Received fatal alert: handshake_failure

Stack Trace:
javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:154)
	at sun.security.ssl.SSLSocketImpl.recvAlert(SSLSocketImpl.java:1991)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1098)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1344)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1371)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1355)
	at org.parosproxy.paros.network.SSLConnector.createSocket(Unknown Source)
	at org.apache.commons.httpclient.HttpConnection.open(HttpConnection.java:707)
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$HttpConnectionAdapter.open(MultiThreadedHttpConnectionManager.java:1361)
	at org.apache.commons.httpclient.HttpMethodDirector.executeWithRetry(Unknown Source)
	at org.apache.commons.httpclient.HttpMethodDirector.executeMethod(Unknown Source)
	at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:397)
	at org.parosproxy.paros.network.HttpSender.executeMethod(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.runMethod(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.send(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAuthenticated(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(Unknown Source)
[...]

PR #1977, opened 2015-10-14
Fixes zaproxy/zaproxy#1975 RC4-SHA SSL cipher suite not supported.

kingthorin:
OpenJDK and Sun/Oracle JDK now default disable some weak ciphers (such as RC4). This change enables all supported cipher suites in ZAP to maximize connectivity success.

### DIFF
--- a/src/org/parosproxy/paros/network/SSLConnector.java
+++ b/src/org/parosproxy/paros/network/SSLConnector.java
@@ -29,6 +29,7 @@
 // ZAP: 2014/08/14 Issue 1184: Improve support for IBM JDK
 // ZAP: 2014/08/14 Issue 1274: ZAP Error [javax.net.ssl.SSLException]: Unsupported record version SSLv2Hello
 // ZAP: 2014/10/28 Issue 1390: Force https on cfu call
+// ZAP: 2015/10/13 Issue 1975: Allow use of default disabled cipher suites (such as RC4-SHA)
 
 package org.parosproxy.paros.network;
 
@@ -102,7 +103,7 @@ public class SSLConnector implements SecureProtocolSocketFactory {
 	private static String[] serverEnabledProtocols;
 
 	private static ServerSslSocketsDecorator serverSslSocketsDecorator;
-	private static ClientSslSocketsDecorator clientSslSocketsDecorator;
+	private ClientSslSocketsDecorator clientSslSocketsDecorator;
 
     /**
      * The maximum time, in minutes, that the cached misconfigured hosts are considered valid.
@@ -549,7 +550,7 @@ public class SSLConnector implements SecureProtocolSocketFactory {
 		return new DecoratedSocketsSslSocketFactory(delegate, serverSslSocketsDecorator);
 	}
 
-	private static SSLSocketFactory createDecoratedClientSslSocketFactory(final SSLSocketFactory delegate) {
+	private SSLSocketFactory createDecoratedClientSslSocketFactory(final SSLSocketFactory delegate) {
 		return new DecoratedSocketsSslSocketFactory(delegate, clientSslSocketsDecorator);
 	}
 
@@ -564,7 +565,7 @@ public class SSLConnector implements SecureProtocolSocketFactory {
 		}
 	}
 
-	private static class ClientSslSocketsDecorator implements DecoratedSocketsSslSocketFactory.SslSocketDecorator {
+	private class ClientSslSocketsDecorator implements DecoratedSocketsSslSocketFactory.SslSocketDecorator {
 
 		@Override
 		public void decorate(SSLSocket sslSocket) {
@@ -572,6 +573,9 @@ public class SSLConnector implements SecureProtocolSocketFactory {
 				readSupportedProtocols(sslSocket);
 			}
 			sslSocket.setEnabledProtocols(getClientEnabledProtocols());
+			if (relaxedTrust) {
+				sslSocket.setEnabledCipherSuites(sslSocket.getSupportedCipherSuites());
+			}
 		}
 	}
 


thc202:
I think the user should be allowed to choose which cipher suites are enabled from the supported ones (at the very least it should be allowed to choose to just use default cipher suites).
We should use the defaults when `relaxedTrust` is `false`, so that "check for updates" does not use weak(er) cipher suites.

ElColmo:
Not a bad idea at all. I was thinking about it from the point of improving our ability to scan sites with "unusual" configurations, not for downloading data (over weak ciphers) which we would run as part of ZAP.  That's a *bad* idea :)

The "relaxedTrust" setting appears to be set to "true" by default, however.  Is this by design?  I don't see any GUI setting for it either, or is it set behind the scenes somewhere?

In theory, we could query the underlying SSL connection for the list of supported+enabled ciphers, and display that to the user.  That would be additionally useful, since the list of supported and enabled ciphers varies by VM, and it might not be the same from VM-to-VM, platform-to-platform, or version to version..  Do we want to abstract away from that, or simply rely on the less granular "VM default enabled ciphers" versus "all ciphers enabled"?

ElColmo:
Another thought.. should we really be using the same configuration of connection to download files, and to scan servers?  Those two use cases have very different requirements (security vs compatibility, respectively). It might be safer to not allow the "download" connection to be downgraded to use weak ciphers. Ever.  It might also be better to only allow "strong" ciphers in that use case, in fact.

kingthorin:
I'd like to focus on proxy use and scan use. If another change is needed for how ZAP checks for updates, etc. that should be handled separately in another issue and PR. But I would agree with the idea that said functionality should just use the default set based on whatever the JVM defaults are. For the time being I'm fairly certain that even if supported server side the preference of our download/update sources is not weak ciphers.

For the change here I'd like to suggest the following:
1. Move forward with this PR as is. It facilitates greater coverage and reduction of user reported issues. We have no documentation [that I'm aware of anyway] stating one way or the other which ciphers we support or why so from my point of view we're not increasing risk or causing harm. 
2. I'll open another issue and tackle a user controlled option in the local proxy settings:
*  A single checkbox or radio button that facilitates defaults vs all supported ciphers. (Probably in the form `"Use JVM Default Cipher Suites?" true/false` with I guess true being the default/base setting?)

If down the road someone asks for a user controlled item by item list it could be added as an advanced option I suppose. I note that Burp does allow this type of control, but at the same time I've also never bothered to change it in years of use always going with default or all (just my 2cents). Allowing a item by item user controlled list just seems to muddy things from my PoV. Allowing control of defaults vs. supported is more control than users get with the preferences in most modern browsers [there are probably "under the hood" options but you get the point]. I'm also 90% certain neither Fiddler nor AppScan allow control of an itemized list.

ElColmo:
It's a can of worms alright! Sorry.  My only other comment would be that if "Use JVM Default Cipher Suites?" is not checked by the user, it's not entirely clear what is or is not being configured.  I'd prefer to see something more explicit that suggests what positive action will be taken, or can be enabled by the user if the box is checked.  For instance: "Enable *all* SSL Cipher Suites supported by the JVM, including cipher suites that are disabled by default".

thc202:
@ElColmo in theory ZAP should already be using stronger (i.e. default enabled) cipher suites for check for updates (CFU) and download of the add-ons.

Yes, that's by design, that's the default behaviour applied to all connections done by ZAP (that use `HttpSender`), it's done that way to improve connectivity (but I guess we could expose that to the user too but the user would have to manage the certs and trusted authorities). The only time the `relaxedTrust` is `false` is when doing CFU.

Yeah, if we agree on allowing the user to choose which cipher suites are enabled, that's the way to go (we are already doing the same for the SSL/TLS protocols).

@kingthorin the problem with the proposed change (as is) is that it's already changing the behaviour of CFU. That's why the comment about using (just) the default cipher suites when `relaxedTrust` is `false`.
Regarding the 2nd suggestion, I'm ok with that and with the proposed option, except for its location. I think it's more appropriate in "Connection" panel, since the option would apply to all connections not just the proxied ones.

kingthorin:
Oh I can wrap the change in a check for `relaxedTrust != false`, will that be sufficient for cfu accommodation?

thc202:
Yes, that should do it.

kingthorin:
K, will update the PR tonight and open an issue for the second item.

@ElColmo yes I'll use something more like the text you suggested.

kingthorin:
Ugh, nothing is ever simple.....ClientSslSocketsDecorator is static but relaxedTrust is not static. I tried switching relaxedTrust to static and things seemed to behave normally or as expected, but before I updated the PR I wanted to ask if there might be something I'm overlooking?

thc202:
That might create problems, checking for updates would change `relaxedTrust`  to `false` which would end up changing the behaviour of "normal" connections.
You can just change `ClientSslSocketsDecorator` to an inner class, or add a constructor to pass the `relaxedTrust` state.

thc202:
Looking at the code again, you would also need to change `clientSslSocketsDecorator` to an instance variable (and `createDecoratedClientSslSocketFactory` to an instance method).

kingthorin:
@thc202 thanks, will tackle it at some point this week.

kingthorin:
Ok I may have lied, this might be beyond me :disappointed: 
> You can just change `ClientSslSocketsDecorator` to an inner class

It already is...

> Looking at the code again, you would also need to change `clientSslSocketsDecorator` to an instance variable (and `createDecoratedClientSslSocketFactory` to an instance method).

Instance variable and method in relationship to what?

thc202:
It is a nested class but not an inner class (unless I'm switching the names?).
Changing from class variable/method to instance variable/method (i.e. just drop the static modifier, the same for inner class).

No, you can remove the merge commit any time. There's two ways merge or rebase, the latter is preferable since it does not add the merge commits.

kingthorin:
@thc202 thanks, will work on it some more over the weekend.

kingthorin:
I need someone else to test. I can't reproduce the error on Windows :disappointed: 

I even setup XAMPP portable locally and configured it for only RSA_WITH_RC4_128_SHA, but with Oracle JDK/JRE it didn't fail. The Docs I reviewed earlier suggested that this "should" fail with the Oracle implementation but it doesn't seem to for whatever reason...... (with or without these changes).

kingthorin:
Seems I had u76 installed not u79. I'll upgrade and see if I can make it fail, then further test the change.

kingthorin:
No luck. 

Switched to u79, still doesn't fail as reported :disappointed: 

The httpsinfo addon verifies that only RSA_WITH_RC4_128_SHA is being offered.

When I connect directly either via IE or Firefox they both confirm RC4_128. If I proxy the browsers get an upgraded connection to ZAP.....

If someone out there has an OpenJDK dev system already setup and can test that would really help. I'm not sure when I'll get a chance to setup a Kali system, get Eclipse installed, etc.

ElColmo:
I have a working Kali system, which uses  openjdk. What precisely do you
want me to check?

It should be possible to test it on windows by installing openjdk windows
binaries, and playing with zap.bat. I don't currently have access to a
windows machine to verify that though..

Colm

kingthorin:
Just need someone to grab this branch and see if it addresses the issue.

ElColmo:
Do you have a link I can use to get the branch?

kingthorin:
I think one of thrse is what you're after: 
*  https://github.com/kingthorin/zaproxy/tree/issue-1975-default-disabled-cipher-suite-support-RC4-SHA-etc
* https://github.com/kingthorin/zaproxy.git

ElColmo:
My local Apache is configured with just the RC4-SHA cipher.

My machine uses OpenJDK, but the version has been updated from the version where I originally noted the issue (this is the latest and greatest on Kali 2.0 as of today): 

# java -version
Java version "1.7.0_85"
OpenJDK Runtime Environment (IcedTea 2.6.1) (7u85-2.6.1-5~deb8u1)
OpenJDK 64-Bit Server VM (build 24.85-b03, mixed mode)
# 

With ZAP *not* proxying, Iceweasel throws the following (which is what I would expect): "An error occurred during a connection to 127.0.0.1. Cannot communicate securely with peer: no common encryption algorithm(s). (Error code: ssl_error_no_cypher_overlap)"

With ZAP proxying, ZAP fails with the same error as before:  

GET https://127.0.0.1/ HTTP/1.1
-->
HTTP/1.1 502 Bad Gateway
Content-Length: 1845
Content-Type: text/plain; charset=UTF-8

ZAP Error [javax.net.ssl.SSLHandshakeException]: Received fatal alert: handshake_failure

Stack Trace:
javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:154)
	at sun.security.ssl.SSLSocketImpl.recvAlert(SSLSocketImpl.java:1991)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1098)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1344)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1371)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1355)
	at org.parosproxy.paros.network.SSLConnector.createSocket(Unknown Source)
	at org.apache.commons.httpclient.HttpConnection.open(HttpConnection.java:707)
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$HttpConnectionAdapter.open(MultiThreadedHttpConnectionManager.java:1361)
	at org.apache.commons.httpclient.HttpMethodDirector.executeWithRetry(Unknown Source)
	at org.apache.commons.httpclient.HttpMethodDirector.executeMethod(Unknown Source)
	at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:397)
	at org.parosproxy.paros.network.HttpSender.executeMethod(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.runMethod(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.send(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAuthenticated(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(Unknown Source)
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(Unknown Source)
	at org.parosproxy.paros.core.proxy.ProxyThread.processHttp(Unknown Source)
	at org.parosproxy.paros.core.proxy.ProxyThread.run(Unknown Source)

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -572,6 +573,9 @@ public void decorate(SSLSocket sslSocket) {
 				readSupportedProtocols(sslSocket);
 			}
 			sslSocket.setEnabledProtocols(getClientEnabledProtocols());
+			if (!relaxedTrust) {
```
thc202: The `relaxedTrust` should not be negated, we want to use all cipher suites when `relaxedTrust` is `true`.

kingthorin:
Ok. I'll work on getting an environment set up.

src/org/parosproxy/paros/network/SSLConnector.java
kingthorin: Fixed, now matches what we talked about back here:
 * kingthorin https://github.com/zaproxy/zaproxy/pull/1977#issuecomment-148732399
 * thc202 https://github.com/zaproxy/zaproxy/pull/1977#issuecomment-148734499

kingthorin:
Let me know if that does it.

ElColmo:
Looking at this now..

ElColmo:
This now allows the user to navigate to the website, using an insecure cipher, while proxying via ZAP.

Java version: 
java version "1.7.0_85"`
OpenJDK Runtime Environment (IcedTea 2.6.1) (7u85-2.6.1-6+deb8u1)
OpenJDK 64-Bit Server VM (build 24.85-b03, mixed mode)

Nice one!

Point for the user to bear in mind: while proxying via ZAP, the HTTPS cipher suite shown in the browser (ie, supported by Zap, and negotiated between Zap and the web browser) is not necessarily one supported by the actual web server, so the user should validate that the cipher suites supported by the web server are secure.

thc202:
@ElColmo thanks for giving it a try.

@kingthorin would you mind fixup the commits for merge of PR?

kingthorin:
@ElColmo thanks for re-testing glad it worked! :smile:
@thc202 yup no problem. It probably also makes more sense to change the check to use positive logic, i.e.: `if (relaxedTrust)` versus `if (relaxedTrust != false)` .

thc202:
@kingthorin yeah, it's easier to read.

kingthorin:
Done.

PR #1985, opened 2015-10-16
Bring up the Modify dialog if the user doubles on a row in a AbstractMultipleOptionsBaseTablePanel

psiinon:


thc202:
I guess we should mention somewhere in the help pages that it's possible to edit the entries with double click?

PR #2018, opened 2015-11-02
Fix issues with installation of passive and active scanners

### LINKED ISSUE #1969: Issues with installation of scanners
Scanners show wrong quality after installation:
Steps to reproduce the issue:
1. Run ZAP (2.4.2), without add-ons with scanners (so that's easier to reproduce the issue);
2. Install add-ons with active and passive scanners (for example, add-ons "Active scanner rules" and "Passive scanner rules");
3. Open the "Scan Policy Manager" dialogue (Ctrl+P), select "Default policy", press "Modify" button and note that the installed scanners shown "Unknown" quality;
4. Open the "Options" dialogue, select "Passive Scan Rules" and note the same issue.

Active scanners added twice on add-on installation:
Steps to reproduce the issue:
1. Run ZAP (2.4.2), without add-ons with active scanners;
2. Install the add-on with active scanners (for example, "Active scanner rules" add-on);
3. Note the messages in the log, which means that the scanners are being loaded twice:
```
38801 [AWT-EventQueue-0] INFO org.parosproxy.paros.core.scanner.PluginFactory  - loaded plugin Path Traversal
38801 [AWT-EventQueue-0] INFO org.parosproxy.paros.core.scanner.PluginFactory  - Ignoring (apparently) less stable scanner version, id=6, ExistingPlugin[Status=release, Class=org.zaproxy.zap.extension.ascanrules.TestPathTraversal], LessStablePlugin[Status=unknown, Class=org.zaproxy.zap.extension.ascanrules.TestPathTraversal]
```
with 2.4.0, the messages are different:
```
79916 [AWT-EventQueue-0] INFO org.parosproxy.paros.core.scanner.PluginFactory  - loaded plugin Path Traversal
79916 [AWT-EventQueue-0] ERROR org.parosproxy.paros.core.scanner.PluginFactory  - Duplicate id 6 org.zaproxy.zap.extension.ascanrules.TestPathTraversal org.zaproxy.zap.extension.ascanrules.TestPathTraversal
```
4\. Uninstall the add-on with active scanners (for example, "Active scanner rules" add-on);
5. Open the "Scan Policy Manager" dialogue (Ctrl+P), modify the "Default policy" and note, by the errors logged, that the scanners of the uninstalled add-on are still loaded (since just the first scanner was removed):
```
151700 [AWT-EventQueue-0] ERROR org.parosproxy.paros.core.scanner.PluginFactory  - Can't find resource for bundle java.util.PropertyResourceBundle, key ascanrules.testpathtraversal.name
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key ascanrules.testpathtraversal.name
    at java.util.ResourceBundle.getObject(ResourceBundle.java:450)
    at java.util.ResourceBundle.getString(ResourceBundle.java:407)
    at org.zaproxy.zap.utils.I18N.getString(Unknown Source)
    at org.zaproxy.zap.extension.ascanrules.TestPathTraversal.getName(TestPathTraversal.java:185)
    at org.parosproxy.paros.core.scanner.PluginFactory.loadAllPlugin(Unknown Source)
    at org.zaproxy.zap.extension.ascan.ScanPolicy.<init>(Unknown Source)
    at org.zaproxy.zap.extension.ascan.PolicyManager.loadPolicy(Unknown Source)
    at org.zaproxy.zap.extension.ascan.PolicyManager.getPolicy(Unknown Source)
    at org.zaproxy.zap.extension.ascan.ExtensionActiveScan.showPolicyDialog(Unknown Source)
    at org.zaproxy.zap.extension.ascan.PolicyManagerDialog$2.actionPerformed(Unknown Source)
```

Remarks:
Although the issues are different they are related (fixing the first should fix the second).

ZAP Version:
Versions >= 2.4.0

thc202:
Fix issues with installation of passive and active scanners

Change to load and set the quality of the passive and active scanners
during installation of the add-on, preventing the quality from being
"Unknown". The loading/instantiation of the scanners is now done by the
same code (refactored to new class AddOnLoaderUtils, also refactored the
code that loads/instantiates a class, or classes, from an add-on). The
methods previously used during the installation, ExtensionPassiveScan.
addPassiveScanner(String) and PluginFactory.loadedPlugin(String), are
now deprecated as the quality is not correctly set. The method
PluginFactory.unloadedPlugin(String) is also deprecated, no longer used
by core code (the active scanners are now removed using the loaded
instances).
Change to load the active scanners just once by checking that the
scanners were not already loaded/added. The active scanners were being
added twice because the method PluginFactory.loadedPlugin(String) called
the method that initialised the list of scanners which was being
initialised with the add-ons already loaded, which included the add-on
being installed. The PluginFactory now checks if a scanner is already
loaded by reference and when removing, to make sure that's checking and
removing the exact instance (added). Previously it was using the equals
of the default implementation (AbstractPlugin) which checked the
equality by using the ID of the scanner, which would lead to unexpected
results when different scanners have (wrongly) the same ID (as it could
remove the other scanner).
The AddOn now keeps references to the scanners loaded, which is used to
prevent loading the scanners more than once and are used during the
uninstallation of the add-on.
Fix #1969 - Issues with installation of scanners

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/PluginFactory.java
+++ b/src/org/parosproxy/paros/core/scanner/PluginFactory.java
@@ -39,6 +39,7 @@
 // ZAP: 2015/01/04 Issue 1486: Add-on components leak
 // ZAP: 2015/07/25 Do not log error if the duplicated scanner is (apparently) a newer/older version
 // ZAP: 2015/08/19 Issue 1785: Plugin enabled even if dependencies are not, "hangs" active scan
+// ZAP: 2015/11/02 Issue 1969: Issues with installation of scanners
 
 package org.parosproxy.paros.core.scanner;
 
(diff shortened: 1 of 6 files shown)


kingthorin:
Yay, I noticed this quality "Unknown" behavior a few weeks ago but hadn't been able to look into it at all or even to open an issue for it. Thanks!

PR #2026, opened 2015-11-04
Report the host:port ZAP is listening on in daemon mode, or exit if it cant

### LINKED ISSUE #1920: Report the host:port ZAP is listening on in daemon mode, or exit if it cant
Following the instructions at https://github.com/zaproxy/zaproxy/wiki/Docker#zap-headless and have the Docker portion installed, but running the following command just stalls "Creating new root CA" (on OS X 10.10.5):

```sh
$ docker run -u zap -p 8080:8080 -i owasp/zap2docker-stable zap.sh -daemon -port 8080

Found Java version 1.7.0_79
Available memory:  996 MB
Setting jvm heap size: -Xmx256m
364 [main] INFO org.zaproxy.zap.DaemonBootstrap  - OWASP ZAP 2.4.2 started.
814 [main] INFO hsqldb.db.HSQLDB379AF3DEBD.ENGINE  - dataFileCache open start
826 [main] INFO hsqldb.db.HSQLDB379AF3DEBD.ENGINE  - dataFileCache open end
1381 [main] INFO org.parosproxy.paros.network.SSLConnector  - Reading supported SSL/TLS protocols...
1382 [main] INFO org.parosproxy.paros.network.SSLConnector  - Using a SSLEngine...
1721 [main] INFO org.parosproxy.paros.network.SSLConnector  - Done reading supported SSL/TLS protocols: [SSLv2Hello, SSLv3, TLSv1, TLSv1.1, TLSv1.2]
1741 [main] INFO org.parosproxy.paros.extension.option.OptionsParamCertificate  - Unsafe SSL renegotiation disabled.

...

5771 [Thread-4] INFO org.zaproxy.zap.extension.dynssl.ExtensionDynSSL  - Creating new root CA
```

It's been stuck on that step for ~10m, so I'm guessing it's stuck and it doesn't take that long to create a cert. I've also tried running the command a few times, to make sure it wasn't a 1-time thing.

psiinon:
Fixes #1920 Report the host:port ZAP is listening on in daemon mode, or exit if it cant

### DIFF
--- a/src/org/parosproxy/paros/control/Control.java
+++ b/src/org/parosproxy/paros/control/Control.java
@@ -56,6 +56,7 @@
 // ZAP: 2015/02/10 Issue 1208: Search classes/resources in add-ons declared as dependencies
 // ZAP: 2015/04/02 Issue 321: Support multiple databases and Issue 1582: Low memory option
 // ZAP: 2015/09/17 Issue 1914: Support multiple add-on directories
+// ZAP: 2015/11/04 Issue 1920: Report the host:port ZAP is listening on in daemon mode, or exit if it cant
 
 package org.parosproxy.paros.control;
 
@@ -107,7 +108,7 @@ public class Control extends AbstractControl implements SessionListener {
 		super(null, null);
 	}
 
-	private void init(ControlOverrides overrides) {
+	private boolean init(ControlOverrides overrides) {
 
 		// Load extensions first as message bundles are loaded as a side effect
 		loadExtension();
@@ -122,7 +123,7 @@ public class Control extends AbstractControl implements SessionListener {
 		    getExtensionLoader().hookSiteMapListener(view.getSiteTreePanel());
 		}
 		
-		proxy.startServer();
+		return proxy.startServer();
     }
 
     public Proxy getProxy() {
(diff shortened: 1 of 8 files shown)


kingthorin:
I just added this type of comment to another issue in the tracker but it's probably also applicable here.

If there user has bound to other than loop local do we want to warn/remind them that exposes ZAP to other network users? (Consider 0.0.0.0 binding, or specifically binding to a non loop local address 1918 or externally routeable)

psiinon:
@kingthorin via a message to stdout? As its not the default, they will have to have chosen to change the configuration..

kingthorin:
Ya. I'm just thinking for instances where someone uses ZAP, doesn't touch it for weeks/months :cry:, then goes back to use it again. We don't need to prevent it, but warning/reminding might be a good move...

psiinon:
I suspect that in daemon mode a log warning will be ignored by everyone. And in UI mode I'm wary of too many popups :/ Maybe we should warn people when they first set the addr via the UI and provide an option to warn them when they restart ZAP?? Might need more thought, so I dont think it should be part of this PR...

kingthorin:
Agreed on all counts

PR #2048, opened 2015-11-12
zaproxy/zaproxy#2000 general fix title caps

kingthorin:
Initial attempt to address issue 2000. 

Might need to be handled as multiple PRs off the same working branch. This is likely all I'll get completed prior to 2.4.3.


thc202:
Link to issue, #2000.

PR #2054, opened 2015-11-15
Faster loading of persisted sessions

NVolcz:
Switched to ConcurrentSkipListSet as internal data structure in SiteNode
due to heavy load on contains method. Have not measured how this affects
other operations.
Changed how alerts are loaded from DB. Instead of fetching all IDs
and iteratively fetch one by one they are now loaded all at once.
Added a flag to addAlert to allow less notifyChange calls which causes redraw

### DIFF
--- a/src/org/zaproxy/zap/authentication/FormBasedAuthenticationMethodType.java
+++ b/src/org/zaproxy/zap/authentication/FormBasedAuthenticationMethodType.java
@@ -379,7 +379,7 @@ public class FormBasedAuthenticationMethodType extends AuthenticationMethodType 
 			values.put("methodName", API_METHOD_NAME);
 			values.put("loginUrl", loginRequestURL);
 			values.put("loginRequestData", this.loginRequestBody);
-			return new ApiResponseSet("method", values);
+			return new ApiResponseSet<String>("method", values);
 		}
 
 		@Override
--- a/src/org/zaproxy/zap/authentication/GenericAuthenticationCredentials.java
+++ b/src/org/zaproxy/zap/authentication/GenericAuthenticationCredentials.java
@@ -63,7 +63,7 @@ public class GenericAuthenticationCredentials implements AuthenticationCredentia
 	public ApiResponse getApiResponseRepresentation() {
 		Map<String, String> values = new HashMap<>(paramValues);
 		values.put("type", API_NAME);
-		return new ApiResponseSet("credentials", values);
+		return new ApiResponseSet<String>("credentials", values);
 	}
 
 	/**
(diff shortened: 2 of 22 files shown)


psiinon:
Just to let you know we're not ignoring this!
My concerns are that the code retrieves all of the alerts in one go from the db - I'm not sure how well this will scale. We'll have to test it on some very large sessions. Paging the alerts (eg in 100s) might be more scalable.

NVolcz:
How would it be less scalable than what we already have? By fetching everything in one go there will be less overhead and according to my testing it significantly increases the performance of loading large session.

psiinon:
Because we might run into memory problems?
Thats why we need to test it.
The aws set up we use for regression testing does now give us a much better environment for this sort of testing, its just a matter of finding the time to do the tests :/
So keep hassling us ;)

NVolcz:
Even if you have the memory it won't work right now (for me the loading went from 30min+ to less than 1min~). This pull requests at least makes huge sessions viable. 
Will run some tests on memory usage. Do ZAP have any existing performance tests that I can take inspiration from? Are there anything more I should be aware of before I continue?

psiinon:
We dont have any performance tests right now, other than running against wavsep and wivet: http://zapbot.github.io/zap-mgmt-scripts/index.html
However ... saving and then restoring (and timing!) a session after each test run would be a really useful thing to do - I'll look into that :)

NVolcz:
I have been trying to write some tests today but is sorry to say that the singleton patterns are killing me! Any plans on killing them (with fire)?
Secondly the code formatting is terrible any plans on doing a whole project format soon?

psiinon:
We have a long running development called ZAP as a Service: http://zaproxy.blogspot.co.uk/2015/05/zap-as-service-zaas.html
In order to support multiple users etc we'll _have_ to kill a load of singletons with fire :)
And we have talked about autoformatting the whole project, that would be a very useful thing to do :)

psiinon:
FYI I've just written a script for testing (and timing) backup and restoring: https://github.com/zapbot/zap-mgmt-scripts/blob/master/zap-backup-test.py
I need to sanity check this vs a big scan (which I'm doing now on aws) and if so we can then use your changes:)

kingthorin:
Great work guys!

NVolcz:
Seems like this issue has been fixed by itself somewhere between 3bb046b-006a35b. 
Should we close this then?

kingthorin:
Re-opening. I'm not convinced this was resolved ...

thc202:
Correct, the issue was not yet fixed in develop branch. There were some changes in SiteNode/Context that conflicted with the changes (NVolcz/zaproxy@7df552d3131899d7323d98adddd5734b4702c5db) though.

Note that this pull request no longer contains the proposed fix.

NVolcz:
Have you been able to replicate the issue? When I tried with my test
session it did not take nearly as long time to load as before. Can take
another look at it.

NVolcz:
Experienced the slow loading today! Could it be that when the proxy is used before the loading has finished it slows down?

thc202:
It really depends on the contents of the session, how many messages (and, also important, its type) and alerts it has.

kingthorin:
I believe this has been addressed by https://github.com/zaproxy/zaproxy/pull/2968

Agreed? Close this one?

thc202:
Agreed, the changes done to develop in the meantime should improve the loading times.

PR #2105, opened 2015-12-04
Fix: Evil NPE

Harinus:
tokensSet is not checked for null

```
Traceback (most recent call last):
  File "<script>", line 77, in responseReceived
  File "<script>", line 133, in authenticate
  File "<script>", line 139, in resendMessageWithSession
	at org.zaproxy.zap.session.CookieBasedSessionManagementHelper.processMessageToMatchSession(CookieBasedSessionManagementHelper.java:64)

	at org.zaproxy.zap.session.CookieBasedSessionManagementHelper.processMessageToMatchSession(CookieBasedSessionManagementHelper.java:49)

	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)

	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)

	at java.lang.reflect.Method.invoke(Unknown Source)
```

### DIFF
--- a/src/org/zaproxy/zap/session/CookieBasedSessionManagementHelper.java
+++ b/src/org/zaproxy/zap/session/CookieBasedSessionManagementHelper.java
@@ -61,6 +61,12 @@ public class CookieBasedSessionManagementHelper {
 
 		// Make a copy of the session tokens set, as they will be modified
 		HttpSessionTokensSet tokensSet = session.getTokensNames();
+		
+		// If no tokens exists create dummy Object -> NPE
+		if (tokensSet == null) {
+			tokensSet = new HttpSessionTokensSet();
+		}
+		
 		Set<String> unsetSiteTokens = new LinkedHashSet<>(tokensSet.getTokensSet());
 
 		// Iterate through the cookies in the request


src/org/zaproxy/zap/session/CookieBasedSessionManagementHelper.java
```diff
@@ -61,6 +61,12 @@ public static void processMessageToMatchSession(HttpMessage message, List<HttpCo
 
 		// Make a copy of the session tokens set, as they will be modified
 		HttpSessionTokensSet tokensSet = session.getTokensNames();
+		
+		// If no tokens exists create dummy Object -> NPE
+		if (tokensSet == null) {
+			tokensSet = new HttpSessionTokensSet();
```
thc202: Since it's empty it could just skip the following two loops (i.e. not being used at all).

thc202:
How did this happen? Did you create the HttpSession "manually"?

Maybe it might be better to change HttpSession to not allow null tokenNames (either throwing an exception or use an empty HttpSessionTokensSet)?
Also, having an HttpSession with no session token names might lead to (re)authentication issues if the message already contained session tokens.

Harinus:
@thc202 Here is my code that made the NPE happen: https://github.com/Harinus/ZAP-Selenium-Auth/blob/master/selenium-auth.py

psiinon:
I also got this NPE during automation, so I think CookieBasedSessionManagementHelper should handle it rather than throwing an exception. I'll close my PR #2747 - should have spotted that this addressed the same issue ;)

thc202:
Looks good to me (even though some of the code could be skipped).

kingthorin:
This is a really old PR. Shall we close (stale), fix, resubmit? Fix and manual merge (however that workflow works)? Merge then address flow control separately?

thc202:
I'm fine with merging as is (we change the flow control after or if even prevent the code from having null HttpSessionTokensSet, per https://github.com/zaproxy/zaproxy/pull/2105#issuecomment-164737440 ).

psiinon:
Agree with merging now - it fixes the NPE and we can make other changes later

PR #2117, opened 2015-12-10
set / update default threshold and strength for a scan policy

wmaintw:
main changes:
- AddScanPolicy API enhanced, 2 optional params added, so that user could set policy's default alert threshold and attack strength
- UpdateScanPolicy API added, so that user could modify the default alert threshold or attack strength for a given scan policy

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
+++ b/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
@@ -33,12 +33,14 @@ import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.apache.commons.lang.StringEscapeUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.core.scanner.Category;
 import org.parosproxy.paros.core.scanner.HostProcess;
 import org.parosproxy.paros.core.scanner.Plugin;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
+import org.parosproxy.paros.core.scanner.Plugin.AttackStrength;
 import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
@@ -128,7 +130,7 @@ public class ActiveScanAPI extends ApiImplementor {
 	public ActiveScanAPI (ExtensionActiveScan controller) {
 		this.controller = controller;
         this.addApiAction(new ApiAction(ACTION_SCAN,
-        		new String[] {PARAM_URL}, 
+        		new String[] {PARAM_URL},
         		new String[] {PARAM_RECURSE, PARAM_JUST_IN_SCOPE, PARAM_SCAN_POLICY_NAME, PARAM_METHOD, PARAM_POST_DATA}));
 		this.addApiAction(new ApiAction(
 				ACTION_SCAN_AS_USER,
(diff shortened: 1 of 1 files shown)


wmaintw:
zap is using apache commons-lang 2.6 version, and previously my code is using 3.4 version, according to the apache official website, the package name is changed since 3.0. 

currently the build is fixed by modify the package name that I'm using, but it's better to use newer version of commons-lang, after all 2.6 is a legacy version. I will make this change in another changeset and submit pull request later.

thc202:
The library Apache Commons Lang version 2.6 is a dependency of dependencies (e.g. Apache Commons Configuration version 1.9, JSON-lib version 2.4...), so there's not much to update (it's already the latest version of 2.x series).
We could add Apache Commons Lang version 3.x but seems overkill for just one method (that's provided by older version).

wmaintw:
Ok, I agree with you, if the commons lang is a dependency of dependencies, there not much we can do for it.

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -128,7 +131,7 @@
 	public ActiveScanAPI (ExtensionActiveScan controller) {
 		this.controller = controller;
         this.addApiAction(new ApiAction(ACTION_SCAN,
-        		new String[] {PARAM_URL},
```
thc202: Indentation/formatting changes should be removed. [1]

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -366,6 +372,8 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 				}
 				policy = controller.getPolicyManager().getTemplatePolicy();
 				policy.setName(newPolicyName);
+				policy.setDefaultThreshold(getAlertThreshold(params));
```
thc202: The alert threshold and attack strength should be set only when those are specified, otherwise we have to maintain the defaults in two places (or expose the defaults, but no point re-setting them in any case).

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -377,6 +385,12 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 				}
 				controller.getPolicyManager().deletePolicy(policy.getName());
 				break;
+			case ACTION_UPDATE_SCAN_POLICY:
+				policy = getScanPolicyFromParams(params);
+				changeAlertThreshold(policy, params);
```
thc202: It could be validated that at least one of the parameters is being changed.

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -377,6 +385,12 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+				changeAttackStrength(policy, params);
+				controller.getPolicyManager().savePolicy(policy);
```
thc202: Scan policy should not be saved if not changed.

src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -386,6 +400,38 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 		return ApiResponseElement.OK;
 	}
 
+	private AlertThreshold getAlertThreshold(JSONObject params) throws ApiException {
+		if (!isParamExists(params.getString(PARAM_ALERT_THRESHOLD))) {
```
thc202: It should be checked that the parameter was really sent (by using `JSONObject.has(String)` or `JSONObject.containsKey(Object)`) before obtaining it's value. Otherwise, this statement (more precisely `params.getString(...)`) will lead to `JSONException` if the parameter was not sent (given that it's optional it's very likely to happen).

wmaintw:
@thc202 thanks for your feedbacks, I've made some modifications to the code

thc202:
@wmaintw changes look good to me. Would you mind fixup the commits?
Also, if you do want to keep the indentation/formatting changes, could you do so in another commit (before or after the commit with the actual changes)?

kingthorin:
@wmaintw is there a fixup coming?

kingthorin:
@wmaintw is there a fixup coming?

thc202:
@wmaintw any news on this?

wmaintw:
@thc202 I'm really sorry that I was quite busy with my delivery project. I will fixup the format, let these commits contain the changes only.

thc202:
@wmaintw no problem. Thanks!

kingthorin:
@wmaintw fixup still coming?

kingthorin:
Now  conflicting and still pending fixup/squash

thc202:
Will fixup/rebase and merge manually (just pending format/indentation changes).

thc202:
Changes applied, thank you!

@wmaintw how would you like to be credited?

Issue #1341, opened 2015-06-04
Update alert XML schema to remove element-level name collision


zapbot:
```
When using the API to fetch the XML for post-scan alerts, the file looks something like:

<?xml version="1.0" encoding="UTF-8" standalone="no"?>
    <alerts type="list">
        <alert type="set">
            <alert>X-Content-Type-Options header missing</alert>
            <reliability>Warning</reliability>
            <solution>.....

The schema contains alerts --> alert --> alert.  The first one is a set of problems.
 The second one is a single alert.  The third one is the title of that alert.  

If you are writing xquery or calling various java/python XML API searches, the name
clash for <alert> adds extra complexity to the code.  So, a search for "//alert" must
also test to see if this is a leaf node or contains additional elements.  The name
collision makes quickly coded scripts harder.  

Junior developers (from my observation) would tend to search by element name and then
add extra "if" statements to make sure it was the correct name.  For one program, I
was told to ignore every-other row. 

We can save headaches by renaming <alert> #2  to <title>.  Yes, this supports lazy
programming, but for quick scripts, it is nicer.

Renaming will break compatibility with existing downstream tools, so I propose a phased
approach.  In the next version of ZAP, include both an <alert> and <title> element,
they would have the same value.  Inform people that <alert> is deprecated.  Modify
the various client python/java API that consume the XML to start reading <title> instead
if ZAP version == this version.  In a later & larger point release, remove <alert>
and make it obsolete.  

If there is already a discussion to have a larger re-vamping of the XML schema which
breaks backward compatibility, then this request is easily accommodated.   In general
for the future, please ensure that we never have element-level name collision.
```

Original issue reported on code.google.com by `veggiespam` on 2014-09-17 15:54:36

PR #2145, opened 2015-12-19
deprecate use of alert vs name in Alert

kingthorin:
Make the name of an alert be a name field (vs.
alert within alert), which is more developer and user friendly. Ensure
web API components updated, and xsl for reporting purposes. Fixes
zaproxy/zaproxy#1341

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/Alert.java
+++ b/src/org/parosproxy/paros/core/scanner/Alert.java
@@ -40,6 +40,7 @@
 // ZAP: 2015/02/09 Issue 1525: Introduce a database interface layer to allow for alternative implementations
 // ZAP: 2015/08/24 Issue 1849: Option to merge related issues in reports
 // ZAP: 2015/11/16 Issue 1555: Rework inclusion of HTML tags in reports 
+// ZAP: 2016/02/26 Deprecate alert as an element of Alert in favour of name
 
 package org.parosproxy.paros.core.scanner;
 
@@ -95,7 +96,7 @@ public class Alert implements Comparable<Object>  {
 	
 	private int		alertId = -1;	// ZAP: Changed default alertId
 	private int		pluginId = 0;
-	private String 	alert = "";
+	private String name = "";
 	private int risk = RISK_INFO;
 	/**
 	 * @deprecated
@@ -131,10 +132,10 @@ public class Alert implements Comparable<Object>  {
 		
 	}
 	
-	public Alert(int pluginId, int risk, int confidence, String alert) {
+	public Alert(int pluginId, int risk, int confidence, String name) {
 		this(pluginId);
 		setRiskConfidence(risk, confidence);
-		setAlert(alert);
+		setName(name);
 	}
 
 	public Alert(RecordAlert recordAlert) {
@@ -185,13 +186,23 @@ public class Alert implements Comparable<Object>  {
 		this.risk = risk;
 		this.confidence = confidence;
 	}
-	
+	/**
+	 * @deprecated (TODO add version) Replaced by {@link #setName}.
+	 * Use of alert has been deprecated in favour of using name.
+	 */
+	@Deprecated
 	public void setAlert(String alert) {
-	    if (alert == null) return;
-	    // ZAP: Changed to not create a new String.
-	    this.alert = alert;
+	    setName(alert);
 	}
-	
+	/**
+	 * Sets the name of the alert to name
+	 * @param name the name to set for the alert
+	 * @since TODO add version
+	 */
+	public void setName(String name) {
+	    if (name == null) return;
+	    this.name = name;
+	}
 	
 	/**
 	 * @deprecated (2.2.0) Replaced by
@@ -304,7 +315,7 @@ public class Alert implements Comparable<Object>  {
 			return 1;
 		}
 		
-		int result = alert.compareToIgnoreCase(alert2.alert);
+		int result = name.compareToIgnoreCase(alert2.name);
 		if (result != 0) {
 			return result;
 		}
@@ -333,7 +344,7 @@ public class Alert implements Comparable<Object>  {
 		Alert item = null;
 		if (obj instanceof Alert) {
 			item = (Alert) obj;
-			if ((pluginId == item.pluginId) && alert.equals(item.alert) && uri.equalsIgnoreCase(item.uri)
+			if ((pluginId == item.pluginId) && name.equals(item.name) && uri.equalsIgnoreCase(item.uri)
 				&& param.equalsIgnoreCase(item.param) && otherInfo.equalsIgnoreCase(item.otherInfo)) {
 				return true;
 			}
@@ -347,7 +358,7 @@ public class Alert implements Comparable<Object>  {
 	public Alert newInstance() {
 		Alert item = new Alert(this.pluginId);
 		item.setRiskConfidence(this.risk, this.confidence);
-		item.setAlert(this.alert);
+		item.setName(this.name);
 		item.setDetail(this.description, this.uri, this.param, this.attack, this.otherInfo, this.solution, this.reference, this.historyRef);
 		return item;
 	}
@@ -356,7 +367,8 @@ public class Alert implements Comparable<Object>  {
 		StringBuilder sb = new StringBuilder(150); // ZAP: Changed the type to StringBuilder.
 		sb.append("<alertitem>\r\n");
 		sb.append("  <pluginid>").append(pluginId).append("</pluginid>\r\n");
-		sb.append("  <alert>").append(replaceEntity(alert)).append("</alert>\r\n");
+		sb.append("  <alert>").append(replaceEntity(name)).append("</alert>\r\n"); //Deprecated in TODO add version, maintain for compatibility with custom code
+		sb.append("  <name>").append(replaceEntity(name)).append("</name>\r\n");
 		sb.append("  <riskcode>").append(risk).append("</riskcode>\r\n");
 		sb.append("  <confidence>").append(confidence).append("</confidence>\r\n");
 		sb.append("  <riskdesc>").append(replaceEntity(MSG_RISK[risk] + " (" + MSG_CONFIDENCE[confidence] + ")")).append("</riskdesc>\r\n");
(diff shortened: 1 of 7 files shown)


kingthorin:
Could use advice on resolution of the current conflict status. It was clean 2 months ago.

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -185,13 +185,18 @@ public void setRiskConfidence(int risk, int confidence) {
 		this.risk = risk;
 		this.confidence = confidence;
 	}
-	
+	/**
+	 * @deprecated (2.4.4) Replaced by {@link #setName}.
```
thc202: Next version should be 2.5.0 (because of the enhancements and deprecations).
I usually add `TODO add version`, that can be later replaced when in the release branch (but the former should be ok too).

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -185,13 +185,18 @@ public void setRiskConfidence(int risk, int confidence) {
...
+	 * Use of alert has been deprecated in favour of using name.
+	 */
+	@Deprecated
 	public void setAlert(String alert) {
-	    if (alert == null) return;
-	    // ZAP: Changed to not create a new String.
-	    this.alert = alert;
+	    setName(alert);
+	}
+	public void setName(String name) {
```
thc202: Consider adding since JavaDoc tag (and document the method while at it ;)

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -356,7 +361,7 @@ public String toPluginXML(String urls) {
 		StringBuilder sb = new StringBuilder(150); // ZAP: Changed the type to StringBuilder.
 		sb.append("<alertitem>\r\n");
 		sb.append("  <pluginid>").append(pluginId).append("</pluginid>\r\n");
-		sb.append("  <alert>").append(replaceEntity(alert)).append("</alert>\r\n");
```
thc202: The "alert" element should still be returned, otherwise older tools/code would stop working.

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -392,12 +397,20 @@ public String replaceEntity(String text) {
 	public String paragraph(String text) {
 		return "<p>" + text.replaceAll("\\r\\n","</p><p>").replaceAll("\\n","</p><p>") + "</p>";
 	}
-    
-    /**
-     * @return Returns the alert.
-     */
+	/**
+	 * @deprecated (2.4.4) Replaced by {@link #getName}.
```
thc202: Same comment as above regarding the version.

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -392,12 +397,20 @@ public String replaceEntity(String text) {
...
+	 */
+	@Deprecated
     public String getAlert() {
-        return alert;
+        return name;
+    }
+	/**
+	 * @return Returns the name of the alert.
+	 */
+    public String getName() {
```
thc202: Same comment as above regarding the since JavaDoc tag.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -706,7 +706,7 @@ public String getXml(SiteNode site) {
     }
     
     private String alertFingerprint(Alert alert) {
-    	return alert.getPluginId() + "/" + alert.getRisk() + "/" + alert.getConfidence();
+    	return alert.getPluginId() + "/" + alert.getName() + "/" + alert.getRisk() + "/" + alert.getConfidence();
```
thc202: Consider adding this in another commit, it's easier to note the (important) change/fix.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -953,7 +953,7 @@ private void getURLs(SiteNode parent, ApiResponseList list) {
 	private ApiResponseSet alertToSet(Alert alert) {
 		Map<String, String> map = new HashMap<>();
 		map.put("id", String.valueOf(alert.getAlertId()));
-		map.put("alert", alert.getAlert());
```
thc202: The "alert" property should still be returned, to keep compatibility with older tools.

thc202:
That's caused by #2194, you need to rebase (and resolve the conflict, it's just a two line change though).

Other review comments:
 - The deprecated getAlert() method is still in use (in AlertTreeModel and AlertViewPanel);
 - It's missing the ZAP comment in Alert class;
 - The client API should not be updated yet (didn't remember of that), we might need to release new versions before releasing "ZAP core" (it already happened to fix an issue).

kingthorin:
Ok so I need to rebase  5878e62 onto my branch to resolve the conflict?

thc202:
That would allow to resolve the conflict but it's preferable to rebase on (latest) develop branch instead (it puts your branch up-to-date with latest changes).

kingthorin:
Ok, I'll tackle all this later in the week.

kingthorin:
Ok still need to address a bunch of feedback but now in a non-conflicting state :smile:

Let me know if you see something wrong...

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Fair enough, reverted

kingthorin:
- [x] The deprecated getAlert() method is still in use (in AlertTreeModel and AlertViewPanel);
- [x] It's missing the ZAP comment in Alert class;
- [ ] The client API should not be updated yet (didn't remember of that), we might need to release new versions before releasing "ZAP core" (it already happened to fix an issue).

So I should revert the client API changes?

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -131,10 +132,10 @@ public Alert(int pluginId) {
 		
 	}
 	
-	public Alert(int pluginId, int risk, int confidence, String alert) {
+	public Alert(int pluginId, int risk, int confidence, String name) {
 		this(pluginId);
 		setRiskConfidence(risk, confidence);
-		setAlert(alert);
+		setAlert(name);
```
thc202: It should also call setName.

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -185,13 +186,24 @@ public void setRiskConfidence(int risk, int confidence) {
...
+	 */
+	@Deprecated
 	public void setAlert(String alert) {
-	    if (alert == null) return;
-	    // ZAP: Changed to not create a new String.
-	    this.alert = alert;
+	    setName(alert);
+	}
+	/**
+	 * @since TODO add version
```
thc202: The description of the method should be before the JavaDoc tags, otherwise it's included in the tag (the since tag can be added after the param/return tags).

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -392,12 +405,22 @@ public String replaceEntity(String text) {
...
+	 * @return Returns the alert.
+	 */
+	@Deprecated
     public String getAlert() {
-        return alert;
+        return name;
+    }
+	/**
+	 * @since TODO add version
+	 * <p>
```
thc202: The paragraph it not needed to separate the tags.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: Are you going to open a pull request instead?

thc202:
No, branch looks good to me :)

Yes, I think it's better. The changes can be applied closer to ZAP's release date.
@psiinon what do you think?

psiinon:
Yeah, I think we should be _very_ wary of changing the API in any way so stashing the API changes is safer.

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

src/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Done

kingthorin:
Ok, I've rewound the clientapi changes.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -975,6 +975,7 @@ private ApiResponseSet alertToSet(Alert alert) {
 		map.put("id", String.valueOf(alert.getAlertId()));
 		map.put("pluginId", String.valueOf(alert.getPluginId()));
 		map.put("alert", alert.getAlert());
```
thc202: This should also use getName() (to not use the deprecated method) and it might worth adding a comment like the one in the Alert.toPluginXML(String). I'd also add a XXX task tag to be more visible, for future removal.

thc202:
I missed the class PopupMenuShowAlerts in the previous review, it's still using getAlert().

btw, there's a question in one of the review comments (https://github.com/zaproxy/zaproxy/pull/2145#discussion_r54261657).

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Yes after this set is accepted i'll submit the fingerprint change in a separate PR.

thc202: OK.

src/org/zaproxy/zap/extension/api/CoreAPI.java
kingthorin: Done

kingthorin:
PopupMenuShowAlerts adjusted.

thc202:
Looks good to me.

PR #2162, opened 2016-01-05
add-marketplace-sorting

kingthorin:
Setup auto row sorter for installed and uninstalled addons tables in the
marketplace dialog. This is most handy in the uninstalled section if you
want to change to name sorting (vs. the default quality-name), or if (in
either table) you want the list reversed.

Add StatusUI component to be leveraged elsewhere for status/quality 
info. View now initializes a map (AddOn.Status, StatusUI), and exposes 
a getStatusUI method.

### DIFF
--- a/src/org/parosproxy/paros/view/View.java
+++ b/src/org/parosproxy/paros/view/View.java
@@ -63,6 +63,7 @@
 // ZAP: 2015/08/11 Fix the removal of context panels
 // ZAP: 2015/09/07 Start GUI on EDT
 // ZAP: 2015/11/26 Issue 2084: Warn users if they are probably using out of date versions
+// ZAP: 2016/03/16 Add StatusUI handling
 // ZAP: 2016/03/23 Issue 2331: Custom Context Panels not show in existing contexts after installation of add-on
 
 package org.parosproxy.paros.view;
@@ -76,8 +77,11 @@ import java.awt.event.ActionListener;
 import java.awt.event.KeyEvent;
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.MissingResourceException;
 import java.util.Vector;
 
 import javax.swing.ImageIcon;
@@ -102,6 +106,8 @@ import org.parosproxy.paros.extension.ViewDelegate;
 import org.parosproxy.paros.extension.option.OptionsParamView;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
+import org.zaproxy.zap.control.AddOn;
+import org.zaproxy.zap.control.AddOn.Status;
 import org.zaproxy.zap.extension.ExtensionPopupMenu;
 import org.zaproxy.zap.extension.help.ExtensionHelp;
 import org.zaproxy.zap.extension.httppanel.HttpPanelRequest;
@@ -121,6 +127,7 @@ import org.zaproxy.zap.view.SessionExcludeFromProxyPanel;
 import org.zaproxy.zap.view.SessionExcludeFromScanPanel;
 import org.zaproxy.zap.view.SessionExcludeFromSpiderPanel;
 import org.zaproxy.zap.view.SplashScreen;
+import org.zaproxy.zap.view.StatusUI;
 import org.zaproxy.zap.view.TabbedPanel2;
 import org.zaproxy.zap.view.ZapMenuItem;
 import org.zaproxy.zap.view.messagelocation.MessageLocationHighlightRenderersEditors;
@@ -168,6 +175,8 @@ public class View implements ViewDelegate {
 
     // ZAP: splash screen
     private SplashScreen splashScreen = null;
+
+    private Map<AddOn.Status, StatusUI> statusMap = new HashMap<>();
 
     private boolean postInitialisation;
 
@@ -223,6 +232,25 @@ public class View implements ViewDelegate {
 
         // adds the Request/Response representation buttons in ZAP toolbar
         getMessagePanelsPositionController().restoreState();
+        
+        String statusString;
+        for(Status status : AddOn.Status.values()) {     	
+        	//Try/catch in case AddOn.Status gets out of sync with cfu.status i18n entries
+        	try {
+        		statusString = Constant.messages.getString("cfu.status." + status.toString());
+        	} catch (MissingResourceException mre) {
+        		statusString = status.toString();
+        		
+        		String errString="Caught " + mre.getClass().getName() + " " + mre.getMessage() + 
+						" when looking for i18n string: cfu.status." + statusString;
+        		if (Constant.isDevBuild()) {
+        			logger.error(errString);
+        		} else {
+        			logger.warn(errString);
+        		}
+        	}
+        	statusMap.put(status, new StatusUI(status, statusString));
+        }
     }
 
     public void postInit() {
@@ -917,5 +945,15 @@ public class View implements ViewDelegate {
      */
     public Component getSplashScreen() {
         return splashScreen;
+    }
+    
+    /**
+     * Returns a StatusUI for the given AddOn.Status
+     * @param status the Status for which a StatusUI is wanted
+     * @return a StatusUI
+     * @since TODO add version
+     */
+    public StatusUI getStatusUI(AddOn.Status status) {
+    	return statusMap.get(status);
     }
 }
--- a/src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
+++ b/src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
@@ -63,6 +63,10 @@ public abstract class AddOnsTableModel extends AbstractTableModel {
 
     private AddOnSearcher addOnSeacher;
 
+	/**
+	 * @deprecated (TODO add version) Replaced by {@link #AddOnsTableModel(AddOnCollection, int)}. It will be removed in a future release.
+	 */
+    @Deprecated
     public AddOnsTableModel(Comparator<AddOnWrapper> comparator, AddOnCollection addOnCollection, int progressColumn) {
         super();
 
@@ -73,6 +77,16 @@ public abstract class AddOnsTableModel extends AbstractTableModel {
         this.addOnCollection = addOnCollection;
     }
 
+    public AddOnsTableModel(AddOnCollection addOnCollection, int progressColumn) {
+        super();
+
+        this.comparator = null;
+        this.wrappers = new ArrayList<>();
+        this.progressColumn = progressColumn;
+
+        this.addOnCollection = addOnCollection;
+    }
+    
     public void setAddOnCollection(AddOnCollection addOnCollection) {
         this.addOnCollection = addOnCollection;
     }
@@ -109,13 +123,18 @@ public abstract class AddOnsTableModel extends AbstractTableModel {
     protected void addAddOnWrapper(AddOn addOn, AddOnWrapper.Status status) {
         AddOnWrapper aow = createAddOnWrapper(addOn, status);
         int idx = 0;
-        for (; idx < getAddOnWrappers().size(); idx++) {
-            if (comparator.compare(aow, getAddOnWrappers().get(idx)) < 0) {
-                break;
-            }
+    	if (comparator != null) {
+    		for (; idx < getAddOnWrappers().size(); idx++) {
+        		if (comparator.compare(aow, getAddOnWrappers().get(idx)) < 0) {
+        			break;
+        		}
+                getAddOnWrappers().add(idx, aow);
+        	}
+        } else {
+        	idx = getAddOnWrappers().size();
+        	getAddOnWrappers().add(aow);
         }
 
-        getAddOnWrappers().add(idx, aow);
         fireTableRowsInserted(idx, idx);
 
         refreshEntries();
--- a/src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
+++ b/src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
@@ -28,6 +28,7 @@ import java.awt.HeadlessException;
 import java.io.File;
 import java.net.URL;
 import java.text.MessageFormat;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -41,6 +42,8 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JTabbedPane;
+import javax.swing.RowSorter;
+import javax.swing.SortOrder;
 import javax.swing.border.TitledBorder;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
@@ -355,7 +358,6 @@ public class ManageAddOnsDialog extends AbstractFrame implements CheckForUpdateC
 	private JXTable getInstalledAddOnsTable () {
 		if (installedAddOnsTable == null) {
 			installedAddOnsTable = new JXTable();
-			installedAddOnsTable.setSortable(false);
 			installedAddOnsModel.addTableModelListener(new TableModelListener() {
 				@Override
 				public void tableChanged(TableModelEvent e) {
@@ -366,17 +368,21 @@ public class ManageAddOnsDialog extends AbstractFrame implements CheckForUpdateC
 				}});
 
 			installedAddOnsTable.setModel(installedAddOnsModel);
-			installedAddOnsTable.getColumnModel().getColumn(0).setPreferredWidth(200);
-			installedAddOnsTable.getColumnModel().getColumn(1).setPreferredWidth(400);
-			installedAddOnsTable.getColumnModel().getColumn(2).setPreferredWidth(60);
-			installedAddOnsTable.getColumnModel().getColumn(3).setPreferredWidth(40);
-			
-			installedAddOnsTable.getColumnModel().getColumn(0).setMaxWidth(20);
-			installedAddOnsTable.getColumnModel().getColumn(1).setPreferredWidth(200);
-			installedAddOnsTable.getColumnModel().getColumn(2).setPreferredWidth(400);
-			installedAddOnsTable.getColumnModel().getColumn(3).setPreferredWidth(60);
+			installedAddOnsTable.getColumnModel().getColumn(0).setMaxWidth(20);//icon
+			installedAddOnsTable.getColumnExt(0).setSortable(false);//icon doesn't need to be sortable
+			installedAddOnsTable.getColumnModel().getColumn(1).setPreferredWidth(200);//name
+			installedAddOnsTable.getColumnModel().getColumn(2).setPreferredWidth(400);//description
+			installedAddOnsTable.getColumnExt(2).setSortable(false);//description doesn't need to be sortable
+			installedAddOnsTable.getColumnModel().getColumn(3).setPreferredWidth(60);//update
+			installedAddOnsTable.getColumnExt(3).setSortable(false);//update doesn't need to be sortable
 			installedAddOnsTable.getColumnModel().getColumn(4).setPreferredWidth(40);
-
+			installedAddOnsTable.getColumnExt(4).setSortable(false);//checkbox doesn't need to be sortable
+          
+            //Default sort by name (column 1)
+            List<RowSorter.SortKey> sortKeys = new ArrayList<RowSorter.SortKey>(1);
+            sortKeys.add(new RowSorter.SortKey(1, SortOrder.ASCENDING));
+            installedAddOnsTable.getRowSorter().setSortKeys(sortKeys);
+			
 			DefaultAddOnToolTipHighlighter toolTipHighlighter = new DefaultAddOnToolTipHighlighter(
 					AddOnsTableModel.COLUMN_ADD_ON_WRAPPER);
 			for (int i = 1; i < installedAddOnsTable.getColumnCount(); i++) {
@@ -404,7 +410,6 @@ public class ManageAddOnsDialog extends AbstractFrame implements CheckForUpdateC
 	private JXTable getUninstalledAddOnsTable () {
 		if (uninstalledAddOnsTable == null) {
 			uninstalledAddOnsTable = new JXTable();
-			uninstalledAddOnsTable.setSortable(false);
 
 			uninstalledAddOnsModel.addTableModelListener(new TableModelListener() {
 				@Override
@@ -417,7 +422,8 @@ public class ManageAddOnsDialog extends AbstractFrame implements CheckForUpdateC
 				public void valueChanged(ListSelectionEvent e) {
 					getAddOnInfoButton().setEnabled(false);
 					if (DesktopUtils.canOpenUrlInBrowser() && getUninstalledAddOnsTable ().getSelectedRowCount() == 1) {
-						AddOnWrapper ao = uninstalledAddOnsModel.getAddOnWrapper(getUninstalledAddOnsTable ().getSelectedRow());
+						//convertRowIndexToModel in-case they sorted
+						AddOnWrapper ao = uninstalledAddOnsModel.getAddOnWrapper(getUninstalledAddOnsTable().convertRowIndexToModel(getUninstalledAddOnsTable().getSelectedRow()));
 						if (ao != null && ao.getAddOn().getInfo() != null) {
 							getAddOnInfoButton().setEnabled(true);
 						}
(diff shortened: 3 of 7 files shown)


src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -366,17 +368,23 @@ public void tableChanged(TableModelEvent e) {
...
+			installedAddOnsTable.getColumnModel().getColumn(1).setPreferredWidth(200);//name
+			installedAddOnsTable.getColumnModel().getColumn(2).setPreferredWidth(400);//description
+			installedAddOnsTable.getColumnExt(2).setSortable(false);//description doesn't need to be sortable
+			installedAddOnsTable.getColumnModel().getColumn(3).setPreferredWidth(60);//update
+			installedAddOnsTable.getColumnExt(3).setSortable(false);//update doesn't need to be sortable
 			installedAddOnsTable.getColumnModel().getColumn(4).setPreferredWidth(40);
-
+			installedAddOnsTable.getColumnExt(4).setSortable(false);//checkbox doesn't need to be sortable
+			
+			installedAddOnsTable.setAutoCreateRowSorter(true);
```
thc202: Is this statement really needed? JXTable has that property enabled by default.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -425,13 +432,26 @@ public void valueChanged(ListSelectionEvent e) {
...
+			uninstalledAddOnsTable.getColumnModel().getColumn(2).setPreferredWidth(150);//Name
+			uninstalledAddOnsTable.getColumnModel().getColumn(3).setPreferredWidth(300);//Description
+			uninstalledAddOnsTable.getColumnExt(3).setSortable(false);//Description doesn't need sorting
+			uninstalledAddOnsTable.getColumnModel().getColumn(4).setPreferredWidth(60);//Update (version number)
+			uninstalledAddOnsTable.getColumnExt(4).setSortable(false);//Update doesn't need sorting
+			uninstalledAddOnsTable.getColumnModel().getColumn(5).setPreferredWidth(40);//Checkbox
+			uninstalledAddOnsTable.getColumnExt(5).setSortable(false);//Checkbox doesn't need sorting
+
+
+			uninstalledAddOnsTable.setAutoCreateRowSorter(true);
```
thc202: Same comment as above.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -425,13 +432,26 @@ public void valueChanged(ListSelectionEvent e) {
...
+           
+            //Default sort by status (column 1) descending (Release, Beta, Alpha), and name (column 2) ascending 
+            List<RowSorter.SortKey> sortKeys = new ArrayList<RowSorter.SortKey>(2);
+            sortKeys.add(new RowSorter.SortKey(1, SortOrder.DESCENDING));
```
thc202: This sorting might not work as expected with all languages and current translations. For example selecting Arabic will show "Beta", "Alpha" and "Release" (this status was not yet translated).
Probably safer to compare using the underlying status than relying on String/Collator comparison?

thc202:
Now that the rows are sorted the indexes of the rows in the view and the model might differ, this means that the indexes must be converted (using `JTable.convertRowIndexToModel(int)` or `JTable.convertRowIndexToView(int)`) before being used in the other "side".
You can check that's using the wrong index with the "More Info" button after sorting by name in the "Marketplace" tab. It opens the info of a different add-on.
I think that the only places that require the conversion (to model) is at the lines 427 and 936.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
kingthorin: Hmmm ok I'll look into that. I was trying to avoid having to build or customize the sorter...but hey new challenges are good right :wink:

kingthorin:
@thc202 thanks for the thorough review. I'll try to tackle this later today/tomorrow.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
kingthorin: My bad, I guess I had modeled it after a JTable implementation I'd done last year.

I'll remove it and do some testing.

thc202: I did remove the statements and didn't see any change, that's why was asking.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
thc202: You don't need to change/customise the sorter, instead of returning a String you could return a wrapper of the status (using a static nested class, in UninstalledAddOnsTableModel, which would implement Comparable using the add-on Status and the toString() return the internationalised name; getColumnClass(int) would have to be changed to reflect the new class returned though).

kingthorin: Isn't this what you're talking about? https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java#L46-L56
Then for getColumnClass I can just insert a check for column index 1 and return int.class?
https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java#L81

kingthorin:
What class are you talking about with lines 427/936?

kingthorin:
I think I figure it out....fingers crossed :)

kingthorin:
Looking better now?

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
thc202: Actually, the comparator that you highlighted could be removed, the elements no longer need to be sorted in the model(s) (they are now sorted by the view :)

I was suggesting create and return a custom static nested class, for example, `StatusUI` with implementation:
```Java
    private static class StatusUI implements Comparable<StatusUI> {

        private final AddOn.Status status;
        private final String stringRepresentation;

        public StatusUI(AddOn.Status status, String stringRepresentation) {
            this.status = status;
            this.stringRepresentation = stringRepresentation;
        }

        @Override
        public int compareTo(StatusUI o) {
            if (o == null) {
                return 1;
            }
            return status.compareTo(o.status);
        }

        @Override
        public String toString() {
            return stringRepresentation;
        }
    }
```
(caching the StatusUI instances for the statuses available and return as needed in the method getValueAt(int, int))

But, this class should be a top level class, there are other UI components that need it (that is, have the same sorting problem, for example, the table in "Passive Scan Rules" options panel).

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -912,8 +929,9 @@ private JButton getAddOnInfoButton() {
 			addOnInfoButton.addActionListener(new java.awt.event.ActionListener() { 
 				@Override
 				public void actionPerformed(java.awt.event.ActionEvent e) {
-					if (getUninstalledAddOnsTable ().getSelectedRow() >= 0) {
-						AddOnWrapper ao = uninstalledAddOnsModel.getAddOnWrapper(getUninstalledAddOnsTable ().getSelectedRow());
+					//convertRowIndexToModel in-case they sorted
+					if (getUninstalledAddOnsTable().convertRowIndexToModel(getUninstalledAddOnsTable().getSelectedRow()) >= 0) {
```
thc202: This statement does not need the conversion, it's just checking that there's at least one row selected.

thc202:
Added some comments.

kingthorin:
Fixed

thc202:
There's another "hidden" comment: https://github.com/zaproxy/zaproxy/pull/2162#discussion_r52984511

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
kingthorin: K, I'm missing something here. Is it nested or top level?

thc202: It would be better if it was added as top level class, it would be half the work done for the other dialogues.
But static nested class would be enough just for this change. Pick one ;)

kingthorin:
Just so you know I haven't forgotten about this, it's still on my radar. Just gotta make time to get back to it.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
kingthorin: Ok I'm finally back to this. I'm left with a ton of questions:

1.  You mentioned dropping the comparator in the model but it seems to be required by a  call to the super constructor at [L62](https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java#L62). Would we drop some part of StatusUI in this place or is there a way to avoid the requirement?
2. Where do you see StatusUI fitting in? I'm thinking either: org.zaproxy.zap.utils or org.zaproxy.zap.view.
3. You mention caching the StatusUI info. What do you mean by "caching".
4. You mention returning as part of getValueAt(int, int). Were you intending that info to replace the [current cfu.status and wrapper/status info](https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java#L98)?
5. In the implementation you handed me....thanks by the way! :thumbsup: Where does stringRepresentation ever get set? Would it be passed in to the constructor based on a reference to an addon or addon wrapper?

kingthorin: Ok I decided to try moving ahead with this anyway. For now I've stuck it in the view location mentioned in #1 above. For now I've got it returning a new StatusUI at getValueAt (#4 above) based on passing the constructor the status and status.toString() from the AddOn, which I doubt was your intent/vision but I'm torn as to where it should be cached/instanced? I've also changed the return in getColumnClass.

Seems to be operating as expected.

Is there a requirement for the enum in AddOn to use lowercase letters, in the table it looks kind of silly to not have Title Caps (IMHO).

thc202: Answering the previous questions:
 1. As the models no longer need to be sorted the base class no longer needs to require a comparator but, to keep the base class binary compatible we should add a new constructor and deprecate the old one (of course addAddOnWrapper(AddOn, AddOnWrapper.Status) needs to be updated to handle a null comparator);
 1. The package org.zaproxy.zap.view seems fine to me;
 1. Create just one instance per each Status and hand those to the callers; I guess we can add that to the View class, "StatusUI getStatusUI(AddOn.Status)"?
 1. Yes, the StatusUI would be returned in the case 1 of the switch statement, we still need to use the wrapper status to obtain the corresponding StatusUI though.

The constructor's parameter "stringRepresentation" should be the internationalised name of the status, the `Constant.messages.getString("cfu.status." + ...`.

Let me know if I missed answering something.

thc202: BTW, the constructor of StatusUI should be updated to validate that the Status and stringRepresentation are not null.

kingthorin: My simple change is failing to be simple :disappointed:  I'm starting to feel like I'm in over my head....

RE: 3, which view are you talking about? The UninstalledAddOns view, which is actually the ManageAddOnsDIalog? Am I looping through the enum to get all the possible statuses? How can I getStatusUI based solely on AddOn.Status if I still want an internationalized stringRepresentation?

null handling has been added to the constructor, it defaults to use of AddOn.Status.unknown if necessary.

thc202: But it's improving the code base, so that's not a bad thing, right?

> which view are you talking about?

I was talking about the View class. [1]
Maybe there's a better place for that?

> Am I looping through the AddOn.Status enum to get all the possible statuses? 

Yeah, that's fine (but we should verify that the internationalised name for the Status(es) exist, otherwise it might start failing if a new Status is added without the corresponding internationalised name).

> How can I getStatusUI based solely on AddOn.Status where am I constructing it so that the internationalized stringRepresentation has been passed? 

I was thinking on the View class to hold a `Map<AddOn.Status, StatusUI>`, so the method "getStatusUI(AddOn.Status)" would just do a get on the map and return the StatusUI (which was previously constructed).


> null handling has been added to the constructor, it defaults to use of AddOn.Status.unknown if necessary.

OK, how about the string representation? Which value are you setting in that case?


[1] https://github.com/zaproxy/zaproxy/blob/61dab8de452dbe445e956484c085d4ec5384cdd0/src/org/parosproxy/paros/view/View.java

kingthorin: Hahahaha, yes I'm good with improving things :smile:

Thanks for the answers.

As for the null handling it uses AddOn.Status.unknown.toString() for the stringRepresentation, which isn't perfect in internationalized cases but it's handling for the minority of cases when something goes wrong so I don't think it's problem.

thc202: You're welcome :)

Yeah, that's a fallback case so it's fine that way.

kingthorin: Does View have access to the i18n strings or is that too early in the startup?

thc202: It does, there are already some methods using it.
(You could also lazy initialise the map.)

kingthorin: Thanks, I actually figured out what the problem was, working through more of this now.

kingthorin:
Don't worry about reviewing 14505e3 there's more coming.

kingthorin:
Ok I think I've addressed everything we've discussed. Let me know what you think of 78577af.

thc202:
Sure, I'll take a look at it tomorrow morning.

src/org/parosproxy/paros/view/View.java
```diff
@@ -168,6 +175,8 @@
     // ZAP: splash screen
     private SplashScreen splashScreen = null;
 
+    private static Map<AddOn.Status, StatusUI> statusMap = new HashMap<AddOn.Status, StatusUI>();
```
thc202: Consider adding the map as an instance variable instead, it would "force" the use of the getSingleton() method which ensures that the map is initialised with the StatusUIs.

src/org/parosproxy/paros/view/View.java
```diff
@@ -220,6 +229,17 @@ public void init() {
 
         // adds the Request/Response representation buttons in ZAP toolbar
         getMessagePanelsPositionController().restoreState();
+        
+        String statusString = "";
```
thc202: Consider removing the initialisation, it's not necessary.

src/org/parosproxy/paros/view/View.java
```diff
@@ -220,6 +229,17 @@ public void init() {
...
         // adds the Request/Response representation buttons in ZAP toolbar
         getMessagePanelsPositionController().restoreState();
+        
+        String statusString = "";
+        for(Status status : AddOn.Status.values()) {     	
+        	//Try/catch in case AddOn.Status gets out of sync with cfu.status i18n entries
+        	try {
+        		statusString = Constant.messages.getString("cfu.status." + status.toString());
+        	} catch (MissingResourceException mre) {
+        		statusString = status.toString();
```
thc202: Consider logging the exception as error, we really don't want to ignore/miss this kind of issues.

src/org/parosproxy/paros/view/View.java
```diff
@@ -877,4 +897,8 @@ public void removeMainToolbarSeparator(JToolBar.Separator separator) {
     public Component getSplashScreen() {
         return splashScreen;
     }
+    
+    public static StatusUI getStatusUI(AddOn.Status status) {
```
thc202: Consider adding JavaDoc, including the "since" tag.

src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
```diff
@@ -53,7 +53,7 @@
 
     protected final Logger logger = Logger.getLogger(this.getClass());
 
-    private final Comparator<AddOnWrapper> comparator;
+    private Comparator<AddOnWrapper> comparator;
```
thc202: This can be kept final, you just need to initialise it to null in the new constructor.

src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
```diff
@@ -63,6 +63,9 @@
 
     private AddOnSearcher addOnSeacher;
 
+	/**
+	 * @deprecated (TODO add version) Replaced by {@link #AddOnsTableModel(AddOnCollection, int)}. It will be removed in a future release.
+	 */
     public AddOnsTableModel(Comparator<AddOnWrapper> comparator, AddOnCollection addOnCollection, int progressColumn) {
```
thc202: Consider adding the Deprecated annotation to the method.

src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
```diff
@@ -110,9 +122,11 @@ protected void addAddOnWrapper(AddOn addOn, AddOnWrapper.Status status) {
         AddOnWrapper aow = createAddOnWrapper(addOn, status);
         int idx = 0;
         for (; idx < getAddOnWrappers().size(); idx++) {
-            if (comparator.compare(aow, getAddOnWrappers().get(idx)) < 0) {
-                break;
-            }
+        	if (comparator != null) {
```
thc202: Consider moving this if statement to outside the loop, no need to loop if there's no comparator (and in that case the idx is just the last position, i.e. getAddOnWrappers().size().)

src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java
```diff
@@ -79,9 +68,12 @@ public int getRowCount() {
 
     @Override
     public Class<?> getColumnClass(int columnIndex) {
-        if (columnIndex == 0) {
+        if (columnIndex == 0) {//Icon
             return Icon.class;
-        } else if (columnIndex == 5) {
+        } else if (columnIndex == 1) {//Status (Quality)
+//        	return int.class;
```
thc202: The commented code should be removed.

src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java
```diff
@@ -95,7 +87,8 @@ public Object getValueAt(int rowIndex, int columnIndex) {
         case 0:
             return Boolean.valueOf(getAddOnWrapper(rowIndex).hasRunningIssues());
         case 1:
-            return Constant.messages.getString("cfu.status." + getAddOnWrapper(rowIndex).getAddOn().getStatus().toString());
+//            return Constant.messages.getString("cfu.status." + getAddOnWrapper(rowIndex).getAddOn().getStatus().toString());
```
thc202: Same here.

src/org/zaproxy/zap/extension/autoupdate/UninstalledAddOnsTableModel.java
```diff
@@ -152,7 +145,7 @@ public boolean canIinstallSelected() {
     public void addAddOn(AddOn addOn) {
         addAddOnWrapper(addOn, null);
     }
-
+
```
thc202: Consider reverting to previous indentation.

src/org/zaproxy/zap/view/StatusUI.java
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.view;
+
+import org.zaproxy.zap.control.AddOn;
+
+public class StatusUI implements Comparable<StatusUI> {
```
thc202: Consider adding the since version JavaDoc tag and a little doc ;)

src/org/parosproxy/paros/view/View.java
kingthorin: Warn level since it's doesn't actually result in a failure?

thc202: I suggested error level because of the extra visibility when in dev mode, it's shown in the output tab and the bug count incremented.
We could also use error in dev mode and warn otherwise.

kingthorin: Ok, thanks.

src/org/parosproxy/paros/view/View.java
kingthorin: Should I suppress the warning that comes from removing the type arguments?

thc202: I mean remove just the type arguments (i.e. using diamond operator). There shouldn't be a warning because of that.

kingthorin: Yup I know.

If I remove from the right I get a type safety warning.
If I also remove from the left I get a raw type warning and then in getStatusUI() I end-up with an error can't convert Object to StatusUI.

thc202: Not remove everything, it should be kept the diamond:
`Map<AddOn.Status, StatusUI> statusMap = new HashMap<>();`

the compiler infers the types for us (in this case.)

kingthorin: Thanks, done.

thc202: Would you mind changing the statusMap to instance variable and the method getStatusUI(...) to instance method? I do think that's better in the long run.

src/org/parosproxy/paros/view/View.java
```diff
@@ -220,6 +229,25 @@ public void init() {
...
+        String statusString;
+        for(Status status : AddOn.Status.values()) {     	
+        	//Try/catch in case AddOn.Status gets out of sync with cfu.status i18n entries
+        	try {
+        		statusString = Constant.messages.getString("cfu.status." + status.toString());
+        	} catch (MissingResourceException mre) {
+        		statusString = status.toString();
+        		
+        		String errString="Caught " + mre.getClass().getName() + " " + mre.getMessage() + 
+						" when looking for i18n string: cfu.status." + status.toString();
```
thc202: It could use statusString instead.

src/org/zaproxy/zap/extension/autoupdate/AddOnsTableModel.java
```diff
@@ -109,10 +123,12 @@ protected AddOn getAddOnForDownload(AddOnWrapper aow) {
...
-            }
+    	if (comparator != null) {
+    		for (; idx < getAddOnWrappers().size(); idx++) {
+        		if (comparator.compare(aow, getAddOnWrappers().get(idx)) < 0) {
+        			break;
+        		}
+        	}
         }
 
         getAddOnWrappers().add(idx, aow);
```
thc202: This statement should be moved to the above if-then block, and in the (new) else block simply add the wrapper:
```Java
idx = getAddOnWrappers().size();
getAddOnWrappers().add(aow);
```
it avoids copying arrays around when there's no comparator, which should be the case from now on.

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -416,8 +421,9 @@ public void tableChanged(TableModelEvent e) {
 				@Override
 				public void valueChanged(ListSelectionEvent e) {
 					getAddOnInfoButton().setEnabled(false);
-					if (DesktopUtils.canOpenUrlInBrowser() && getUninstalledAddOnsTable ().getSelectedRowCount() == 1) {
-						AddOnWrapper ao = uninstalledAddOnsModel.getAddOnWrapper(getUninstalledAddOnsTable ().getSelectedRow());
+					if (DesktopUtils.canOpenUrlInBrowser() && getUninstalledAddOnsTable().getSelectedRowCount() == 1) {
```
thc202: A space was removed :P

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
```diff
@@ -912,8 +929,9 @@ private JButton getAddOnInfoButton() {
 			addOnInfoButton.addActionListener(new java.awt.event.ActionListener() { 
 				@Override
 				public void actionPerformed(java.awt.event.ActionEvent e) {
-					if (getUninstalledAddOnsTable ().getSelectedRow() >= 0) {
-						AddOnWrapper ao = uninstalledAddOnsModel.getAddOnWrapper(getUninstalledAddOnsTable ().getSelectedRow());
+					//convertRowIndexToModel in-case they sorted
```
thc202: Should this code comment be moved to inside the if-then block?

src/org/zaproxy/zap/extension/autoupdate/ManageAddOnsDialog.java
kingthorin: Uh oh, my OCD crept in....

thc202: ;)

src/org/parosproxy/paros/view/View.java
kingthorin: No problem. Done.

thc202:
Looks good to me. Thank you.

kingthorin:
Rebased, conflict resolved.

Issue #2168, opened 2016-01-12
Issues with ZAP PHP API client


fulgore83:
Hello,

when I try use php API to call scan i got error:

GET http://zap/JSON/spider/action/scan/?zapapiformat=JSON&apikey=12345&url=http%3A%2F%2Fexample.com&maxChildren=&recurse=&contextName= HTTP/1.1

HTTP/1.1 400 Bad Request
Pragma: no-cache
Cache-Control: no-cache
Access-Control-Allow-Methods: GET,POST,OPTIONS
Access-Control-Allow-Headers: ZAP-Header
X-Clacks-Overhead: GNU Terry Pratchett
Content-Length: 50
Content-Type: application/json; charset=UTF-8

{"code":"does_not_exist","message":"Nie istnieje"}

Can You help me?

Best regards
Grzegorz

thc202:
The problem is caused by the empty "contextName" (#2122).

What version of the PHP API are you using?

fulgore83:
I use in composer on files from repo which downloaded today:

"zaproxy/php-owasp-zap-v2": "=dev-develop"

This problem is occure when i use connection to new ZAP 2.4.3. On Zap 2.4.1 is fine.

fulgore83:
and my additional question:

when i use Yours requires with composer:

"zaproxy/php-owasp-zap-v2": ">=0.1.0",

i got :+1: 
Problem 1
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.1, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.2, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.3, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.4, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.5, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.6, dev-develop].
    - Can only install one of: zaproxy/php-owasp-zap-v2[0.1.7, dev-develop].
    - Installation request for zaproxy/php-owasp-zap-v2 dev-develop -> satisfiable by zaproxy/php-owasp-zap-v2[dev-develop].
[...]

Issue #2108, opened 2015-12-07
ZAP API clients should not send optional parameters unless specified


continuumsecurity:
In 2.4.3:
Have to specify a context name when calling spider.scan from the API, if no context is entered then ZAP responds with "Does not exist".

To workaround this, I entered a context of "Default Context"
1. Enter context of "Default Context",
2. enter the API key 
4. enter true for recurse
3. and the URL: http://localhost:/9110/ropeytasks

The API request looks like this:
http://zap/JSON/spider/action/scan/?zapapiformat=JSON&apikey=apisecret&url=http%3A%2F%2Flocalhost%3A9110%2Fropeytasks&maxChildren=&recurse=&contextName=Default+Context

When submitting this request, the spider does not perform any action, it just immediately changes to FINISHED.  Reading the results from the spider with spider.results returns an empty array:
{"results":[]}

Performing these exact same actions in 2.4.0 worked (which did not request a context name), and the spider ran and produced results.

In case this had to do with recurse and maxchildren, I set both to 10 and resubmitted the request, with the same empty results list.

continuumsecurity:
False positive, using the wrong URL.

thc202:
Thanks for letting us know.

thc202:
This is an issue, the context name is not mandatory but the API client sends always something even if the context name is not provided (i.e. null), which ends up being used and validated, when the API request is processed in ZAP.
The API client needs to be changed to not send the optional parameters if not specified, the Spider API should be changed as well to ignore empty context names, since most of the API clients send the parameters with empty values.

thc202:
A new version (9) of the Java ZAP API Client has been released to address this issue.

LisinIwan:
In 2017-10-02 (Weekly-Release) i faced the same Problem:
Always getting the ERROR-message that "Default Context" does not exist. It took me nearly 1 week to figure out what actually caused the error. 
First of all i tried to modify the Code, the most simple solution was just to comment out the try-catch-block inside the  includeInContext-Method, in the ZapHelper class. 
But i geuss it is not satisfying to modify a plugin yourself, so i was looking further into the  issue and figured out that actually the language Setup of ur PC caused the error. 
In my case (Germany) i could also set the 
private static final String ZAP_DEFAULT_CONTEXT_NAME = "Default Context"; to 
[...]

PR #2170, opened 2016-01-12
Fix typo in PhpAPIGenerator, to use a property defined in class Zapv2

thc202:
Correct the name of a Zapv2 property ("base_other") used by the
generated API classes when calling OTHER endpoints.
Regenerate the PHP API client to fix the typo (and to not send the
parameters that were not specified, #2108).
Reported in #2168 - Problem with API on new ZAP 2.4.3


thc202:
Closes #2168.

PR #2184, opened 2016-01-18
Create default policy in "weekly" and "stable" Docker images

thc202:
When initialising ZAP does not create the "Default Policy" if at least
one policy was already created, the case with the docker images which
already have custom policies (e.g. St-High-Th-Med.policy,
St-Low-Th-Med.policy...).
The default policy is needed/used for ZAP scan tests.


thc202:
Issue reported in OWASP ZAP Developer Group: https://groups.google.com/d/msg/zaproxy-develop/kpxnenhuoJA/HYFfCyoJDAAJ

PR #2221, opened 2016-02-03
Skip "Script Active Scan Rules" scanner if there are no scripts enabled

thc202:
Change class ScriptsActiveScanner to skip the scan if the
ExtensionScript is not enabled, there are no (active) scripts or if
there are none is enabled. Remove, now, redundant if statements that
check if the ExtensionScript is enabled, the scanner is no longer run if
that's not the case.


thc202:
Based on zaproxy/zap-extensions#208 ;)

Issue #382, opened 2015-06-04
Add option to zap.sh to allow the memory settings to be overridden


zapbot:
```
Getting the memory settings right for any java app can be tricky, and especially for
apps like ZAP which can be used in very different circumstances.
We try to guestimate the optimum memory settings, but to override those you currently
have to manually edit zap.sh (on non Windows platforms)
Be useful to have command line options to allow a user to override the defaults without
having to edit this file.
```

Original issue reported on code.google.com by `psiinon` on 2012-09-24 08:13:11

thc202:
Closing as duplicate of #2237 (see also #1795, for GUI).

Issue #1795, opened 2015-08-17
Allow JVM options to be configured via GUI


psiinon:
As per https://groups.google.com/d/msg/zaproxy-develop/9iMqEgzG1nE/xzJTrE0mBwAJ

PR #2237, opened 2016-02-15
Allow java max mem to be overridden on the cmdline

psiinon:


kingthorin:
Is this meant to address https://github.com/zaproxy/zaproxy/issues/382? If so should it include a similar mod for zap.bat?

On it's own this change looks OK to me.

thc202:
Works fine here.

Though this addresses #382, I'd close #382 as duplicate of #1795 (which applies to zap.sh and zap.bat).

psiinon:
Is it possible to do something similar in zap.bat? Not with my limited knowledge ;)
Not sure how many people actually use the bat file, I'd expect most people to use the exe. And the bat file doesnt have any equivalent code for determining the memory available.
This change helps with the docker images, as you can then set the memory without having to create a new file on the docker images.

kingthorin:
Ok, I'm good with that reasoning.

PR #2238, opened 2016-02-15
spider: fallback to parse HTML comments as plain text if no URL found

thc202:
Change class SpiderHtmlParser to try to find URLs in HTML comments if
the comment, parsed as HTML, does not have any, allowing to find URLs
in "plain" comments.
With the change it finds URLs with schemes HTTP and HTTPS and with no
explicit scheme ("//example.com/") all with (possibly) path and query
components. For example, WIVET test case 8_1b6e1
("http://example.com/innerpages/8_1b6e1.php") would be found while test
case 8_2b6f1 ("/innerpages/8_2b6f1.php") would not. The latter is not
included as it might lead to too many false positives.
Add tests to assert the expected behaviour of SpiderHtmlParser.
Other minor changes were done to "normalise" the behaviour/expectations
of class SpiderHtmlParser:
 it's required to check if HTML comments should be parsed;
 - Expect always a message when parsing (letting a NullPointerException
 be thrown if not);
 - Require the HTML base element to have a non empty value to be used;
 - Document in canParseResource(...) that a message is expected (that
 is, a NullPointerException is thrown if not provided).

### DIFF
--- a/src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
+++ b/src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
@@ -42,6 +42,8 @@ public class SpiderHtmlParser extends SpiderParser {
 
 	/** The Constant urlPattern defining the pattern for a meta url. */
 	private static final Pattern urlPattern = Pattern.compile("url\\s*=\\s*([^;]+)", Pattern.CASE_INSENSITIVE);
+	private static final Pattern PLAIN_COMMENTS_URL_PATTERN = Pattern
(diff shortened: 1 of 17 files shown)


kingthorin:
Looks good to me. Regex101 wasn't able to figure out the Regex so i'd have to trust on that detail.

Not specific to this commit, but to this code, why don't we include embed src? Might be missing some newer html 5 tags too.

thc202:
You need to remove the Java string escapes, the following should work:
`(?:http(?:s?):)?\/\/[^\x00-\x1f\"'\s<>#]+`

ZAP might not like that?? But, yeah, worth taking a look at that.

kingthorin:
Tks, looks good.

psiinon:
Looks good to me too :)

PR #2245, opened 2016-02-17
Convert options to not use extensions' names as XML element names

thc202:
Prevent exceptions when the name of the extensions are not valid XML
element names.
Add class ExtensionParam that takes care to save/load the configurations
instead of doing that in the class OptionsExtensionPanel (now using the
former).
Change Constant class to convert to the new options when updating from
version 2.4.3.
Change class ExtensionExtension to install/register ExtensionParam.
The configurations now use specific elements, "name" and "enabled", to
save the name and enabled state of the extension, each one in its own
"extension" element (which are all under an "extensions" element).


kingthorin:
Looks good.

PR #2248, opened 2016-02-18
Change "sessions" API view to not use tokens' names as XML element names

### LINKED ISSUE #2246: Error in sessions view with tokens
The sessions view throws an exception, when trying to show a session which contains a token that starts with a number. This only happens for the XML output format, JSON and HTML work properly.

### Exception:
ERROR API - INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified.
org.w3c.dom.DOMException: INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified. 
	at com.sun.org.apache.xerces.internal.dom.CoreDocumentImpl.createElement(CoreDocumentImpl.java:622)
	at org.zaproxy.zap.extension.api.ApiResponseSet.toXML(Unknown Source)
	at org.zaproxy.zap.extension.api.ApiResponseList.toXML(Unknown Source)
	at org.zaproxy.zap.extension.api.ApiResponseList.toXML(Unknown Source)
	at org.zaproxy.zap.extension.api.API.responseToXml(Unknown Source)
	at org.zaproxy.zap.extension.api.API.handleApiRequest(Unknown Source)
	at org.parosproxy.paros.core.proxy.ProxyThread.processHttp(Unknown Source)
	at org.parosproxy.paros.core.proxy.ProxyThread.run(Unknown Source)

### Steps to reproduce:
1. Start ZAP and a browser with ZAP's proxy settings (e.g. localhost, port: 8090)
2. Open a page in the browser, which contains a session token starting with a number
	* Open a web page of your choice (e.g. http://example.com)
	* [Add a session token](http://zap/UI/httpSessions/action/addSessionToken) that starts with a number (e.g. 1234567890)
	* [Set the session token value](http://zap/UI/httpSessions/action/setSessionTokenValue)
3. Open the [sessions view](http://zap/UI/httpSessions/view/sessions) and select XML as output format

==> The error occurs, see log file [zap.zip](https://github.com/zaproxy/zaproxy/files/135853/zap.zip).

### Possible solutions:
1. Don't use the token names as element names.
2. Add a prefix to the element names, to prevent that they start with a number, e.g. token_{tokenName}.

thc202:
The "sessions" API view will now return as tokens of a session:
```XML
 <tokens type="set">
   <token type="set">
     <name>SESSID</name>
     <value>123</value>
     <domain>example.com</domain>
     <path>/</path>
     <secure>false</secure>
   </token>
 </tokens>
```
instead of:
```XML
 <tokens type="set">
   <SESSID>SESSID=123</SESSID>
 <tokens>
```
preventing exceptions when the name of the session tokens are not valid
XML element names and normalising its structure (no longer dependent on
tokens' names).
Change class HttpSessionsAPI to use a custom ApiResponseSet, to produce
the newer XML structure (class TokenValuesResponseSet), also extract a
method that creates a response with session's data, to reduce code
duplication.
Fix #2246 - Error in sessions view with tokens

### DIFF
--- a/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
+++ b/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
@@ -1,6 +1,10 @@
 package org.zaproxy.zap.extension.httpsessions;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 
 import net.sf.json.JSONObject;
@@ -8,6 +12,8 @@ import net.sf.json.JSONObject;
 import org.apache.commons.httpclient.Cookie;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
 import org.zaproxy.zap.extension.api.ApiAction;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiImplementor;
@@ -16,6 +22,8 @@ import org.zaproxy.zap.extension.api.ApiResponseElement;
 import org.zaproxy.zap.extension.api.ApiResponseList;
 import org.zaproxy.zap.extension.api.ApiResponseSet;
 import org.zaproxy.zap.extension.api.ApiView;
+import org.zaproxy.zap.utils.Pair;
+import org.zaproxy.zap.utils.XMLStringUtil;
 
 /**
  * The Class HttpSessionsAPI.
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
```diff
@@ -338,4 +344,45 @@ static String getAuthority (String site) {
...
+			List<List<Pair<String, String>>> tokens = new ArrayList<>();
+			for (Entry<String, Cookie> token : values.entrySet()) {
+				Cookie cookie = token.getValue();
+				List<Pair<String, String>> fields = new ArrayList<>();
+				fields.add(new Pair<>("name", token.getKey()));
+				fields.add(new Pair<>("value", cookie.getValue()));
+				fields.add(new Pair<>("domain", cookie.getDomain()));
+				fields.add(new Pair<>("path", cookie.getPath()));
+				fields.add(new Pair<>("secure", Boolean.toString(cookie.getSecure())));
+				tokens.add(fields);
```
kingthorin: Name, value, domain, path, secure. What about httponly?

thc202: This cookie class does not support that attribute.

kingthorin: Fair enough :)

kingthorin:
Looks good to me.

PR #2254, opened 2016-02-19
Revert HSQLDB to version 2.2.9 because of issues with newer version

thc202:
With the newer version, 2.3.3, sometimes the persistence of history
records and alerts fails (e.g. from proxying and passive scanning) as
can be seen in logs of zapbot's WIVET scans, following stack trace
excerpt:
[...]
Caused by: org.hsqldb.HsqlException: transaction rollback: serialization
failure
 at org.hsqldb.error.Error.error(Unknown Source)
 at org.hsqldb.error.Error.error(Unknown Source)
 at org.hsqldb.Session.executeCompiledStatement(Unknown Source)
 at org.hsqldb.Session.execute(Unknown Source)
 ... 9 more
[...]

https://github.com/zapbot/zap-mgmt-scripts/blob/3b7de55a075d05aefe9343d3aeeab381c9434cbc/reports/wivet-weekly.logs.txt#L154-L243


kingthorin:
Looks ok to me.

PR #2255, opened 2016-02-19
Change "params" API view to filter the results with the given site

thc202:
Change class ParamsAPI to use the optional "site" parameter to filter
the results returned in the API response. Extract a method that creates
the API response to reduce code duplication, it's used for "all results"
and the filtered ones.
Change class ExtensionParams to allow to check if a site already exists
and if it has parameters, to allow the API to return an error if it
doesn't and to not create the (empty) response if it does not have any
parameters (also, change class SiteParameters to allow to check if it
has or not parameters).
Move the method HttpSessionsApi.getAuthority(String) to class ApiUtils,
it's now used by several classes (HttpSessionsApi and ParamsAPI) and
change it to return default port for HTTP scheme to not require the
callers to check and add the port. Also, move the tests to a new class,
ApitUtilsUnitTest, and update them to conform to new behaviour.

---

### DIFF
--- a/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
+++ b/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
@@ -2,7 +2,6 @@ package org.zaproxy.zap.extension.httpsessions;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


kingthorin:
Looks ok to me.

One question though (kind of generic), with things like: https://github.com/thc202/zaproxy/blob/48fce442bff789751d5f0296faab3046f36941bc/src/org/zaproxy/zap/extension/params/SiteParameters.java#L63 how to other input vectors fit in (i.e.: JSON, GWT, AMF, etc) are they mapped to one of the types referred to here?

thc202:
Those would be mapped to "form" (post? ;) parameters (but currently that's just for "x-www-form-urlencoded" parameters).

PR #2259, opened 2016-02-22
Fix NullPointerException in VariantCookie implementation

thc202:
Change class VariantCookie to conform to the expected behaviour when the
parameter's (that is, the cookie) name and value are set to null, which
indicates the parameter should be removed, preventing
NullPointerExceptions, for example:
>ERROR org.zaproxy.zap.extension.ascanrules.TestParameterTamper -
> Error occurred while scanning a message:
>[...]
>java.lang.NullPointerException
> at java.net.HttpCookie.<init>(HttpCookie.java:143)
> at java.net.HttpCookie.<init>(HttpCookie.java:139)
> at VariantCookie.setParameter(VariantCookie.java:125)
> at VariantCookie.setParameter(VariantCookie.java:92)
> at AbstractAppParamPlugin.setParameter(AbstractAppParamPlugin.java:259)
> at TestParameterTamper.scan(TestParameterTamper.java:129)
>[...]

Also, change the class to:
 - Manually parse and create the cookie header to correctly handle null
 names and values (which would lead to similar exceptions if used with
 HttpCookie class);
 - Remove (now) redundant null checks from methods getEscapedValue and
 getUnescapedValue;
 - Update/add JavaDoc to class and methods.

Change class NameValuePair to add the methods hashCode(), equals(Object)
for proper use in collections and comparisons, and method toString() for
help with debug and remove redundant instance variable initialisations.
Add tests to assert the expected behaviour of classes VariantCookie and
NameValuePair.

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/NameValuePair.java
+++ b/src/org/parosproxy/paros/core/scanner/NameValuePair.java
@@ -21,7 +21,7 @@
 // ZAP: 2013/07/02 Changed API to public because future extensible Variant model
 // ZAP: 2014/01/06 Issue 965: Support 'single page' apps and 'non standard' parameter separators
 // ZAP: 2014/02/08 Used the same constants used in ScanParam Target settings
-//
+// ZAP: 2016/02/22 Add hashCode, equals and toString methods. Remove redundant instance variable initialisations.
 package org.parosproxy.paros.core.scanner;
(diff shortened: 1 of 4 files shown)


psiinon:
Looks good to me

kingthorin:
Thanks

PR #2261, opened 2016-02-23
Making SimpleExample work

volkovs:
- changing optional parameter values from empty string to null (so they are not added to the map)
- successful response is "1" instead of "OK"

### DIFF
--- a/src/org/zaproxy/clientapi/core/SimpleExample.java
+++ b/src/org/zaproxy/clientapi/core/SimpleExample.java
@@ -41,7 +41,7 @@ public class SimpleExample {
 			
 			// Start spidering the target
 			System.out.println("Spider : " + target);
-			ApiResponse resp = api.spider.scan(apikey, target, "", "", "");
+			ApiResponse resp = api.spider.scan(apikey, target, null, null, null);
 			String scanid;
 			int progress;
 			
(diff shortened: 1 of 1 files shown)


psiinon:
Should maybe check for not getting an error instead of "1" for OK - the second time you start a scan you'll get 2, then 3 etc
Its probably ok for this case but examples should show best practice, and this wont work if you make multiple ascan calls.
Although it works better than now of course;)

volkovs:
@psiinon Thanks for noticing it (my understanding of response value was some kind of constant error code :) )

I have removed response check at all (like there is no check in spider scan in the same example), since ApiResponseFactory just throws ClientApiException instead of returning response with according value.

Once proper response is returned I will return response checking back. It will be possible to encapsulate isOk() / hasError() right into response class. What do you think?

thc202:
Probably better to extract the scan ID from the response and use it in the status check (it's more reliable to use scan's ID).

kingthorin:
@volkovs is there an update coming?

volkovs:
@kingthorin thanks for reminding, will send in the evening

volkovs:
@thc202 done

kingthorin:
Looks ok to me

thc202:
Code changes look good to me but the commit message says "however it is not taken into account by ZAP server", isn't it returning the status of the corresponding scan?

volkovs:
@thc202 it does return the status. But scanid is not taken into account in line:
`progress = Integer.parseInt(((ApiResponseElement)api.ascan.status(scanid)).getValue());`
It was just hardcoded empty string before...

thc202:
Right, but an empty string (i.e. no scan ID) means return the status of the last started scan. Otherwise, if you specify an ID it will/should return the status of that scan.

volkovs:
@thc202 right, server tries to parse integer from scanId parameter and implicitly falls back to last scan - the same logic as in spider. So, my comment in the commit message is wrong.

IMO if API accepts (not existing) scanId it should either return null or throw exception, instead of silently falling back to some other scanId. That would make sense for me. But might be API backward compatibility issues...

thc202:
If the provided scan ID does not exist the API throws an error. Only when the scan ID is not provided that the API returns the status of the last scan, which is more or less to be compatible with old behaviour.

Will you amend the commit message?

volkovs:
@thc202 do you mean to correct just commit message? Or you would like me to correct the behaviour? May be in another pull request?

Anyway, will update the PR in a day or two.

thc202:
Yes, just the commit message.

Is there any behaviour to correct/improve? You mean how the scan ID is used?

volkovs:
thc202 Yeah, I mean if you pass any string (not parseable to integer) as scan ID it still falls back silently to last scan ID. Not critical of course and might be never an issue in real life.

Amended and rebased.

kingthorin:
Looks good

thc202:
@volkovs Right, worth fixing that case anyway (in another pull request, though).

Pull request looks good to me, thanks!

PR #2262, opened 2016-02-23
 spider: use the broadest target component for scan's display name

thc202:
Change spider to create the display name for the scans using all the
components of the target and custom configurations, with the following
order:
 - Context;
 - Just in scope;
 - Start node;
 - Start URL.

So, for example, if a start node is specified along with a context it
will display the context name instead of the start node (which is more
specific, narrow target when spidering).
Change class ExtensionSpider to follow the above rules using a new
method to start the scans, to ensure that all scans use the same display
name rules (the old method that allows to define the display name is
still available but its use is discouraged in favour of using the new
method).
Change classes ExtensionSpider, SpiderAPI and SpiderDialog to use the
new method, for the latter class also remove the creation of the custom
display name.

---
Old title/message (for the record):
Change Target display name to use the broadest target component first

Change the method Target.getDisplayName() to check first if there's a
context defined or if the target is "in scope only" to use that as the
display name, only if those are not defined use the start node. So, if a
start node is specified along with a context it will display the context
name instead of the start node (which is more specific, narrow target).
(Also, remove unnecessary Object array instantiation.)

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
+++ b/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
@@ -36,6 +36,7 @@ import java.util.List;
 import javax.swing.KeyStroke;
 
 import org.apache.commons.httpclient.URI;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
@@ -309,7 +310,7 @@ public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedL
 	public void startScanNode(SiteNode node) {
 		Target target = new Target(node);
 		target.setRecurse(true);
-		this.startScan(target.getDisplayName(), target, null, null);
+		this.startScan(target, null, null);
 	}
 	
 	/**
@@ -320,7 +321,7 @@ public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedL
 	public void startScanNode(SiteNode node, User user) {
 		Target target = new Target(node);
 		target.setRecurse(true);
-		this.startScan(target.getDisplayName(), target, user, null);
+		this.startScan(target, user, null);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


kingthorin:
Looks good.

psiinon:
Is this the most useful behaviour?
Say I'm scanning a load of individual pages, such as http://localhost:8080/bodgeit/login.jsp which are all in the BodgeIt context.
Wont this mean that in the Active Scan tab they are all labelled as "BodgetIt"? I'd have thought that more specific names would be more useful (in this case).

thc202:
It is for the spider (it is, right?), which does not have the concept of scanning/spidering just a page. I didn't think on the active scanner :/
Will change the code/pull request to just affect the spider.

psiinon:
+1
On a related note, maybe we should be able to optionally constrain the spider to a subtree?
This would actually be v useful, and in which case the more specific name would again be more useful.

thc202:
Right, raised an issue (#2274).
Yeah, more tweaks for the display name then :)

thc202:
Pull request and commit updated.

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
```diff
@@ -426,14 +427,81 @@ public void addCustomParseFilter(ParseFilter filter) {
...
+	 * @see #startScan(String, Target, User, Object[])
+	 * @throws IllegalStateException if the target or custom configurations are not allowed in the current
+	 *             {@link org.parosproxy.paros.control.Control.Mode mode}.
+	 */
+	public int startScan(Target target, User user, Object[] customConfigurations) {
+		return startScan(createDisplayName(target, customConfigurations), target, user, customConfigurations);
+	}
+
+	/**
+	 * Creates the display name for the given target and, optionally, the give custom configurations.
```
kingthorin: Should be "given" custom configuration.

kingthorin:
Looks good, one minor typo.

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
thc202: Fixed, thanks!

PR #2263, opened 2016-02-23
spider: use all information/data available when seeding

thc202:
Change SpiderThread to use the start URI and start node as well as the
context nodes or all nodes in scope.

The spider will be seeded with following constraints:
 - If a context is provided:
  - Start URI, if in context;
  - Start node, if in context;
  - All nodes in the context;
 - If spidering just in scope:
  - Start URI, if in scope;
  - Start node, if in scope;
  - All nodes in scope;
 - If there's no context/scope restriction:
  - Start URI;
  - Start node, also:
    - Child nodes, if spidering "recursively".

Differences to previous seeding behaviour:
 - The start URI and start node are always used (if constraints met);
 - The nodes of the provided context are always used (previously, the
 nodes might not be used if a start node was also provided);

Seeding with context and "just in scope" are mutually exclusive, taking
precedence the former (same behaviour as before this change).
 ---

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/SpiderThread.java
+++ b/src/org/zaproxy/zap/extension/spider/SpiderThread.java
@@ -19,6 +19,7 @@
 package org.zaproxy.zap.extension.spider;
 
 import java.awt.EventQueue;
+import java.util.Collections;
 import java.util.Date;
 import java.util.Enumeration;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good.

One quick question on the JavaDoc content containing HTML, is that valid? I have little experience with such.

psiinon:
Looks good to me. Just @kingthorin 's question outstanding?

thc202:
@kingthorin yeah, it's valid.

PR #2273, opened 2016-02-24
Support site level stats

psiinon:
Also included in this change is the recording of tag, response code and contentType site level stats.

### DIFF
--- a/src/org/zaproxy/zap/extension/anticsrf/AntiCsrfDetectScanner.java
+++ b/src/org/zaproxy/zap/extension/anticsrf/AntiCsrfDetectScanner.java
@@ -23,12 +23,18 @@ import java.util.List;
 
 import net.htmlparser.jericho.Source;
 
+import org.apache.commons.httpclient.URIException;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.pscan.PassiveScanThread;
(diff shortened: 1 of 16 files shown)


src/org/zaproxy/zap/extension/anticsrf/AntiCsrfDetectScanner.java
```diff
@@ -86,4 +87,20 @@ public void setLevel(AlertThreshold level) {
 		// Ignore
 	}
 
+	private boolean tagHistoryType(int historyType) {
```
kingthorin: This code is repeated in a bunch of places, both as `tagHistoryType` and `appliesToHistoryType`, perhaps it could be centralized?

Or perhaps a constant definition of the applicable history types could be added to HistoryReference and then referenced in all these places, that way if it needs maintenance down the road it's only a single location that has to be modified. (Maybe within the zap.utils.Stats or zap.utils.Statistics ... would be a better home than HistoryReference?)

psiinon: True, this was originally hard coded for all rules ;)
The tags only make sense for records that appear in the History tab .. but I was thinking of adding support for displaying other record types in the History tab :/ 
But yeah, good suggestion, will look into it

psiinon: That better?

kingthorin:
Yup that looks good, thanks.

thc202:
Looks good to me.

PR #2275, opened 2016-02-25
spider: do not require a URL, through the API, if the context has seeds

thc202:
Change the SpiderAPI to not require a URL if the context provided
already has seeds. Also, if provided validate always that the URL is in
context and when in protected mode that the context is in scope.
Add helper method to Context class to check if there's any site tree
node (that is, seeds) in context.
 ---


kingthorin:
Looks good to me.

PR #2277, opened 2016-02-26
spider: preserve query parameters with same name when canonicalising URL

thc202:
Change class URLCanonicalizer to keep query parameters with same name
to preserve (or attempt to) the meaning of the original URL.
For example, with old behaviour URLs like:
 http://example.com/?array[]=1&array[]=2
 http://example.com/?IDs=1&IDs=2
would loose one of the query parameters, which could result in different
content being served.

### DIFF
--- a/src/org/zaproxy/zap/spider/URLCanonicalizer.java
+++ b/src/org/zaproxy/zap/spider/URLCanonicalizer.java
@@ -25,10 +25,9 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -153,7 +152,7 @@ public final class URLCanonicalizer {
 			path = path.trim();
 
 			/* Process parameters and sort them. */
-			final SortedMap<String, String> params = createParameterMap(canonicalURI.getRawQuery());
+			final SortedSet<QueryParameter> params = createSortedParameters(canonicalURI.getRawQuery());
 			final String queryString;
 			String canonicalParams = canonicalize(params);
 			queryString = (canonicalParams.isEmpty() ? "" : "?" + canonicalParams);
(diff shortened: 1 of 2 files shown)


kingthorin:
Wasn't there a thread in the user group or an issue related to this?

src/org/zaproxy/zap/spider/URLCanonicalizer.java
```diff
@@ -446,4 +450,83 @@ private static String normalizePath(final String path) {
...
+			}
+			int result = name.compareTo(other.name);
+			if (result != 0) {
+				return result;
+			}
+			return value.compareTo(other.value);
+		}
+
+		@Override
+		public int hashCode() {
```
kingthorin: I think I saw this code in another PR this week, should it be standardized in a utility class?

thc202:
Related to spider? I'm not recalling any.
(There are issues related to mishandling of parameters with same name in the active scanner, maybe were those?)

kingthorin:
Could be... no big deal.
I'm good with this change, just the one question which isn't really a blocking issue.

src/org/zaproxy/zap/spider/URLCanonicalizer.java
thc202: The hashCode() implementation?
(NameValuePair perhaps?)

thc202:
Added a comment to your question.

Updated the class QueryParameter to remove redundant null checks.

src/org/zaproxy/zap/spider/URLCanonicalizer.java
kingthorin: hashCode()....it was over here: https://github.com/thc202/zaproxy/commit/f2fb322ad21f0a5072accb0663426df18c2106c3#diff-24025444157a86c23540fcbb1c660144R106

thc202: Right, there are not much to standardise unfortunately, it really depends on the state of each class.
There are utility classes that help reduce "boiler plate" code though (e.g. Commons Lang's HashCodeBuilder).

kingthorin: Fair enough.

kingthorin:
Looks good

PR #2291, opened 2016-03-02
Change "context" view to wrap the context data with an object

### LINKED ISSUE #2115: Python API context.context("MyContext") is broken
The python API context.context("MyContext") is broken, possibly due to encoding issues. It is not returning proper output.

>>> zap.context.context("MyContext")
u'\\Q{"user":{"\\E.*'

thc202:
Change class ContextAPI to wrap the context data with an object (as all
views do), called "context", instead of returning the context data
directly, so that the Python ZAP API client is able to correctly read
the data, which expects (and removes) the wrapper object.

 {"context":{"id":"1", ..., "inScope":"true","loggedOutPattern":""}}

instead of:
 {"id":"1", ..., "inScope":"true","loggedOutPattern":""}

This change breaks existing custom clients that consume the JSON format.
The structure of XML format is unchanged.

Fix #2115 - Python API context.context("MyContext") is broken

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ContextAPI.java
+++ b/src/org/zaproxy/zap/extension/api/ContextAPI.java
@@ -260,7 +260,7 @@ public class ContextAPI extends ApiImplementor {
             result = new ApiResponseElement(name, contextNames.toString());
             break;
         case VIEW_CONTEXT:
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good.

PR #2292, opened 2016-03-03
Enhance alertFingerPrint

kingthorin:
In order to accommodate scanners which report alerts via multiple names
update alertFingerprint within ExtensionAlert to include the name of the
alert as an element.


thc202:
Looks good to me.

PR #2307, opened 2016-03-07
Add missing optional parameter "scanPolicyName" to ascan API actions

thc202:
Change class ActiveScanAPI to add the missing optional parameter
"scanPolicyName" to the declaration of the actions "enableScanners",
"disableScanners" and "setEnabledPolicies".


kingthorin:
Looks ok to me.

Issue #646, opened 2015-06-04
Password outgoing proxy in plain text on screen


zapbot:
```
What steps will reproduce the problem?
1.Go to Tools>Options>Connection
2.When setting your password for the outgoing proxy, the password is not hidden from
shoulder surfers by **** asterisks. 

What is the expected output? What do you see instead?
I think it would be good to at least hide them from direct view, preferably even encrypt
it. I know you can let ZAP ask for the credentials at startup, but it would be nice
not to have them in direct view. 


What version of the product are you using? On what operating system?
Windows 7, version 2.1.0

```

Original issue reported on code.google.com by `hugo.valk` on 2013-04-19 09:36:09
<hr>
 * *Attachment: zap_proxy_password.png<br>![zap_proxy_password.png](https://storage.googleapis.com/google-code-attachments/zaproxy/issue-646/comment-0/zap_proxy_password.png)*

Harinus:
A fix for this would be desirable. Use JPasswordField like in the "Add a new User" Dialog in Context Options.

kingthorin:
Also:
[OptionsConnectionPanel.java#L652-L662](https://github.com/zaproxy/zaproxy/blob/29a69eaddfdf02e1df8fb922d1c58213fb66327f/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java#L652-L662)
[OptionsConnectionPanel.java#L496](https://github.com/zaproxy/zaproxy/blob/29a69eaddfdf02e1df8fb922d1c58213fb66327f/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java#L496)

PR #2309, opened 2016-03-09
Fix outgoing proxy pass in plain text

kingthorin:
Change org.parosproxy.paros.extension.option.OptionsConnectionPanel to
use a JPasswordField instead of a ZapTextField for the outgoing proxy
password field. JPasswordFields show pips (bullets) to cut down on
concerns surrounding shoulder surfing.

Fixes zaproxy/zaproxy#646

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
@@ -30,6 +30,7 @@
 // ZAP: 2014/03/23 Issue 968: Allow to choose the enabled SSL/TLS protocols
 // ZAP: 2015/02/10 Issue 1528: Support user defined font size
 // ZAP: 2015/08/07 Issue 1768: Update to use a more recent default user agent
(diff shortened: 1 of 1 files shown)


psiinon:
The changes look good. 
Do we also want an option to show the field in the clear?

kingthorin:
While we do see such functionality on a lot of interfaces these days it is kind of contrary to preventing the shoulder surfing the reporting user(s) expressed concern about.

thc202:
This introduces a usability problem (IMHO), how do you know if the password was entered correctly? It would be great if it was possible to verify that there are no typos.
Adding a "clear option", or a second password field, or a button to test the authentication credentials... would address that problem.
I guess the latter is the better option which ensures everything is working as expected.

kingthorin:
Even in clear text someone could typo the entry ... same as Context configuration ...

thc202:
Well, in clear text you can actually see it, it's a lot easier to spot the problem.
With context, you mean the users' password? Yes, that has exactly the same problem.

thc202:
For the record, that's not a blocker, code changes look good to me.

psiinon:
Agreed - I'll merge but I think an option for showing password fields is always a good idea, ie not just this case;)

PR #2312, opened 2016-03-09
Give focus to "Edit Keyboard Shortcut" dialogue

thc202:
Use the "Options" dialogue (instead of main frame) as owner of the
dialogue "Edit Keyboard Shortcut" to allow the latter dialogue to
receive focus and be shown, always, in front of the former dialogue,
moreover change the "Edit Keyboard Shortcut" dialogue to be modal to
prevent navigation and modifications in the "Options" dialogue while
having it open.


kingthorin:
Looks good

PR #2313, opened 2016-03-09
Properly convert (old) excluded proxy domains

thc202:
Use the correct variable when converting (to Pattern) the old excluded
domains that use wildcard char '*', otherwise the conversion would not
be properly done thus not matching the expected domains.


kingthorin:
Seems good

PR #2323, opened 2016-03-11
Fix (secure) ZAP API request loop

thc202:
Change API class to consider insecure requests to secure ZAP API as
handled, otherwise the requests would be forward by the proxy which
would end up again (if local addresses) in the ZAP API, leading to a
request loop.
With the change insecure requests to secure API are simply dropped.


kingthorin:
Looks good

Issue #2183, opened 2016-01-18
API: Include quality in the ascan/progress response


psiinon:
As per https://groups.google.com/d/msg/zaproxy-develop/kpxnenhuoJA/WiN28ZNRDAAJ

alienatix:
+1

PR #2324, opened 2016-03-11
Include scanners' quality in API scanners views

thc202:
Change classes ActiveScanAPI and PassiveScanAPI to include the quality
of the scanners in the "scanners" view.
The response will now include a field called "quality" for each scanner,
with the corresponding quality ("alpha", "beta" and "release").

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
+++ b/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
@@ -798,6 +798,7 @@ public class ActiveScanAPI extends ApiImplementor {
 			scannerData.put("alertThreshold", String.valueOf(scanner.getAlertThreshold(true)));
 			scannerData.put("policyId", String.valueOf(scanner.getCategory()));
 			scannerData.put("enabled", String.valueOf(scanner.isEnabled()));
+			scannerData.put("quality", scanner.getStatus().toString());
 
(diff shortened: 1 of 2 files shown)


kingthorin:
Does this fix https://github.com/zaproxy/zaproxy/issues/2183, or does this facilitate the fix for that issue?

kingthorin:
Looks good

thc202:
It addresses it indirectly, whether it fixes it or not, it really depends on the use case.
For example, in the zapbot's wavsep script it would require another API call instead of simply use the status returned by the progress call. [1]

It makes things a lot easier to also return the status in progress call. I can update the PR (or better in another?) to do that.

[1] https://github.com/zapbot/zap-mgmt-scripts/blob/b7c7d32f26936d6b4735a344d03a6092f93ba55a/wavsep/wavsep-score.py#L518

thc202:
Should the field be called quality instead, as mentioned in #2183?
For the active/passive scanners it's used "quality" in the UI (but "status" for the add-ons).

kingthorin:
I personally prefer quality, but I can't really explain why specifically. Neither is wrong technically.

psiinon:
I dont really mind if its quality or status, but ideally we should be consistent :)
Shall we go for 'quality' ?

thc202:
Pull request updated to use quality.
I'll open another pull request to fix #2183.

thc202:
For the record, it was pull request #2326.

PR #2325, opened 2016-03-12
Reserved a Plugin ID for the Active Scan Rule HttpOnlySite

DarkPrince304:
As mentioned by @kingthorin in zaproxy/zap-extensions#277 I have reserved a Plugin ID in alerts.xml file for the Active Scan Rule "Http Only Site"


kingthorin:
Looks good.

PR #2328, opened 2016-03-15
Fix issue during uninstallation of named extension

thc202:
Obtain the UI name of the extension before uninstalling it, otherwise it
might lead to an exception (MissingResourceException, if the UI name of
the extension is internationalised) preventing the corresponding add-on
from being successfully uninstalled, dynamically.


kingthorin:
Looks good.

PR #2329, opened 2016-03-15
Filter seeds immediately before running the spider

thc202:
Filter all the seeds, using the current fetch filters, immediately
before running the spider to prevent any non-valid seed from being
accessed.
For example, auto generated seeds (e.g. "robots.txt") or previously
accessed messages would be used as seeds even if its URLs were excluded
from the spider.


psiinon:
Looks good to me

kingthorin:
Thanks

PR #2333, opened 2016-03-16
Allow to spider a site's subtree

### LINKED ISSUE #2274: Allow to spider just a site's subtree
From https://github.com/zaproxy/zaproxy/pull/2262#issuecomment-188296963

Allow to spider just a site's subtree, for example, for the given URLs:
http://example.com/path1
http://example.com/path1/path2
http://example.com/path1/path2/path3
http://example.com/path1/path2/path4/path5
http://example.com/path1/path6

it should allow to spider just pages under the subtree/path "path2", for example:
http://example.com/path1/path2
http://example.com/path1/path2/page.asp
http://example.com/path1/path2/path3
http://example.com/path1/path2/path3/page.html
http://example.com/path1/path2/path4/path5
...

thc202:
Add a new option to Spider dialogue ("Spider Subtree Only") and a
parameter to spider ZAP API ("subtreeOnly") to allow to spider just a
site's subtree. In both cases it's required a URI (manually set or
previously accessed), that's used as the subtree to spider. It's only
used the scheme, host, port and path component of the URI when
considering if a resource found while spidering is under or not the
specified subtree. For example, if the seed URI is:
 http://example.com:8080/path/?a=b#c
it's used only:
 http://example.com:8080/path/

The spider will exclude any resources found not under the subtree by
using a new FetchFilter, HttpPrefixFetchFilter, that allows to filter
URIs based on a URI prefix, in this use case the subtree URI.
Add tests to assert the expected behaviour of HttpPrefixFetchFilter.
Fix #2274 - Allow to spider just a site's subtree

### DIFF
--- a/src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
+++ b/src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
@@ -1,0 +1,320 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2016 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.spider.filters;
+
+import java.util.Arrays;
+import java.util.Locale;
+
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.apache.log4j.Logger;
+
+/**
+ * A {@code FetchFilter} that filters based on a HTTP or HTTPS {@code URI}.
+ * <p>
+ * The filtered {@code URI}s are required to start with the {@code URI} (the prefix) to be considered valid.
+ * 
+ * @since TODO add version
+ * @see #checkFilter(URI)
+ */
+public class HttpPrefixFetchFilter extends FetchFilter {
+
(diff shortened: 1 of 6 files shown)


src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
```diff
@@ -0,0 +1,320 @@
...
+
+        if (!startsWith(uri.getRawPath(), path)) {
+            return FetchStatus.OUT_OF_SCOPE;
+        }
+
+        return FetchStatus.VALID;
+    }
+
+    /**
+     * Tells whether or not the given {@code uri} has the same host as required by this prefix.
```
kingthorin: Shouldn't the code reference here be uppercase? (Matching L277)

src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
```diff
@@ -0,0 +1,320 @@
...
+        if (array == null) {
+            return false;
+        }
+
+        int length = prefix.length;
+        if (array.length < length) {
+            return false;
+        }
+
+        for (int i = 0; i < length; i++) {
```
kingthorin: The `i=length` element doesn't need to be checked?

thc202: There's no character at that index in the prefix (last character is at length -1, because of the array index starting at zero).

src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
thc202: In this case it's referring to the parameter in the other the class.

src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
kingthorin: K

src/org/zaproxy/zap/spider/filters/HttpPrefixFetchFilter.java
kingthorin: K

kingthorin:
Looks good to me.

psiinon:
Functionalities good, but maybe should persist the 'Spider Subtree Only' checkbox state, at least during the same run?

thc202:
Yeah, will change that.

thc202:
Change done.

psiinon:
Looks good :)

PR #2335, opened 2016-03-18
Add Outbound Proxy Password Peak

kingthorin:
Add a checkbox to allow the user to show their password.
Password is currently shown as bullets/pips by default
to prevent shoulder surfing. The checkbox setting is 
temporary and always defaults to unselected/disabled
on each display of the dialog(s). Changes applied to 
both the connection options panel and prompt at 
startup dialog.

Related to zaproxy/zaproxy#646, and [pull 2309](https://github.com/zaproxy/zaproxy/pull/2309).

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
@@ -31,6 +31,7 @@
 // ZAP: 2015/02/10 Issue 1528: Support user defined font size
 // ZAP: 2015/08/07 Issue 1768: Update to use a more recent default user agent
 // ZAP: 2016/03/08 Issue 646: Outgoing proxy password as JPasswordField (pips) instead of ZapTextField
+// ZAP: 2016/03/18 Add checkbox to allow showing of the password
 
 package org.parosproxy.paros.extension.option;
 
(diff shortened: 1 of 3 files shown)


psiinon:
Lgtm

src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
```diff
@@ -85,6 +85,7 @@
 	private ZapTextField txtProxyChainRealm = null;
 	private ZapTextField txtProxyChainUserName = null;
 	private JPasswordField txtProxyChainPassword = null;
+	private JCheckBox chkShowPassword = null;
```
thc202: Should this change have a "ZAP comment" too?

src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
```diff
@@ -508,6 +546,7 @@ private void setProxyChainEnabled(boolean isEnabled) {
 	    spinnerProxyChainPort.setEnabled(isEnabled);
 	    getProxyExcludedDomainsPanel().setComponentEnabled(isEnabled);
 	    chkProxyChainAuth.setEnabled(isEnabled);
+	    chkShowPassword.setEnabled(isEnabled);
```
thc202: This statement can be removed, the enabled stated is being set in setProxyChainAuthEnabled.

src/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
kingthorin: Done.

thc202:
Thank you.

PR #2343, opened 2016-03-21
Restore message auto tagging

thc202:
Change the method RegexAutoTagScanner.matched(HttpMessage, int) to use
the type of the message instead of its ID, when checking if the type
matches the allowed types.

Reported by @kingthorin.


kingthorin:
Looks good.

PR #2346, opened 2016-03-22
Add default implementations to AbstractPlugin

thc202:
Change class AbstractPlugin to add (default) implementations to methods
init() (changed from abstract) and getDependency(), implementation of a
method from Plugin interface. The init() method does nothing by default
and the method getDependency() returns an empty array, that is, no
dependencies.

The methods init() and getDependency() are not usually needed by the
scanners which leads to forced implementation of empty methods. Given
that most of the scanners do not need to override those methods the base
class should implement them, reducing boilerplate code overall.

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -47,6 +47,7 @@
 // ZAP: 2015/03/26 Issue 1573: Add option to inject plugin ID in header for all ascan requests
 // ZAP: 2015/07/26 Issue 1618: Target Technology Not Honored
 // ZAP: 2015/08/19 Issue 1785: Plugin enabled even if dependencies are not, "hangs" active scan
+// ZAP: 2016/03/22 Implement init() and getDependency() by default, most plugins do not use them
 
 package org.parosproxy.paros.core.scanner;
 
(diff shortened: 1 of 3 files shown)


src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -148,7 +159,17 @@ public void init(HttpMessage msg, HostProcess parent) {
         init();
     }
 
-    public abstract void init();
+    /**
+     * Finishes the initialisation of the plugin, subclasses should add any initialisation logic/code to this method.
+     * <p>
+     * Called after a message has been set. By default it does nothing.
```
kingthorin: This is confusing. The plugins init after a message is sent?

thc202: "Set" not "sent", but maybe it should say "Called after the plugin has been initialised with the message being scanned." instead? That might be more clear on when is called and how the message is used.

kingthorin: Ya perfect!

thc202: Updated the JavaDoc.

kingthorin:
Looks good to go.

PR #2349, opened 2016-03-23
Use Context factories when installed

### LINKED ISSUE #2331: Custom Context Panels not show in existing contexts after installation of add-on
If you have ZAP running and install the Context Alert Filters addon then open the properties of your Default Context (or any pre-existing context) it doesn't show the Alert Filter category. However, if you create a new context after adding the Alert Filter Addon you'll see it. OR, if you restart ZAP it shows for the Default Context.

thc202:
Use existing custom context data when adding a ContextDataFactory and
add custom context panels when adding ContextFactoryPanel, after full
initialisation (for example, when an add-on is installed), so that
existing Contexts show the correct data and custom panels.

Change class Control to notify the class Model once the initialisation
is done, by calling the (new) instance method Model.postInit().
Change class Model to call the method loadContextData(Session, Context)
on the added ContextDataFactory for all existing contexts, if it's being
added after the initialisation so that any existing custom context data
that's in the session is used/loaded, otherwise the data would be
missing when re-installing/updating an add-on.
Change class View to add the panels to all existing contexts after
initialisation, when a ContextPanelFactory is added to the view. Also,
some other refactorings were made:
 - Extract a method that obtains the ContextGeneralPanel of a context,
 used to create the path of the panels added;
 - Extract a method that adds the panels from a ContextPanelFactory for
 a context;
 - Reuse the same panel path (and contexts node name) when adding panels
 to an added Context, in method addContext(Context).

Fix #2331 - Custom Context Panels not show in existing contexts after
installation of add-on

### DIFF
--- a/src/org/parosproxy/paros/control/Control.java
+++ b/src/org/parosproxy/paros/control/Control.java
@@ -57,6 +57,7 @@
 // ZAP: 2015/04/02 Issue 321: Support multiple databases and Issue 1582: Low memory option
 // ZAP: 2015/09/17 Issue 1914: Support multiple add-on directories
 // ZAP: 2015/11/04 Issue 1920: Report the host:port ZAP is listening on in daemon mode, or exit if it cant
+// ZAP: 2016/03/23 Issue 2331: Custom Context Panels not show in existing contexts after installation of add-on
 
 package org.parosproxy.paros.control;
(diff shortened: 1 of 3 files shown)


kingthorin:
Looks good.

PR #2350, opened 2016-03-24
Do not access the EDT in daemon mode

thc202:
Change class SpiderThread to not add the messages found wile spidering
to the sites tree through the EDT and change class SiteNode to skip the
notification of changes if there's no view to notify to (that is, when
running in daemon mode), preventing the EDT accesses.


kingthorin:
Looks good

PR #2351, opened 2016-03-25
Fix concurrent modification in spider API

thc202:
Correct the object used in the synchronized block, by using
SpiderScan.getResourcesFound() instead of getResults(), when iterating
the resources found by the spider, to fix (a possible)
ConcurrentModificationException.


kingthorin:
Looks good

PR #2361, opened 2016-03-31
Rename tool bar option Expand Full to Full Layout

thc202:
Rename main tool bar option "Expand Full" to "Full Layout", the latter
name is the one used in the Options dialogue. Also, rename references
to that option to use the "new" name.


kingthorin:
Looks good

PR #2362, opened 2016-03-31
Allow to select all Spider results

thc202:
Allow to select all (or, simply more than one) results shown in the
Spider tab, so the user can copy them at once, instead of one at a time.

Change class SpiderPanel to allow to select, in the table, multiple non-
-consecutive rows, also change to use a highlighter for the "Processed"
column to copy true/false, not the paths to the icons.
Change class SpiderPanelTableModel to no longer return Icons for the
"Processed" column but the actual boolean value. Also, rename instance
variable of SpiderScanResult from "skipped" to "processed" and rename
the parameter "inScope" to "skipped" in method addScanResult(...) to
clarify its use and be easier to read the code.


kingthorin:
Looks good

PR #2363, opened 2016-03-31
Keep break buttons in sync when changing mode

thc202:
Change class BreakPanelToolbarFactory to keep the break buttons "All"
and "Request"/"Response" in sync, that is, when changing between the two
types of buttons set them selected (and old buttons unselected, to not
interfere with others buttons' state) depending on the selection state
of the other ones, otherwise while changing modes the button(s) would
still be selected and would trigger the breakpoints without the newly
shown button(s) being selected.
The logic is as follow:
 - When changing to "Request"/"Response" buttons set both selected, if
 the "All" button is selected;
 - When changing to "All" set it selected, if at least one of the
 "Request"/"Response" buttons is selected;
 - If none of the button(s) is selected there's no change in its states.


kingthorin:
Looks good

PR #2369, opened 2016-04-01
Fix issues related to break buttons

### LINKED ISSUE #2366: Inconsistencies in break buttons shown
Steps to reproduce the issue(s):
1. Run ZAP;
2. Change the option ("Options" > "Display" >) "Show break buttons:" to "Only in the break panel" and restart ZAP;
3. Note that in the "Break" panel it's missing the button (or buttons, if "Separate Request and Response buttons" is set) to enable the breakpoints;
4. Enable the breakpoints through the menu "Tools" > "Toggle Break on All Requests";
5. Proxy a request, forward the request and once the response is received note that the button is now shown, as well the "separate" buttons.

ZAP Versions:
Versions >= 2.4.0

### LINKED ISSUE #2367: Change break buttons location without restarting ZAP
From #2366, it should be possible to change the break buttons location (option "Options" > "Display" > "Show break buttons:") without restarting ZAP.

thc202:
Change the location of the break buttons without restarting ZAP and fix
break buttons inconsistencies.

Change class ExtensionBreak to set the location of the break buttons
when the options are loaded and when changed, through a (new) helper
method applyViewOptions(OptionsParam).
Change class BreakPanel to show/hide the tool bars/buttons depending on
the option set, also show corresponding break buttons in all tool bars
(using helper class BreakButtonsUI).

Fix #2366 - Inconsistencies in break buttons shown
Fix #2367 - Change break buttons location without restarting ZAP

### DIFF
--- a/src/org/zaproxy/zap/extension/brk/BreakPanel.java
+++ b/src/org/zaproxy/zap/extension/brk/BreakPanel.java
@@ -28,6 +28,8 @@ import java.awt.event.KeyEvent;
 
 import javax.swing.BorderFactory;
 import javax.swing.ImageIcon;
+import javax.swing.JButton;
+import javax.swing.JComponent;
 import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JToggleButton;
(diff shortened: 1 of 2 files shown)


kingthorin:
Looks good

PR #2370, opened 2016-04-03
Implemented deletion of nodes via API. Fixes #2342

### LINKED ISSUE #2342: Ability to manage Sites in ZAP via API
It would be great to be able to manage Sites in ZAP via API. Right now once site is added (while proxying requests through ZAP), it's only possible to remove it via GUI.

DarkPrince304:
This patch will fix the issue #2342 . The user will now be able to delete the Site Nodes of the scanned site via the API too and not just the GUI. The functionality for now is just similar to the GUI that is if a parent is chosen then all of its children are deleted as well without the choice of leaving the children and just deleting the parent. The user can also specify the method as a parameter.

Would like your reviews @thc202 , @kingthorin , @psiinon  :)

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -854,6 +854,7 @@ context.api.action.excludeAllContextTechnologies = Excludes all built in technol
 copy.copy.popup = Copy
 copy.desc       = Provides a right click option to copy highlighted text to the clipboard
 
+core.api.action.deleteSiteNode = Deletes the site node found in the Sites Tree on the basis of the URL, HTTP method, and post data (if applicable and specified). 
 core.api.action.loadSession = Loads the session with the given name. If a relative path is specified it will be resolved against the "session" directory in ZAP "home" dir.
 core.api.action.newSession = Creates a new session, optionally overwriting existing files. If a relative path is specified it will be resolved against the "session" directory in ZAP "home" dir.
 core.api.action.saveSession = Saves the session with the name supplied, optionally overwriting existing files. If a relative path is specified it will be resolved against the "session" directory in ZAP "home" dir.
--- a/src/org/zaproxy/clientapi/gen/Core.java
+++ b/src/org/zaproxy/clientapi/gen/Core.java
@@ -532,6 +532,21 @@ public class Core {
 		return api.callApi("core", "action", "setOptionUseProxyChainAuth", map);
 	}
 
+	public ApiResponse deleteSiteNode(String apikey, String url, String method, String postData) throws ClientApiException {
+		Map<String, String> map = new HashMap<String, String>();
+		if (apikey != null) {
+			map.put("apikey", apikey);
+		}
+		map.put("url", url);
+		if (method != null) {
+			map.put("method", method);
+		}
+		if (postData != null) {
+			map.put("postData", postData);
+		}
+		return api.callApi("core", "action", "deleteSiteNode", map);
+	}
+
 	public byte[] proxypac(String apikey) throws ClientApiException {
 		Map<String, String> map = null;
 		map = new HashMap<String, String>();
--- a/src/org/zaproxy/zap/extension/api/CoreAPI.java
+++ b/src/org/zaproxy/zap/extension/api/CoreAPI.java
@@ -67,6 +67,7 @@ import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.model.SessionListener;
 import org.parosproxy.paros.model.SiteNode;
+import org.parosproxy.paros.model.SiteMap;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
@@ -79,6 +80,7 @@ import org.w3c.dom.Element;
 import org.w3c.dom.Text;
 import org.zaproxy.zap.extension.alert.ExtensionAlert;
 import org.zaproxy.zap.extension.dynssl.ExtensionDynSSL;
+import org.zaproxy.zap.extension.history.PopupMenuPurgeSites;
 import org.zaproxy.zap.model.SessionStructure;
 import org.zaproxy.zap.model.SessionUtils;
 import org.zaproxy.zap.utils.HarUtils;
@@ -113,6 +115,7 @@ public class CoreAPI extends ApiImplementor implements SessionListener {
 	private static final String ACTION_COLLECT_GARBAGE = "runGarbageCollection";
 	private static final String ACTION_CLEAR_STATS = "clearStats";
 	private static final String ACTION_SET_MODE = "setMode";
+	private static final String ACTION_DELETE_SITE_NODE = "deleteSiteNode";
 	
 	private static final String VIEW_ALERT = "alert";
 	private static final String VIEW_ALERTS = "alerts";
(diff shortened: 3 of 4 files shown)


src/org/zaproxy/clientapi/gen/Core.java
```diff
@@ -532,6 +532,21 @@ public ApiResponse setOptionUseProxyChainAuth(String apikey, boolean bool) throw
 		return api.callApi("core", "action", "setOptionUseProxyChainAuth", map);
 	}
 
+	public ApiResponse deleteSiteNode(String apikey, String url, String method) throws ClientApiException {
+		Map<String, String> map = null;
+		map = new HashMap<String, String>();
```
kingthorin: These can be combined I believe...

kingthorin: These can be combined I believe...

I realize it's reused code, but map = new HashMap should be fine without first being null.

DarkPrince304: Fixed!

kingthorin:
I think this may actually need to include checking/specifying parameters. Specifically I'm thinking of single page apps where
`GET http://example.org/page?fred=bob` could be different than `GET http://example.org/page`. If only considering URL and method too much might be deleted.

DarkPrince304:
@kingthorin : Yeah makes sense. Can you guide me where to look at to implement this? Does it have something related to the `String postData` parameter in `SiteMap.findNode` method?

kingthorin:
Yes I believe it should accommodate postData.

org.zaproxy.zap.extension.alert.ExtensionAlert has a couple good examples of use (for example findNode being used based on alert details).

DarkPrince304:
@kingthorin But I believe that findNode already takes care of GET parameters. What do you think?

kingthorin:
I believe so. Should be fairly simple for you to test with your code. Visit a page without params and one with, then use your api call to delete the one with. Check that the one without still exists. It doesn't matter if the page actually uses the params just that they're part of the uri. So `example.org` and `example.org?foo=bar` or something like that should work.

src/lang/Messages.properties
```diff
@@ -854,6 +854,7 @@ context.api.action.excludeAllContextTechnologies = Excludes all built in technol
 copy.copy.popup = Copy
 copy.desc       = Provides a right click option to copy highlighted text to the clipboard
 
+core.api.action.deleteSiteNode = Deletes the site node found in the Sites Tree on the basis of the URL and the Method passed.
```
kingthorin: I'd suggest "...of the URL, HTTP method, and post data (if applicable and specified)."

kingthorin:
One minor suggestion, otherwise looks good to me.

DarkPrince304:
I have made the appropriate changes and will also be testing it today. You can also test it at your end to see if it works on your end too :)

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -176,6 +182,7 @@ public CoreAPI() {
 		this.addApiAction(new ApiAction(ACTION_DELETE_ALL_ALERTS));
 		this.addApiAction(new ApiAction(ACTION_COLLECT_GARBAGE));
 		this.addApiAction(new ApiAction(ACTION_CLEAR_STATS, null, new String[] {PARAM_KEY_PREFIX}));
+		this.addApiAction(new ApiAction(ACTION_DELETE_SITE_NODE, null, new String[] {PARAM_URL, PARAM_METHOD, PARAM_POST_DATA}));
```
thc202: The PARAM_URL should be mandatory.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -453,6 +460,22 @@ public void process(HttpMessage msg) {
 			Stats.clear(this.getParam(params, PARAM_KEY_PREFIX, ""));
 			return ApiResponseElement.OK;
 			
+		} else if (ACTION_DELETE_SITE_NODE.equals(name)) {
+			try {
+				String url = params.getString(PARAM_URL);
+				String method = getParam(params, PARAM_METHOD, "GET");
```
thc202: Maybe it should not default to GET? We need to handle the case where there's no method being sent (i.e. thus having null value), to allow to delete directory nodes.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -453,6 +460,22 @@ public void process(HttpMessage msg) {
...
+				String postData = params.getString(PARAM_POST_DATA);
```
thc202: Since it's optional it should default to an empty string.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -453,6 +460,22 @@ public void process(HttpMessage msg) {
...
+				URI uri = new URI(url, true);
+				SiteMap siteMap = session.getSiteTree();
+				SiteNode siteNode = siteMap.findNode(uri, method, postData);
+				if(siteNode == null) {
+					throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, PARAM_URL);
```
thc202: `DOES_NOT_EXIST` instead of `ILLEGAL_PARAMETER` might describe better this case.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -453,6 +460,22 @@ public void process(HttpMessage msg) {
...
+				}
+				PopupMenuPurgeSites.purge(siteMap, siteNode);
+				return ApiResponseElement.OK;
+			} catch (URIException e) {
+				throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, PARAM_URL);
```
thc202: The URIException should be added as cause, it provides more details when the ApiException is logged.

src/org/zaproxy/zap/extension/api/CoreAPI.java
thc202: OK, let it be GET but then we need to change an if statement in [`SiteMap.getLeafName(String, URI, String, String)`](https://github.com/zaproxy/zaproxy/blob/a48e956541defb4d5124fb83821a332900d4a1c0/src/org/parosproxy/paros/model/SiteMap.java#L549) to also consider an empty `String` as no method.
So, when the user specifies an empty method (as opposed to no method, which defaults to GET) it's able to find the leaf (directory) nodes.

kingthorin:
Build with this code is failing

DarkPrince304:
I'll check. What are you getting?

kingthorin:
Just check the Travis CI build associated with this PR, you should see it at the bottom of the page.

alienatix:
Hey @DarkPrince304 thanks for implementing this! I'd be happy to test it in the nearest weekly build, so if you could close the issue shortly, that would be perfect.

alienatix:
Btw, will this API request delete corresponding alerts as well? It happens in UI right now and seems reasonable.

thc202:
Yes, corresponding alerts will be deleted too.

alienatix:
Fantastic! 🎯

DarkPrince304:
@alienatix : Hello, I have made the appropriate changes and I think it should all be fine now so lets just wait for this to get merged now. :)

alienatix:
waiting! 🕙

thc202:
Looks good to me, thank you.

kingthorin:
Thanks

PR #2371, opened 2016-04-04
Use StatusUI in scanners' dialogues

thc202:
Change all dialogues that show the Quality of the active and passive
scanners to use StatusUI, for proper sorting of the values of the
Quality column.

Change class CategoryTableModel to use StatusUI for its Quality column
and change to use a List/ArrayList instead of a Vector to keep the table
entries, synchronisation provided by Vector class is not required.
Change class PolicyAllCategoryPanel to use StatusUI's string
representation in the quality combo boxes and extracted helper methods
to reduce code duplication.
Change class PolicyPassiveScanPanel to use StatusUI's string
representation in the quality combo box.
Change PolicyPassiveScanTableModel to use StatusUI for its Quality
column.


psiinon:
Looks good

PR #2372, opened 2016-04-04
Do not require a restart to show/hide the tool bar

### LINKED ISSUE #2368: Do not require a restart to show/hide the main tool bar
As with the break buttons location (#2367), do not require ZAP's restart to show/hide the main tool bar.

thc202:
Change class MainFrame to always add the main tool bar, so that it can
be shown/hidden at any time, also add a method to allow to change its
visible state.
Change class View to apply the main tool bar visibility option when it
is initialised and add a method to allow to change the visible state of
the main tool bar (which calls the method of MainFrame class).
Change class MenuToolsControl to update the View with the main tool bar
visibility option, after the options are set.
Change class BreakPanel to take into account the changes done to
visibility of the main tool bar (to show/hide the buttons in the break
panel if the main tool bar is hidden/shown and the option is to show or
not the buttons in the tool bar).
Change class OptionsParamView to use boolean (for external code) instead
of an int for the visibility option of the main tool bar and deprecate
the methods that use int.
Change class OptionsViewPanel to use the boolean methods of class
OptionsParamView.

Fix #2368 - Do not require a restart to show/hide the main tool bar

### DIFF
--- a/src/org/parosproxy/paros/control/MenuToolsControl.java
+++ b/src/org/parosproxy/paros/control/MenuToolsControl.java
@@ -22,6 +22,7 @@
 //      options have changed.
 // ZAP: 2013/01/23 Clean up of exception handling/logging.
 // ZAP: 2013/03/03 Issue 546: Remove all template Javadoc comments
+// ZAP: 2016/04/04 Do not require a restart to show/hide the tool bar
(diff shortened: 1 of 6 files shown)


kingthorin:
Looks good

Issue #2127, opened 2015-12-14
Warn if the report generation failed


sbunde:
Hi,
- I am running a spider scan on an intranet site. 
- Generation XML and HTML report from 'Report-> generate XML/HTML report' option of ZAP GUI.
- Saving the report.
- Open the report (xml and html) it gives blank page.
- Please let me know what wrong am I doing.

thc202:
What ZAP version?
Would you mind checking ZAP's log file to see if there's any error? (file zap.log located in ZAP's default directory or the directory manually specified [1]).


[1] https://github.com/zaproxy/zaproxy/wiki/FAQconfig

sbunde:
ZAP version is 2.4.3 and I could not find the ZAP directory (windows 7 C:\Users\<username>\OWASP ZAP)
Also, I don't remember specifying it manually.

thc202:
Did you replace `<username>` with the actual user that you're using?

sbunde:
yes

sbunde:
![image](https://cloud.githubusercontent.com/assets/16288184/11779764/875220c4-a25d-11e5-9d2b-de892bc0daed.png)

thc202:
OK, might be hidden?
An alternative is to start ZAP with zap.bat file. [1]
It would output the error to the console, when generating the report.

[1] https://github.com/zaproxy/zaproxy/wiki/FAQinstall

sbunde:
Hi,
I was now able to check the logs and it showed error
2015-12-14 13:22:32,628 [pool-2-thread-1] WARN  ENGINE - data file enlarge failed 
java.io.IOException: There is not enough space on the disk

So I changed the drive to save the report and they are saved successfully.

Thanks for your guidance.

kingthorin:
@sbunde thanks for letting us know. 

I'm going to re-open this one as we should probably handle that condition more gracefully.

DarkPrince304:
@kingthorin: How do you think this should be handled?

kingthorin:
So originally I had thought this was about report generation failing, so I envisioned it going here:
https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/view/widgets/WritableFileChooser.java

[...]

PR #2376, opened 2016-04-06
Issue 2127 Fixes IOException

mustafa421:
This is a rather small and simple fix for #2127 . However, it checks if there is available disk space prior to saving any results. Previously, ZAP attempts to save the file anyway which lead to an IOException being thrown that wasn't properly handled. I haven't internationalized the text just yet, because I'd like to get opinions. Please let me know if I'm missing something or should look into something else as well. Thoughts :) @psiinon @thc202 @kingthorin


mustafa421:
#2127

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -24,6 +24,11 @@ public WritableFileChooser(File currentDirectory) {
 	@Override
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
+		
+		if (selectedFile.length() > selectedFile.getUsableSpace()) {
+			JOptionPane.showMessageDialog(this, "Not enough disk space!");
```
kingthorin: The message should be internationalized (as others are in WritableFileChooser).

The change should align with the commit message (or the commit message should be reworded. via interactive rebase to correctly portray the change).

mustafa421:
@kingthorin Fixed!

src/lang/Messages_uk_UA.properties
```diff
@@ -1553,6 +1553,7 @@ report.write.overwrite.dialog.title=Existing File
 report.write.overwrite.dialog.message=The selected file already exists. Do you want to replace it?
 report.write.permission.dialog.title=Permissions Failure
 report.write.permission.dialog.message=File or Directory not writable\: \n{0}\n Please select a different location.
+report.write.diskspace.error.dialog.message=Not enough disk space to export file
```
thc202: This should be added to Messages.properties file (which is the source of translations, like this file).
The message should probably say, save/write than export.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -24,6 +24,11 @@ public WritableFileChooser(File currentDirectory) {
 	@Override
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
+		
+		if (selectedFile.length() > selectedFile.getUsableSpace()) {
```
thc202: I'm not sure this check would prevent the errors about disk space (when writing the report or other info).
The selected file might not exist, which would return 0L for both methods, even if it does exist it's not necessarily the size that the final file will have (in most cases the data was not yet generated so the final size is unknown, at least by this class).
I'm not sure the issue is fixable in this class.

This is what get from those calls for a file that exists and another that doesn't:
```
/exists           length: 511906 getUsableSpace: 16297455616
/does_not_exist   length: 0      getUsableSpace: 0
```

mustafa421: That makes sense. I initially imagined a check to verify that at least 5-10MB (possible user set limit?) exists on a hard disk before proceeding, but I figured that method would be sloppy. I guess for now I could do more research into it unless someone else has an idea.

kingthorin: I think doing it here was my suggestion. If that's not workable then the next best bet for this specific case is to add a try/catch for ioexception (is it always ioexception when full?) To ReportLastScan, showing a dialog that suggests the disk may be full and the file empty or incomplete... @thc202 what do you think?

thc202: We could have both, warn about low disk space available (or if it's full) when choosing the file and warn when there's an exception during the generation of the report, the latter definitely. Currently ReportGenerator just catches the IOExceptions and logs them, which is good, but other than the log messages there's no indication that the report generation failed:
```
660393 [AWT-EventQueue-0] ERROR org.parosproxy.paros.extension.report.ReportGenerator  - com.sun.org.apache.xalan.internal.xsltc.TransletException: org.xml.sax.SAXException: No space left on device
java.io.IOException: No space left on device
javax.xml.transform.TransformerException: com.sun.org.apache.xalan.internal.xsltc.TransletException: org.xml.sax.SAXException: No space left on device
java.io.IOException: No space left on device
	at com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl.transform(TransformerImpl.java:752)
	at com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl.transform(TransformerImpl.java:359)
	at org.parosproxy.paros.extension.report.ReportGenerator.stringToHtml(ReportGenerator.java:118)
	at org.parosproxy.paros.extension.report.ReportLastScan.generate(ReportLastScan.java:74)
	at org.parosproxy.paros.extension.report.ReportLastScan.generateReport(ReportLastScan.java:190)
	at org.parosproxy.paros.extension.report.ExtensionReport$1.actionPerformed(ExtensionReport.java:98)
```

```
781904 [AWT-EventQueue-0] ERROR org.parosproxy.paros.extension.report.ReportGenerator  - No space left on device
java.io.IOException: No space left on device
	at java.io.FileOutputStream.writeBytes(Native Method)
	at java.io.FileOutputStream.write(FileOutputStream.java:345)
	at sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:221)
	at sun.nio.cs.StreamEncoder.implWrite(StreamEncoder.java:282)
	at sun.nio.cs.StreamEncoder.write(StreamEncoder.java:125)
	at java.io.OutputStreamWriter.write(OutputStreamWriter.java:207)
	at java.io.BufferedWriter.flushBuffer(BufferedWriter.java:129)
	at java.io.BufferedWriter.write(BufferedWriter.java:230)
	at java.io.Writer.write(Writer.java:157)
	at org.parosproxy.paros.extension.report.ReportGenerator.stringToHtml(ReportGenerator.java:181)
	at org.parosproxy.paros.extension.report.ReportLastScan.generate(ReportLastScan.java:74)
	at org.parosproxy.paros.extension.report.ReportLastScan.generateReport(ReportLastScan.java:190)
	at org.parosproxy.paros.extension.report.ExtensionReport$2.actionPerformed(ExtensionReport.java:116)
```

kingthorin:
Ok I like that idea. In WritableFileChooser let's add a 5MB check (?) and warn, don't prevent selection of the file/destination but add a warning. Then in ReportGenerator as @thc202 suggested catch the applicable exception and state that the file creation may have failed or content might be incomplete.

thc202:
Sounds good to me.

mustafa421:
Changes made in both WritableFileChooser & ReportGenerator classes.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -126,7 +127,7 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
 					bw = new BufferedWriter(new FileWriter(outfilename + "-orig.xml"));
 					bw.write(inxml);
 				} catch (IOException e2) {
-					logger.error(e.getMessage(), e);
+					logger.error(Constant.messages.getString("report.write.diskspace.error.log.message"), e);
```
kingthorin: From a usability perspective should we also pop a message dialog here? There is a small chance that a huge report might be written where the 5MB warning wouldn't be seen...

mustafa421: Good point, I'll include that into the next commit

mustafa421:
Added a message dialog to alert user that a report may have failed after catching the IOException. Thanks @kingthorin .

kingthorin:
Looks good

src/lang/Messages.properties
```diff
@@ -1629,6 +1629,8 @@ report.unknown.error=Failed to write to file {0}, see log for detail
 report.write.overwrite.dialog.title=Existing File
 report.write.overwrite.dialog.message= The selected file already exists. Do you want to replace it?
 report.write.permission.dialog.title=Permissions Failure
+report.write.diskspace.warning.dialog.message=Warning: There may not be enough disk space to continue operation.
```
thc202: - "the operation"?
 - It should probably say how much space may not be enough (i.e. the 5MB).
 - The "Warning: " could be dropped if it's shown a warning dialogue (currently it's showing an information dialogue).

src/lang/Messages.properties
```diff
@@ -1629,6 +1629,8 @@ report.unknown.error=Failed to write to file {0}, see log for detail
...
+report.write.diskspace.error.log.message=File creation may have failed or is not complete.
```
thc202: The "diskspace" and "log" should be removed, the message applies to other errors and it's no longer just logged.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
thc202: I don't think the log message should be internationalised, that makes it harder to (immediately) understand the logs (although in this case it might not be that bad).

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -126,7 +129,8 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
 					bw = new BufferedWriter(new FileWriter(outfilename + "-orig.xml"));
 					bw.write(inxml);
 				} catch (IOException e2) {
-					logger.error(e.getMessage(), e);
+					logger.error(Constant.messages.getString("report.write.diskspace.error.log.message"), e);
+					JOptionPane.showMessageDialog(View.getSingleton().getMainFrame(), Constant.messages.getString("report.write.diskspace.error.log.message"));
```
thc202: - There might not be a view (i.e. when generating the report through the command line), View.isInitialised() should be checked first;
 - Probably better to leave the parent null, it might be other dialogue;
 - The other IOException catch statements (the ones when writing) should have this warning too.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -24,6 +24,10 @@ public WritableFileChooser(File currentDirectory) {
 	@Override
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
+		//Warns if there is less than 5MB available
+		if (selectedFile.getUsableSpace() < 5000000) {
```
thc202: If the file does not exist getUsableSpace() returns 0.

5000000 could/should be extracted to a constant, in which case the comment could be removed (if the code is clearer).

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -24,6 +24,10 @@ public WritableFileChooser(File currentDirectory) {
...
+			JOptionPane.showMessageDialog(this, Constant.messages.getString("report.write.diskspace.warning.dialog.message"));
```
thc202: Allowing the user to cancel the selection might be better than just warn, as it allows the user to choose other destination.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
thc202: (Also, the function should return, no point trying to continue with the I/O operations if the previous failed.)

src/org/parosproxy/paros/extension/report/ReportGenerator.java
mustafa421: Did you want me to change it, or simply put a comment? Ofcourse, I can always leave it alone :)

thc202: I think it should be changed to the exception message, but looking again it should be another message, something like "Failed to write debug XML file:", which is more informative.
Also, the view warn should be in the outer catch block, otherwise there's no warn if the "debug" file is correctly saved.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
mustafa421: The method expects a File to be returned. Should I simply return null in the catch block?

thc202: That was a side comment, didn't mean you should fix/change that.
But if you want, yeah, it should probably return null. Some of the callers ignore the return value, but one is already showing a custom error message if the file is null. [1] That caller would have to be changed to not show the message.

[1] https://github.com/zaproxy/zaproxy/blob/c678ac66c740290df8956c835596acecc2198ed8/src/org/parosproxy/paros/extension/report/ReportLastScan.java#L190-L194

mustafa421: Oh, now I understand you :). I was wondering as well why the code attempts to continue the I/O operation.

mustafa421:
Fixed the changes. It was a bit confusing to fix the conflict, but I think I got it. @thc202 what do you think?

src/lang/Messages.properties
```diff
@@ -1625,6 +1625,10 @@ report.unknown.error=Failed to write to file {0}, see log for detail
 report.write.overwrite.dialog.title=Existing File
 report.write.overwrite.dialog.message= The selected file already exists. Do you want to replace it?
 report.write.permission.dialog.title=Permissions Failure
+report.write.diskspace.warning.dialog.title=Low Disk Space
+report.write.diskspace.warning.dialog.message=There may not be enough disk space to continue the operation, would you like to continue?
+report.write.dialog.message=File creation may have failed or is not complete.
+report.write.permission.dialog.message=File or Directory not writable: \n{0}\n Please select a different location.
```
thc202: This message should be deleted, it was replaced with the following two.

src/lang/Messages.properties
```diff
@@ -1625,6 +1625,10 @@ report.unknown.error=Failed to write to file {0}, see log for detail
 report.write.overwrite.dialog.title=Existing File
 report.write.overwrite.dialog.message= The selected file already exists. Do you want to replace it?
 report.write.permission.dialog.title=Permissions Failure
+report.write.diskspace.warning.dialog.title=Low Disk Space
+report.write.diskspace.warning.dialog.message=There may not be enough disk space to continue the operation, would you like to continue?
```
thc202: We should tell how much space we check, so the user can judge better if it's enough or not.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -121,12 +124,12 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
 				logger.error(e.getMessage(), e);
 				// Save the xml for diagnosing the problem
 				BufferedWriter bw = null;
-	
+				JOptionPane.showMessageDialog(null, Constant.messages.getString("report.write.dialog.message"));
```
thc202: - This should be shown only when there's view/GUI (i.e. View.isInitialised() returns true);
 - To reduce code duplication this (the view check and show of the dialogue) should be extracted to a method;
 - Now that the message is shown in several catch blocks we must return, otherwise ZAP might end up showing the warning dialogue twice (returning "File (outfilename)" at the end of the catch block is fine, it keeps the same behaviour).

There's one IOException catch block that's missing the message (the else "// No XSLT file specified, just output the XML straight to the file").

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -13,6 +13,8 @@
 public class WritableFileChooser extends JFileChooser {
 
 	private static final long serialVersionUID = -8600149638325315049L;
+	
+	private static final int minimumSpaceRequirement =  5000000;
```
thc202: Constant names should be all upper case and separated with underscores (instead of camel-case).
Consider adding the postfix MB to the name.
There's an extra space after the equal that should be removed.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -25,6 +27,16 @@ public WritableFileChooser(File currentDirectory) {
 	@Override
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
+		
+		if (selectedFile.getUsableSpace() < minimumSpaceRequirement) {
```
thc202: This check needs changes, if the file does not exist the message is always shown (if it does not exist it should use the parent file instead).

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -49,6 +50,8 @@
 
 import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.view.View;
```
thc202: This import is not used (but, it should/will, so not a problem).

thc202:
There's just one message that should be deleted (that resulted from the conflict).

The merge commit should be removed, it's better/easier to rebase than merge, to not introduce merge commits.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
mustafa421: It's there at line 182?

mustafa421:
I made the changes as asked. I also deleted the message after fixing my local branch (it was a mess) :). In the WritableFileChooser, I added a check to see if selectedFile exists, and changed it to its parent if it didn't.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -121,12 +124,13 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
...
 				BufferedWriter bw = null;
-	
+				showDialogForGUI();
 				try {
 					bw = new BufferedWriter(new FileWriter(outfilename + "-orig.xml"));
 					bw.write(inxml);
 				} catch (IOException e2) {
-					logger.error(e.getMessage(), e);
+					logger.error("Failed to write debug XML file", e);
+					return outfile;
```
thc202: - It should return "new File(outfilename)" instead, to keep the old behaviour.
- In the outer catch block it should return the same.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -158,7 +162,9 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
 				}
 	
 			} catch (IOException e) {
+				showDialogForGUI();
 				logger.error(e.getMessage(), e);
+				return outfile;
```
thc202: This return is not necessary (it can use the existing return).

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -180,7 +186,9 @@ public static File stringToHtml(String inxml, String infilexsl, String outfilena
 				bw = new BufferedWriter(new FileWriter(outfilename));
 				bw.write(inxml);
 			} catch (IOException e2) {
+				showDialogForGUI();
 				logger.error(e2.getMessage(), e2);
+				return new File (outfilename);
```
thc202: This one is not necessary, it will return this anyway.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -219,7 +227,9 @@ public static String stringToHtml(String inxml, String infilexsl) {
 			transformer.transform(source, result);
 
 		} catch (TransformerException | SAXException | ParserConfigurationException | IOException e) {
+			showDialogForGUI();
 			logger.error(e.getMessage(), e);
+			return inxml;
```
thc202: This one is not necessary, the next statement will return (without doing any I/O operation).

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -259,7 +269,9 @@ public static File fileToHtml(String infilexml, String infilexsl, String outfile
 			transformer.transform(source, result);
 
 		} catch (TransformerException | SAXException | ParserConfigurationException | IOException e) {
+			showDialogForGUI();
 			logger.error(e.getMessage(), e);
+			return outfile;
```
thc202: This one is not necessary, the next statement already returns.

src/org/parosproxy/paros/extension/report/ReportGenerator.java
```diff
@@ -313,4 +325,10 @@ public static String getDebugXMLString(Document doc) throws TransformerException
 		transformer.transform(new DOMSource(doc), new StreamResult(writer));
 		return writer.getBuffer().toString().replaceAll("\n|\r", "");
 	}
+	
+	public static void showDialogForGUI() {
```
thc202: It should be private.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,6 +28,19 @@ public WritableFileChooser(File currentDirectory) {
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
 
+		if (!selectedFile.exists()) {
```
thc202: The check that the file exists could be saved in a temporary variable to use in the other if statement.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,6 +28,19 @@ public WritableFileChooser(File currentDirectory) {
...
+			selectedFile = selectedFile.getParentFile();
```
thc202: It should not be saved in this variable, otherwise the following checks would not be correct (e.g. would be checking that the grand-parent directory is writable instead of the parent directory).

src/org/parosproxy/paros/extension/report/ReportGenerator.java
thc202: The commit that I checked didn't have it (I guess it could be outdated).

thc202:
The class ReportGenerator should have a "ZAP comment" at the beginning of the file. [1]

mustafa421:
Changes made as asked. I realized that some of the return statements were indeed pointless 👍 . I added a temp variable checkFile to be used when checking for available space.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,6 +28,21 @@ public WritableFileChooser(File currentDirectory) {
 	public void approveSelection() {
 		File selectedFile = getSelectedFile();
 
+        File checkFile = selectedFile;
```
thc202: This and following line should use tabs for indentation (same chars as surrounding code).

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,6 +28,21 @@ public WritableFileChooser(File currentDirectory) {
...
+        boolean fileExists = checkFile.exists() ? true : false;
```
thc202: It can just use the return value of method exists().
The condition "selectedFile.exists()" in a following if statement can/should be replaced with this variable.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,6 +28,21 @@ public WritableFileChooser(File currentDirectory) {
...
+		}
+		if (checkFile.getUsableSpace() < MINIMUM_SPACE_REQUIREMENT_MB) {
+				int result = JOptionPane.showConfirmDialog(this,
+						Constant.messages.getString("report.write.diskspace.warning.dialog.message"),
+						Constant.messages.getString("report.write.diskspace.warning.dialog.title"),
+						JOptionPane.YES_NO_OPTION);
+				if (result == JOptionPane.NO_OPTION) {
+					return;
+				}
+			}
```
thc202: There's one indentation level that should be removed in the above statements.

thc202:
Just minor comments.

Consider adding a comment to close the issue. [1]

[1] https://help.github.com/articles/closing-issues-via-commit-messages/

mustafa421:
Changes made as requested :) thanks for the feedback. I added the closing comment in my commit message, but doesn't 2127 have 2 pull requests referencing it? Wouldn't this prematurely close the issue if merged?

thc202:
It's still missing the indentation changes (commented in https://github.com/zaproxy/zaproxy/pull/2376#discussion-diff-60186644).

There's just this pull request, the other reference is an issue.

Shouldn't the "fix issue comment" be in the commit message body? In the issue there's no indication that it will be closed by the commit nor the pull request.

mustafa421:
Removed the indented line. I also put the fix issue comment in the message body

thc202:
The other line(s) was fine. I mean the lines (in the current commit) 36 to 43, they are indented as if there was another code block there.

There's still no indication that the issue will be closed... :/
Lets see if it's closed once merged :)

mustafa421:
I think I understand you now. I left-shifted some of the lines of code you mentioned and made sure to mention in the commit message body that this pull request closes issue 2127. I hope I got it right :)

thc202:
Lines 42 to 44 should also be shifted ;)

Yeah, that should be good to close it.

mustafa421:
I can't believe I forgot about that! Nice catch tho :) . All is fixed!

thc202:
Line 41 and 42 were shifted one level too much.

mustafa421:
Done.

src/org/zaproxy/zap/view/widgets/WritableFileChooser.java
```diff
@@ -26,11 +28,26 @@ public WritableFileChooser(File currentDirectory) {
...
+		boolean fileExists = checkFile.exists();
+		if (!fileExists) {
+			checkFile = checkFile.getParentFile();
+		}
+		if (checkFile.getUsableSpace() < MINIMUM_SPACE_REQUIREMENT_MB) {
+			int result = JOptionPane.showConfirmDialog(this,
+					Constant.messages.getString("report.write.diskspace.warning.dialog.message"),
+					Constant.messages.getString("report.write.diskspace.warning.dialog.title"),
+					JOptionPane.YES_NO_OPTION);
+			if (result == JOptionPane.NO_OPTION) {
```
thc202: Maybe this should check that's not `YES_OPTION` instead? That would handle the option `CLOSED_OPTION`, which should behave as `NO_OPTION`.
Thoughts?

mustafa421:
Good idea, forgot that I should cover that case as well. Changed.

thc202:
Looks good to me, thank you!

kingthorin:
Thanks

PR #2378, opened 2016-04-07
Ignore non-public option methods for ZAP API

thc202:
Change the method ApiImplementor.isIgnored(Method) to also ignore
non-public methods, if the method is not public it is not expected to be
called be "external" code, including the ZAP API.

The change prevents the new options related private method,
OptionsParamCheckForUpdates.getSdf(), from being added to the ZAP API.


kingthorin:
Looks good

PR #2381, opened 2016-04-07
Do not allow to save/write to read-only files

thc202:
Change class WritableFileChooser to validate, when the file exists, that
it is writable, otherwise the process could continue and end up in an
exception: "java.io.FileNotFoundException: /file (Permission denied)".
Add a message that indicates that the file is not writable and tweak
existing message (and path used in the warning dialogue) to only mention
directory, which is what was being tested.
Extract a method to show the warning dialogue for the write permissions.


kingthorin:
Looks good

PR #2081, opened 2015-11-25
Issue 1874 : 2 cookies line header

michaelcourcy:
When using "forced user mode" with "Sigle Cookie Request Header" checked
the cookies added by ZAP are always put in a new line. Even if there
was a cookie line already in the request. It resulted in two cookies
line header, for instance :

Cookie: toolbar.collapsed=0; has_js=1
Connection: keep-alive
Content-Length: 0
Host: 172.17.0.3
Cookie: SESS8bcff576a76286b5afa8392e2841f07f=9344OUeBNCCDF8YKkTni2SKSpFuzx5mNulec1n_sXnA

This is not compliant to the RFC 6265 https://tools.ietf.org/html/rfc6265#section-5.4
and to the intent to put all the cookies in a single header line.

PR #2383, opened 2016-04-08
Correct class loading order when compiling tests

thc202:
Change "compile" target to include, first, the core classes in the class
path when compiling the test classes, to force the load of forked
classes.
Change "test" target to include core/test classes first and then the
core and test libraries, for same reason.


thc202:
See also #2081.

kingthorin:
Looks good.

Issue #2380, opened 2016-04-07
Dynamically unload "Access Control Testing" add-on


thc202:
The add-on can not be dynamically unloaded (correctly, see zaproxy/zap-extensions#306), thus requiring a restart of ZAP when uninstalling/updating it.

The following changes need to be done:
- Unload the panels ContextAccessControlPanel, by setting a new model to [`usersComboBox`](https://github.com/zaproxy/zap-extensions/blob/9c3cbc83417de623367390d1899e3612150b039d/src/org/zaproxy/zap/extension/accessControl/view/ContextAccessControlPanel.java#L88) (the model that it has is a core object, which would keep references to the combo box, the panel and ExtensionAccessControl) when [`discardContexts()`](https://github.com/zaproxy/zap-extensions/blob/9c3cbc83417de623367390d1899e3612150b039d/src/org/zaproxy/zap/extension/accessControl/ExtensionAccessControl.java#L440) and [`discardContext(Context)`](https://github.com/zaproxy/zap-extensions/blob/9c3cbc83417de623367390d1899e3612150b039d/src/org/zaproxy/zap/extension/accessControl/ExtensionAccessControl.java#L448) are called, although the latter is not strictly needed, it ensures that the there's no references from core objects;
- When unloading the add-on:
  - Remove the [`ContextDataFactory`](https://github.com/zaproxy/zap-extensions/blob/9c3cbc83417de623367390d1899e3612150b039d/src/org/zaproxy/zap/extension/accessControl/ExtensionAccessControl.java#L122);
  - Remove the [`ContextPanelFactory`](https://github.com/zaproxy/zap-extensions/blob/9c3cbc83417de623367390d1899e3612150b039d/src/org/zaproxy/zap/extension/accessControl/ExtensionAccessControl.java#L128), if it has view.

This issue depends on #2347 to remove `ContextDataFactory` and `ContextPanelFactory`, changes expected to be available in ZAP v2.5.0.

Add-on:
"Access Control Testing" version 1.

thc202:
Fixed in zaproxy/zap-extensions#1521.

Issue #2379, opened 2016-04-07
Dynamically unload "Context Alert Filters" add-on


thc202:
The add-on can not be dynamically unloaded, requiring a restart of ZAP when uninstalling/updating it.

To dynamically unload the add-on the following changes need to be done:
- Unregister the [consumer of events](https://github.com/zaproxy/zap-extensions/blob/257a5e75a217feeea58eb8d298b7ba11ab7dbf83/src/org/zaproxy/zap/extension/alertFilters/ExtensionAlertFilters.java#L176-L178);
- Remove the [`ContextDataFactory`](https://github.com/zaproxy/zap-extensions/blob/257a5e75a217feeea58eb8d298b7ba11ab7dbf83/src/org/zaproxy/zap/extension/alertFilters/ExtensionAlertFilters.java#L188);
- Remove the [`ContextPanelFactory`](https://github.com/zaproxy/zap-extensions/blob/257a5e75a217feeea58eb8d298b7ba11ab7dbf83/src/org/zaproxy/zap/extension/alertFilters/ExtensionAlertFilters.java#L192), if it has view;
- Remove the [`ApiImplementor`](https://github.com/zaproxy/zap-extensions/blob/257a5e75a217feeea58eb8d298b7ba11ab7dbf83/src/org/zaproxy/zap/extension/alertFilters/ExtensionAlertFilters.java#L196).

This issue depends on #2347 to remove `ContextDataFactory` and `ContextPanelFactory`, changes expected to be available in ZAP v2.5.0.

Add-on:
"Context Alert Filters" versions >= 1.

thc202:
Forgot this issue... This was fixed in zaproxy/zap-extensions#939. Released in version 5.

PR #2384, opened 2016-04-08
Allow to hook context factories

thc202:
Change classes ExtensionHook and ExtensionHookView to allow to add
ContextDataFactory and ContextPanelFactory, respectively, so the
factories can be automatically added/removed by core code.
Change ExtensionLoader to add/remove the context factories when the
extensions are initialised/unloaded.

### DIFF
--- a/src/org/parosproxy/paros/extension/ExtensionHook.java
+++ b/src/org/parosproxy/paros/extension/ExtensionHook.java
@@ -25,10 +25,12 @@
 // ZAP: 2013/05/02 Re-arranged all modifiers into Java coding standard order
 // ZAP: 2014/03/23 Issue 1022: Proxy - Allow to override a proxied message
 // ZAP: 2014/10/25 Issue 1062: Added scannerhook to be added by extensions. 
+// ZAP: 2016/04/08 Allow to add ContextDataFactory
 
 package org.parosproxy.paros.extension;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


thc202:
Convenience changes for #2379 and #2380.

src/org/parosproxy/paros/extension/ExtensionHookView.java
```diff
@@ -80,4 +94,32 @@ public void addOptionPanel(AbstractParamPanel panel) {
         return optionPanelList;
     }
     
+    /**
+     * Adds the given {@link ContextPanelFactory} to the view hook, to be latter added to the
```
kingthorin: I think you want later not latter

src/org/parosproxy/paros/extension/ExtensionHook.java
```diff
@@ -189,4 +202,32 @@ public ViewDelegate getView() {
     public List<OverrideMessageProxyListener> getOverrideMessageProxyListenerList() {
         return overrideMessageProxyListenersList;
     }
+
+    /**
+     * Adds the given {@link ContextDataFactory} to the extension hook, to be latter added to the {@link Model}.
```
kingthorin: I think you want later not latter

kingthorin:
Two minor java doc comments, otherwise looks good.

thc202:
Fixed, thank you.

PR #2388, opened 2016-04-11
Remove unused spider's tool bar resource messages

thc202:
The following spider's tool bar resource messages are no longer in use:
 spider.toolbar.site.label
 spider.toolbar.site.select
 spider.toolbar.button.start

First two were superseded by "spider.toolbar.progress.*" resource
messages and the last one is no longer needed, when the class
SpiderPanel was changed to implement ScanPanel2 (per issue #1176).


kingthorin:
Looks good

PR #2389, opened 2016-04-11
Include authorization API in ZAP API clients

thc202:
Change class ApiGeneratorUtils to add the AuthorizationAPI so that it is
automatically generated/included in ZAP API clients (and wiki).
 ---
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/7-DZTINjsuM/discussion


kingthorin:
Looks good

Issue #2311, opened 2016-03-09
Alert Filter Addon: HTML report reports the filtered false positives as true results


hackthem:
Hello guys,

I made several tests and realised that the tagged issues as false positives (through the api call or through the GUI) are reported as true positives.

Can you please deal with this bug ?

Best regards,

Abdellah AZOUGARH

hackthem:
This is a reference to the opened thread : https://groups.google.com/forum/#!topic/zaproxy-develop/PvsBdMvY0-Q

thc202:
How are you asserting that the alerts are tagged as false positive? Through the "Alerts" tab (i.e. shown with green flag)?

psiinon:
All bugs will be dealt with in time .. the question is when ;)
At the moment I'm (probably) the only person paid to work on ZAP (and I'm ridiculously busy) and so much of the work on ZAP is done by volunteers.
Individual bugs typically get addressed because a contributor thinks they are serious (or interesting) enough and are prepared to spend the time looking into them. So its probably better to try to justify why the bug is serious and needs fixing that just telling us it must be fixed. Or you could jump in and have a look at solving it yourself;)
It so happens that I'm planning on using the Context Alert Filters in the near future, in which case I'll have a strong incentive to make sure this does get fixed.
The fact that you also want to put this in the Jenkins plugin is also a good 'selling' point :)

hackthem:
I'm sure that dealing with this one will help by using ZAP Proxy through jenkins plugin. Most of companies will use ZAP Proxy through automated tools as jenkins plugin.
So this is my selling points ;)  : 
 -All the developers to home i presented ZAP Proxy and how to use with jenkins told me that they have to obtain a delta between two or three jobs and need to tag false positives.

-As you know the managers need also the information to see how the issues are dealt with and the only thing we can get when using ZAP as daemon is the report.

I hope you understand the importance of the fix ;).

thc202 : the alerts are shown with green flag .

Thank you all for your help.

Best regards,

Abdellah AZOUGARH

kingthorin:
I'm confused what the issue is here. If I edit alerts and set Confidence as "False Positive" I do not see them in the HTML report. [2.4.3]

What version of ZAP are you using and how are you generating the report?

hackthem:
Okay this is my scenario : 
-I create a context and specify all necessary parameters
-I create a set of false positives (via alert filters) before launching the scan as i identified them before 
-I launch the spidering and the scan 
[...]

PR #2393, opened 2016-04-12
Correctly propagate Alert changes

thc202:
Make sure that the corresponding StructuralNode (Sites tree) of an Alert
exists when adding alerts and that the changes done to an alert are
correctly propagated to the corresponding StructuralNode and
HistoryReference. Also, publish events on Alert changes (so other
extensions can react accordingly).

Following more detailed explanation of the changes done to each class:
 - ExtensionAlert:
  - When adding alerts, alertFound(Alert, HistoryReference):
    - Make sure the Alert instance has the correct message ID by calling
    Alert.setSourceHistoryId(int), previously done in PassiveScanThread;
    - Add the StructuralNode, if not added already, to ensure  that it
    will have the Alert being raised;
    - Remove (now) redundant notification of changes in SiteNode.
  - Extracted a method to publish alert events;
  - Remove method no longer needed, siteNodeChangedEventHandler(...),
  notifications are no longer required (done by HistoryReference);
  - In addAlertToTreeEventHandler(Alert), remove the code that added the
  SiteNode to Sites tree, that's now done (with StructuralNode) in the
  method alertFound(...);
  - When modifying alerts, updateAlert(Alert), update the Alert in the
  corresponding HistoryReference and in the Alerts tree and publish an
  event about the change done;
  - In updateAlertInTree(...), change to return immediately if low
  memory option is set, to prevent adding the alerts to the tree;
  - Change deleteAlert(Alert), to use the extracted method to publish
  the events;
  - In deleteAlertFromDisplayEventHandler(Alert), remove notifications
  of changes and direct manipulation of SiteNode, that's done by
  HistoryReference;
 - ExtensionHistory, change to listen to Alert events to update the
 entries of the table model (PassiveScanThread no longer needs to do
 that when alerts are raised);
 - HistoryReference, change method deleteAlert(Alert) to remove the
 Alert from corresponding SiteNode;
 - SiteNode, change method updateAlert(Alert) to notify of changes;
 - AlertEventPublisher, add new event "alert.changed" and a new field,
 "historyId", that indicates the ID of the message that has the Alert;
 - AlertTreeModel, change method updatePath(...) to not access the EDT
 if the view is not initialised;
 - AlertAddDialog, update logic to no longer add/update the Alert to the
 HistoryReference and no longer notify ExtensionHistory of changes, if
 the ExtensionAlert is available (which takes care to do so);
 - PassiveScanThread, change method raiseAlert(int, Alert), to no longer
 add the Alert to the HistoryReference and no longer notify
 ExtensionHistory of changes, no longer needed as per previous changes;
 - DefaultHistoryReferencesTableEntry, change to cache the highest alert
 of the HistoryReference, the table entry is updated as needed (through
 alert change events);
 - DefaultHistoryReferencesTableModel, add a method to allow to refresh
 all entries (needed when all alerts are deleted).

Core changes for #2311 - HTML report reports the filtered false
positives as true results
 ---
Edited to add the changes of `DefaultHistoryReferencesTableModel`, done in the amended commit to fix the `NullPointerException`.

### DIFF
--- a/src/org/parosproxy/paros/extension/history/ExtensionHistory.java
+++ b/src/org/parosproxy/paros/extension/history/ExtensionHistory.java
@@ -62,6 +62,7 @@
 // ZAP: 2015/07/16 Issue 1617: ZAP 2.4.0 throws HeadlessExceptions when running in daemon mode on headless machine
 // ZAP: 2015/09/16 Issue 1890: ZAP can't completely scan OWASP Benchmark
 // ZAP: 2016/01/26 Fixed findbugs warning
+// ZAP: 2016/04/12 Listen to alert events to update the table model entries
 
 package org.parosproxy.paros.extension.history;
 
@@ -89,6 +90,10 @@ import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HttpMessage;
 import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
+import org.zaproxy.zap.eventBus.EventConsumer;
+import org.zaproxy.zap.extension.alert.AlertEventPublisher;
 import org.zaproxy.zap.extension.help.ExtensionHelp;
 import org.zaproxy.zap.extension.history.AlertAddDialog;
 import org.zaproxy.zap.extension.history.HistoryFilterPlusDialog;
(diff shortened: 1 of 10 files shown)


kingthorin:
Looks good

thc202:
Amended the commit to fix a `NullPointerException` in `ExtensionHistory.AlertEventConsumer` when the event was `ALL_ALERTS_REMOVED_EVENT` (which does not have a `HistoryReference` ID).

PR #2394, opened 2016-04-12
Change API authorization view to wrap its object

thc202:
Change class AuthorizationAPI to wrap the authorization data with an
object (as all views do), called "authorizationDetectionMethod", instead
of returning the authorization data directly, so that the Python ZAP API
client is able to correctly read the data, which expects (and removes)
the wrapper object.

 {"authorizationDetectionMethod":{"statusCode":"-1"..."headerRegex":""}}

instead of:
 {"statusCode":"-1"..."headerRegex":""}

This change breaks existing custom clients that consume the JSON format.
The structure of XML format is unchanged.
 ---
Issue reported in OWASP ZAP User Group: https://groups.google.com/d/msg/zaproxy-users/7-DZTINjsuM/kB9xTJoBDAAJ


kingthorin:
Looks good

Issue #2390, opened 2016-04-12
HeadlessException should be handled more gracefully & README needs Headless details


tschernobyl:
Hi,
I'm trying to use ZAP for the first time on a virtual Fedora machine inside the Qubes OS. According to the README, I should just run the zap.sh, but this doesn't work. Next I tried zap.sh -.daemon (please add that instruction to README) and this time I get a lot further, but still not the whole way.

kingthorin:
There's some info on the Docker page about starting ZAP headless which may help you: https://github.com/zaproxy/zaproxy/wiki/Docker

kingthorin:
Reference: https://groups.google.com/forum/#!topic/zaproxy-users/OEvQgs2kNNw

> [user@ZAP ZAP_2.4.3]$ ./zap.sh

> Found Java version 1.8.0_60
> Available memory: 2901 MB
> Setting jvm heap size: -Xmx512m
> 0 [AWT-EventQueue-0] INFO org.zaproxy.zap.GuiBootstrap  - OWASP ZAP 2.4.3 started.
> 108 [AWT-EventQueue-0] ERROR org.zaproxy.zap.ZAP$UncaughtExceptionLogger  - Exception in thread "AWT-EventQueue-0"
> java.awt.HeadlessException
>     at java.awt.GraphicsEnvironment.checkHeadless(GraphicsEnvironment.java:204)
>     at java.awt.Window.<init>(Window.java:536)
>     at java.awt.Frame.<init>(Frame.java:420)
>     at java.awt.Frame.<init>(Frame.java:385)
>     at javax.swing.JFrame.<init>(JFrame.java:189)
>     at org.parosproxy.paros.view.AbstractFrame.<init>(Unknown Source)
>     at org.zaproxy.zap.view.LicenseFrame.<init>(Unknown Source)
>     at org.zaproxy.zap.GuiBootstrap.showLicense(Unknown Source)
>     at org.zaproxy.zap.GuiBootstrap.startImpl(Unknown Source)
>     at org.zaproxy.zap.GuiBootstrap.access$000(Unknown Source)
>     at org.zaproxy.zap.GuiBootstrap$1.run(Unknown Source)
>     at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
>     at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:756)
>     at java.awt.EventQueue.access$500(EventQueue.java:97)
[...]

PR #2395, opened 2016-04-12
Inform if GUI is started on headless environment

thc202:
Change class GuiBootStrap to check and inform if ZAP (GUI) is being
started in a headless environment (preventing the HeadlessException).

Code changes for #2390 - HeadlessException should be handled more
gracefully & README needs Headless details


kingthorin:
Looks good

psiinon:
Worth mentioning that you can also use technologies like XVFB?

thc202:
Would that allow to run/interact with ZAP's GUI? Or, it's for effective use of other add-ons (e.g. Selenium)?

psiinon:
Dont know if you can interact with the GUI, but it does allow you to run add-ons which use Selenium which is very uiseful in a headless environment.

thc202:
OK, I'll try add something about that.

thc202:
Added a note about the use of Xvfb.

kingthorin:
Looks good

PR #2403, opened 2016-04-13
Allow Active Rules to have just one scan method

### LINKED ISSUE #1786: Activescan Scripts Silent Failure
ZAP 2.4.0, ScriptConsole v14.

Create a new javascript active scan script based on the template. Set a scan policy that runs only active scan scripts. Scan a single static page (index.html).

In scanNode just add a println (below I just print fred)...
``` javascript
function scanNode(as, msg) {
	// Debugging can be done using println like this
	println('scan called for url=' + msg.getRequestHeader().getURI().toString());

	// Copy requests before reusing them
	msg = msg.cloneRequest();
	
	// sendAndReceive(msg, followRedirect, handleAntiCSRFtoken)
	as.sendAndReceive(msg, false, false);
	println("Fred");
	// Test the responses and raise alerts as below
}

function scan(as, msg, param, value) {
	// Debugging can be done using println like this
	println('scan called for url=' + msg.getRequestHeader().getURI().toString() + 
		' param=' + param + ' value=' + value);
	
	// Copy requests before reusing them
	msg = msg.cloneRequest();
	
	// setParam (message, parameterName, newValue)
	as.setParam(msg, param, 'Your attack');
	
	// sendAndReceive(msg, followRedirect, handleAntiCSRFtoken)
	as.sendAndReceive(msg, false, false);
	
	// Test the response here, and make other requests as required
	if (true) {	// Change to a test which detects the vulnerability
		// raiseAlert(risk, int confidence, String name, String description, String uri, 
		//		String param, String attack, String otherInfo, String solution, String evidence, 
		//		int cweId, int wascId, HttpMessage msg)
		// risk: 0: info, 1: low, 2: medium, 3: high
		// confidence: 0: falsePositive, 1: low, 2: medium, 3: high, 4: confirmed
		as.raiseAlert(1, 1, 'Active Vulnerability title', 'Full description', 
		msg.getRequestHeader().getURI().toString(), 
			param, 'Your attack', 'Any other info', 'The solution ', '', 0, 0, msg);
	}
}
```

Run an active scan and note things behave properly.

Remove the scan function completely (thinking you don't need it):
``` javascript
function scanNode(as, msg) {
	// Debugging can be done using println like this
	println('scan called for url=' + msg.getRequestHeader().getURI().toString());

	// Copy requests before reusing them
	msg = msg.cloneRequest();
	
	// sendAndReceive(msg, followRedirect, handleAntiCSRFtoken)
	as.sendAndReceive(msg, false, false);
	println("Fred");
	// Test the responses and raise alerts as below
}
```
Run an active scan and note nothing happens, and there is no error. (I don't think it even issues a request.)

works
fails

thc202:
Change interface ActiveScript2 to not extend ActiveScript, to allow
Active Rules (scripts) to implement just one of the scan methods.
Change class ScriptsActiveScanner to warn if the scripts do not
implement any of the interfaces (the warn is only issued if the message
being scanned contains "parameters", thus requiring both interfaces).

Fix #1786 - Activescan Scripts Silent Failure

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ActiveScript2.java
+++ b/src/org/zaproxy/zap/extension/ascan/ActiveScript2.java
@@ -23,7 +23,7 @@ import javax.script.ScriptException;
 
 import org.parosproxy.paros.network.HttpMessage;
 
-public interface ActiveScript2 extends ActiveScript {
+public interface ActiveScript2 {
 
 	void scanNode(ScriptsActiveScanner sas, HttpMessage msg) throws ScriptException;
 }
--- a/src/org/zaproxy/zap/extension/ascan/ScriptsActiveScanner.java
+++ b/src/org/zaproxy/zap/extension/ascan/ScriptsActiveScanner.java
@@ -19,8 +19,10 @@
 package org.zaproxy.zap.extension.ascan;
 
 import java.io.IOException;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Set;
 
 import org.apache.commons.httpclient.HttpException;
 import org.apache.log4j.Logger;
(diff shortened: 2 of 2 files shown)


kingthorin:
Looks good

PR #2406, opened 2016-04-14
Change ProxyThread to write the response "later"

thc202:
Change method ProxyThread.processHttp(HttpRequestHeader, boolean) to
forward/write the HTTP response to the client in another catch block,
after reading/writing the request and reading the response, to prevent
exceptions that might happen while reading/writing the request or
reading the response from being mixed with the exceptions that happen
while writing the response, moreover, it does not attempt to write the
response twice (the actual writing and ~~in the catch blocks~~ the write in
the catch blocks) which, most likely, would lead to another exception, resulting
in two notifications.


kingthorin:
Change looks good. Is there a word missing from this part of the commit message?

> the actual writing and in the catch blocks

thc202:
No? :) I guess it could be "the actual writing and the write in the catch blocks", does that sound better?

kingthorin:
Now that I've read that I'll agree it can read the way you're intending...I do think your revision is more clear though....it's a commit message no big deal.

Looks good to me.

thc202:
Commit message amended.

PR #2407, opened 2016-04-14
Change View to allow to display a message

thc202:
Change interface ViewDelegate to allow to display a Message, through the
method displayMessage(Message), to reduce code duplication and
standardise how the messages are displayed.
Change class View to implement the method, current implementation just
supports HttpMessage (the only Message supported by core).
Change all classes to use the new method, instead of manually setting
the message to request/response panels.


kingthorin:
Looks good

PR #2408, opened 2016-04-15
Set focus back to table after displaying a message

thc202:
Change class HistoryReferencesTable.DisplayMessageOnSelectionValueChange
to request focus back to the table after displaying a message to allow
navigate the table entries with the keyboard more easily. The table
could lose focus if the message panel displayed a big request/response
body, requiring the user to transfer the focus through all the
components back to the table.


kingthorin:
Looks good

PR #2409, opened 2016-04-15
Use tool bar separators instead of vertical bars

thc202:
Change classes ActiveScanPanel and SpiderPanel to use tool bar
separators to separate the labels that show the number of requests sent
and number of URLs found, respectively, from the existing labels in the
tool bar (and remove hardcoded border space from one of them). Also, set
those labels to zero when initialised/reset instead of leaving them
empty.
Remove the vertical bars (and unnecessary extra whitespaces) from the
resource messages used in the labels.


kingthorin:
Looks good

Issue #2354, opened 2016-03-28
Add "diff" functionality to scan results to assist in finding vulnerability specifics


wadeschofield:
Currently, when running a SQL Injection scan, ZAP claims that there is a difference when a normal request is send vs when an "injection" request is sent. The problem is that it is very difficult to find this difference. Providing a "diff" or even a line/index indicating the beginning of the difference would be extremely helpful for decoding scan results.

kingthorin:
Ref: https://groups.google.com/forum/m/#!topic/zaproxy-users/6TqI5_XHab8

psiinon:
We do have diff functionality, but maybe its not available where you need it.
To use it select 2 lines from the History tab (or other ones like it), right click and "Compare 2 responses" (or requests). 
So .. where would you like that?

thc202:
The problem is that the alert just contains one message, we need to either change core to allow to raise the alert with all messages used to assert the issue (which would allow to "manually" diff the messages) or change the scanners to add the diff to the alert.

psiinon:
We should probably change the core to support multiple messages, although that will probably be non trivial :/ I've been meaning to add arbitrary binary blobs to alerts for ages, eg things like screenshots and Zest scripts. In theory Zest scripts could handle the multiple message problem, but I suspect it would be cleaner to support both. Thoughts?

thc202:
Yeah, both, it's (probably) better in the long run.

wadeschofield:
The problem with the history tab is that it doesn't have the injection
alerts. It places a flag with the packets for which there are injection
alerts. I.e. I don't see how I can select a "normal" and an "injected"
response message and do a compare.

Now I have manually saved a number of the flagged messages using the "save
as raw" option and manually compared them using a text editor. The only
difference between them is the keep-alive value in the response header.
Could ZAP be flagging based on that?

thc202:
SQL Injection scanner seems to compare just the response body.

What's exactly the name of the alert and the other info? (the other info can be obfuscated)

wadeschofield:
The alert info is:
SQL Injection
Risk: High
Confidence: Medium
Parameter: dataTimeStamp
Attack: 14603779660575' OR '1'='1'
Evidence:
CWE ID: 89
WASC ID: 19

In scanning our application, we picked up 10 of these. I've manually
compared half of them in a text editor and the only difference is the max=
parameter in the http response.

It seems to me this is a false positive but it would give me more
[...]

PR #2413, opened 2016-04-18
Disable "Show in History Tab" if message not valid

thc202:
Disable the menu item "Show in History Tab" if the type of the selected
message is not valid for the History tab (only manual or proxied
requests are shown in that tab). Prevents the menu item from being
enabled for other message types (for example, scanner, spider) which
would have no effect other than showing/focusing the History tab.
 ---
From #2354.


kingthorin:
Looks good

PR #2416, opened 2016-04-19
Remove unnecessary log warnings

thc202:
Change method ExtensionGlobalExcludeURL.optionsLoaded() to remove the
existing log warning, not needed.
Change method OptionsGlobalExcludeURLPanel.saveParam(Object) to remove
the log warning (by moving the logged message, the ignored regular
expressions, to existing debug log statement and wrap it with a
"isDebugEnabled"), as with the previous warning it's not needed (as
warning).
 ---
From https://github.com/jenkinsci/zaproxy-plugin/issues/19#issuecomment-211820358.


kingthorin:
Looks good

PR #2418, opened 2016-04-20
Spider: support maximum duration in mins option

psiinon:
Associated help changes will be in another PR


src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -672,6 +678,13 @@ public boolean isPaused() {
 	 * @return true, if is stopped
 	 */
 	public boolean isStopped() {
+		if (! stopped && this.spiderParam.getMaxDuration() > 0) {
+			// Check to see if the scan has exceeded the specified maxDuration
+			if ((new Date().getTime() - this.timeStarted.getTime()) > (this.spiderParam.getMaxDuration() * 60000)) {
```
kingthorin: Would it be better to use TimeUnit.MILLISECONDS.toMinutes here?

psiinon:
Yeah, should have guessed there would be something useful like that. Fixed.

kingthorin:
One more quick question (because I'm not in a position to checkout the code and try it right now): What happens if someone enters their duration as something greater than `Integer.MAX_VALUE` ? (That might be handled elsewhere in ZAP or somewhere that isn't shown in the changes view).

src/lang/Messages.properties
```diff
@@ -1773,8 +1773,9 @@ spider.custom.button.reset	= Reset
 spider.custom.button.scan	= Start Scan
 spider.custom.label.adv		= Show Advanced options
 spider.custom.label.context	= Context:
-spider.custom.label.maxChildren		= Maximum children to crawl:
-spider.custom.label.maxDepth		= Maximum depth to crawl:
+spider.custom.label.maxChildren		= Maximum children to crawl (0 is unlimited):
+spider.custom.label.maxDepth		= Maximum depth to crawl (0 is unlimited):
```
thc202: In the UI it does not allow to select 0, the minimum is 1 (the actual check expects a value greater than 0).

src/lang/Messages.properties
```diff
@@ -1801,7 +1802,7 @@ spider.custom.targetNotInScope.error = The following target is not allowed in 'P
 spider.desc                     = Spider used for automatically finding URIs on a site
 spider.label.inScope            = URI found during crawl:
 spider.label.outOfScope         = URI found but out of crawl scope:
-spider.options.label.depth      = Maximum depth to crawl:
+spider.options.label.depth      = Maximum depth to crawl (0 is unlimited):
```
thc202: Same comment as above.

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
```diff
@@ -254,6 +267,13 @@ private JSlider getSliderThreads() {
 		}
 		return sliderThreads;
 	}
+	
+	private ZapNumberSpinner getDurationNumberSpinner() {
+		if (durationNumberSpinner == null) {
+			durationNumberSpinner = new ZapNumberSpinner(0, 60, Integer.MAX_VALUE);
```
thc202: Maybe the default should be 0, to be consistent with the default of the persisted option?

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -116,6 +117,9 @@
 
 	/** The scan user. */
 	private User scanUser;
+	
+	/** The time the scan was started */
+	private Date timeStarted;
```
thc202: It could use `long` instead of `Date`.

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -423,6 +427,8 @@ protected synchronized void submitTask(SpiderTask task) {
 	public void start() {
 
 		log.info("Starting spider...");
+		
+		this.timeStarted = new Date();
```
thc202: If changed to long, `System.currentTimeMillis()` instead of `new Date()`.

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -672,6 +678,14 @@ public boolean isPaused() {
 	 * @return true, if is stopped
 	 */
 	public boolean isStopped() {
+		if (! stopped && this.spiderParam.getMaxDuration() > 0) {
+			// Check to see if the scan has exceeded the specified maxDuration
+			if (TimeUnit.MILLISECONDS.toMinutes(new Date().getTime() - this.timeStarted.getTime()) >
```
thc202: Could use `System.currentTimeMillis()` instead of `new Date().getTime()`.

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -672,6 +678,14 @@ public boolean isPaused() {
 	 * @return true, if is stopped
 	 */
 	public boolean isStopped() {
+		if (! stopped && this.spiderParam.getMaxDuration() > 0) {
+			// Check to see if the scan has exceeded the specified maxDuration
+			if (TimeUnit.MILLISECONDS.toMinutes(new Date().getTime() - this.timeStarted.getTime()) > 
+					this.spiderParam.getMaxDuration()) {
+				log.info("Spidering process has exceeded maxDuration of " + this.spiderParam.getMaxDuration() + " minute(s)");
+				this.stop();
```
thc202: It should not call `stop()` directly because `stop()` tries to wait for the worker threads to finish, which are one of the threads calling this method.
It could call `complete()`, if changed to specify if "successful" when notifying the listeners.

src/org/zaproxy/zap/spider/SpiderParam.java
```diff
@@ -152,6 +153,8 @@ public String getName() {
 	private HandleParametersOption handleParametersVisited = HandleParametersOption.USE_ALL;
 	/** Defines if we take care of OData specific parameters during the visit in order to identify known URL **/
 	private boolean handleODataParametersVisited = false;
+	/** The maximum duration in minutes that the spider is allowed to run for, 0 meaning no limit */
+	private int maxDuration = 0;
```
thc202: It's missing the read from the configuration file, in parse() method?

thc202:
@kinthoring the UI component does not allow values greater than the maximum defined (either resets the current value to an acceptable value or does nothing). Does it answer the question?

kingthorin:
Perfect, I figured something like that :smile:

Thanks!

psiinon:
Hopefully fixed all of the comments...

thc202:
Looks good to me, thank you.

Issue #2308, opened 2016-03-08
Support Python 3.5 in API


psiinon:
As per https://groups.google.com/d/msg/zaproxy-users/pWaukbxtG-M/9OQz2__8CQAJ
Known changes required: 
- print statements to use parentheses

thc202:
Also:
https://groups.google.com/d/topic/zaproxy-users/JxUJpC2wz5k/discussion

Completely rewrite how the requests (to ZAP API) are done?
https://docs.python.org/3/library/urllib.request.html

you1anna:
@thc202 I hope we can get a fix for this. Our project is eager to use Zap for automated testing, however our build servers (and auto test suites) all use Python 3. Do you estimate the changes required for Urllib to be extensive?

psiinon:
Quick update to let you know we're looking into this.
There arent many print statements, so those seem pretty straightforward.
The requests are another matter - I'm gradually solving each problem as I hit them, but its slow going :/

Woolworths:
+1 just saw this after I submitted the request

jitendriya09:
Hi
Is this issue fixed?
I checked the link "Improved Python API library #2419" above
It was said that support for python3 is added.

but while trying to install the zap API(python api) i am facing same issue
pip install python-owasp-zap-v2.4

below is the error message i am getting.
Downloading/unpacking python-owasp-zap-v2.4
  Cannot fetch index base URL https://pypi.python.org/simple/
  Downloading python-owasp-zap-v2.4-0.0.8.tar.gz
  Running setup.py (path:/tmp/pip-build-l7azoeh2/python-owasp-zap-v2.4/setup.py) egg_info for package python-owasp-zap-v2.4
    Traceback (most recent call last):
      File "<string>", line 17, in <module>
      File "/tmp/pip-build-l7azoeh2/python-owasp-zap-v2.4/setup.py", line 12
        print "You must have setuptools installed to use setup.py. Exiting..."
                                                                             ^
    SyntaxError: Missing parentheses in call to 'print'
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):

  File "<string>", line 17, in <module>

  File "/tmp/pip-build-l7azoeh2/python-owasp-zap-v2.4/setup.py", line 12

    print "You must have setuptools installed to use setup.py. Exiting..."

                                                                         ^

SyntaxError: Missing parentheses in call to 'print'

----------------------------------------
Cleaning up...
[...]

PR #2419, opened 2016-04-20
Improved Python API library

cr0hn:
Add: support for Python 3 and some improvements. Now the API runs in Python 2 and 3
Add: high performance json library to improve coding/encoding
Add: request library to simplify handling of proxies
Add: Transparent transparent in-memory cache for queries
Fix: setup.py - Now the dist generation, dependecies and upload to pypi is more simple.
Imp: Improved source code to be more PEP-8 compilant
Imp: Changes in code to be more "pythonic". Some parts of code smells Java :)

### DIFF
--- a/python/api/MANIFEST.in
+++ b/python/api/MANIFEST.in
@@ -1,0 +1,15 @@
+include LICENSE requirements.txt
+
+recursive-exclude * __pycache__
+recursive-exclude * *.pyc
+recursive-exclude * *.pyo
+recursive-exclude * *.orig
+recursive-exclude * .DS_Store
+global-exclude __pycache__/*
+global-exclude .deps/*
+global-exclude *.so
+global-exclude *.pyd
+global-exclude *.pyc
+global-exclude .git*
+global-exclude .DS_Store
+global-exclude .mailmap
--- a/python/api/requirements.txt
+++ b/python/api/requirements.txt
@@ -1,0 +1,4 @@
+six
+ujson
+requests
+requests-cache
--- a/python/api/setup.py
+++ b/python/api/setup.py
@@ -1,22 +1,30 @@
 #!/usr/bin/env python
+
 """
 Standard build script.
 """
 
-__docformat__ = 'restructuredtext'
+from __future__ import print_function
 
+from os.path import dirname, join
 
 try:
     from setuptools import setup, find_packages
 except ImportError:
-    print "You must have setuptools installed to use setup.py. Exiting..."
+    print("You must have setuptools installed to use setup.py. Exiting...")
     raise SystemExit(1)
 
+__docformat__ = 'restructuredtext'
+
+# Import requirements
+with open(join(dirname(__file__), 'requirements.txt')) as f:
+    required = f.read().splitlines()
 
 setup(
     name="python-owasp-zap-v2.4",
-    version="0.0.7",
+    version="0.0.8",
     description="OWASP ZAP 2.4 API client",
+    install_requires=required,
     long_description="OWASP Zed Attack Proxy 2.4 API python client",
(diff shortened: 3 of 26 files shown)


psiinon:
Ha! Would have been good to mention you were looking at this - I've also been changing the API to support Python 2 and 3! And its been taking me too long as I'm no python expert ;)
Unfortunately you've manually changed a load of files marked "This file was automatically generated." - you need to change the generator file: https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/api/PythonAPIGenerator.java - thats the one that creates the generated files.
But you changes sound really good :)
You feel up to changing the generator?

cr0hn:
I'm starting to use the Python library for a personal project, and probably I'll include more improvements (like coroutines in request). 

Ups, I doesn't knew you're improving the library :(

I can try to take a look to the generator, but I'm not a Java developer. I'll see the code and I say something.

psiinon:
See #2308 
I _am_ a java dev, so I can help with that :) 
Changing __init__.py is fine.
Let me know if you have problems understanding / changing the generator, I wront that so between us we should be able to get it all working correctly:)

thc202:
@psiinon a good opportunity to move the Python API to other repo?

cr0hn:
Seen the issue :)

The patch I sent, made the code runs in Python 2 and 3. And, coming soon with new features of Python 3.4 and greater.

Taking a look to the code, I understand that the main function of it is the generation of end points, right?

Maybe we can adopt an hybrid approach, generating with the generator the end points, and using the rest .py as static. Doing so, I can send more updated without change the generator code.

If you like the idea, the changes in the endpoint generation are:

**Import Six**
Importing six library in each file:

`import six`

**Iteritems**
`return next(....).itervalues())`
to
`return next(six.itervalues(...))`

Ensuring the compatibility with Python 2 & 3

**Remove parenthesis**
` return (self.zap._request_other(...))`
to
` return self.zap._request_other(...)`

The parenthesis here are redundant.

The rest of modifications are to improve PEP-8, but it's not modify the behavior of code, so it could be as the generator generates.

psiinon:
@cr0hn exactly - the API is defined via Java classes, so we can use code generation to create clients in java, python etc etc. Ok, so we could have used Swagger or something like that, but the API _was_ defined a while ago ;)
The more code we can move into common static classes the better imo - so your suggested approach sounds good to me :)

psiinon:
@cr0hn - is there a std format for python projects? I like @thc202 's suggestion of moving the python API to a new repo now as it will make everything cleaner and easier to manage.

cr0hn:
If we move the code to the new project I also think It could be more clean. I suppose the Zap API is stable and It will not change often. In this case this approach has more sense for me also.

@psiinon In Python the std defined in [PEP-8](https://www.python.org/dev/peps/pep-0008/)

psiinon:
https://github.com/zaproxy/zap-api-python :)
Although I cant actually see the recommendations on directory structure in that link :/
You can submit a PR for the static code whenever its ready - we'll generate the generated code once the generator has been correctly updated (which again, I can help with)

cr0hn:
Ah, ok ok, I don't understood. There are not a STD for structure. The most common are some similar to:

http://docs.python-guide.org/en/latest/writing/structure/

If you want, I can structure the project and send a new update I don't mind. I'll send a PR with the static code.

psiinon:
That would be great :)

cr0hn:
Done :)

kingthorin:
Conflicts need resolution.

kingthorin:
@cr0hn's contributions now in https://github.com/zaproxy/zap-api-python/pulls/cr0hn

PR #2424, opened 2016-04-21
Fix issues related to active scanner request count

### LINKED ISSUE #2420: Allow active scanners to notify that a message was sent
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/AlT3gpxale8/discussion
 ---
Allow active scanners to notify that a message was sent, so that scanners that send other messages than HTTP (like "Heartbleed OpenSSL Vulnerability") can also be counted and reported to the user.

### LINKED ISSUE #2421: Active scanner request count mismatch
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/AlT3gpxale8/discussion
 ---
Steps to reproduce the issue:
1. Run ZAP with "Active scanner rules (beta)" add-on (v19) installed (easier to reproduce the issue);
2. Change the "Default Policy" to disable all scanners except "Heartbleed OpenSSL Vulnerability", "Source Code Disclosure - /WEB-INF folder" and "Cross-Domain Misconfiguration";
3. Start the active scan against a target;
4. Once finished see the number of requests sent by each scanner (using the GUI, "Scan Progress" dialogue, or the ZAP API, <http://zap/HTML/ascan/view/scanProgress/?scanId=0>) and note that the count does not correspond to the messages sent by each scanner. For example, in the attached screen capture, the scanners "Source Code Disclosure - /WEB-INF folder" and "Cross-Domain Misconfiguration" have 0 messages while "Heartbleed OpenSSL Vulnerability" has 9, but from the messages sent:
```
http://example.com/WEB-INF/web.xml
http://example.com/WEB-INF/applicationContext.xml
http://example.com/WEB-INF/classes/2/0.class
http://example.com/WEB-INF/classes/web/xml.class
http://example.com/WEB-INF/classes/2/4/7.class
http://example.com/WEB-INF/classes/example/com.class
http://example.com/WEB-INF/classes/applicationContext/xml.class
http://example.com/crossdomain.xml
http://example.com/clientaccesspolicy.xml
```
the scanner "Source Code Disclosure - /WEB-INF folder" sent 7 and "Cross-Domain Misconfiguration" sent 2.

Remarks;
Messages sent during the analysis step, before any active scanner has been run, will be counted as sent by the first scanner run.

ZAP Versions:
Version >= 2.4.3.
 ---
Request Count Mismatch

thc202:
Allow scanners to notify that a (non-HTTP) message was sent and change
HostProcess to keep track of the requests sent (that is, notified) by
each scanner, so that the request count is more accurate.

More detailed changes:
 - HostProcess:
   - Add methods notifyNewMessage(Plugin plugin, HttpMessage), to know
   which scanner sent the HTTP message, and notifyNewMessage(Plugin),
   for the same purpose but for non-HTTP messages;
   - Add method getRequestCount(), to obtain the request count (of all
   scanners and Analyser);
   - Refactor the management of the stats collected from each scanner
   (start time, progress and request count), by using a nested class
   (PluginStats) and replacing the existing maps (plugin ID to each
   stat) with only one (plugin ID to PluginStats).
   - Change processPlugin(Plugin) to initialise the PluginStats of the
   scanner being processed;
   - Change scanSingleNode(Plugin, StructuralNode) to increase the
   progress of Plugin;
   - Change getTestCurrentCount(Plugin), getPluginRequestCount(int) and
   pluginCompleted(Plugin) to use the new stats map/class;
   - Deprecate setTestCurrentCount(Plugin, int), no longer needed the
   progress is now maintained internally;
   - Deprecate setPluginRequestCount(int, int), no longer needed the
   request count is now maintained internally;
 - AbstractAppPlugin, remove the manual setting of Plugin's progress, no
 longer needed, HostProcess takes care of keeping track of progress;
 - AbstractPlugin, change to notify the HostProcess with Plugin itself,
 so the requests sent are correctly counted;
 - Analyser, change to allow to obtain the request count, included in
 the total request count of the HostProcess;
 - ActiveScan, change to obtain the request counts from the HostProcess
 instead of calculating manually (and deprecate the method that was
 doing that);
 - ActiveScanAPI and ScanProgressTableModel, remove calls to ActiveScan
 to update the request counts.

Fix #2420 - Allow active scanners to notify that a message was sent
Fix #2421 - Active scanner request count mismatch

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/HostProcess.java
+++ b/src/org/parosproxy/paros/core/scanner/HostProcess.java
@@ -55,6 +55,7 @@
 // ZAP: 2015/11/27 Issue 2086: Report request counts per plugin
 // ZAP: 2015/12/16 Prevent HostProcess (and plugins run) from becoming in undefined state
 // ZAP: 2016/01/27 Prevent HostProcess from reporting progress higher than 100%
+// ZAP: 2016/04/21 Allow scanners to notify of messages sent (and tweak the progress and request count of each plugin)
 
 package org.parosproxy.paros.core.scanner;
 
@@ -98,16 +99,17 @@ public class HostProcess implements Runnable {
     private User user = null;
     private TechSet techSet = null;
 
-    // time related 
-    // ZAP: changed to Integer because the pluginId is int
-    private final Map<Integer, Long> mapPluginStartTime = new HashMap<>();
+    /**
+     * A {@code Map} from plugin IDs to corresponding {@link PluginStats}.
+     * 
+     * @see #processPlugin(Plugin)
+     */
+    private final Map<Integer, PluginStats> mapPluginStats = new HashMap<>();
     private final Set<Integer> listPluginIdSkipped = new HashSet<>();
     private long hostProcessStartTime = 0;
 
     // ZAP: progress related
     private int nodeInScopeCount = 0;
-    private final Map<Integer, Integer> mapPluginProgress = new HashMap<>();
-    private final Map<Integer, Integer> mapPluginReqCounts = new HashMap<>();
     private int percentage = 0;
     
     /**
@@ -217,8 +219,9 @@ public class HostProcess implements Runnable {
     }
 
     private void processPlugin(final Plugin plugin) {
-        mapPluginStartTime.put(plugin.getId(), System.currentTimeMillis());
-        mapPluginProgress.put(plugin.getId(), 0);
+        synchronized (mapPluginStats) {
+            mapPluginStats.put(plugin.getId(), new PluginStats());
+        }
 
         if (techSet != null && !plugin.targets(techSet)) {
             listPluginIdSkipped.add(plugin.getId());
@@ -387,6 +390,7 @@ public class HostProcess implements Runnable {
             
         } while (thread == null);
 
+        mapPluginStats.get(plugin.getId()).incProgress();
         return true;
     }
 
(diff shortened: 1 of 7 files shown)


src/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -404,16 +408,21 @@ public int getTestTotalCount() {
...
+            return 0;
+        }
+        return pluginStats.getProgress();
     }
 
     /**
-     * ZAP: method to set the current progress status for a specific plugin
-     * @param plugin the plugin we're setting the progress
-     * @param value the value that need to be set
+     * @deprecated (TOD add version) No longer used/needed, Plugin's progress is automatically updated/maintained by
```
kingthorin: Typo on TODO

thc202: Fixed, thank you.

src/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -646,18 +701,50 @@ public String getHostAndPort() {
 		return this.hostAndPort;
 	}
 	
+	/**
+	 * @deprecated (TOD add version) No longer used/needed, Plugin's request count is automatically updated/maintained by
```
kingthorin: Typo on TODO

thc202: Also fixed, thanks.

psiinon:
Looks good

PR #2430, opened 2016-04-22
Fix memory leak on session creation/loading

### LINKED ISSUE #2428: Memory leak on session creation/loading
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/R9CAhLwa37c/discussion
 ---
Steps to reproduce the issue:
1. Start ZAP;
2. Create a new session several times (>= 5 to be easier to spot the issue);
3. Create a heap dump, analyse it (for example, using Java VisualVM) and note that the instance variable `listenerList` of `ParosDatabase` has many instances of `ClientTable`, `MessageTable` and `TableWebSocket`, one per session opened (the exact types depend on the installed add-ons);
4. The more sessions are created the more items the `listenerList` has (and the more memory is used by `ParosDatabase`).

ZAP:
Versions >= 2.4.0

thc202:
Remove the database listeners added by add-ons, once the database is
closed (database listeners are added by add-ons when a new database is
opened, so removing when closed ensures that there are no more database
listeners than it is necessary and preventing the memory leak).

More detailed changes:
 - Control, change to notify that a database was opened before notifying
 that the session was changed, some add-ons might need a database
 connection to handle the changed session (not a problem before because
 several database listeners were kept in memory);
 - Add AbstractDatabase class to manage the DatabaseListeners and to
 remove them when the database is closed;
 - Change ParosDatabase, HsqlbDatabase and SqlDatabase to use/extend the
 class AbstractDatabase.

Fix #2428 - Memory leak on session creation/loading

### DIFF
--- a/src/org/parosproxy/paros/control/Control.java
+++ b/src/org/parosproxy/paros/control/Control.java
@@ -58,6 +58,7 @@
 // ZAP: 2015/09/17 Issue 1914: Support multiple add-on directories
 // ZAP: 2015/11/04 Issue 1920: Report the host:port ZAP is listening on in daemon mode, or exit if it cant
 // ZAP: 2016/03/23 Issue 2331: Custom Context Panels not show in existing contexts after installation of add-on
+// ZAP: 2016/04/22 Issue 2428: Memory leak on session creation/loading
 
 package org.parosproxy.paros.control;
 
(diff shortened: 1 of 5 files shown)


kingthorin:
Looks good

PR #2435, opened 2016-04-25
Clean up spider task resources, when not consumed

thc202:
Clean up spider tasks' resources (persisted HTTP message) in all cases,
when it is and it isn't consumed. If the spider is stopped with pending
tasks it would leave a HTTP message per task in the database that would
not be deleted, ever.

Changes:
 - SpiderTask:
  - Change method run() to delete the message when stopped before
  consuming the message, also reorder the statements to prevent a
  (potential) NullPointerException when logging debug data;
  - Extracted a method, deleteHistoryReference(), to delete the
  persisted message, called when the message is consumed and when not;
  - Add a method to allow to cleanup its resources (called by Spider
  class);
 - Spider, call the method SpiderTask.cleanup() when the tasks are not
 executed.


kingthorin:
Looks good

PR #2441, opened 2016-04-27
Reduce the number of session/db notifications

thc202:
Reduce the number of session/db notifications (sessionChangedAllPlugin,
sessionAboutToChangeAllPlugin and databaseOpen) when creating a new
session by creating the database at the same time as the session (which
no longer requires calling the previously mentioned methods twice).

More detailed changes:
 - Control:
  - Change methods newSession() and newSession(String, SessionListener)
  to also create the empty database and no longer notify of
  sessionAboutToChangeAllPlugin (done by new method);
  - Deprecate the method createAndOpenUntitledDb(), no longer needs to
  be called by external code;
  - Add helper method that creates the database and only notifies of
  sessionAboutToChangeAllPlugin;
 - CoreAPI and MenuFileControl, remove the calls to the method
 Control.createAndOpenUntitledDb().

### DIFF
--- a/src/org/parosproxy/paros/control/Control.java
+++ b/src/org/parosproxy/paros/control/Control.java
@@ -363,9 +363,9 @@ public class Control extends AbstractControl implements SessionListener {
 		// The session is opened in a thread, so notify the listeners via the callback
     }
 
-	public Session newSession() {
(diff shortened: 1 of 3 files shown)


kingthorin:
Looks good

PR #2442, opened 2016-04-27
Fix exception while opening Active Scan dialogue

### LINKED ISSUE #2440: Exception while opening the Active Scan dialogue
Steps to reproduce the issue:
1. Run ZAP with a new home directory (-dir command line argument);
2. Open the Active Scan dialogue;
3. Check the log file, there should be an exception:
```
30261 [AWT-EventQueue-0] ERROR org.parosproxy.paros.common.AbstractParam  - 'view.locale' doesn't map to a String object
org.apache.commons.configuration.ConversionException: 'view.locale' doesn't map to a String object
	at org.apache.commons.configuration.AbstractConfiguration.getString(AbstractConfiguration.java:1050)
	at org.parosproxy.paros.extension.option.OptionsParamView.parse(Unknown Source)
	at org.parosproxy.paros.common.AbstractParam.load(Unknown Source)
	at org.parosproxy.paros.model.OptionsParam.parse(Unknown Source)
	at org.parosproxy.paros.common.AbstractParam.load(Unknown Source)
	at org.zaproxy.zap.extension.ascan.CustomScanDialog.reset(Unknown Source)
	at org.zaproxy.zap.extension.ascan.CustomScanDialog.<init>(Unknown Source)
	at org.zaproxy.zap.extension.ascan.ExtensionActiveScan.showCustomScanDialog(Unknown Source)
	at org.zaproxy.zap.extension.ascan.ExtensionActiveScan$3.actionPerformed(Unknown Source)
```

ZAP Versions:
Versions >= 2.4.0

thc202:
Change method OptionsParamView.setLocale(Locale) to save the locale as
String instead of Locale (call existing method setLocale(String),
preventing the issue if the OptionsParamView is (re)loaded.
Remove instance variable CustomScanDialog.optionsParam and corresponding
load of configurations (which was triggering the issue) when the
dialogue is reset, the optionsParam is not being used. Also, change to
use the method ScannerParam.clone() instead of doing the clone manually.

Fix #2440 - Exception while opening the Active Scan dialogue

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsParamView.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsParamView.java
@@ -32,6 +32,7 @@
 // ZAP: 2015/03/04 Added dev build warning option
 // ZAP: 2016/04/04 Do not require a restart to show/hide the tool bar
 // ZAP: 2016/04/06 Fix layouts' issues
+// ZAP: 2016/04/27 Save, always, the Locale as String
 
 package org.parosproxy.paros.extension.option;
 
@@ -218,8 +219,7 @@ public class OptionsParamView extends AbstractParam {
 			sb.append(locale.getLanguage());
 			if (locale.getCountry().length() > 0) sb.append("_").append(locale.getCountry());
 			if (locale.getVariant().length() > 0) sb.append("_").append(locale.getVariant());
-			this.locale = sb.toString();
-			getConfig().setProperty(LOCALE, locale);
+			setLocale(sb.toString());
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


kingthorin:
I know this is a fix to existing code but isn't this a rather kludgey way to get locale? Shouldn't there be some system or vm constant that can just be tostringed directly instead of assembling the locale in pieces?

I don't have a problem with the change, just wonder if there's a better way. Something that would address the issue and improve the code.

thc202:
We are already obtaining the default Locale (see Constant.SYSTEMS_LOCALE) and that's used to determine which language, from the ones available in ZAP, should be shown/selected. The conversion to String is done manually because Locale.toString() returns more than we need (e.g. script, extensions).

kingthorin:
Ok. Looks good.

Issue #1171, opened 2015-06-04
Sites tree to raise 'new site / new node' events


zapbot:
```
The sites tree should raise events for events like:
* Node added
* Site added
* Node removed
* Site removed
It would probably be best to plug this into the ExtensionHook mechanism, but other
options can be proposed.

This is one of a series of issues raise based on a thread in the dev group: https://groups.google.com/d/msg/zaproxy-develop/GKSHtogTNP4/Q0-_xUbCX-wJ
```

Original issue reported on code.google.com by `psiinon` on 2014-05-01 12:35:48

kingthorin:
Is the expectation that this be done in site or node constructors, or in any call that adds a site or node to the tree? If the later, how would we ensure that Extensions notify (publish events) of additions?

Was this already tacked and not closed? I do see that SiteMap publishes `SITE_NODE_ADDED_EVENT` events, etc. via https://github.com/zaproxy/zaproxy/blob/296801bb838ae1ceca102a6be5b5ed2e8c29e097/src/org/parosproxy/paros/model/SiteMapEventPublisher.java

thc202:
The events should be published when the nodes/sites are actually added/removed to/from the tree (as currently done in SiteMap for "nodes added" event and should be done for "nodes removed", removeNodeFromParent(...)?).

Not everything, it's missing the events "site added/removed" (if they are really needed) and the notification of "nodes removed".

kingthorin:
Ok, I'll start working this shortly. Thanks for confirming.

kingthorin:
Ok so I've put together a solution for the `SITE_NODE_REMOVED_EVENT` by overriding `removeNodeFromParent` in SiteMap, calling the super method and generating the event.

Any idea what the original distinction between node and site was? I think covering nodes is all that's necessary but maybe I'm missing some part of the original vision?

psiinon:
Quite a few extensions maintain a list of sites in the UI and currently they have to extend SiteMapListener and look at all events and then work out if a site was added or removed. Specific events for just those things would make the code simpler and easier to understand.

PR #2443, opened 2016-04-28
Raise SiteMap Add or Remove Events

kingthorin:
Raise SiteMap Add or Remove Events

SiteMap 
- Add private method handleEvent: Called from findAndAddChild,
findAndAddLeaf, removeNodeFromParent. 
- Add private convenience method publishEvent that accepts an event and
node to publish the necessary event (leveraging
SiteMapEventPublisher). Node events are always published,
site events are also published when the parent is the root of the tree.
- New overridden removeNodeFromParent method that extracts the
node's parent, calls the super method, and makes the call for the event
to be handled.

SiteMapEventPublisher
- Added site events (node events already existed).

Fixes zaproxy/zaproxy#1171

### DIFF
--- a/src/org/parosproxy/paros/model/SiteMap.java
+++ b/src/org/parosproxy/paros/model/SiteMap.java
@@ -48,6 +48,7 @@
 // ZAP: 2015/11/05 Change findNode(..) methods to match top level nodes
 // ZAP: 2015/11/09 Fix NullPointerException when creating a HistoryReference with a request URI without path
 // ZAP: 2016/04/21 Issue 2342: Checks non-empty method for deletion of SiteNodes via API 
+// ZAP: 2016/04/28 Issue 1171: Raise site and node add or remove events
 
 package org.parosproxy.paros.model;
 
@@ -62,6 +63,7 @@ import java.util.SortedSet;
 import java.util.TreeSet;
 
 import javax.swing.tree.DefaultTreeModel;
+import javax.swing.tree.MutableTreeNode;
 import javax.swing.tree.TreeNode;
 
 import org.apache.commons.httpclient.URI;
@@ -84,6 +86,8 @@ import org.zaproxy.zap.view.SiteTreeFilter;
 public class SiteMap extends DefaultTreeModel {
 
 	private static final long serialVersionUID = 2311091007687218751L;
+	
+	private enum EventType {ADD, REMOVE};
 	
 	private static Map<Integer, SiteNode> hrefMap = new HashMap<>();
 
@@ -441,9 +445,7 @@ public class SiteMap extends DefaultTreeModel {
             hrefMap.put(result.getHistoryReference().getHistoryId(), result);
 
             applyFilter(newNode);
-
-            ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
-            		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, new Target(result)));
+            handleEvent(parent, result, EventType.ADD);
 
         }
         // ZAP: Cope with getSiteNode() returning null
@@ -503,8 +505,7 @@ public class SiteMap extends DefaultTreeModel {
 
             this.applyFilter(node);
 
-            ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
-            		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, new Target(node)));
+            handleEvent(parent, node, EventType.ADD);            
         } else {
            
             // do not replace if
@@ -745,5 +746,48 @@ public class SiteMap extends DefaultTreeModel {
 			clearParentFilter(parent.getParent());
 		}
 	}
+	
+	@Override
+	public void removeNodeFromParent(MutableTreeNode node) {
+		SiteNode parent=(SiteNode)node.getParent();
+		super.removeNodeFromParent(node);
+		handleEvent(parent, (SiteNode)node, EventType.REMOVE);
+	}
 
+	/**
+	 * Handles the publishing of the add or remove event. Node events are always published.
+	 * Site events are only published when the parent of the node is the root of the tree.
+	 * 
+	 * @param parent relevant parent node
+	 * @param node the site node the action is being carried out for
+	 * @param eventType the type of event occurring (ADD or REMOVE)
+	 * @see EventType
+	 * @since TODO add version
+	 */
+	private void handleEvent(SiteNode parent, SiteNode node, EventType eventType) {
+		switch (eventType) {
+		case ADD:
+			publishEvent(SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, node);
+			if (parent == getRoot()) {
+				publishEvent(SiteMapEventPublisher.SITE_ADDED_EVENT, node);
+			}
+			break;
+		case REMOVE:
+			publishEvent(SiteMapEventPublisher.SITE_NODE_REMOVED_EVENT, node);
+			if(parent == getRoot()) {
+				publishEvent(SiteMapEventPublisher.SITE_REMOVED_EVENT, node);
+			}
+		}
+	}
+	
+	/**
+	 * Publish the event being carried out.
+	 * 
+	 * @param event the event that is happening
+	 * @param node the node being acted upon
+	 * @since TODO add version
+	 */
+	private static void publishEvent(String event, SiteNode node) {
+		ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), event, new Target(node)));
+	}
 }
--- a/src/org/parosproxy/paros/model/SiteMapEventPublisher.java
+++ b/src/org/parosproxy/paros/model/SiteMapEventPublisher.java
@@ -28,6 +28,8 @@ public class SiteMapEventPublisher implements EventPublisher {
 	private static SiteMapEventPublisher publisher = null;
 	public static final String SITE_NODE_ADDED_EVENT	= "siteNode.added";
 	public static final String SITE_NODE_REMOVED_EVENT	= "siteNode.removed";
+	public static final String SITE_ADDED_EVENT	= "site.added";
+	public static final String SITE_REMOVED_EVENT	= "site.removed";
 	
 	@Override
 	public String getPublisherName() {
@@ -38,7 +40,7 @@ public class SiteMapEventPublisher implements EventPublisher {
 		if (publisher == null) {
 			publisher = new SiteMapEventPublisher(); 
 	        ZAP.getEventBus().registerPublisher(publisher, 
-	        		new String[] {SITE_NODE_ADDED_EVENT, SITE_NODE_REMOVED_EVENT});
+	        		new String[] {SITE_NODE_ADDED_EVENT, SITE_NODE_REMOVED_EVENT, SITE_ADDED_EVENT, SITE_REMOVED_EVENT});
 
 		}
 		return publisher;


src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +743,26 @@ private void clearParentFilter (SiteNode parent) {
 			clearParentFilter(parent.getParent());
 		}
 	}
+	
+	@Override
+	public void removeNodeFromParent(MutableTreeNode node) {
+		super.removeNodeFromParent(node);
+		if(node.getParent() == getRoot()) {
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+	        		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_REMOVED_EVENT, new Target((SiteNode)node)));
```
thc202: This can be extracted to a method (one that takes the SiteNode and event type), it will reduce  code duplication a little further.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +743,26 @@ private void clearParentFilter (SiteNode parent) {
...
+	
+	@Override
+	public void removeNodeFromParent(MutableTreeNode node) {
+		super.removeNodeFromParent(node);
+		if(node.getParent() == getRoot()) {
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+	        		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_REMOVED_EVENT, new Target((SiteNode)node)));
+		} else {
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+	        		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_REMOVED_EVENT, new Target((SiteNode)node)));
```
thc202: This event should be published in both cases, if it is a site or not, otherwise listeners that check only the SITE_NODE events (like the Attack mode) would not know that the node (that represents the site) was added.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +743,26 @@ private void clearParentFilter (SiteNode parent) {
...
+		}
+	}
 
+	private void publishAddEvent(SiteNode parent, SiteNode item) {
+        if (parent==(SiteNode) getRoot()) {
```
thc202: The cast can be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +743,26 @@ private void clearParentFilter (SiteNode parent) {
...
+        	ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+        			new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_ADDED_EVENT, new Target(item)));
+        } else {
+        	ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+        			new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, new Target(item)));
+        }
```
thc202: Same comment (that it needs to be published in both cases).

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +743,26 @@ private void clearParentFilter (SiteNode parent) {
...
+		if(node.getParent() == getRoot()) {
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+	        		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_REMOVED_EVENT, new Target((SiteNode)node)));
+		} else {
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), 
+	        		new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_REMOVED_EVENT, new Target((SiteNode)node)));
+		}
+	}
 
+	private void publishAddEvent(SiteNode parent, SiteNode item) {
```
thc202: btw, some statements/lines (in both methods) are indented with spaces and others with tabs, this should be normalised.

thc202:
Shouldn't the ZAP comment be added to SiteMap class?

kingthorin:
It seems there's no way to identify that a node might be a site during removal (I had really thought that I tested that in the original iteration, but now for removal all details seem to be the same [null parent, index -1, isLeaf always true, etc]...maybe I saw what I wanted to see originally.)

Is there any point in raising Add events for "site" if there will never be a corresponding Remove event?

thc202:
That's a side effect of calling the method removeNodeFromParent, obtaining the parent before calling that method should do it.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -85,6 +87,8 @@
 
 	private static final long serialVersionUID = 2311091007687218751L;
 	
+	private static enum EventType {ADD, REMOVE};
```
thc202: The static modifier can be removed (it's redundant).

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,37 @@ private void clearParentFilter (SiteNode parent) {
...
+	 * Site events are only published when the parent of the node is the root of the tree.
+	 * 
+	 * @param parent relevant parent node
+	 * @param node the site node the action is being carried out for
+	 * @param event the type of event occurring (ADD or REMOVE)
+	 * 
+	 * @see {@code EventType}
+	 * @since (TODO add version)
+	 */
+
```
thc202: This empty line could be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,37 @@ private void clearParentFilter (SiteNode parent) {
...
+	 * @param event the type of event occurring (ADD or REMOVE)
+	 * 
+	 * @see {@code EventType}
+	 * @since (TODO add version)
+	 */
+	
+	private void publishEvent(SiteNode parent, SiteNode node, EventType eventType) {
+		switch (eventType) {
+		case ADD:
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, new Target(node)));
```
thc202: This statement(s) can be extracted to another method, one that just does:
```Java
ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), event, new Target(node)));
```
We would just need to change one place, if anything needs to be changed in how the events are published.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,37 @@ private void clearParentFilter (SiteNode parent) {
...
+			if (parent == getRoot()) {
+				ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_ADDED_EVENT, new Target(node)));
+			}
+		case REMOVE:
+			ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), SiteMapEventPublisher.SITE_NODE_REMOVED_EVENT, new Target((SiteNode)node)));
+			if(node.getParent() == getRoot()) {
```
thc202: `parent` instead of `node.getParent()`?

thc202:
The pull request or commit message should say that it fixes the issue.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,49 @@ private void clearParentFilter (SiteNode parent) {
 			clearParentFilter(parent.getParent());
 		}
 	}
+	
+	public void removeNodeFromParent(SiteNode node) {
```
thc202: Probably safer to override the method removeNodeFromParent(MutableTreeNode) (we could validate that the node is a SiteNode).

kingthorin: Will do. Removing the override and switching back to SiteNode for the param was just a change that I guess I lost track off while dealing with the null parent issue. Which as you pointed out was a time of use issue (can't get the parent from the node if the node has already been removed oops).

kingthorin:
Good catch

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
 			clearParentFilter(parent.getParent());
 		}
 	}
+	
+	@Override
+	public void removeNodeFromParent(MutableTreeNode node) {
+		SiteNode parent=(SiteNode)node.getParent();
+		super.removeNodeFromParent((MutableTreeNode) node);
```
thc202: This cast can be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+		handleEvent(parent, (SiteNode)node, EventType.REMOVE);
+	}
 
+	/**
+	 * Handles the publishing of the add or remove event. Node events are always published.
+	 * Site events are only published when the parent of the node is the root of the tree.
+	 * 
+	 * @param parent relevant parent node
+	 * @param node the site node the action is being carried out for
+	 * @param event the type of event occurring (ADD or REMOVE)
```
thc202: The parameter name should be `eventType`.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+	 *
```
thc202: This empty line could be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
 
+	/**
+	 * Handles the publishing of the add or remove event. Node events are always published.
+	 * Site events are only published when the parent of the node is the root of the tree.
+	 * 
+	 * @param parent relevant parent node
+	 * @param node the site node the action is being carried out for
+	 * @param event the type of event occurring (ADD or REMOVE)
+	 * 
+	 * @see {@code EventType}
```
thc202: Should be just `@see EventType` (it will link automatically to the enum).

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+	 * @since (TODO add version)
```
thc202: The parenthesis can be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+			publishEvent(SiteMapEventPublisher.SITE_NODE_ADDED_EVENT, node);
+			if (parent == getRoot()) {
+				publishEvent(SiteMapEventPublisher.SITE_ADDED_EVENT, node);
+			}
+			break;
+		case REMOVE:
+			publishEvent(SiteMapEventPublisher.SITE_NODE_REMOVED_EVENT, node);
+			if(parent == getRoot()) {
+				publishEvent(SiteMapEventPublisher.SITE_REMOVED_EVENT, node);
+				System.out.println("Site - Remove");
```
thc202: This statement should be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+			}
+		}
+	}
+	
+	/**
+	 * Publish the event being carried out.
+	 * 
+	 * @param event the event that is happening
+	 * @param node the node being acted upon
+	 *
```
thc202: This empty line could be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+		}
+	}
+	
+	/**
+	 * Publish the event being carried out.
+	 * 
+	 * @param event the event that is happening
+	 * @param node the node being acted upon
+	 * 
+	 * @since (TODO add version)
```
thc202: The parenthesis can be removed.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+	 */
+	private void publishEvent(String event, SiteNode node) {
```
thc202: Consider adding the static modifier.

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -745,5 +746,51 @@ private void clearParentFilter (SiteNode parent) {
...
+		ZAP.getEventBus().publishSyncEvent(SiteMapEventPublisher.getPublisher(), new Event(SiteMapEventPublisher.getPublisher(), event, new Target(node)));
```
thc202: The tabs at the end of the line should be removed.

kingthorin:
Tks, all done.

thc202:
Looks good to me, thank you.

PR #2444, opened 2016-04-28
Generate Authorization Java API class

thc202:
Generate Authorization Java API class to already include it in the
ClientApi class (and not be forgotten).


kingthorin:
Looks good

PR #2447, opened 2016-04-29
Allow to copy active scan's progress details

### LINKED ISSUE #1864: Copy scan progress as text to clipboard
Its nice that you can see the process of a scan by rule in the scan progress dialog. But it would be even nicer if you could copy a text version of that status into the clipboard so you could easily use it/send it to someone else as text without having to take a screenshot and send it that way.

thc202:
Change class ScanProgressDialog to allow to copy (to the clipboard) the
contents of the table that shows the progress of the active scan. The
change adds a button (to the footer of the dialogue, same place as the
close button) that does the actual copy of contents to the clipboard.
Change classes ScanProgressActionIcon and ScanProgressItem to implement
the method toString(), used when copying the data.

Fix #1864 - Copy scan progress as text to clipboard


kingthorin:
Looks good

PR #2448, opened 2016-04-29
Change README to inform about Xfvb, when headless

thc202:
Update the section "Headless Environment" of README (and README.weekly)
to mention the use of applications like Xfvb when running in a headless
environment to use functionalities/add-ons that require a display.

Related to changes for #2390 - HeadlessException should be handled more
gracefully & README needs Headless details


kingthorin:
Looks good

PR #2438, opened 2016-04-26
Show more informative error on proxy read timeout

thc202:
Change class ProxyThread to set a more informative error message into
the body of the HTTP response, indicating that a read timeout occurred
while reading the response after the seconds defined in the Connection
options panel (as proposed in the issue report).

Fix #2400 - Show a more informative message on read timeouts through the
proxy

PR #2449, opened 2016-04-29
Adjust ProxyThread log levels and log read timeout

thc202:
Change class ProxyThread to adjust the log level of exceptions caused by
listeners (a probable issue/bug) from WARN to ERROR and change some log
WARN levels to DEBUG (ones that are not an issue/bug). Also, change to
log the timeout with WARN level, per discussion in #2438.
Change log4j.properties file to remove the override that changed the
default log level (of the class ProxyThread) to ERROR.


kingthorin:
Looks good

PR #2453, opened 2016-05-02
Include all DDNs when exporting a context

### LINKED ISSUE #2451: Only a single Data Driven Node can be saved in a context
OWASP ZAP 2.4.3 running on a Kali Linux VM.

The UI lets you define multiple data driven nodes for a context, but only one (I think the last one edited) is saved in the context when you export it. I first noticed this when importing a context into a new session, at which point I examined the contents of the .context file, where I confirmed that only one node was saved.

thc202:
Change method Session.exportContext() to add the DDNs to the file,
instead of set (which would overwrite the previously set/added DDN).

Fix #2451 - Only a single Data Driven Node can be saved in a context


kingthorin:
Looks good

PR #2454, opened 2016-05-03
Remove unnecessary exceptions' stack trace prints

thc202:
Change class AbstractPlugin to remove unnecessary exceptions' stack
trace prints, the exception UnsupportedEncodingException should not
happen for "UTF-8", which is one of the standard charsets (moreover the
exception should have been logged not printed).


kingthorin:
Looks good

PR #2455, opened 2016-05-04
Set existing Plugin instance as completed

thc202:
Change PluginFactory to set the existing Plugin instance as completed
instead of the instance passed as parameter, to ensure that the Plugin
that's returned as completed is the correct instance and not one of the
instances created during and for the active scan (which might not have
all its state up-to-date).
The change prevents completed AbstractHostPlugin scanners from having
unknown status when returning scans' progress through the active scan
API.


kingthorin:
Looks good

PR #2456, opened 2016-05-04
Fix issues related to ParameterParser

### LINKED ISSUE #1801: URL StandardParameterParser not working correctly with QueryString
If a URL GET parameter is parsed by the  StandardParameterParser which actually splits on '&' and '=' , the result is not correct if the parameter value includes a '=" as value. 
With SAML this is quite often the case as SAML requests are BASE64 encoded.

Digging into the problem it turned out that the Querystring is read several times either in or in HttpMessage The problem is generated by 

https://github.com/zaproxy/zaproxy/blob/708b5470446a2f522162093a621481b492116f5f/src/org/zaproxy/zap/model/StandardParameterParser.java#L108
https://github.com/zaproxy/zaproxy/blob/develop/src/org/parosproxy/paros/network/HttpMessage.java#L564

...getURI().getQuery();

which should be 

...getURI().getRawQuery());

Example SAML Request with an Encoded "=" at the end:

https://idp.ssocircle.com/sso/SSORedirect/metaAlias/ssocircle?SAMLRequest=fZHbTsJAEEB%2FZbPvbRcM2kwoptCYNCohQIzhbW1X2aR7cWeK6Ne7gBjjA49zPWcyY%2FRQ9rS1S%2FXeKyS2N51FQF%2FwPlhwEjWClUYhUAOr8vEBhqkAHxy5xnWclYgqkHZ25iz2RoWVCjvdqNq2al%2FwIWd1VfD59Ob%2B2i2JYojYxyqStBTrYjBKRJ4M8rXIYTQCcbXhbBHcTrcqzCO44JvnumKVMo6tFpw9qYARF0dTwSdjlHDcGM7m8rK5PPvyyZbIQ5Y1bzrBk3Tif8ApStMN08YZyMWx40Xb7Guv21tXTMfZLzUKeDho1tXCdbr5ZGXXuY9ZUJKiOoVecXbngpF02euQ0W3yemwFfzgSSVmKJ2b%2FEKfM369NvgE%3D

Querystring decoded looks like:
SAMLRequest=fZHbTsJAEEB/ZbPvbRcM2kwoptCYNCohQIzhbW1X2aR7cWeK6Ne7gBjjA49zPWcyY/RQ9rS1S/XeKyS2N51FQF/wPlhwEjWClUYhUAOr8vEBhqkAHxy5xnWclYgqkHZ25iz2RoWVCjvdqNq2al/wIWd1VfD59Ob+2i2JYojYxyqStBTrYjBKRJ4M8rXIYTQCcbXhbBHcTrcqzCO44JvnumKVMo6tFpw9qYARF0dTwSdjlHDcGM7m8rK5PPvyyZbIQ5Y1bzrBk3Tif8ApStMN08YZyMWx40Xb7Guv21tXTMfZLzUKeDho1tXCdbr5ZGXXuY9ZUJKiOoVecXbngpF02euQ0W3yemwFfzgSSVmKJ2b/EKfM369NvgE=

The parser(splitting parameters at '=') would generate an incorrect value of SAMLRequest missing the last '='. Besides being incorrect the base64 decoding will fail due to missing padding.

### LINKED ISSUE #1848: VariantURLQuery throwing exceptions on active scan
When scanning this error occurs. 

Error like this occurs for different scan rules 

769590 [ZAP-ActiveScanner-1] ERROR org.zaproxy.zap.extension.ascanrules.TestServerSideInclude  - Error occurred while scanning with variant org.parosproxy.paros.core.scanner.VariantURLQuery
java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern - For input string: "{p"
	at java.net.URLDecoder.decode(URLDecoder.java:192)
	at org.parosproxy.paros.core.scanner.AbstractPlugin.getURLDecode(Unknown Source)
	at org.parosproxy.paros.core.scanner.VariantURLQuery.getUnescapedValue(Unknown Source)
	at org.parosproxy.paros.core.scanner.VariantAbstractQuery.setParams(Unknown Source)
	at org.parosproxy.paros.core.scanner.VariantURLQuery.setMessage(Unknown Source)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scan(Unknown Source)
	at org.parosproxy.paros.core.scanner.AbstractPlugin.run(Unknown Source)

### LINKED ISSUE #2153: 2.4.3 failed parse the POST Data containts bracket([])
When the POST Data containts bracket([]), ZAP 2.4.3 puts that parameters has same name into one parameter.

1) Submit this form.

[posttest.txt](https://github.com/zaproxy/zaproxy/files/70463/posttest.txt)

2) ZAP 2.4.3 gets data below.

POST http://localhost/test.php HTTP/1.1
param%5B%5D=value1&param%5B%5D=value2&param%5B%5D=value3&param%5B%5D=value4&param%5B%5D=value5

3) Do Attack - Active Scan.
The sending data is :

POST http://localhost/test.php HTTP/1.1
param%5B%5D=0W45pz4p

POST http://localhost/test.php HTTP/1.1
param%5B%5D=%27

POST http://localhost/test.php HTTP/1.1
param%5B%5D=value5%27
  .
  .

thc202:
Parse form/query parameters in encoded form, so they are correctly split
into name/value pairs even if the names and/or values have separator
characters (normally, '&' and '='), also allow to keep all parameters,
so requests that require multiple values with same name (e.g. arrays)
are still sent with all of them (instead of just one).
When rebuilding the query component of the URL, during active scan, make
sure to also encode parameters' names not just its values, otherwise the
query could be malformed if the names contain characters that should be
encoded (common when dealing with multiple vales which might use the
characters '[' and ']'), leading to exceptions and failing to properly
attack those parameters. (The same logic applies for form parameters.)

More detailed changes:
 - Variant, document the method getParamList();
 - VariantAbstractQuery:
  - Add method, getEscapedName(HttpMessage, String), to encode/escape
  parameter's name;
  - Add method, setParameters(int, List\<NameValuePair>), to allow
  extending classes to set (and finally send) all parameters even if
  they have the same name;
  - Deprecate the method setParams(int, Map\<String, String>) in favour
  of the new method.
 - VariantFormQuery and VarianURLQuery, change to obtain all the
 parameters using a new method in Session class and to set all of them
 using the new method of the class VariantAbstractQuery;
 - Session:
  - Add method, getParameters(HttpMessage, HtmlParameter.Type), that
  allows to obtain (all) URL or form parameters from a message;
  - Change method getUrlParams(URI) to parse the query in escaped form
   (and use the new method added to ParameterParser, which decodes the
   parameters names and values when parsing, as expected by callers);
 - Add interface NameValuePair and default implementation, class
 DefaultNameValuePair;
 - ParameterParser, add methods getParameters(...) and
 parseParameters(String) which return List\<NameValuePair>;
 - StandardParameterParser:
  - Add implementations of the methods added to the interface
  ParameterParser;
  - Change method getParams, getTreePath and getAncestorPath to use
  parseParameters(String) when parsing the (now escaped) query, so the
  parameters are correctly separated (and decoded).

Fix #1801 - URL StandardParameterParser not working correctly with
QueryString
Fix #1848 - VariantURLQuery throwing exceptions on active scan
Fix #2153 - 2.4.3 failed parse the POST Data containts bracket([])

### DIFF
--- a/src/org/parosproxy/paros/model/Session.java
+++ b/src/org/parosproxy/paros/model/Session.java
@@ -56,6 +56,7 @@
 // ZAP: 2015/12/14 Issue 2119: Context's description not imported
 // ZAP: 2016/02/26 Issue 2273: Clear stats on session init
 // ZAP: 2016/05/02 Issue 2451: Only a single Data Driven Node can be saved in a context
+// ZAP: 2016/05/04 Changes to address issues related to ParameterParser
 
 package org.parosproxy.paros.model;
 
@@ -66,6 +67,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -90,6 +92,7 @@ import org.zaproxy.zap.control.ExtensionFactory;
 import org.zaproxy.zap.extension.ascan.ExtensionActiveScan;
 import org.zaproxy.zap.extension.spider.ExtensionSpider;
 import org.zaproxy.zap.model.Context;
+import org.zaproxy.zap.model.NameValuePair;
 import org.zaproxy.zap.model.ParameterParser;
 import org.zaproxy.zap.model.StandardParameterParser;
 import org.zaproxy.zap.model.StructuralNodeModifier;
@@ -1390,6 +1393,38 @@ public class Session {
 	}
 
 	/**
+	 * Gets the parameters of the given {@code type} from the given {@code message}.
+	 * <p>
+	 * Parameters' names and values are in decoded form.
+	 *
+	 * @param msg the message whose parameters will be extracted from
+	 * @param type the type of parameters to extract
+	 * @return a {@code List} containing the parameters
+	 * @throws IllegalArgumentException if any of the parameters is {@code null} or if the given {@code type} is not
+	 *			 {@link org.parosproxy.paros.network.HtmlParameter.Type#url url} or
+	 *			 {@link org.parosproxy.paros.network.HtmlParameter.Type#form form}.
+	 * @since TODO add version
+	 * @see StandardParameterParser#getParameters(HttpMessage, org.parosproxy.paros.network.HtmlParameter.Type)
+	 */
+	public List<NameValuePair> getParameters(HttpMessage msg, HtmlParameter.Type type) {
+		if (msg == null) {
+			throw new IllegalArgumentException("Parameter msg must not be null.");
+		}
+		if (type == null) {
+			throw new IllegalArgumentException("Parameter type must not be null.");
+		}
+
+		switch (type) {
+		case form:
+			return this.getFormParamParser(msg.getRequestHeader().getURI().toString()).getParameters(msg, type);
+		case url:
+			return this.getUrlParamParser(msg.getRequestHeader().getURI().toString()).getParameters(msg, type);
+		default:
+			throw new IllegalArgumentException("The provided type is not supported: " + type);
+		}
+	}
+
+	/**
 	 * Returns the URL parameters for the given URL based on the parser associated with the
 	 * first context found that includes the URL, or the default parser if it is not
 	 * in a context
@@ -1398,7 +1433,11 @@ public class Session {
 	 * @throws URIException
 	 */
 	public Map<String, String> getUrlParams(URI uri) throws URIException {
-		return this.getUrlParamParser(uri.toString()).parse(uri.getQuery());
+		Map<String, String> map = new HashMap<>();
+		for (NameValuePair parameter : getUrlParamParser(uri.toString()).parseParameters(uri.getEscapedQuery())) {
+			map.put(parameter.getName(), parameter.getValue());
+		}
+		return map;
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


kingthorin:
Build is failing, but I'm guessing that the test suite just needs updating to account for the new code: https://travis-ci.org/zaproxy/zaproxy/builds/127931164#L247

thc202:
It was a split limit in the wrong place (should be in name/value but it was in the pairs), should be fixed now.
Thanks!

psiinon:
Looks good

src/org/parosproxy/paros/model/Session.java
```diff
@@ -1390,6 +1393,38 @@ public ParameterParser getFormParamParser(String url) {
...
 
 	/**
+	 * Gets the parameters of the given {@code type} from the given {@code message}.
+	 * <p>
+	 * Parameters' names and values are in decoded form.
+	 *
+	 * @param msg the message whose parameters will be extracted from
+	 * @param type the type of parameters to extract
+	 * @return a {@code List} containing the parameters
+	 * @throws IllegalArgumentException if any of the parameters is {@code null} or if the given {@code type} is not
```
kingthorin: Should that last word be "not" or "null"?

src/org/zaproxy/zap/model/StandardParameterParser.java
```diff
@@ -104,17 +107,44 @@ public String getConfig() {
...
-		} catch (URIException e) {
-			log.error(e.getMessage(), e);
+			return map;
+		default:
+					throw new InvalidParameterException("Type not supported: " + type);
+		}
+	}
+
+	/**
+	 * @throws IllegalArgumentException if any of the parameters is {@code null} or if the given {@code type} is not
```
kingthorin: Should that last word be "not" or "null"?

PR #2466, opened 2016-05-09
Allow to access a URL through the ZAP API

thc202:
Add a convenient and simple action to CoreAPI, called "accessUrl", that
allows to access a URL, optionally following redirections. Since the URL
is accessed from within ZAP it does not require external applications to
trust ZAP's root CA certificate when accessing HTTPS URLs. As with the
other actions that allow to send requests this one also returns the
request sent and response received (and redirections followed, if any).
 ---
From OWASP ZAP Developer Group: https://groups.google.com/d/topic/zaproxy-develop/PPqg7QCaSmM/discussion


kingthorin:
Looks good

PR #2467, opened 2016-05-09
Normalise the case of HTTP/S

thc202:
Change all resource messages that were using Http and Https to HTTP and
HTTPS, respectively (in one message there was "HttpSessions" which was
split into "HTTP sessions").
 ---
From review comments in zaproxy/zap-core-help#24.


kingthorin:
Looks good

PR #2469, opened 2016-05-10
Always return 100% when spider stopped

psiinon:
If the maxDuration is set and takes effect the state stays at <100 and so the client cant tell its finished


kingthorin:
Looks good

thc202:
Thanks!

PR #2470, opened 2016-05-10
Use empty String for "old" parameters

thc202:
Change classes Session and StandardParameterParser to use an empty
String if the parameter does not have a value when converting the List
of parameters to a Map, keeping the old behaviour (some add-ons and some
core code does not handle parameters without value, leading to
NullPointerException).

Related to #2456 - Fix issues related to ParameterParser


kingthorin:
Looks good

PR #2472, opened 2016-05-11
Use file name case when loading policies from file

thc202:
Change method PolicyManager.loadPolicy(File) to obtain the name of the
file in correct case. Prevents the default scan policy, which name is
read from resource files, from having its name with a different case
than the (file) name read directly from the (case insensitive) file
system, when the policies are enumerated. This would lead to scan policy
selection issues in Policy tab of Active Scan dialogue.


kingthorin:
Looks good

PR #2474, opened 2016-05-11
Add multiple context URL in/exclusions at once

thc202:
Change pop up menu items Exclude from Context and Include In Context to
add the selected URLs at once and show the dialogue just once (instead
of showing the dialogue multiple times for each URL selected).


psiinon:
looks good

PR #2475, opened 2016-05-11
Correct help page for Passive Scan Rules options

thc202:
Change the help index of PolicyPassiveScanPanel (Passive Scan Rules
options panel) to link to "Options Passive Scan Rules Screen" help page.
(It was linking to "Options Passive Scan Tags screen" help page.)


kingthorin:
Looks good

PR #2478, opened 2016-05-12
Use dev plugin dir when copying (old) config file

thc202:
Use dev/weekly directory for plugin downloads when copying the existing
'release' config file. Prevents dev add-ons from polluting the 'release'
version of ZAP, moreover it prevents an exception during initialisation
of dev/weekly version of ZAP because of mismatch plugin download dir.
Following the message of the exception:
 `InvalidParameterException: Directory must be the default one or one of
 the addonDirectories /home/<username>/.ZAP/plugin`


kingthorin:
Looks good

PR #2479, opened 2016-05-12
Show User Guide even if a search view has errors

thc202:
Add a SearchView whose underlying BasicSearchNavigatorUI keeps merging
(search) views even if one of them is invalid (that is, has errors). A
search view is invalid when, for example, it has no search indexes which
are required by the search engine implementation.
Core help pages (helpset) need to use the new SearchView implementation
to work as intended.


psiinon:
Looks good

PR #2481, opened 2016-05-13
Set correct application name in Linux

thc202:
Change class GuiBootstrap to set the correct application name in Linux
(if Toolkit is XToolkit), instead of showing "org-zaproxy-zap-ZAP" it
shows "OWASP ZAP".


kingthorin:
Looks good

PR #2482, opened 2016-05-13
Normalise Session dialogue tables "Exclude from"

thc202:
Change the Session dialogue panels "Exclude from Proxy", "Exclude from
Scanner" and "Exclude from Spider" to use the same tables as other
dialogues to manage multiple options (that is, table shown in one side
and buttons to add/modify/remove at the other).


kingthorin:
Looks good

PR #2485, opened 2016-05-16
Do not ignore exception when building the request

thc202:
Change class HttpMethodHelper to throw a URIException when failed to set
the URI to the request method, instead of logging as error and continue.
Ignoring the exception just leads to other (unexpected) exceptions later
on when processing the request, for example:
   IllegalArgumentException: host parameter is null

The exception URIException is already expected to be handled by its
callers (either directly or through IOException).
Based on results from zapbot's WAVSEP scans.


kingthorin:
Looks good

PR #2486, opened 2016-05-16
Clear ScriptVars on session change

thc202:
Change class ScriptVars to allow to clear the variables and change class
ExtensionScript to call that method when the session is about to change.
Ensures that the scripts and its variables are in the expected state
when a new session is created or an existing session loaded.


kingthorin:
Looks good

PR #2493, opened 2016-05-18
Always use charset set when changing the HTTP body

### LINKED ISSUE #2487: Wrong charset used in HTTP body
ZAP version 2.4.3
These characters are used in a replay request: ÃÂÃÂÃ in a text field.
Every time I press SEND, the characters duplicate within the request.

thc202:
Change class HttpBody to always use the charset set when creating the
String representation of the body (it could happen in some cases that it
would use the default charset instead of the one set, leading to wrong
encoding of the characters). Also, fix undocumented exceptions/behaviour
and change to use StandardCharsets/Charset (instead of String) when
obtaining the bytes from the String or converting the bytes to String.
Change class HttpResponseBody to also use StandardCharsets/Charset and
use the new code to create the String representation.
Add tests to assert the expected behaviour of classes HttpBody and
HttpResponseBody.

Fix #2487 - Wrong charset used in HTTP body

### DIFF
--- a/src/org/parosproxy/paros/network/HttpBody.java
+++ b/src/org/parosproxy/paros/network/HttpBody.java
@@ -21,189 +21,276 @@
  */
 // ZAP: 2012/03/15 Changed to use byte[] instead of StringBuffer.
 // ZAP: 2014/11/26 Issue: 1415 Fixed file uploads > 128k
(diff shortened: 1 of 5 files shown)


kingthorin:
Looks good

PR #2501, opened 2016-05-23
Regenerated API code

psiinon:
Note that this PR doesnt include changes to update any of the packages.
The plan is to move the generated code to new repos (like zaproxy/zap-api-java) prior to releasing them, but I thought it would be better to do that as separate PRs.


kingthorin:
Looks good

PR #2502, opened 2016-05-23
Set AWT as headless when ZAP is in a headless mode

thc202:
Change class HeadlessBootstrap to set AWT as headless, to ensure that
daemon and command line modes fail always to use view components that
require a display, not just when there's no display available. Allows to
detect (wrong) use of view components more easily (HeadlessException is
thrown even if there's a display available) when in daemon and command
line modes.


kingthorin:
Looks good

PR #2504, opened 2016-05-23
Explicitly add manual HTTP requests to Sites tree

thc202:
Change class HttpPanelSender to also add the HTTP requests (sent with
Manual Request Editor or Resend dialogues) to the Sites tree, so that
the Sites tree can be manually populated (for use with active scan),
always, not just when an alert is raised on the request sent.


kingthorin:
Looks good

PR #2505, opened 2016-05-24
Proxy WebSocket traffic through outgoing proxy set

### LINKED ISSUE #2463: Websocket not proxied when outgoing proxy is set
I have an issue with testing websockets in ZAP when outgoing proxy is set in Tools->Options->Connections. The usual web traffic works fine, but for some reason the websocket (e.g. http://www.websocket.org/echo.html) fails to connect through ZAP in such case.

thc202:
Change class HttpSender to also use the custom connection manager (one
that does not close the connection and exposes its socket) when sending
requests that have a connection upgrade header (e.g. WebSocket) through
an outgoing proxy.
Change HttpMethodDirector to establish a tunnel if the request has a
connection upgrade header.
Fork HttpConnection to allow to establish a tunnel when requested by
calling code (in this case when the request has a connection upgrade).

Fix #2463 - Websocket not proxied when outgoing proxy is set

### DIFF
--- a/src/org/apache/commons/httpclient/HttpConnection.java
+++ b/src/org/apache/commons/httpclient/HttpConnection.java
@@ -1,0 +1,1407 @@
+/*
+ * HeadURL: https://svn.apache.org/repos/asf/httpcomponents/oac.hc3x/trunk/src/java/org/apache/commons/httpclient/HttpConnection.java 
+ * Revision: 832758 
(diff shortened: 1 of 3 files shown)


kingthorin:
What will happen if the proxy is put in place after the initial websocket (upgrade) connection?

thc202:
Those changes have no effect on established connections.

kingthorin:
Perfect, looks good to me.

PR #2509, opened 2016-05-25
Normalise Alert.equals/hashCode/compareTo

### LINKED ISSUE #2132: Zap Report Counting Bug
Report creation generates a different amount of alerts.

Scan revealed 7 High, 6 Med, 10 Low, 3 Notice

Export Counts
7H, 7M, 19L, 10N

Please refer to [1] for more details!

[1] https://groups.google.com/forum/#!topic/zaproxy-users/dpOMYG3nPhQ

thc202:
Normalise the implementation of Alert's methods equals and compareTo and
implement the method hashCode for consistency when used in different
collections.
Two alerts are considered the same if the following properties are
equal:
 - Plugin ID;
 - Name;
 - Risk;
 - Confidence;
 - URI;
 - Parameter;
 - Evidence;
 - Attack;
 - Other Info.

Change class AlertTreeModel to use the method Alert.compareTo instead of
comparing manually the alerts to ensure that the alerts shown in the
Alerts tab are the same as the ones included in the report.

Fix #2132 - Zap Report Counting Bug

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/Alert.java
+++ b/src/org/parosproxy/paros/core/scanner/Alert.java
@@ -41,6 +41,7 @@
 // ZAP: 2015/08/24 Issue 1849: Option to merge related issues in reports
 // ZAP: 2015/11/16 Issue 1555: Rework inclusion of HTML tags in reports 
 // ZAP: 2016/02/26 Deprecate alert as an element of Alert in favour of name
+// ZAP: 2016/05/25 Normalise equals/hashCode/compareTo
 
 package org.parosproxy.paros.core.scanner;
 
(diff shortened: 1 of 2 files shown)


kingthorin:
Looks good

src/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -332,7 +337,33 @@ public int compareTo(Object o2) throws ClassCastException {
...
 		
-		return otherInfo.compareToIgnoreCase(alert2.otherInfo);
+		result = otherInfo.compareToIgnoreCase(alert2.otherInfo);
+		if (result != 0) {
+			return result;
+		}
+
+		if (evidence == null) {
+			if (alert2.evidence == null) {
+				return 0;
```
psiinon: Shouldnt we carry on here instead of returning 0? We still need to check the attack etc

thc202: Fixed.

PR #2510, opened 2016-05-26
Updated relevant 'since' versions to 2.5.0

psiinon:
OK, so in theory this should be done when we create 2.5.0, but I'm just trying to tie up loose ends ;)


kingthorin:
Looks good

PR #2511, opened 2016-05-26
Delete temporary history types sequentially

thc202:
Change class ParosTableHistory to delete the temporary history types
sequentially instead of all at once (using an array), the former
requires less memory (e.g. under the same conditions, it's able to
process 3M messages without leading to an OutOfMemoryError) and it's
faster to delete all temporary types.


kingthorin:
Looks good

PR #2512, opened 2016-05-26
Update HSQLDB library to version 2.3.4

thc202:
It's faster than currently used version (2.2.9) and has the interrupt
issue fixed (that led to revert of the previous update to 2.3.3).

Related to #2134 and #2254 (update and revert).


kingthorin:
Looks good

PR #2514, opened 2016-05-26
Use empty String in query/post variant parameters

thc202:
Change class VariantAbstractQuery to use an empty String, when creating
the parameters, if the parameter does not have a name/value, keeping the
old behaviour (some scanners do not handle parameters without value,
leading to NullPointerException).
Based on results from zapbot's WAVSEP scans.

Related to #2456 - Fix issues related to ParameterParser


kingthorin:
Looks good

PR #2516, opened 2016-05-27
Add auth and spider task types to temporary types

thc202:
Add authentication and spider task types to the set of default temporary
types, types of message that are removed once the session is closed.
The authentication messages are shown temporarily in the History tab,
once the session is closed they become inaccessible (they are no longer
shown in any UI component), which would leak messages in the database as
it was not possible to delete them.
The spider task messages are also temporary and are automatically
removed by the spider when no longer needed, added the type to prevent
database leaks in case of exceptions in the spider execution.
Move the management of the history types to class HistoryReference,
since the types are "now" used by more than one TableHistory
implementation.


kingthorin:
Looks good

PR #2518, opened 2016-05-30
Reset the charset of the body with null/empty name

### LINKED ISSUE #2517: Wrong charset used in resent response
1) startup zaproxy
2) in shell, make a request with curl, command is:   curl -x localhost:8080 xx.fang.com

In history panel, we can see its response panel show is OK.
Then, we make a resent to this url, the response panel of the rent will show: 
<title>�����緿�ز��Ż�|���緿�ز�����- �����ѷ��� ������</title> 

but return to history panel, its response still OK, so, I think there may be something wrong with reponse panel in resent window.

thc202:
Change class HttpBody to reset the charset when it's set a null/empty
charset name, to ensure that it always uses the charset set from the
header (if reset it defaults to default charset or the one determined
internally by HttpBody implementations, like HttpResponseBody).
Change tests to assert the (new) expected behaviour of class HttpBody.

Related to #2493 - Always use charset set when changing the HTTP body
(Since those changes it no longer determines, always, the charset from
the response body when the cached string is re-created.)

Fix #2517 - Wrong charset used in resent response

### DIFF
--- a/src/org/parosproxy/paros/network/HttpBody.java
+++ b/src/org/parosproxy/paros/network/HttpBody.java
@@ -340,7 +340,10 @@ public abstract class HttpBody {
      * Sets the charset used for {@code String} related operations, for example, {@link #append(String)},
      * {@link #setBody(String)}, or {@link #toString()}.
      * <p>
-     * The charset is ignored if {@code null}, empty, or not valid (either the name is not valid or is unsupported).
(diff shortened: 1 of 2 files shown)


kingthorin:
Looks good

zdzhjx:
YES, IT IS OK NOW

THANKS.

PR #2519, opened 2016-05-30
Enable menu/button when help add-on is installed

thc202:
Enable/disable the main menu item (Help > OWASP ZAP User Guide) and main
tool bar help button when the help add-on (for the selected language) is
installed/uninstalled. Also, make sure to clean up any reference to the
help contents when uninstalling to ensure that the correct/latest help
contents are shown when the help add-on is updated and keep track of
components that have a help page assigned so that they can be set/unset
when the help is installed/uninstalled.

As part of the changes it's now possible for extensions to listen for
the installation status of the add-ons (required to dynamically enable
and disable the help related functionality when the help add-on is
installed/uninstalled).


kingthorin:
Looks good

PR #2521, opened 2016-05-31
Add hashCode implementation to HttpMessage

thc202:
Since the method equals(Object) is overridden, hashCode() should as
well.


kingthorin:
Looks good

PR #2527, opened 2016-06-02
Guard against exceptions from ScriptUI

thc202:
Change ExtensionScript to catch exceptions (possibly) thrown by ScriptUI
implementation, when notifying of engines added or removed, preventing
the exceptions from escalating and causing issues in other areas of ZAP.
For example, an issue in ScriptUI implementation [1] prevented the Zest
add-on from being successfully uninstalled.

[1] zaproxy/zap-extensions#388


kingthorin:
Looks good

PR #2544, opened 2016-06-07
Add options for config url, xml report and flagging alerts as INFO

psiinon:
Will also update the one in zap-api-python

### DIFF
--- a/build/docker/zap-baseline.py
+++ b/build/docker/zap-baseline.py
@@ -31,10 +31,13 @@
 #	2:	At least one WARN and no FAILs
 #	3:	Any other failure
 # By default all alerts found by ZAP will be treated as WARNings.
-# You can use the -c parameter to specify a configuration file to override this.
+# You can use the -c or -u parameters to specify a configuration file to override
+# this.
 # You can generate a template configuration file using the -g parameter. You will
-# then need to change 'WARN' to 'FAIL' or 'IGNORE' for the rules you want to be
-# handled differently.
+# then need to change 'WARN' to 'FAIL', 'INFO' or 'IGNORE' for the rules you want
+# to be handled differently.
+# You can also add your own messages for the rules by appending them after a tab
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

build/docker/zap-baseline.py
```diff
@@ -120,7 +135,7 @@ def main(argv):
     usage()
     sys.exit(3)
     
-  if len(config) > 0 or len(generate) > 0 or len(report) > 0:
+  if len(config_file) > 0 or len(config_url) > 0 or len(generate) > 0 or len(report) > 0:
```
thc202: - Is the check of `config_url` really needed? (It does not seem to use the directory.)
 - The variable `report` is no longer defined, should be `report_html`/`report_xml` instead?

build/docker/zap-baseline.py
```diff
@@ -270,12 +310,18 @@ def main(argv):
...
+      if len(report_html) > 0:
+        # Save the report
+        with open('/zap/wrk/' + report_html, 'w') as f:
+          f.write (zap.core.htmlreport())
+
+      if len(report_xml) > 0:
         # Save the report
-        with open('/zap/wrk/' + report, 'w') as f:
+        with open('/zap/wrk/' + report_xml, 'w') as f:
           f.write (zap.core.htmlreport())
```
thc202: Core call needs to be updated to XML report.

build/docker/zap-baseline.py
psiinon: Ah yea, it should be report_html and report_xml (not config_url) as they are both written to the mapped drive

build/docker/zap-baseline.py
psiinon: yep

psiinon:
Hopefully fixed those

thc202:
Thanks!

PR #2547, opened 2016-06-07
Remove directories no longer used

thc202:
Remove the directories "log" and "session" as they are no longer used,
the logging is done in ZAP's home directory and the file sessions are
saved (by default) in the home "session" directory.

Change class Constant to:
 - Remove commented constants;
 - Remove a statement that sets the directory used to save the log that
 had no (actual) effect, the directory set is overwritten by a following
 statement (which sets it to the home directory);
 - Add JavaDoc to FOLDER_SESSION_DEFAULT;
 - Initialise the variable FOLDER_SESSION with FOLDER_SESSION_DEFAULT.

Remove log and session directories from the build file.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -157,9 +157,6 @@
 		<copy todir="${dist}/scripts">
 			<fileset dir="${src}/scripts" excludes="**/.svn/** **/_svn/**" />
 		</copy>
-		<copy todir="${dist}/session">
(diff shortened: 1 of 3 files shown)


kingthorin:
Build failing

thc202:
Fixed. (That reminds me of the fuzzers directory...)

kingthorin:
Looks good

PR #2548, opened 2016-06-07
Change filters to use "filter" home directory

thc202:
Change the classes FilterLogGetQuery, FilterLogPostQuery and
FilterLogRequestResponse to use the filter directory in ZAP's home
directory instead of installation directory. Depending on how ZAP is
installed it might not even be writable, leading to, for example:
   java.io.FileNotFoundException: filter/get.xls (Permission denied)

Change class Constant to create a filter directory in ZAP's home
directory if it does not exist already, also document the constant
FOLDER_FILTER to indicate that it should be used along the ZAP's home
directory.

Remove the (installation) filter directory, since it would no longer be
needed, this might affect 3rd party filters (although very unlikely,
given it's use is discouraged, scripts are recommended instead). The
track of the IDs that was done in the dummy.txt is lost, it was outdated
moreover the filters are no longer actively maintained.
Remove the filter directory from the build file.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -148,9 +148,6 @@
 		<copy todir="${dist}/db">
 			<fileset dir="${src}/db" excludes="**/.svn/** **/_svn/**" />
 		</copy>
-		<copy todir="${dist}/filter">
-			<fileset dir="${src}/filter" excludes="**/.svn/** **/_svn/**" />
-		</copy>
 		<copy todir="${dist}/plugin">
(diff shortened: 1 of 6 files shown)


kingthorin:
Build failing

thc202:
Fixed.

kingthorin:
Looks good

psiinon:
Looks good but has a conflict now.

thc202:
Conflict resolved.

thc202:
Conflict with #2570 resolved.

Issue #2526, opened 2016-06-02
CA Certs errors


ucugo:
HI, I am getting errors each time i try to import my root CA certs into zap 2.4.3. This is required in order to test in a mutual ssl communication. the stack trace can be seen below...

`java.lang.ArrayIndexOutOfBoundsException: 1
    at org.zaproxy.zap.extension.dynssl.SslCertificateUtils.parseDERFromPEM(Unknown Source)
    at org.zaproxy.zap.extension.dynssl.SslCertificateUtils.pem2Keystore(Unknown Source)
    at org.zaproxy.zap.extension.dynssl.DynamicSSLPanel.doImport(Unknown Source)
    at org.zaproxy.zap.extension.dynssl.DynamicSSLPanel.access$500(Unknown Source)
    at org.zaproxy.zap.extension.dynssl.DynamicSSLPanel$5.actionPerformed(Unknown Source)
    at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
    at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2346)
    at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
    at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
    at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
    at java.awt.Component.processMouseEvent(Component.java:6525)
    at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
    at java.awt.Component.processEvent(Component.java:6290)
    at java.awt.Container.processEvent(Container.java:2234)
    at java.awt.Component.dispatchEventImpl(Component.java:4881)
    at java.awt.Container.dispatchEventImpl(Container.java:2292)
    at java.awt.Component.dispatchEvent(Component.java:4703)
    at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4898)
    at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4533)
    at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4462)
    at java.awt.Container.dispatchEventImpl(Container.java:2278)
    at java.awt.Window.dispatchEventImpl(Window.java:2750)
    at java.awt.Component.dispatchEvent(Component.java:4703)
    at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
[...]

PR #2549, opened 2016-06-07
Tweak import ZAP Root CA certificate error message

thc202:
Change the error message shown when the import of ZAP Root CA
certificate failed to explicitly mention that's attempting to import ZAP
Root CA certificate (not other certificates). Also, merge the resource
messages shown in the error (2nd part is not used in other error
messages).

Related to #2526 - CA Certs errors


kingthorin:
Looks good

PR #2552, opened 2016-06-08
Use default User-Agent for CONNECT requests

thc202:
Change class HttpSender to set the default User-Agent defined in the
options as the default User-Agent for internal CONNECT requests (that
is, CONNECT requests sent by ZAP to the configured outgoing proxy).
Change class HttpMethodDirector to allow to set the default User-Agent
for CONNECT requests.


kingthorin:
Looks good

Issue #265, opened 2015-06-04
Allow to configure outgoing proxy with PAC file/URL


zapbot:
```
Be nice for ZAP to support Proxy Auto Config files.

This might well do the job: http://blogger.ziesemer.com/2009/05/markutils-pacproxyselector-java.html

It requires Mozilla Rhino, but there are plenty of other things we could use a JavaScript
engine for!


```

Original issue reported on code.google.com by `psiinon` on 2012-01-20 17:31:03

afeld:
Seems that this issue should be closed...

![screen shot 2016-01-21 at 11 24 41 am](https://cloud.githubusercontent.com/assets/86842/12486655/949a6592-c031-11e5-8cd2-8da1332a3710.png)

psiinon:
You're right :)

aine-rb:
I guess that we talk about "client-side" of ZAP in this issue. The issue https://github.com/zaproxy/zaproxy/issues/621 talks about "server-side".
What we're looking for here is an engine inside ZAP capable of interpreting javascript function FindProxyForURL(), and a new option in the Connection Options panel allowing the user to configure the outgoing proxy through a Proxy auto-config file accessed by URL (like Firefox does).

I would like to develop this functionality. It will be my first issue on ZAP, if you agree.

thc202:
Updated the issue for that use case.

aine-rb:
The PAC script evaluation functionality is written (see the PACScript class of the org.zaproxy.zap.extension.script package). I would like to let someone else develop the activation of this class, allowing ZAP to configure its outgoing proxy server dynamically through PAC files.

Here is what needs to be done with the PACScript class in order to evaluate PAC files:

    Create a new PACScript object
    Evaluate its content every time it's needed through the evaluate() method.

Examples can be found in the Unit Test class org.zaproxy.zap.extension.script.PACScriptUnitTest of the test directory.

kingthorin:
https://github.com/zaproxy/zaproxy/pull/2556

sudpan:
Hello, I wish to use a pac script file as outgoing proxy in ZAP. I came across this issue and see that parts of it are done. I am not a great coder, so cannot really figure out if the functionality is completely done in the back-end.

[...]

PR #2556, opened 2016-06-09
Add PAC file function evaluation

aine-rb:
This commit adds the PACScript class, which should be used to evaluate Proxy Auto Config files in order to setup dynamically the outgoing proxy server. To do so, it's necessary to:

```
Create a new PACScript object
Evaluate its content every time it's needed through the evaluate() method Examples can be found in the Unit Test file of the class named PACScriptUnitTest.
```

This class depends on the Rhino external library.

### DIFF
--- a/LEGALNOTICE.md
+++ b/LEGALNOTICE.md
@@ -49,6 +49,7 @@ and subject to their respective licenses.
 | commons-text-1.6.jar                | Apache 2.0                |
 | commons-validator-1.6.jar           | Apache 2.0                |
 | diffutils-1.2.1.jar                 | Apache 2.0                |
+| delight-nashorn-sandbox-0.1.25.jar  | Apache 2.0                |
 | ezmorph-1.0.6.jar                   | Apache 2.0                |
 | harlib-1.1.2.jar                    | Apache 2.0                |
 | hsqldb-2.4.1.jar                    | BSD                       |
--- a/zap/src/main/java/org/zaproxy/zap/extension/script/PacScript.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/script/PacScript.java
@@ -1,0 +1,864 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2016 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.script;
+
+import delight.nashornsandbox.NashornSandbox;
+import delight.nashornsandbox.NashornSandboxes;
+import delight.nashornsandbox.exceptions.ScriptAbuseException;
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.nio.charset.StandardCharsets;
+import java.time.Clock;
+import java.time.DayOfWeek;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.WeekFields;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import javax.script.Invocable;
+import javax.script.ScriptException;
+import org.apache.commons.io.IOUtils;
+
+/**
+ * Represents a PAC script file and methods to evaluate its content.
+ *
+ * @author aine-rb
+ */
+public class PacScript {
+
+    private static final String GMT_TIME_ZONE = "GMT";
+    private static final List<String> MONTHS =
+            Collections.unmodifiableList(
+                    Arrays.asList(
+                            "JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT",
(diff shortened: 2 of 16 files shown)


src/org/zaproxy/zap/extension/script/PACScript.java
```diff
@@ -0,0 +1,849 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.zaproxy.zap.extension.script;
+
+import java.io.IOException;
+import java.net.*;
```
kingthorin: Specific imports are preferred over wildcards

src/org/zaproxy/zap/extension/script/PACScript.java
```diff
@@ -0,0 +1,849 @@
...
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.*;
```
kingthorin: Specific imports are preferred over wildcards

src/org/zaproxy/zap/extension/script/PACScript.java
```diff
@@ -0,0 +1,849 @@
...
+            return Context.toString(scriptRes);
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
+            return "";
+        } finally {
+            Context.exit();
+        }
+    }
+
+    private static boolean isDay (String value)
```
kingthorin: I suspect java already has such a mechanism?

For a bunch of the buiLatin date related functionality couldn't a DateValidator or some existing java functionality be used?

src/org/zaproxy/zap/extension/script/PACScript.java
```diff
@@ -0,0 +1,849 @@
...
+        } catch (NumberFormatException e) {
+            return false;
+        }
+
+        return day > 0 && day <= 31;
+    }
+
+    private static boolean isMonth (String value)
+    {
+        String[] months = {"JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC"};
```
kingthorin: Thus series of strings seems to be declared more than once. Would a constant be more appropriate?

test/org/zaproxy/zap/extension/script/PACScriptUnitTest.java
```diff
@@ -0,0 +1,265 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.zaproxy.zap.extension.script;
+
+import java.net.URL;
+import java.nio.file.Paths;
+import java.util.*;
```
kingthorin: Specific imports are preferred over wildcards

kingthorin:
@aine-rb fixup coming?

src/org/zaproxy/zap/extension/script/PACScript.java
aine-rb: What we want here is exactly what is written in the code : some simple functions which try to find out if the String given in argument is a day, a month, or a year. We do not want a DateValidator object or any other Java Date mechanism. It' mainly because of the PAC file standard: The function dateRange() can have different type of values for a same parameter (day, month, year or timezone), but it's never a Java friendly Date format value. It depends on the number of argument values passed to the function and the order of the arguments. It's not really developer-friendly but this standard is a very old one, created for Netscape Navigator back in 1996 and he didn't really change since this moment. Then, we have to deal with it.

aine-rb:
Sorry for the delay, fixup done.
Thank you for your comments kingthorin.

thc202:
Probably better to wait for Java 8 move, that way we just need to support one JavaScript engine (instead of adding a 3rd one) and we can properly sandbox the PAC script.

@aine-rb would you mind allowing modification by maintainers? [1]

[1] https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

kingthorin:
https://github.com/zaproxy/zaproxy/pull/2556

aine-rb:
Allowing edits from maintainers is done.

thc202:
@aine-rb thanks!

psiinon:
Just had a request re #265 and saw the link to this.
We've moved to Java 8, so can we now update and merge this PR?

kingthorin:
Should be reworked for Nashorn.

thc202:
- Rebased to include the build changes (#5369):
   - Moved the files to new locations;
   - Normalised the indentation;
   - Tweaked how the test files are loaded (load as resource instead of using `Paths`);
   - Removed the local Rhino library and use the (latest) one from Maven Central.
 - Removed version JavaDoc tag from the new classes, as we don't maintain that;
 - Changed the JavaDoc of the test class to be the same as the other tests;
 - Addressed deprecation on IOUtils usage in `PACScript`.

thc202:
- Changed to use Nashorn sandboxed (it still needs to be tuned).
   - Remove Rhino dependency and references in JavaDoc.
   - Add dependency on Nashorn sandbox.
 - Throw exceptions when failed to obtain/parse the script instead of just logging and continue.
 - Parse the script once and invoke as needed.
 - Extract constant for the days and use `List`s always for months and days.
 - Make everything that can be private and static.

thc202:
Investigating a test failure:
```
org.zaproxy.zap.extension.script.PACScriptUnitTest > test_timeRange FAILED
    org.junit.ComparisonFailure: expected:<[SUCCESS]> but was:<[FAILURE]>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at org.zaproxy.zap.extension.script.PACScriptUnitTest.test_timeRange(PACScriptUnitTest.java:238)
```
seems unrelated to the above changes, tests were passing some hours ago (locally) and now failing...

thc202:
Updated #265 with remaining tasks (which IMO can be done in follow up PRs, once this is merged).

thc202:
The test is failing because the time checks are incorrect for some hours (e.g. it sums two hours to current hour which at the end of the day gives an invalid hour).

kingthorin:
So the functionality it technically incorrect? Fix inbound?

thc202:
Sorry for not being clear, it's the time checks in the test that are not correct. The implementation of the PAC function is ok (afaict).

Yes, was thinking on changing the tests to use hardcoded values to avoid these cases.

kingthorin:
Okay thanks!

thc202:
Changes done, now using a fixed clock for the tests.

thc202:
Did some cosmetic changes:
 - Rename `PACScript` to `PacScript`;
 - Rename variables to start with lower case;
 - Remove `test_` prefix from test files (that's already implied by being in the test resources dir).

thc202:
@aine-rb thanks!

aine-rb:
Thank you for your modifications and updates.

PR #2557, opened 2016-06-09
Do not add a User-Agent header by default

thc202:
Change class HttpMethodBase to not add a User-Agent header by default,
thus reducing the (unwanted) changes done to the original HTTP message
being sent.


kingthorin:
Looks good

PR #2558, opened 2016-06-09
Fix NPE when invoking context menu in Alerts tab

thc202:
Fix NullPointerException when attempting to invoke the context menu of
Alerts tab using the keyboard when no entry was yet selected, following
excerpt of the exception's stack trace:
[AWT-EventQueue-0] ERROR org.zaproxy.zap.ZAP$UncaughtExceptionLogger  -
 Exception in thread "AWT-EventQueue-0"
 java.lang.NullPointerException
  at o.z.z.extension.alert.AlertPanel$5.keyPressed(AlertPanel.java:485)

the exception happened because there was no entry selected, therefore no
location existed (that is, null) when calculating where to show the pop
up menu.

Change class AlertPanel to set a JPopupMenu to the Alerts tree, which
already takes care to show the pop up menu in the most appropriate
location (when invoked with keyboard and with the mouse).

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/AlertPanel.java
+++ b/src/org/zaproxy/zap/extension/alert/AlertPanel.java
@@ -21,15 +21,16 @@
 package org.zaproxy.zap.extension.alert;
 
 import java.awt.CardLayout;
+import java.awt.Component;
 import java.awt.Dimension;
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

PR #2562, opened 2016-06-10
Merge docker and non docker versions, add OOS patterns and sort alerts

psiinon:


psiinon:
I'll raise a PR to copy this to the zap-api-python repo once its been merged

kingthorin:
Looks good

build/docker/zap-baseline.py
```diff
@@ -266,7 +342,7 @@ def main(argv):
 
       if len(generate) > 0:
         # Create the config file
-        with open('/zap/wrk/' + generate, 'w') as f:
+        with open(generate, 'w') as f:
```
thc202: Should use `base_dir +` too? (like config and XML report are using)

build/docker/zap-baseline.py
```diff
@@ -289,35 +365,50 @@ def main(argv):
...
+          if key in config_msg:
+            user_msg = config_msg[key]
+          print_rule(config_dict[key], alert_list, detailed_output, user_msg)
+          fail_count += 1
+          
       if len(report_html) > 0:
-        # Save the HTML report
-        with open('/zap/wrk/' + report_html, 'w') as f:
+        # Save the report
+        with open(report_html, 'w') as f:
```
thc202: Should use `base_dir +` too?

build/docker/zap-baseline.py
```diff
@@ -289,35 +365,50 @@ def main(argv):
...
-        else:
-          action = 'WARN'
+
+      # print out the warning rules
+      for key, alert_list in sorted(alert_dict.iteritems()):
+        if (not config_dict.has_key(key)) or (config_dict[key] == 'WARN'):
+          user_msg = ''
+          if key in config_msg:
+            user_msg = config_msg[key]
+          print_rule('WARN', alert_list, detailed_output, user_msg)
```
thc202: It could use `config_dict[key]` instead of the harcoded value like the other loops? Or, a function extracted? (It's the same logic for all, count and print.)

psiinon: It cant use config_dict[key] as it needs to cope with the case where its not present in the dictionary - (not config_dict.has_key(key))

psiinon:
Updated for all apart from the last comment, happy to address that in another way :)

build/docker/zap-baseline.py
thc202: Right. missed than detail.

thc202:
Thanks!

PR #2563, opened 2016-06-10
Updated Weekly Dockerfile to reduce size

raesene:
So this makes some changes to the Dockerfile syntax to try and reduce the size without (hopefully!) breaking anything.

I noticed from https://imagelayers.io/?images=owasp%2Fzap2docker-weekly:latest that quite a bit of the size was coming from chown and chgrp operations which were essentially recreating a load of files (due to the copy-on-write filesystem that docker uses) so used the USER directive to go to the zap user early.

The only bit that doesn't work for is COPY operations (this is an open issue with Docker), so had to do some limited chown/chmod stuff there.

Also collapsed all the apt-get items to one layer to make the clean up work well.

As a first pass this gets the Docker image down from 1.724GB to 1.069GB.  There's probably more that could be done but this looked like a decent quick win.

### DIFF
--- a/build/docker/Dockerfile-weekly
+++ b/build/docker/Dockerfile-weekly
@@ -2,29 +2,62 @@
 FROM ubuntu:15.04
 MAINTAINER Simon Bennetts "psiinon@gmail.com"
 
-RUN apt-get update && apt-get clean
-RUN apt-get install -q -y make automake autoconf gcc g++ openjdk-7-jre-headless openjdk-7-jdk ruby wget curl xmlstarlet unzip git x11vnc xvfb openbox xterm net-tools ruby-dev && apt-get clean
+
+RUN apt-get update && apt-get install -q -y --fix-missing \
+	make \
+	automake \
+	autoconf \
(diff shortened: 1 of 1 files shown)


psiinon:
Nice! Fraid its a bit too late to test it today, and I'm probably going to be very busy next week (its a Mozilla work week) but I'll do my best to look at it asap. Hopefully @thc202 will also be able to check it as well ;)

thc202:
Looks good to me. Tested some of the functionalities (zap.sh, zap-baseline.py, zap-webswing.sh, zap-cli) and, as expected, all worked fine.

psiinon:
@raesene many thanks :) How would you like to be included in the credits? 

raesene:
cool, glad it worked :)

For credits - Rory McCune - NCC Group PLC

Would be cool, thanks :)

thc202:
Done in zaproxy/zap-core-help@ccfededea39ed432a5479de2b38d919fb0d6b0a5. Thanks!

PR #2564, opened 2016-06-10
Honour Mode and scope when following redirections

### LINKED ISSUE #2546: ZAP access URLs which are out of scope
Hello ZAP team,

I've found a case that ZAP access URLs which are out of scope on 'protected mode'.

Steps to reproduce:

1) Run ZAP 2.5.0(or ZAP 2.4.3)

2) Run Fiddler or Burp Suite

3) Set up proxy network configration to:
ZAP - Fiddler or Burp Suite - Target Site
(The reason of use Fiddler or Burp Suite is to logging ZAP raw access.)

4) Make this php (or same logic other program), and put it on webserver.
(In this example, this php is: http://localhost/zaptest/redirect.php)

[redirect.php]
```
<?php
header('Location: https://www.*****.com/');
exit;
?>
```
In this source, https://www.*****.com/ is example. Please change this URL to your own site that can attack by ZAP.

- CASE-A  - ZAP does not access URL in out of scope (Normal Case)

5-A) Access the URL below

`http://localhost/zaptest/redirect.php`

- CASE-B  - ZAP access the URL that out of scope (Probrem Case)

5-B) Access the URL below (NOTICE: In this case, URL parameter (?a=b) is needed)

`http://localhost/zaptest/redirect.php?a=b`

6) Set ZAP mode to 'protected mode', and set 'http://localhost/' into default context.

7) Right click the URL in history that 

CASE-A:
`http://localhost/zaptest/redirect.php`

CASE-B:
`http://localhost/zaptest/redirect.php?a=b`

And choice Active scan.Set 'Injectable Targets' to All,
And set scan policy to all threshold off except 'Server Side Include'.
Set threshold of 'Server Side Include' to default (execute this test only).
Then Start active scan.

Then, duaring activescan, check the log of Fiddler(or Burp Suite).
In CASE-A, ZAP does not access to 'https://www.*****.com/'.
There is no probrem.

But in CASE-B, ZAP access to 'https://www.*****.com/' like this log below.

> GET http://localhost/zaptest/redirect.php?a=%22%3E%3C%21--%23EXEC+cmd%3D%22ls+%2F%22--%3E%3C
> GET https://www.*****.com/   <<< access to out of scope url by redirect
> GET http://localhost/zaptest/redirect.php?a=%22%3E%3C%21--%23EXEC+cmd%3D%22dir+%5C%22--%3E%3C
> GET https://www.*****.com/
>  .
>  .
> GET http://localhost/zaptest/redirect.php?a=b  (* User-Agent:)
> GET https://www.*****.com/  (* User-Agent: )
>  .
>  .
> GET http://localhost/zaptest/redirect.php?a=b   (* User-Agent:)
> GET https://www.*****.com/  (* User-Agent: )
>  .
>  .

(The ZAP active scan log doesn't show access to https://www.*****.com.
Fiddler or Burp log shows that ZAP access to https://www.*****.com. )

So in CASE-B, ZAP Attacks the URL that out of scope.
Is this bug or by design? 

Regards.

thc202:
Change classes AbstractPlugin (active scanners) and Analyser (pre active
scanner task) to honour scan's scope when following redirections (which
already includes the Mode).
Change class HttpPanelSender ("Manual Request Editor" and "Resend"
dialogues) to honour the Mode when following redirections.
Change class CoreApi to honour the Mode when sending manual requests and
following redirections, either plain text or in HAR format (as with UI
the requests are not allowed if in 'Safe' mode or in 'Protected' mode if
out of scope).
Change class HttpSender to allow to validate the redirections before
following them.

Fix #2546 - ZAP access URLs which are out of scope


kingthorin:
Looks good

PR #2570, opened 2016-06-13
Change proxy option to remove unsupported encoding

thc202:
Change Local Proxy option Modify/Remove "Accept-Encoding" request-header
to Remove Unsupported Encodings, as that prevents unsupported encodings
from being used, the previous behaviour just removed specific encodings
(for example, deflate, compress, gzip) allowing to pass/use the ones
that were also not supported like SDCH, brotli and others.

More detailed changes:
 - Messages.properties, update label and tool tip to match the behaviour
 of the new option;
 - Constant, convert the old option to be under a new key configuration
 when updating from 2.5.0;
 - ProxyParam, add methods to get/set the new option and deprecate the
 methods of the old option and rename the existing variables and its
 configuration key;
 - ProxyThread, change to remove the unsupported encodings instead of
 just some (currently all encodings are removed as the HttpBody
 implementations do not handle the encodings);
 - OptionsLocalProxyPanel, rename the variable/methods of the old option
 and change to use the new methods of ProxyParam class.


kingthorin:
Looks good

PR #2571, opened 2016-06-13
Restore the name of main/repo license file

thc202:
The LICENSE file was renamed to LICENSE-2.0.txt to avoid issues in case
insensitive file systems, clashed with a directory named license used by
the ZAP API Java Client. With the move of the API client to another repo
the license directory no longer exists and the license file can have the
previous name.


kingthorin:
Looks good

PR #2579, opened 2016-06-14
Select options panel by clicking in its row

### LINKED ISSUE #2578: Minor Usability Issue: Must click on text in Options column to select row
When you bring up the options dialog, if you click on a row outside of where the text in that row is displayed, nothing happens. You have to actually click on the text, which is annoying, particularly for short option names like JVM. Please fix it so if you click anywhere in that row that row is selected.

thc202:
Change class AbstractParamContainerPanel to select (thus show) the
options panel by clicking in its row (or in a closer location), not just
when clicked its node.
This change affects the Options dialogue and all other dialogues that
use the above class, like Session Properties dialogue.

Fix #2578 - Must click on text in Options column to select row


kingthorin:
Looks good

PR #2585, opened 2016-06-15
Remove temp Sequence requests on session clean up

thc202:
When creating the structure of the sequence to be attacked some
temporary HTTP requests are created, although they are never removed,
being temporary they should be (at the very least) removed when the
session is cleaned up.

More detailed changes:
 - HistoryReference, introduce a new type TYPE_SEQUENCE_TEMPORARY,
 deprecate TYPE_RESERVED_11 (both have the same number) and add the new
 type to set of temporary types.


kingthorin:
Looks good

psiinon:
TYPE_RESERVED_11 was actually reserved for Zest - its hard coded to 11 in https://github.com/zaproxy/zap-extensions/blob/beta/src/org/zaproxy/zap/extension/zest/ZestZapRunner.java#L430 and it looks like this has at some point been incorrectly changed to 12 :/ We'll need to sort this out..

thc202:
OK, changed to use another value for temporary sequence requests.

Will take a look at the other type/values.

kingthorin:
Fixup into a single commit?

PR #2586, opened 2016-06-15
Use option All Requests from Active Scan dialogue

thc202:
The option All Requests (sub-option of HTTP Headers input vector) shown
in Active Scan dialogue was not being used, because the HTTP Header
variant implementation was checking the global options instead of the
given scan options, which includes the overrides done in the Active Scan
dialogue.

More detailed changes:
 - AbstractAppParamPlugin, change to evaluate if the HTTP Header variant
 should be included or not based on the given scan options (the logic
 was moved from the variant to this class which already has the scan
 options);
 - VariantHeader, remove the logic used to check if a message should be
 processed or not, that's now done in and by AbstractAppParamPlugin.
 Document the class and some methods and normalise its behaviour.
 - VariantHeaderUnitTest, add tests to assert the expected behaviour of
 class VariantHeader.


kingthorin:
Looks good

PR #2589, opened 2016-06-16
Add cookie handling when a redirect occurs

aine-rb:
Set-Cookie header is now handled when receiving a redirect response.

Cookies and session are correctly updated when sending a POST request with redirect enabled.
Session is updated at the end of the redirect process for GET requests with redirect enabled.

### DIFF
--- a/src/org/apache/commons/httpclient/HttpMethodDirector.java
+++ b/src/org/apache/commons/httpclient/HttpMethodDirector.java
@@ -31,9 +31,12 @@
 package org.apache.commons.httpclient;
 
 import java.io.IOException;
+import java.net.HttpCookie;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -54,6 +57,7 @@ import org.apache.commons.httpclient.params.HttpMethodParams;
 import org.apache.commons.httpclient.params.HttpParams;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.parosproxy.paros.network.HttpHeader;
 
 /*
  * Forked class...
@@ -104,6 +108,8 @@ public class HttpMethodDirector {
 
     private static final Log LOG = LogFactory.getLog(HttpMethodDirector.class);
 
+    private static List<HttpCookie> redirectCookies = new ArrayList<>();
+
     private ConnectMethod connectMethod;
     
     private HttpState state;
@@ -137,8 +143,7 @@ public class HttpMethodDirector {
         this.state = state;
         this.authProcessor = new AuthChallengeProcessor(this.params);
     }
-    
-	
+
     /**
      * Executes the method associated with this method director.
      * 
@@ -201,7 +206,6 @@ public class HttpMethodDirector {
                     fakeResponse(method);
                     break;
                 }
-                
                 boolean retry = false;
                 if (isRedirectNeeded(method)) {
                     if (processRedirectResponse(method)) {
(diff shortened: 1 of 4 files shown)


src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -31,11 +31,8 @@
 package org.apache.commons.httpclient;
 
 import java.io.IOException;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.Map;
-import java.util.Set;
+import java.net.HttpCookie;
+import java.util.*;
```
kingthorin: Please use specific imports not wildcards

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -253,6 +250,66 @@ public void executeMethod(final HttpMethod method) throws IOException, HttpExcep
...
+    }
+
+    private void setRedirectCookies(HttpMethod method) {
+        // This function does not set the session. It saves cookies received from the Set-Cookie headers in the static
+        // redirect cookie list and creates a new Cookie header. These cookies are then correctly processed in the
+        // HttpSender code. This should fix the problem of Set-Cookie headers received during redirects in most of cases
+
+        // Retrieve cookies from Set-Cookie headers
+        ArrayList<HttpCookie> cookiesToSet = new ArrayList<>();
+        Header[] setCookieHeaders = method.getResponseHeaders("Set-Cookie");
```
kingthorin: This should use a constant, perhaps even those defined in HttpHeader.

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -253,6 +250,66 @@ public void executeMethod(final HttpMethod method) throws IOException, HttpExcep
...
+        redirectCookies.addAll(cookiesToSet);
+
+        // Retrieve new cookie names
+        ArrayList<String> cookieNames = new ArrayList<>();
+        for (HttpCookie cookie : cookiesToSet) {
+            cookieNames.add(cookie.getName());
+        }
+
+        // Add cookies from the request Cookie header in the list if their name is not already present
+        Header cookieHeader = method.getRequestHeader("Cookie");
```
kingthorin: Same comment regarding constant use.

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -253,6 +250,66 @@ public void executeMethod(final HttpMethod method) throws IOException, HttpExcep
...
+
+        // Construct the Cookie header value
+        String cookieHeaderValue = "";
+        for (HttpCookie cookie : cookiesToSet) {
+            cookieHeaderValue += cookie.toString() + "; ";
+        }
+        if (cookieHeaderValue.length() > 2) {
+            cookieHeaderValue = cookieHeaderValue.substring(0, cookieHeaderValue.length() - 2);
+            // Set the new Cookie header in the request
+            method.setRequestHeader("Cookie", cookieHeaderValue);
```
kingthorin: Same comment regarding constant use.

src/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -58,14 +58,12 @@
...
+// ZAP: 2016/06/15 Update session and Cookie header if a Set-Header header has been received in a redirect after a POST
 
 package org.parosproxy.paros.network;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
+import java.util.*;
```
kingthorin: This should  use specific imports not wildcards

src/org/zaproxy/zap/extension/httpsessions/ExtensionHttpSessions.java
```diff
@@ -20,14 +20,10 @@
 import java.net.HttpCookie;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
kingthorin: This should use specific imports not wildcards

kingthorin:
Pull request has conflicts

aine-rb:
Conflicts resolved.
Thank you for your comments kingthorin !

kingthorin:
@psiinon @thc202 separate from the conflict do you have any feedback on this PR?

thc202:
Yes, I've some concerns regarding the implementation, will add the exact comments soon.

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -104,6 +108,8 @@
 
     private static final Log LOG = LogFactory.getLog(HttpMethodDirector.class);
 
+    private static List<HttpCookie> redirectCookies = new ArrayList<>();
```
thc202: This will not work properly when multiple (independent) requests are sent at the same time.

src/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -384,24 +386,35 @@ public void sendAndReceive(HttpMessage msg, boolean isFollowRedirect) throws IOE
 				sendAuthenticated(msg, isFollowRedirect);
 				return;
 			}
+			// POST/PUT method cannot be redirected by library. Need to follow by code
 
 			// ZAP: Reauthentication when sending a message from the perspective of a User
 			sendAuthenticated(msg, false);
 
+			// Update session and Cookie header if a Set-Cookie header has been received
+			notifyResponseListeners(msg);
```
thc202: This assumes that it's being used one of the `HttpSession`s, but that might not be the case, for example, when sending a message with a user.

src/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -384,24 +386,35 @@ public void sendAndReceive(HttpMessage msg, boolean isFollowRedirect) throws IOE
...
 			}
+			// POST/PUT method cannot be redirected by library. Need to follow by code
 
 			// ZAP: Reauthentication when sending a message from the perspective of a User
 			sendAuthenticated(msg, false);
 
+			// Update session and Cookie header if a Set-Cookie header has been received
+			notifyResponseListeners(msg);
+			HttpSession session = msg.getHttpSession();
+			if (session != null) {
```
thc202: This should be done by `ExtensionHttpSessions`, this class shouldn't need to set/use the `HttpSession` directly.

thc202:
Clearly this is not an easy problem to fix, based on how the redirects are being handled (i.e. processed internally by the HTTP client implementation) and how multiple components use/set the HTTP state (e.g. `User` or `HttpSession`). Still checking how to best solve this...

smichka:
And? Was the issue solved yet?

aine-rb:
I agree with you thc202 : This is not really clean code and it certainly doesn't handle every possible scenario.
However, the way I implemented the fix was the best I found regarding how it already worked. At least it deals with the main cases, as I mentioned in my first post.

kingthorin:
@aine-rb could you set allow modification by maintainers on this one too?

aine-rb:
Done kingthorin.

kingthorin:
Thanks

vegardvaage:
I've bumped into this when fuzzing a web site that gives a 302 redirect (either to the protected content or back to the login page) and sets a cookie that determines your access level in the 302 response header. The "follow redirect" option in the fuzzing options works, but the new cookie is not added when the redirect is followed. 

Is this fix any closer to a resolution?

kingthorin:
There's been a bit of progress but there's no ETA. Some research/planning has been done but no further coding yet.

thc202:
This ended up being a problem in `HttpSender` which was not setting the user to the redirect messages, thus ignoring cookies set in following redirections (or not sending cookies at all).

Closing, superseded by #4983.

PR #2596, opened 2016-06-16
Remove SessionChangedListener from ExtensionAPI

thc202:
The notifications provided by SessionChangedListener are not in
effective use.


kingthorin:
Looks good

PR #2597, opened 2016-06-17
Be lenient when getting charset from Content-Type

thc202:
Change the regular expression used to get the charset defined in the
Content-Type header to also accept/ignore single quote chars around the
charset value (even if that's not the expected quoting, which should use
double quote chars).
Some content producers might use single quotes instead of double which
would not work as expected when proxying through ZAP.
For example, the charset extracted from the following Content-Type
header value:
 `text/xml;charset='utf-8'`

would be exactly `'utf-8'` which is not a valid name leading to an
exception:
 `java.nio.charset.IllegalCharsetNameException: 'utf-8'`

and no charset being set to the body.


thc202:
An example in the wild:
https://github.com/sympmarc/SPServices/blob/5780ebad09fb2cd2aa7abb7709540f3ad5e0fee0/src/core/SPServices.core.js#L1269

kingthorin:
Looks good

PR #2600, opened 2016-06-17
Remove redundant inits in HTTP message classes

thc202:
Remove redundant instance variable initialisations in HTTP message
related classes:
 - HttpHeader:
  - Remove initialisations done in the instance variables as they are
  already being initialised either in the constructors or other methods
  called by the constructor;
  - HttpHeader(String), do not call the default constructor it would
  lead to redundant initialisations, since the variables are already
  being initialised in the method called, setMessage(data);
  - setMessage(data), change to call clear() instead of init() which
  also clears the state of extending classes and remove the
  initialisation of mMsgHeader, which is done in the following method,
  parse(String).
 - HttpRequestHeader:
  - Remove initialisations done in the instance variables as they are
  already being initialised either in the constructors or other methods
  called by the constructors;
  - HttpRequestHeader(), change to not call clear() and initialise only
  the required instance variables;
  - HttpRequestHeader(String) and HttpRequestHeader(String, boolean), do
  not call the default constructor it would lead to redundant
  initialisations, since the variables are already being initialised in
  the method called, setMessage(data);
 - HttpResponseHeader:
  - Remove initialisations done in the instance variables as they are
  already being initialised either in the constructors or other methods
  called by the constructors;
  - HttpResponseHeader(), change to not call clear() and initialise only
  the required instance variables;
  - HttpResponseHeader(String), do not call the default constructor and
  use base constructor, HttpHeader(String), which already takes care to
  call the appropriate method.

### DIFF
--- a/src/org/parosproxy/paros/network/HttpHeader.java
+++ b/src/org/parosproxy/paros/network/HttpHeader.java
@@ -33,6 +33,7 @@
 // the request body is not "x-www-form-urlencoded"
 // ZAP: 2015/03/26 Issue 1573: Add option to inject plugin ID in header for all ascan requests
 // ZAP: 2016/06/17 Be lenient when parsing charset and accept single quote chars around the value
+// ZAP: 2016/06/17 Remove redundant initialisations of instance variables
 
 package org.parosproxy.paros.network;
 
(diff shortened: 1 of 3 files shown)


src/org/parosproxy/paros/network/HttpHeader.java
```diff
@@ -104,13 +105,13 @@
...
-    protected Hashtable<String, Vector<String>> mHeaderFields = new Hashtable<>();
-    protected int mContentLength = -1;
-    protected String mLineDelimiter = CRLF;
-    protected String mVersion = HttpHeader.HTTP10;
+    protected String mStartLine;
+    protected String mMsgHeader;
+    protected boolean mMalformedHeader;
+    protected Hashtable<String, Vector<String>> mHeaderFields;
+    protected int mContentLength;
+    protected String mLineDelimiter;
```
kingthorin: If CRLF is declared then this variable is later set to that const, why not just use the const throughout?

thc202: As is now, it might be changed by extending classes.

kingthorin:
Ok, now the conflict just needs to be addressed.

thc202:
Should be resolved.

kingthorin:
Looks good

PR #2603, opened 2016-06-20
Add new history type for Zest HTTP messages

thc202:
Add new type to HistoryReference called TYPE_ZEST_SCRIPT to be used by
HTTP messages sent by Zest scripts, with same value as the (now)
deprecated type TYPE_RESERVED_11 (the type originally intended to be
used by Zest scripts but its value was incorrectly changed to other
value).

Related to #2585 (Remove temp Sequence requests on session clean up)
which adds new type for temporary sequence messages, replacing the use
of the (old) value for Zest scripts.


kingthorin:
Looks good

psiinon:

thc202:
Conflicts resolved.

PR #2604, opened 2016-06-20
Remove unnecessary extensions' constructors

thc202:
Remove unnecessary extensions' constructors, they are not required nor
used by core code to start/load the extensions (it's enough with the
no-arg constructor moreover add-ons do not need nor should start other
extensions). Also, in the cases where an auxiliary "initialize" method
was in use it was merged into the no-arg constructor (in most of the
cases it was only being called by the no-arg constructor).


kingthorin:
Looks good

Issue #1871, opened 2015-09-06
Scan rules don't manage correctly gzip compressed response content


yhawke:
While doing a pentest I discovered that ZAP was unable to find an existing ExternalRedirect vulnerability.
after troubleshooting I discoverde that the original request is set using the Accept-Encoding: gzip header, so when the redirect occurs, the content given back by www.google.it (the site used for testing) is gzipped and the regexes don't work because the content isn't the real one.

Checking inside  the ZAP panel implementation, I saw that the response visualization panel perform the ungzipping of the content while showing it... so everything seems working, but inside the core all plugins contine to working with an encoded content.

This mean that all the plugins could suffer about this implementation if the original request has the Accept-Encoding header set to gzip and the server supports it. Every time a msg.getResponseBody().toString() is performed the resulting string cannot be used correctly to search for regex...

kingthorin:
Good find.

psiinon:
Whats the best way to handle this?
Some options that spring to mind:
* Always removing 'Accept-Encoding: gzip' in reqs from the active scanner
* Automatically replacing the gzipped content with the uncompressed content on ascan responses
* Providing a ResponseBody.toUncompressedString() or similar method and checking all ascan rules use it

kingthorin:
This isn't necessarily limited to active scan. Passive scanners may also be missing findings in compressed responses.

psiinon:
Ugh :(
Maybe we should always return uncompressed content in ResponseBody.toString() and add a ResponseBody.getUncompressedContent() for code that needs it?

thc202:
I'd go with toString() return the content uncompressed, always (and throw an exception if the encoding is not supported).
I don't think we need another method to obtain uncompressed content, it can be used getBytes().

BTW, this duplicates #408.

thc202:
@yhawke was the option "Modify/Remove 'Accept-Encoding' request-header" [1] enabled when that happened?

[1] https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsLocalproxy#modifyremove-accept-encoding-request-header

kingthorin:
I like the suggestion that we `always return uncompressed content in ResponseBody.toString()`

ElColmo:
If ResponseBody.toString() always returns the decompressed/inflated content, how would a plugin get the compressed/deflated content?  Think about BREACH detection, for instance.   Does "getBytes()" not return the inflated content?  I thought it did.

[...]

PR #2605, opened 2016-06-21
Prevent GUI hang when adding messages to History

thc202:
Change ExtensionHistory to add the HTTP messages (HistoryReference)
asynchronously in the EDT to prevent a deadlock between the EDT and
thread(s) adding messages to the History tab. The ExtensionHistory
synchronises on the history table model object when adding the messages
but it would also add the message synchronously in the EDT, so it would
wait for the EDT while holding the lock, if the EDT tried to add a
message in the meantime a deadlock would happen (since it would try to
sync with the history table model object, already locked).
  ---
Issue mentioned in https://github.com/zaproxy/zaproxy/issues/1871#issuecomment-226903369 (a thread dump was provided through IRC).

### DIFF
--- a/src/org/parosproxy/paros/extension/history/ExtensionHistory.java
+++ b/src/org/parosproxy/paros/extension/history/ExtensionHistory.java
@@ -68,6 +68,7 @@
 // ZAP: 2016/05/20 Moved purge method to here from PopupMenuPurgeSites
 // ZAP: 2016/05/30 Issue 2494: ZAP Proxy is not showing the HTTP CONNECT Request in history tab
 // ZAP: 2016/06/20 Removed unnecessary/unused constructor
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good.

As a personal education point, when using invokeAndWait why is it always in a try/catch whereas invokeLater never seems to be? Is that just a style choice inherited from the Paros roots of the project? Or is it that "later" it's impractical to figure out what went wrong?

psiinon:

thc202:
Conflicts resolved.

@kingthorin the last statement/question, the runnable of the invokeLater is executed sometime in the future so the executing thread does not have to wait for the EDT (thus no interrupted exception) nor is the EDT able to propagate any exception thrown by the code being executed (while the invokeAndWait might lead to those exceptions, requiring usually the try-catch statements).

kingthorin:
Thanks

PR #2606, opened 2016-06-21
Tweak how Java version is obtained in zap.sh

### LINKED ISSUE #958: Java version identification when an environment variable for Java is exported
```
zaproxy cannot get Java version when you export an environment variable for Java (e.g.
"export _JAVA_OPTIONS='-Dawt.useSystemAAFontSettings=on'" in ~/.bashrc) on Linux.
This is because the line where the version string appear in 'java -version' command's
output is not the first line but zap.sh use the first line. So I think we'd better
use the following.

JAVA_VERSION=$(java -version 2>&1 | awk -F\" '/version/ { print $2 }')
(This is based on line 4 of svn/trunk/src/zap.sh)

I hope this helps
```

Original issue reported on code.google.com by `redhouse2005` on 2013-12-20 09:30:26

thc202:
Apply the command suggested in the issue report to obtain the Java
version, it's more robust when the environment variable _JAVA_OPTIONS is
defined (that is, Java version is correctly obtained independently of
the values that the variable has).

Fix #958 - Java version identification when an environment variable for
Java is exported

### DIFF
--- a/src/zap.sh
+++ b/src/zap.sh
@@ -39,10 +39,9 @@ if [ "`echo ${JAVA_OUTPUT} | grep "continuing with system-provided Java"`" ] ; t
   echo "    " $JAVA_OUTPUT
   echo "Unsetting JAVA_HOME and continuing with ZAP start-up"
   unset JAVA_HOME
-  JAVA_OUTPUT=$(java -version 2>&1)
 fi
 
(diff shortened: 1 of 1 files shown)


psiinon:
Looks good

Issue #2602, opened 2016-06-20
Java 9 (9-ea)


alfsb:
Zap fails with new Java 9:

    ~/ZedAttackProxy/zap250$ ./zap.sh
    ./zap.sh: linha 49: [: 9-ea: esperado expressão de número inteiro
    Exiting: ZAP requires a minimum of Java 7 to run, found 9-ea

alfsb:
May be related. Removing the `if` on `zap.sh` line 49 results in:

    ~/ZedAttackProxy/zap250$ ./zap.sh 
    Available memory: 16004 MB
    Setting jvm heap size: -Xmx512m
    Can't find bundle for base name Messages, locale en_GB

mozfreddyb:
I have a similar problem, but the version reported is "9-ubuntu".
> zap.sh: line 48: [: 9-Ubuntu: integer expression expected
> Exiting: ZAP requires a minimum of Java 7 to run, found 9-Ubuntu

thc202:
All Java 9 related issues reported so far should be fixed. Leaving this open for now.

thc202:
Per #3919, ZAP does not compile with Java 9.

thc202:
As reported in #3934, the add-on Access Control Testing does not work correctly with Java 9:
```
120727 [AWT-EventQueue-0] ERROR org.zaproxy.zap.ZAP$UncaughtExceptionLogger  - Exception in thread "AWT-EventQueue-0"
java.lang.ExceptionInInitializerError
        at org.zaproxy.zap.extension.accessControl.view.ContextAccessControlPanel$AccessRuleNodeCellRenderer.getTreeCellRendererComponent(ContextAccessControlPanel.java:300)
        at org.jdesktop.swingx.JXTree$DelegatingRenderer.getTreeCellRendererComponent(JXTree.java:1269)
        at java.desktop/javax.swing.plaf.basic.BasicTreeUI$NodeDimensionsHandler.getNodeDimensions(Unknown Source)
        at java.desktop/javax.swing.tree.AbstractLayoutCache.getNodeDimensions(Unknown Source)
        at java.desktop/javax.swing.tree.VariableHeightLayoutCache$TreeStateNode.updatePreferredSize(Unknown Source)
        at java.desktop/javax.swing.tree.VariableHeightLayoutCache.rebuild(Unknown Source)
        at java.desktop/javax.swing.tree.VariableHeightLayoutCache.setModel(Unknown Source)
[...]

PR #2607, opened 2016-06-21
Add lang directory to manifest of ZAP JAR

thc202:
Change build.xml file to include the "lang" directory in the manifest of
ZAP jar, so that it's automatically included in the class path. The
"lang" directory contains the resource messages required to run ZAP.
In newer Java 9 (EA) versions the application class loader is no longer
an instance of URLClassLoader ("JEP 261: Module System") making ZAP to
fail (skip) to dynamically add the "lang" directory to the class path.

Related to #2602 - Java 9 (9-ea)

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -213,7 +213,7 @@
 				<fileset dir="${dist}/lib">
 					<include name="**/JBroFuzz.jar"/>
 				</fileset>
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

PR #2611, opened 2016-06-22
Change HTTP breakpoint dialogues to modal

thc202:
Change the (HTTP) breakpoint dialogues to be modal, to prevent
interactions with other UI components while the dialogues are shown.
Reduces code complexity in core (and add-ons), it's also less error
prone, since it no longer requires the break implementations and core
code to track/tell which type of dialogues (add/edit/remove) are
currently shown.
Deprecate methods used to track/tell if the dialogues are shown in
favour of using modal dialogues.


kingthorin:
Looks good

PR #2612, opened 2016-06-22
Remove deprecation in ExtensionBreak

thc202:
The separator being added to "Analyse" top level menu is not necessary,
by default it's just shown one menu item and it's unrelated to
breakpoint functionalities.


kingthorin:
Looks good

PR #2621, opened 2016-06-24
Minor tidy up of SiteTreeFilter class

thc202:
Remove unused instance variable ("logger") and unnecessary instance
variable initialisation (it's already initialised to false).


kingthorin:
Looks good

PR #2632, opened 2016-06-27
Reduce log level in PluginFactory

thc202:
Change method PluginFactory.loadAllPlugin(Configuration) to reduce the
log level from INFO to DEBUG when a scanner is loaded/initialised. The
log statement is now done with an existing DEBUG statement that
indicates the threshold and strength of the scanner being loaded.
Logging with INFO level is too verbose, the log is filled with not so
useful statements each time the active scan policies are loaded/used,
moreover the active scanner already indicates (as INFO) which scanners
are being used from the selected scan policy.


kingthorin:
Looks good

PR #2633, opened 2016-06-28
Enhance Client Cert File Chooser

kingthorin:
This is a minor enhancement to the client certificate file chooser,
which set the filter for the file chooser to be directory, .p12 files,
and now .pfx files.

.pfx is a PKCS#12 archive.

See also: https://en.wikipedia.org/wiki/PKCS_12


thc202:
It's missing the ZAP comment, other than that looks good,

kingthorin:
Done

thc202:
Looks good to me, thank you!

PR #2635, opened 2016-06-28
Do not start filter thread by default

thc202:
Change ExtensionFilter to only start the timer thread when really
needed, that is, only if at least one filter is enabled.
Filters are not enabled by default so the thread does not need to be
started by default, when ZAP starts, moreover filters are/were replaced
by scripts so its probably not needed in most of the times.


kingthorin:
Looks good

PR #2639, opened 2016-06-29
Do not log when cloning PluginFactory

thc202:
Change method PluginFactory.addPlugin(Plugin) to not do any logging, by
moving the logging to a more appropriate method, addPlugin(String).
The former method is also called when cloning the PluginFactory, which
does not require logging that information once again.

Related to #2632 - Reduce log level in PluginFactory


kingthorin:
Looks good

PR #2640, opened 2016-06-29
Provide the reason why scanners were skipped

### LINKED ISSUE #2415: Show the reason why an active scanner was skipped
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/AlT3gpxale8/discussion
 ---
ZAP shows that the active scanner was skipped but there's no indication why. It should be possible to know, through the GUI and the ZAP API, why an active scanner was skipped.
There are several reasons why an active scanner can be skipped, for example:
 - The target technologies were not included in the active scan (e.g. SQL Injection when DB is not selected);
 - Prerequisites of the scanner were not fulfilled (e.g. Possible Username Enumeration when no login URL was defined);
 - There's no reason to run the scanner (e.g. Script Active Scan Rules when there's no Active Rules enabled). 

Other than the core changes to show the reason it requires changes to allow scanners to report a custom reason (when using prerequisites) and change the scanners (mostly in add-ons) to specify the reason.

thc202:
Change class HostProcess to allow to specify (and obtain) the reason why
a scanner was skipped, to specify the reason why the scanner was skipped
when it does not target the selected technologies and to log the reason
when the scanner finishes (if available).
Change class ScanProgressItem to set a reason when the scanner is
skipped by user action.
Change ScanProgressActionIcon to show the reason why a scanner was
skipped (in the tool tip text of the skipped icon), also include the
reason when copying the scanners progress to the clip board.

Fix #2415 - Show the reason why an active scanner was skipped


src/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -555,22 +556,66 @@ public boolean handleAntiCsrfTokens() {
     }
 
     /**
-     * Skip the current executing plugin
-     * @param plugin the plugin instance that need to be skipped
+     * Skips the given plugin.
+     * <p>
+     * <strong>Note:</strong> Whenever possible it should be used {@link #pluginSkipped(Plugin, String)} instead.
```
kingthorin: How about a minor tweak here, like: Whenever possible implementors shold use...
Or callers (vs implementors)

src/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -555,22 +556,66 @@ public boolean handleAntiCsrfTokens() {
...
+            return false;
+        }
+        return pluginStats.isSkipped();
+    }
+
+    /**
+     * Gets the reason why the given plugin was skipped.
+     * 
+     * @param plugin the plugin that will be checked
+     * @return the reason why the given plugin was skipped, might be {@code null} if not skipped or there's no reason
```
kingthorin: I think we should have a default skip reason, even if "Unknown" to differentiate not skipped vs no skip reason.

kingthorin:
Looks good

src/org/parosproxy/paros/core/scanner/HostProcess.java
thc202: Updated.

src/org/parosproxy/paros/core/scanner/HostProcess.java
thc202: Do we really need to? We can use isSkipped to differentiate that.

kingthorin: I was thinking more from a user perspective than code..... "It didn't run, but I don't know why."

thc202: Ah, I see, in that case it just shows "Skipped", as before.
I can tweak the default skipped text to "Skipped (unknown reason)" (or something like that), if that's better.

kingthorin: Oh ok, leave it as is. I was only consindering the return comment in the javadoc, if null becomes generic "Skipped" elsewhere then the condition is handled. From a support perspective that gives the differentiation I was concerned about.

kingthorin:
Looks good

src/org/parosproxy/paros/core/scanner/HostProcess.java
thc202: OK.

PR #2641, opened 2016-06-30
Add recommended security headers to the ZAP API UI

### LINKED ISSUE #2637: Prevent API UI from being loaded in a frame
Prevent the API UI from being loaded in a frame using a suitable X-Frame-Options header.
This wont work with some older browsers, but I dont think we need to support them - maybe we should recommend only using modern browsers with ZAP somewhere?
Reported via Bugcrowd - thanks to the relevant security researcher

psiinon:
Fixes #2637


kingthorin:
Looks good

psiinon:
Updated a couple of places that set headers directly and added a 'fixes' comment

kingthorin:
Looks even better :)

thc202:
Thanks!

PR #2645, opened 2016-06-30
Changes for bug tracker extension: PopupMenuItemAlert for access to Alerts

DarkPrince304:
This is in reference to zaproxy/zap-extensions#441 .

This changes the scope of two methods getAlertPanel (in ExtensionAlert) and getTreeAlert (in AlertPanel) to public, in order to retrieve the information of the alert.

This is temporary until some other way to get the alerts is devised.

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
+++ b/src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
@@ -1,0 +1,122 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0 
+ *   
+ * Unless required by applicable law or agreed to in writing, software 
+ * distributed under the License is distributed on an "AS IS" BASIS, 
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+ * See the License for the specific language governing permissions and 
+ * limitations under the License. 
+ */
+
+package org.zaproxy.zap.extension.alert;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.awt.Component;
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.swing.tree.DefaultMutableTreeNode;
+import javax.swing.tree.TreePath;
+
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.zap.extension.alert.ExtensionAlert;
+
+
+public abstract class PopupMenuItemAlert extends ExtensionPopupMenuItem {
+
+    private static final long serialVersionUID = 1L;
+    private final boolean multiSelect;
+    private final ExtensionAlert extAlert;
+
+    private static final Logger log = Logger.getLogger(PopupMenuItemAlert.class);
+
+    public PopupMenuItemAlert(String label) {
+        this(label, false);
+    }
+
+    public PopupMenuItemAlert(String label, boolean multiSelect) {
+        super(label);
+        this.multiSelect = multiSelect;
+        addActionListener(new PerformActionsActionListener());
+        this.extAlert = Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class);
+	}
+
+    public final boolean isMultiSelect() {
+        return multiSelect;
+    }
+
+    private Set<Alert> getAlertNodes() {
+        TreePath[] paths = this.extAlert.getAlertPanel().getTreeAlert().getSelectionPaths();
+        HashSet<Alert> alertNodes = new HashSet<Alert>();
+        for(int i = 0; i < paths.length; i++ ) {
+            DefaultMutableTreeNode alertNode = (DefaultMutableTreeNode)paths[i].getLastPathComponent();
+            if(alertNode.getChildCount() == 0) {
+                alertNodes.add((Alert)alertNode.getUserObject());
+                continue;
+            }
+            for(int j = 0; j < alertNode.getChildCount(); j++ ) {
+                DefaultMutableTreeNode node = (DefaultMutableTreeNode)alertNode.getChildAt(j);
+                alertNodes.add((Alert)node.getUserObject());
+            }
+        }
+        return alertNodes;
+    }
+
+    protected abstract void performAction(Alert alert);
+
+    protected void performActions(Set<Alert> alerts) {
+        for(Alert alert: alerts) {
+            performAction(alert);
+        }
+    }
+
+    protected boolean isButtonEnabledForNumberOfSelectedAlerts(int count) {
+        if(count == 0 ) {
+            return false;
+        } else if(!isMultiSelect() && count>1 ) {
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public boolean isEnableForComponent(Component invoker) {
+        if(this.extAlert == null) {
+            return false;
+        }
+        if ("treeAlert".equals(invoker.getName())) {
+            int count = this.extAlert.getAlertPanel().getTreeAlert().getSelectionCount();
+            setEnabled(!this.extAlert.getAlertPanel().getTreeAlert().isRowSelected(0) && isButtonEnabledForNumberOfSelectedAlerts(count));
+            return true;
+        }
+        return false;
+    }
+
+    private class PerformActionsActionListener implements ActionListener {
+
+        @Override
+        public void actionPerformed(ActionEvent evt) {
+            try {
+                Set<Alert> alerts = getAlertNodes();
+                performActions(alerts);
+            } catch (Exception e) {
+                log.error(e.getMessage(),e);
+            }
+        }
+    }
+
+}


psiinon:
I dont think this is the right approach.
I think we should have a PopupMenuAlert abstract class in the core which can then access these methods without having to change their scope.
We can then put this class in a temporary jar which the bug tracker extension can use prior to a new core being released.
Do you fancy having a go at implementing that @DarkPrince304 or would you like one of us to do it?

DarkPrince304:
Hi, I'd like to have a go at it first with a little guidance :)

psiinon:
Cool :)
So you need to create a new PopupMenuAlert abstract class in the org.zaproxy.zap.extension.alert package.
This should access the extension to get the Alerts selected.
I can then provide abstract methods like:
```
void performAction(Alert alert)
void performActions(List<Alert> alerts)
```
The reason for the 2 methods is that this class should be used by the existing popup menus in that package and some of those apply to single alerts while others apply to multiple ones. The constructor can include a param so that the class knows which one it should handle.
The ActionListener can then call the relevant abstract method which will then be implemented by you class which extends this one.
That make sense?

DarkPrince304:
@psiinon @thc202 @kingthorin would like you all to have a look at this. Suggest some improvements if any. This way it works without needing to change the scope of any other methods :)

src/org/zaproxy/zap/extension/alert/PopupMenuAlertBugTracker.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2010 psiinon@gmail.com
```
thc202: The year/name needs to be updated.

src/org/zaproxy/zap/extension/alert/PopupMenuAlertBugTracker.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.swing.tree.DefaultMutableTreeNode;
+import javax.swing.tree.TreePath;
+
+import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.zap.extension.alert.AlertPanel;
+import org.zaproxy.zap.extension.alert.ExtensionAlert;
+
+
+public class PopupMenuAlertBugTracker extends ExtensionPopupMenuItem {
```
thc202: The class should have a generic name, like PopupMenuItemAlert.

src/org/zaproxy/zap/extension/alert/PopupMenuAlertBugTracker.java
```diff
@@ -0,0 +1,59 @@
...
+
+	public PopupMenuAlertBugTracker() {
+		super();
+	}
+
+	public PopupMenuAlertBugTracker(String label) {
+		super(label);
+	}
+
+    public Alert[] getAlertNodes(ExtensionAlert extAlert) {
```
thc202: It should be an abstract class that calls the action methods instead (it reduces the boiler plate code in extending classes, as they just need to implement the corresponding actions). In that case this method does not need to be public.

DarkPrince304:
@psiinon @thc202 @kingthorin please do have a look and review when free :)

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.swing.tree.TreePath;
+
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.zap.extension.alert.AlertPanel;
+import org.zaproxy.zap.extension.alert.ExtensionAlert;
+
+
+public abstract class PopupMenuItemAlert extends ExtensionPopupMenuItem {
```
thc202: - The class should override `boolean isEnableForComponent(Component)` and check that the component is the Alerts tree, to automatically show or not the menu item.
 - Also, good to allow to specify if the pop up menu item is to be used with multiple alerts or not, to have it automatically enabled or disabled (for example, see description/behaviour of `multiSelect` in `PopupMenuHttpMessageContainer(String label, boolean multiSelect)`).

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,80 @@
...
+
+	private static final long serialVersionUID = 1L;
+
+	public PopupMenuItemAlert(String label) {
+		super(label);
+        addActionListener(new PerformActionsActionListener());
+	}
+
+    private Alert[] getAlertNodes() {
```
thc202: This should be changed to `List<Alert>` instead.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private class PerformActionsActionListener implements ActionListener {
+
+        @Override
+        public void actionPerformed(ActionEvent evt) {
+            try {
+                Alert[] alerts = getAlertNodes();
+                performActions(alerts);
+            } catch (Exception e) {
+                // TODO: Will add an exception message later
```
thc202: This should be changed now, it just needs to log the exception as error ;)

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2016 sanchitlucknow@gmail.com
```
thc202: I was expecting something like `The ZAP Development Team`.

thc202:
Some questions regarding the enabled state of the pop up menu item and selection of the alerts:
 - Should the pop up menu item be enabled when the root node ("Alerts") and some other alert node is selected? If so, should it include all alerts for further processing?
 - Should all (child) alerts be included when a main alert node (the ones that show the alert name) is selected?

For the latter I'd say yes, for the former, probably not.

@kingthorin, @psiinon thoughts?

Depending on the answers the pop up menu item might need further changes.

psiinon:
Agree - the pop up menu shoud not be enabled when the root node is selected, and all child nodes should be selected for main alert nodes.

kingthorin:
> Agree - the pop up menu shoud not be enabled when the root node is selected, and all child nodes should be selected for main alert nodes.

Logically and UX wise I think this makes sense. However, technically, I wonder if this will present some kind of memory use issue? Or some kind of input truncation issue. (When raising an bug for an alert that contains 100s or 1000s of entries...)

DarkPrince304:
Should we have some kind of limit for this? Also should this be included as a part of this PR or for the next one?

thc202:
@kingthorin as is now there's no memory issue by just "collecting" the selected alerts. Custom pop up menus do have to take care on how the alerts are used though.

@DarkPrince304 probably better to change that now, the root selection check and get all the selected alerts (current node and child nodes, which should probably use a Set instead of a List to avoid alert duplicates).

kingthorin:
Ok.

DarkPrince304:
@thc202 @psiinon @kingthorin Please do review. I have made all the appropriate changes tested them and they are working :)

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.control.Control;
+import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
+import org.parosproxy.paros.core.scanner.Alert;
+import org.zaproxy.zap.extension.alert.ExtensionAlert;
+
+
+public abstract class PopupMenuItemAlert extends ExtensionPopupMenuItem {
+
+	private static final long serialVersionUID = 1L;
```
thc202: There are some tabs that should be replaced with spaces (the indentation char used throughout the class).

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+        this.multiSelect = multiSelect;
+        addActionListener(new PerformActionsActionListener());
+        this.extAlert = Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class);
+	}
+
+    public final boolean isMultiSelect() {
+        return multiSelect;
+    }
+
+    private HashSet<Alert> getAlertNodes() {
```
thc202: It should return Set<Alert> instead.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+    	TreePath[] paths = this.extAlert.getAlertPanel().getTreeAlert().getSelectionPaths();
+        HashSet<Alert> alertNodes = new HashSet<Alert>();
+        for(int i = 0; i < paths.length; i++ ) {
+            DefaultMutableTreeNode alertNode = (DefaultMutableTreeNode)paths[i].getLastPathComponent();
+            System.out.println((Alert)alertNode.getUserObject());
```
thc202: This println should be removed.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+            if(alertNode.getChildCount() == 0) {
+                alertNodes.add(((Alert)alertNode.getUserObject()));
```
thc202: There's a pair of parentheses that could be removed.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+                continue;
+            }
+            for(int j = 0; j < alertNode.getChildCount(); j++ ) {
+                DefaultMutableTreeNode node = (DefaultMutableTreeNode)alertNode.getChildAt(j);
+                alertNodes.add(((Alert)node.getUserObject()));
```
thc202: There's a pair of parentheses that could be removed.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        return alertNodes;
+    }
+
+    protected abstract void performAction(Alert alert);
+
+    protected void performActions(HashSet<Alert> alerts) {
```
thc202: Should be Set<Alert> instead.

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+        } else if(!isMultiSelect() && count>1 ) {
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public boolean isEnableForComponent(Component invoker) {
+        TreePath[] paths = this.extAlert.getAlertPanel().getTreeAlert().getSelectionPaths();
+        for(int i = 0; i < paths.length; i++ ){
```
thc202: The only node that does not have a user object is the root node so it can be directly checked, for example:
```Java
this.extAlert.getAlertPanel().getTreeAlert().isRowSelected(0)
```

This check should be done for the enabled state of the menu item (below when calling setEnabled) not to decide whether the menu is shown or not (they should still be shown in the alerts tree but be disabled).

src/org/zaproxy/zap/extension/alert/PopupMenuItemAlert.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private class PerformActionsActionListener implements ActionListener {
+
+        @Override
+        public void actionPerformed(ActionEvent evt) {
+            try {
+                HashSet<Alert> alerts = getAlertNodes();
+                performActions(alerts);
+            } catch (Exception e) {
+                log.debug(e.getMessage(),e);
```
thc202: This should be logged as error, if an exception gets this far it's most likely a bug in the custom menu items.

thc202:
Looks good to me, thank you!

kingthorin:
Thanks

PR #2654, opened 2016-07-05
Check all nodes when restructuring Sites tree

### LINKED ISSUE #2608: Removing a DDN from a Context Does Not Appear to Trigger an Update to the Sites Tab
If I define a data driven node (DDN) in the Structure dialog of a context, I see that the corresponding components in the Sites tab are updated to reflect this new DDN.  However, if I then delete the DDN, the components in the Sites tab remain unchanged.

thc202:
Change the method Context.checkNode(SiteNode) to obtain all the child
nodes of the node being checked prior checking them, as the check might
shuffle them (by adding or removing some of nodes) causing some child
nodes to not being checked (its index would no longer be the expected).

Fix #2608 - Removing a DDN from a Context Does Not Appear to Trigger an
Update to the Sites Tab


psiinon:
Looks good

PR #2655, opened 2016-07-05
Provide skip reason for Script Active Scan Rules

thc202:
Change class ScriptsActiveScanner to provide the reason why the scanner
is skipped (simply, when there are no Script Rules enabled).

Related to issues:
 - #2221 - Skip "Script Active Scan Rules" scanner if there are no
 scripts enabled;
 - #2415 - Show the reason why an active scanner was skipped.

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -295,6 +295,7 @@ ascan.progress.table.status		= Status
 ascan.progress.title            = {0} Scan Progress
 ascan.scripts.activescanner.title	= Script Active Scan Rules
 ascan.scripts.interface.active.error = The provided Active Rules script ({0}) does not implement the required interface.\nPlease refer to the provided templates for examples.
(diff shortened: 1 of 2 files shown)


src/lang/Messages.properties
```diff
@@ -295,6 +295,7 @@ ascan.progress.table.status		= Status
 ascan.progress.title            = {0} Scan Progress
 ascan.scripts.activescanner.title	= Script Active Scan Rules
 ascan.scripts.interface.active.error = The provided Active Rules script ({0}) does not implement the required interface.\nPlease refer to the provided templates for examples.
+ascan.scripts.skip.reason = no scripts enabled
```
kingthorin: Should this have title or leading/initial caps?

thc202: It's added to an existing sentence: `Skipped, no scripts enabled.`.

kingthorin:
Looks good

PR #2657, opened 2016-07-06
Correct persistence of disabled extensions

thc202:
Change method ExtensionParam.setExtensions(List<ExtensionState>) to use
the correct index when persisting the data of each extension, otherwise
it would create two extension elements one for the name of the extension
and the other for the enabled state, instead of just one extension
element. Also, make minor changes to not require to wrap the returned
list in an unmodifiable list each time the list is required and when
cloning return the expected type (ExtensionParam).
Change class ExtensionState to implement equals(), hashCode() and
toString() and change to require the name of the extension when creating
an instance.
Add tests to assert the expected behaviour of classes ExtensionParam and
ExtensionState.

Issue introduced in #2245 - Convert options to not use extensions' names
as XML element names


kingthorin:
Looks good

PR #2658, opened 2016-07-06
Log an error if a PluginPassiveScanner has no ID

thc202:
Change ExtensionPassiveScan to log an error if the PluginPassiveScanner
being added does not have a defined ID, that is it has the value -1. The
ID is required to uniquely identify the scanners therefore it should be
used a specific one, always.


psiinon:
Looks good

PR #2659, opened 2016-07-07
Do not add same message to SiteNode's past history

thc202:
Change SiteMap class to not add the message (HistoryReference) to the
past history of the SiteNode if it already belongs to the node,
preventing adding the same message multiple times to its past history.
This could happen when adding a message to the Sites tree that had been
added previously (for example, when raising or filtering alerts).


kingthorin:
Looks good

PR #2660, opened 2016-07-08
Use the ID to save configs of passive scanners

thc202:
Change how the configurations of the passive scanners are saved by using
its ID instead of the classname. The ID is used to (uniquely) identify a
scanner so it's more suitable than the classname which gives problems
when the passive scanners are promoted or demoted (since it, usually,
requires a change in the package name).
The configurations now use the elements, "id", "level" and "enabled", to
save the ID, alert threshold and enabled state of a scanner, each one in
its own "pscanner" element (which are all under an "pscan" element).
While the options are migrated some will have an element called
"classname" containing the classname of the scanner, which is removed
once the configurations are saved again (then starting to use the ID).

Change class PluginPassiveScanner to save the configurations with the
new structure, also, document and normalise its behaviour.
Change Constant class to convert to the new options when updating from
version 2.5.0.
Change class PassiveScanParam to expose the root element of the
configurations.
Add tests to assert the expected behaviour of the class
PluginPassiveScanner.

### DIFF
--- a/test/org/zaproxy/zap/extension/pscan/PluginPassiveScannerUnitTest.java
+++ b/test/org/zaproxy/zap/extension/pscan/PluginPassiveScannerUnitTest.java
@@ -1,28 +1,26 @@
 package org.zaproxy.zap.extension.pscan;
 
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+
 import net.htmlparser.jericho.Source;
 import org.apache.commons.configuration.Configuration;
 import org.junit.Before;
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.control.AddOn;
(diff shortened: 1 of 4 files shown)


test/org/zaproxy/zap/extension/pscan/PluginPassiveScannerUnitTest.java
```diff
@@ -4,78 +4,431 @@
...
 import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.mockito.Mock;
-import org.mockito.runners.MockitoJUnitRunner;
 import org.parosproxy.paros.core.scanner.Plugin.AlertThreshold;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.control.AddOn;
+import org.zaproxy.zap.utils.ZapXmlConfiguration;
 
 import static org.hamcrest.Matchers.*;
```
kingthorin: Should this be switched to specific imports?

kingthorin:
Looks good. Though honestly it was a lot to absorb, and I don't feel the greatest tonight (fighting a cruddy cold)

test/org/zaproxy/zap/extension/pscan/PluginPassiveScannerUnitTest.java
thc202: Done.

thc202:
Take your time (and get well!), there's no rush on these.

PR #2661, opened 2016-07-08
Remove use of BAD_TYPE in API implementations

thc202:
Change API implementations to use ILLEGAL_PARAMETER instead of BAD_TYPE,
the latter is used to inform that the requested API type (view, action
or other) is not valid (which should be used only by core API classes).


kingthorin:
Looks good

PR #2665, opened 2016-07-11
Disable scan menus if ZAP is started in Safe mode

thc202:
Change classes ExtensionActiveScan and ExtensionSpider to disable the
scan menu items, "Tools" > "Active Scan..." and "Spider...", if ZAP is
started in Safe mode.


kingthorin:
Looks good

PR #2672, opened 2016-07-12
Delay log to include session cookies

thc202:
Change class HttpSessionsSite to delay the log that the session was
created to ensure that is fully initialised (that is, its cookies were
already set), otherwise from the logs it seems that all the sessions are
empty (even if that's not the case).

### DIFF
--- a/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsSite.java
+++ b/src/org/zaproxy/zap/extension/httpsessions/HttpSessionsSite.java
@@ -421,6 +421,7 @@ public class HttpSessionsSite {
 			}
 		}
 
+		boolean newSession = false;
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

src/org/zaproxy/zap/extension/httpsessions/HttpSessionsSite.java
```diff
@@ -462,6 +463,10 @@ public void processHttpResponseMessage(HttpMessage message) {
 			}
 		}
 
+		if (newSession) {
+			log.info("Created a new session as no match was found: " + session);
```
psiinon: Should this even be logged at info level? Maybe debug would be better?

PR #2673, opened 2016-07-12
Use all technologies by default in active scans

thc202:
Change classes HostProcess, Plugin, AbtractPlugin and Scanner to not
allow to set a null TechSet and to use all technologies by default
(TechSet.AllTech instead of null).
The change ensures that the TechSet being used in the active scans is
never null (thus preventing NullPointerException in scanners).
Add tests to AbtractPluginUnitTest to assert the expected behaviour.
Update JavaDoc with the new behaviour (and add since tags to technology
related methods).


kingthorin:
Looks good

PR #2676, opened 2016-07-13
Tweak log in SpiderTask to include more info

thc202:
Change SpiderTask to include the URL of the resource being fetched when
logging the Exception.


kingthorin:
Looks good

PR #2677, opened 2016-07-13
Do not process auth messages for HTTP Sessions tab

### LINKED ISSUE #2674: Automated authentication requests shown in HTTP Sessions tab
Issue reported in OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/uLceCjtc8wc/discussion

Steps to reproduce the issue:
1. Run ZAP;
2. Set up a context with users (Cookie-based session management and the session token must match one defined in Options > HTTP Sessions);
3. Spider (or active scan) the context with the defined user;
4. Note that for each authentication attempt it's created a new entry in HTTP sessions tab. As with other automated requests (spider, active scanner, fuzzer...) the authentications requests should not be (processed and) shown in HTTP Sessions tab.

ZAP Versions:
Versions >= 2.0.0

thc202:
Change class ExtensionHttpSessions to skip/ignore requests and responses
from authentications attempts (as with other automated requests they
should not be processed, nor changed since they are already trying to
authenticate a user).

Fix #2674 - Automated authentication requests shown in HTTP Sessions tab

### DIFF
--- a/src/org/zaproxy/zap/extension/httpsessions/ExtensionHttpSessions.java
+++ b/src/org/zaproxy/zap/extension/httpsessions/ExtensionHttpSessions.java
@@ -573,7 +573,7 @@ public class ExtensionHttpSessions extends ExtensionAdaptor implements SessionCh
 
 	@Override
 	public void onHttpRequestSend(HttpMessage msg, int initiator, HttpSender sender) {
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

PR #2680, opened 2016-07-14
Remove Ant library, no longer needed/used

thc202:
The Ant library is used by the Java ZAP API client implementation for
Ant tasks, which are no longer included in ZAP. Also, no add-on seems to
be depending on it.


kingthorin:
Looks good

PR #2681, opened 2016-07-14
Fix exception while adding script through the API

thc202:
Change ScriptAPI to obtain the optional parameter "scriptDescription"
with a default value (i.e. empty string), otherwise it would lead to an
exception if it was not present in the API request:
 Exception while handling API request:
  JSONException: JSONObject["scriptDescription"] not found.

(Since the parameter is optional it's likely to happen.)
 ---
From zaproxy/zap-api-python#5.


kingthorin:
Looks good

PR #2682, opened 2016-07-15
Sort (main) help add-on TOC entries

thc202:
Change class ZapTocMerger to sort the entries under the "Add Ons" TOC
entry, so the add-ons are shown in alphabetic order (instead of the
order they were added/loaded).


kingthorin:
Looks good

PR #2685, opened 2016-07-18
Ignore UnsupportedEncodingException in Form auth

thc202:
Ignore UnsupportedEncodingException in FormBasedAuthenticationMethod,
when encoding user data for the authentication message as it's using one
of the standard charsets (which are guaranteed to exist).


kingthorin:
Looks good

PR #2687, opened 2016-07-18
Tidy up Constant.isKali()

thc202:
Change to use try-with-resource statement when opening the file and use
Boolean values to set the value of the flag.


kingthorin:
Looks good

PR #2688, opened 2016-07-18
Remove code duplication in HttpUserAgent

thc202:
Extract a method that searches the version of a User-Agent in a given
file, also use try-with-resource statement to open/close the file.
Change the other methods to use the extracted methods.


kingthorin:
Looks good

PR #2691, opened 2016-07-19
Restore the enabled state of passive scanners

thc202:
The changes done in #2660 ("Use the ID to save configs of passive
scanners") wrongly changed the default state of passive scanners to
disable (as that's the default state of Enableable, which was no longer
being overridden either when instantiating the passive scanners or
when loading from the configurations).

Change PluginPassiveScanner to be enabled by default and update test to
reflect correct behaviour.


kingthorin:
Looks goodLeave a comment

PR #2692, opened 2016-07-19
Fix elements name in drivers.xml

thc202:
The code was changed to read the slot list index of the drivers using
the element name "slotListIndex" (instead of "slotList") but the file
was not updated accordingly leading to (silent) NullPointerException
in DriverConfiguration when attempting to read them.
Change class DriverConfiguration to prevent the NullPointerException
when reading the slot or slot list index of the drivers if they were not
provided.


kingthorin:
Looks good

PR #2695, opened 2016-07-20
Correctly identify fuzzer tab in HTTP pop up menus

thc202:
Change classes PopupMenuItemHttpMessageContainer and
PopupMenuHttpMessageContainer to correctly identify the Fuzzer tab, so
custom pop up menus can be correctly shown (or not) in that tab.


kingthorin:
Looks good

PR #2696, opened 2016-07-20
Enable Copy URLs pop up menu item for all messages

thc202:
Change pop up menu item "Copy URLs to Clipboard" (PopupMenuCopyUrls) to
be enabled for all HTTP message types, so it can be used in all GUI
components that show HTTP messages.
 ---
Issue reported in the IRC channel.


kingthorin:
Looks good

PR #2702, opened 2016-07-22
Remove non-default constructors from AboutPanel

thc202:
Remove all constructors except default constructor as they are not used
(nor they would work correctly, the LayoutManager would be overridden by
the initialize method which sets another LayoutManager again).
Remove initialize method, merged with default constructor.

Though the changes are not binary compatible the class is not expected
to be used by non-core code.


kingthorin:
Looks good

PR #2703, opened 2016-07-22
Tidy up ScanPanel2 class

thc202:
Class ScanPanel2:
 - Deprecate current constructor, it has more parameters than the ones
 required (AbstractParam);
 - Add a constructor that just has the required parameters;
 - Change visibility of instance variable "prefix" to private (from
 public);
 - Set final modifier to instance variables prefix and controller.

Change classes SpiderPanel and ActiveScanPanel to use the new
constructor.


kingthorin:
Looks goodLeave a comment

PR #2710, opened 2016-07-25
Remove unused menu constructors

thc202:
Remove unused (String) constructors from pop up and top level menu
classes, the constructors are not being used moreover they would not
work as expected since the action listeners were not being added (thus
not doing anything by default). Also, for pop up menus that have an
"initialize" method, merge it with default constructor.

### DIFF
--- a/src/org/zaproxy/zap/extension/params/PopupMenuRemoveSession.java
+++ b/src/org/zaproxy/zap/extension/params/PopupMenuRemoveSession.java
@@ -36,26 +36,7 @@ public class PopupMenuRemoveSession extends ExtensionPopupMenuItem {
      * 
      */
     public PopupMenuRemoveSession() {
-        super();
- 		initialize();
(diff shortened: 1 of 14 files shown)


src/org/zaproxy/zap/extension/params/PopupMenuRemoveSession.java
```diff
@@ -68,6 +49,11 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
 			
 	}
 
+	public void setExtension(ExtensionParams extension) {
+		this.extension = extension;
+	}
+
+
```
kingthorin: This extra line could be removed

thc202: Done, thank you.

kingthorin:
Looks good

PR #2711, opened 2016-07-25
Replace unused parameter in AllCategoryTableModel

thc202:
Change AllCategoryTableModel's constructor to have another parameter
type (PluginFactory) which is used to initialise the model.
Change class PolicyAllCategoryPanel to use the new constructor.

Though the changes are not binary compatible the class is not expected
to be used by non-core code.


kingthorin:
Any concerns about compatability?

thc202:
No, updated the commit message and pull request to mention that.

kingthorin:
Looks good

PR #2714, opened 2016-07-26
Implement -i option and add -l level option

psiinon:


build/docker/zap-baseline.py
```diff
@@ -376,16 +391,16 @@ def main(argv):
 
       # print out the info rules
       for key, alert_list in sorted(alert_dict.iteritems()):
-        if config_dict.has_key(key) and config_dict[key] == 'INFO':
+        if (config_dict.has_key(key) and config_dict[key] == 'INFO')  or (not config_dict.has_key(key)) and info_unspecified:
```
kingthorin: There's a double space leading the `or`.

build/docker/zap-baseline.py
```diff
@@ -376,16 +391,16 @@ def main(argv):
...
           if key in config_msg:
             user_msg = config_msg[key]
-          print_rule(config_dict[key], alert_list, detailed_output, user_msg)
+          print_rule('INFO', alert_list, detailed_output, user_msg)
           info_count += 1
 
       # print out the warning rules
       for key, alert_list in sorted(alert_dict.iteritems()):
-        if (not config_dict.has_key(key)) or (config_dict[key] == 'WARN'):
+        if (not config_dict.has_key(key) and not info_unspecified) or (config_dict.has_key(key) and config_dict[key] == 'WARN'):
```
kingthorin: Instead of using literal strings all over the place would it be better for example to replace `'WARN'` with `levels[3]`? Or maybe define `LEVEL_WARN="WARN"` (etc) which could then be used to build the `levels` array, and referenced throughout?

kingthorin:
This PR doesn't seem to implement -i, I guess that was done earlier?

psiinon:
The -i option was present in the command line options but hadnt actually been implemented ;) The change to line 394 implements it.

build/docker/zap-baseline.py
psiinon: I dont like levels[3] - that means nothing to me when I read the code. I'm ok with either 'WARN' or LEVEL_WARN, but I dont really know whats considered best practice for python code :/

kingthorin:
Ok perfect.

build/docker/zap-baseline.py
kingthorin: I don't "really" know either, this is the closet advice I can find/give: https://www.python.org/dev/peps/pep-0008/#constants

psiinon: Thats all I've found so far as well :) The trouble is, as python is not compiled I dont think constants actually help that much. If I mistyped LEVE_WARN then it would still run, and only error if it went through the relevant path. So you just have to try to test all of the paths anyway...

kingthorin: I was just thinking that maintenance wise it might be easier to have constants defined. But, I do see your point a typo will always be a typo and error out either way. I'm ok with it being left...

psiinon:
Fixed the double space, but have left the 'constants' as I think they are slightly easier to read. But happy  to change those if anyone feels strongly about them...

thc202:
Looks good to me.

psiinon:
Ah, just spotted a bug in the description - it should depend on https://pypi.python.org/pypi/python-owasp-zap-v2.4/ :/ I'll fix that

psiinon:
Fixed

PR #2719, opened 2016-07-26
Minor tidy up of ScanProgressDialog class

thc202:
- Remove commented statement;
 - Remove unnecessary instance variable initialisations (already
 initialised by default to null/false);
 - Remove instance variable "site" needed only once to set the title
 (now done in the constructor);
 - Document the constructor;
 - Remove unnecessary JavaDoc comments (incomplete ones from overridden
 methods);
 - Rename the method showProgress to updateProgress (and update its JavaDoc);
 - Change ScanProgressActionListener to:
  - Extend MouseAdapter instead of implementing the interfaces directly,
  as it doesn't require to override unused methods;
  - Use the methods JTable.columnAtPoint/rowAtPoint instead of
  calculating the position of the column/row "manually".

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ScanProgressDialog.java
+++ b/src/org/zaproxy/zap/extension/ascan/ScanProgressDialog.java
@@ -25,13 +25,14 @@ import java.awt.Dimension;
 import java.awt.FlowLayout;
 import java.awt.Frame;
 import java.awt.GridBagLayout;
+import java.awt.Point;
+import java.awt.HeadlessException;
 import java.awt.Toolkit;
 import java.awt.datatransfer.Transferable;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/ascan/ScanProgressDialog.java
```diff
@@ -360,11 +357,8 @@ private JTable getMainPanel() {
     }
     
     /**
-     * Update the current Scan progress and
-     * prepare actions and scan summary
+     * Updates the current Scan progress and prepare actions and scan summary.
```
kingthorin: And and, seems a little confusing.

kingthorin:
One minor thing, otherwise looks good.

src/org/zaproxy/zap/extension/ascan/ScanProgressDialog.java
thc202: Renamed the method (to `updateProgress`) and rewrote its JavaDoc.

kingthorin:
Looks goodLeave a comment

PR #2724, opened 2016-07-27
Document constructors of ZAP text components

thc202:
Document the constructors of ZapTextArea and ZapTextField. Also, add
"since" JavaDoc tags to those classes and ZapTextComponentUndoManager
and for the latter class remove unnecessary closing paragraph tag in
class doc.


kingthorin:
Looks good

PR #2728, opened 2016-07-28
Allow to remove spider parsers and filters

thc202:
Change ExtensionSpider to allow to remove SpiderParser, FetchFilter and
ParseFilter. Also, validate that the added parsers/filters are not null.

The change allows add-ons (that install custom parsers/filters) to be
correctly uninstalled/updated.


kingthorin:
Looks good

PR #2731, opened 2016-08-01
Support POST requests for API actions. Fixes #2723

### LINKED ISSUE #2723: Support POST requests for API actions
Using GET request everywhere (instead of POSTs for actions) makes certain API calls quite tricky to get right because one has to remember to url-encode any special characters - e.g. (python API):

```
zap.authentication.set_authentication_method(ctx_id, "formBasedAuthentication", "loginUrl=https://www.example.com/signin.html&loginRequestData=email={%25username%25}%26password={%25password%25}")
```

(Notice the required & -> %26 and % -> %25 escaping)  
`loginRequestData` is a url query parameter and therefore it's value cannot contain any special characters (like an ampersand)
Changing API GET requests to POST would most likely solve the problem, but changing the way authentication methods are set might suffice.

Dalimil:


src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -289,13 +289,12 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
 
 					ApiResponse res;
 					switch (reqType) {
-					case action:	
-						// TODO Handle POST requests - need to read these in and then parse params from POST body
-						/*
-						if (getOptionsParamApi().isPostActions()) {
-							throw new ApiException(ApiException.Type.DISABLED);
+					case action:
+						if (requestHeader.getMethod().toUpperCase().equals("POST")) {
```
thc202: It could use `HttpRequestHeader.POST` and it could use `equalsIgnoreCase` instead (in worst case it does not create a new String).

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -289,13 +289,12 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
...
+							String body = httpIn.readRequestBody(requestHeader).toString();
```
thc202: It should be checked that the Content-Type is the one that we expect, otherwise it should throw an ApiException (with type `UNSUPPORTED_REQUEST`?).

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -315,7 +314,7 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
 								}
 							}
 						}
-						
+
 						res = impl.handleApiOptionAction(name, params);
```
thc202: This indentation change should be reverted. [1]

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -42,11 +42,7 @@
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.core.proxy.ProxyParam;
 import org.parosproxy.paros.model.Model;
-import org.parosproxy.paros.network.HttpInputStream;
-import org.parosproxy.paros.network.HttpMalformedHeaderException;
-import org.parosproxy.paros.network.HttpMessage;
-import org.parosproxy.paros.network.HttpOutputStream;
-import org.parosproxy.paros.network.HttpRequestHeader;
+import org.parosproxy.paros.network.*;
```
kingthorin: Please use specific imports.

kingthorin:
Could you please rebase interactive and fix-up into a single commit?

https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase
https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -289,13 +290,17 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
...
 					ApiResponse res;
 					switch (reqType) {
-					case action:	
-						// TODO Handle POST requests - need to read these in and then parse params from POST body
-						/*
-						if (getOptionsParamApi().isPostActions()) {
-							throw new ApiException(ApiException.Type.DISABLED);
+					case action:
+						if (requestHeader.getMethod().equalsIgnoreCase(HttpRequestHeader.POST)) {
+							if(requestHeader.getHeader(HttpHeader.CONTENT_TYPE)
```
thc202: This will lead to a NullPointerException if the request does not contain the Content-Type request header. Might worth inverting the equals check.

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -289,13 +290,17 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
...
+									.equals(HttpHeader.FORM_URLENCODED_CONTENT_TYPE)) {
+								String body = httpIn.readRequestBody(requestHeader).toString();
+								params = getParams(body);
+							} else {
+								throw new ApiException(ApiException.Type.DISABLED);
```
thc202: It should use another type (`UNSUPPORTED_CONTENT_TYPE`?), it might be a little confusing using the one that indicates the API is disabled.

src/org/zaproxy/zap/extension/api/WebUI.java
```diff
@@ -166,6 +166,7 @@ public String handleRequest(String component, ApiImplementor impl,
 			sb.append("var url = '/' + format + '/" + component + "/" + reqType.name() + "/" + name + "/'\n");
 			sb.append("var form=document.getElementById('zapform');\n");
 			sb.append("form.action = url;\n");
+			sb.append("form.method = form.elements[\"formMethod\"].value;\n");
```
thc202: This needs changes to match the latest code (related to #2750).

kingthorin:
@Dalimil any chance you can address @thc202's comments? As well as the conflict?

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -310,13 +311,18 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
...
-						// TODO Handle POST requests - need to read these in and then parse params from POST body
-						/*
-						if (getOptionsParamApi().isPostActions()) {
-							throw new ApiException(ApiException.Type.DISABLED);
+					case action:
+						if (requestHeader.getMethod().equalsIgnoreCase(HttpRequestHeader.POST)) {
+							String contentTypeHeader = requestHeader.getHeader(HttpHeader.CONTENT_TYPE);
+							if(contentTypeHeader != null &&
+									contentTypeHeader.equals(HttpHeader.FORM_URLENCODED_CONTENT_TYPE)) {
+								String body = httpIn.readRequestBody(requestHeader).toString();
```
thc202: Per latest changes (#2937) this is no longer needed, it can just obtain the request body from the message.

src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -310,13 +311,18 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
...
+								params = getParams(body);
+							} else {
+								throw new ApiException(ApiException.Type.BAD_TYPE);
```
thc202: I still think it would be better to use another (new) type to indicate that the content type is not supported. But lets see what others think.

kingthorin: If we're talking exception type then yes I agree, a "new" more specific type would be great.

thc202: Yes, was referring to the `ApiException.Type`.

thc202:
Looks good to me, thank you!

PR #2733, opened 2016-08-01
Check older add-ons on manual installations

### LINKED ISSUE #2707: Manual add-on installation needs more meaningful dialog messages
Attached is a screenshot of a dialog that I encountered when upgrading ContextAlertFilters (in the past). I've also encountered similar with other addons in the past. 

nonspecific-requirements

This doesn't tell the user what the actual requirement is. In this case it was that the addon needs to be manually removed (Marketplace Uninstall) prior to trying to install (or Load). IIRC selecting "Yes" resulted in another error, selecting "No" did nothing (as expected). 
_[Sorry I can't be more specific, I saved the image back in March and forgot to open the ticket at the time.]_

thc202:
Change ExtensionAutoUpdate to check if the add-on being manually
installed is a newer or an older version, to correctly inform about
required changes during the installation (for example, if other add-ons
need to be uninstalled, updated or installed). Also, inform with better
message if the same add-on (version and status) is already installed.

Fix #2707 - Manual add-on installation needs more meaningful dialog
messages


kingthorin:
Looks good

PR #2737, opened 2016-08-02
Correct API error message on missing script params

thc202:
Change ScriptBasedAuthenticationMethodType to catch and re-throw the
ApiException caused by missing parameters, so the error shown is, for
example:
 Missing Parameter (missing_parameter) : LoginURL

instead of:
 An error has occurred when loading the provided script
  (bad_script_format) : missing_parameter

which happened because the ApiException was being treated as an error
in the script.

Related to issue jenkinsci/zaproxy-plugin#31.


kingthorin:
Looks good

PR #2738, opened 2016-08-03
Add support for proxy url whitelisting

Dalimil:
Users may only be interested in a few urls and with this patch can exclude everything from proxy and then only whitelist those few urls.

### DIFF
--- a/src/org/parosproxy/paros/core/proxy/ProxyServer.java
+++ b/src/org/parosproxy/paros/core/proxy/ProxyServer.java
@@ -76,6 +76,7 @@ public class ProxyServer implements Runnable {
     protected boolean enableCacheProcessing = false;
     protected Vector<CacheProcessingItem> cacheProcessingList = new Vector<>();
     private List<Pattern> excludeUrls = null;
+    private List<Pattern> includeUrls = null;
     private static Logger log = Logger.getLogger(ProxyServer.class);
 
     /**
@@ -404,6 +405,14 @@ public class ProxyServer implements Runnable {
         }
     }
 
+    public void setIncludeList(List<String> urls) {
+        includeUrls = new ArrayList<>(urls.size());
+        for (String url : urls) {
+            Pattern p = Pattern.compile(url, Pattern.CASE_INSENSITIVE);
+            includeUrls.add(p);
+        }
+    }
+
     public boolean excludeUrl(URI uri) {
         boolean ignore = false;
         if (excludeUrls != null) {
@@ -419,8 +428,29 @@ public class ProxyServer implements Runnable {
                 }
             }
         }
-        
-        return ignore;
+
+        // Ignore it only if it isn't white-listed.
+        return ignore && !includeUrl(uri);
+
+    }
+
+    private boolean includeUrl(URI uri) {
+        boolean include = false;
+        if (includeUrls != null) {
+            String uriString = uri.toString();
+            for (Pattern p : includeUrls) {
+                if (p.matcher(uriString).matches()) {
+                    include = true;
+                    if (log.isDebugEnabled()) {
+                        log.debug("URL whitelisted: " + uriString + " Regex: " + p.pattern());
+                    }
+
+                    break;
+                }
+            }
+        }
+
+        return include;
     }
 
     // ZAP: Added the method.
--- a/src/org/parosproxy/paros/model/Session.java
+++ b/src/org/parosproxy/paros/model/Session.java
@@ -130,6 +130,7 @@ public class Session {
 	private String fileName = "";
 	private String sessionDesc = "";
 	private List<String> excludeFromProxyRegexs = new ArrayList<>();
+	private List<String> includeInProxyRegexs = new ArrayList<>();
 	private List<String> excludeFromScanRegexs = new ArrayList<>();
 	private List<String> excludeFromSpiderRegexs = new ArrayList<>();
 	// ZAP: Added globalExcludeURLRegexs code.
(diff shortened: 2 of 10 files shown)


src/org/parosproxy/paros/core/proxy/ProxyServer.java
```diff
@@ -399,13 +408,34 @@ public boolean excludeUrl(URI uri) {
                     if (log.isDebugEnabled()) {
                         log.debug("URL excluded: " + uriString + " Regex: " + p.pattern());
                     }
-                    
+
```
kingthorin: Formatting changes (if really necessary) should occur in a separate PR.

https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

src/org/parosproxy/paros/model/Session.java
```diff
@@ -52,7 +52,7 @@
 // ZAP: 2015/04/02 Issue 321: Support multiple databases and Issue 1582: Low memory option
 // ZAP: 2015/08/19 Change to use ZapXmlConfiguration instead of extending FileXML
 // ZAP: 2015/08/19 Issue 1789: Forced Browse/AJAX Spider messages not restored to Sites tab
-// ZAP: 2015/10/21 Issue 1576: Support data driven content
+// ZAP: 2015/10/21 Issue 1576: Support data driven content
```
kingthorin: What was changed here?

src/org/parosproxy/paros/view/View.java
```diff
@@ -121,20 +121,7 @@
...
-import org.zaproxy.zap.view.ContextPanelFactory;
-import org.zaproxy.zap.view.ContextStructurePanel;
-import org.zaproxy.zap.view.ContextTechnologyPanel;
-import org.zaproxy.zap.view.SessionExcludeFromProxyPanel;
-import org.zaproxy.zap.view.SessionExcludeFromScanPanel;
-import org.zaproxy.zap.view.SessionExcludeFromSpiderPanel;
-import org.zaproxy.zap.view.SplashScreen;
-import org.zaproxy.zap.view.StatusUI;
-import org.zaproxy.zap.view.ZapMenuItem;
+import org.zaproxy.zap.view.*;
```
kingthorin: Specific imports should be used.

src/org/zaproxy/zap/view/SessionIncludeInProxyPanel.java
```diff
@@ -0,0 +1,137 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2010 psiinon@gmail.com
```
kingthorin: This should be set as 2016 and "the ZAP Development Team" (check other recent PRs if you want to copy/paste)

src/org/zaproxy/zap/view/SessionIncludeInProxyPanel.java
```diff
@@ -0,0 +1,137 @@
...
+package org.zaproxy.zap.view;
+
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.Session;
+import org.parosproxy.paros.view.AbstractParamPanel;
+import org.parosproxy.paros.view.View;
+
+import javax.swing.*;
+import java.awt.*;
```
kingthorin: Should use specific imports.

src/org/zaproxy/zap/view/SessionIncludeInProxyPanel.java
```diff
@@ -0,0 +1,137 @@
...
+	private MultipleRegexesOptionsPanel regexesPanel;
+
+    public SessionIncludeInProxyPanel() {
+        super();
+ 		initialize();
+	}
+
+	/**
+	 * This method initializes this
+	 */
```
kingthorin: Is this comment really needed?

src/org/zaproxy/zap/view/SessionIncludeInProxyPanel.java
```diff
@@ -0,0 +1,137 @@
...
+	private JPanel getPanelSession() {
+		if (panelSession == null) {
+
+			panelSession = new JPanel();
+			panelSession.setLayout(new GridBagLayout());
+			panelSession.setName("Includeproxy");
+
+			GridBagConstraints gridBagConstraints2 = new GridBagConstraints();
+	        GridBagConstraints gridBagConstraints1 = new GridBagConstraints();
```
kingthorin: There seems to be some inconsistency in tabs/spaces here vs the previous few lines.

kingthorin:
Thanks for tackling this. I've added a few quick hit comments, I haven't been able to do a detailed review.

Note any edited Paros file should be appropriately commented/documented: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines

Also the PR should contain a single commit. Which can be accomplished through interactive rebase and fixup or squash.
https://www.kernel.org/pub/software/scm/git/docs/git-rebase.html#_interactive_mode
https://wiki.eclipse.org/EGit/User_Guide#Interactive_Rebase

kingthorin:
Thanks for making this into a single commit. Could you address the feedback?

Should you also let us know what test cases you've conducted with the code? i.e.: 

1. Includes work when no excludes are defined.
2. Includes and excludes work when there's no overlap.
3. Includes work when they're a subset of excludes.
4. etc.

kingthorin:
@Dalimil any update?

kingthorin:
@Dalimil any update?

Dalimil:
So the behaviour was that the include list can overrule the exclude list, so whenever it checks if a URL is excluded in excludeUrl(uri) it would check the exclude list for any matches as before but now whenever it is about to return true (as in 'ignore this url') it would also check the include list for possible matches and might eventually return false ('don't ignore this url') from the original function if it found any.

But looking at the code now, the functionality might be a bit confusing for many users...  
I might just drop this PR because I haven't touched the codebase in a long time and there have been many changes to the project since August...

kingthorin:
Closing per @Dalimil

psiinon:
For info just added the 'invalid' tag so that it doesnt get picked up for the release notes ;)

PR #2746, opened 2016-08-05
Add http session API sites view

psiinon:


psiinon:
I had problems working out what format the site name was, so this call shows you all of the valid ones :)

thc202:
It should support any HTTP/S URL (it's extracted the authority). Do you recall which site names were failing?

psiinon:
They werent failing, I just had to try a load of different variations of the site name before I hit the one that worked. And I thought that if I dont know what its supposed to look like how would anyone else?

thc202:
OK, could you provide the site names that didn't work? Just want to check if the code supposed to create the expected name is not working correctly (as it seems).

psiinon:
I think I started off trying names like https://www.example.com which is how the site is refered to in other places in the api (I think) when in fact it needed to be like www.example.com:443 (from memory)

thc202:
Odd, that name should have worked, do you recall with which action/view was that happening? (AFAICT all should work fine with the URL you provided.)

Pull request looks good to me.

src/org/zaproxy/zap/extension/httpsessions/HttpSessionsAPI.java
```diff
@@ -231,6 +235,13 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
 
 		HttpSessionsSite site;
 		switch (name) {
+		case VIEW_SITES:
+			// Get existing sessions for all sites
```
kingthorin: Thus comment is a bit misleading, sites are returned not sessions.

psiinon: Doh! Will fix..

psiinon:
Fixed comment

kingthorin:
Looks good

PR #2750, opened 2016-08-05
Add a reasonably strong CSP to the API

psiinon:


src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -964,6 +967,41 @@ public void process(HttpMessage msg) {
...
+					"  var form=document.getElementById('zapform');\n" +
+					"  form.action = url;\n" +
+					"  form.submit();\n" +
+					"}\n" +
+					"document.addEventListener('DOMContentLoaded', function () {\n" +
+					"  var button=document.getElementById('button');\n" +
+					"  if (button) {\n" +
+					"    document.getElementById('button').addEventListener('click',  function(e) {submitScript();}, false);\n" +
+					"  }\n" +
+					"});\n";
```
kingthorin: Wouldn't it be more efficient to use a stringbuilder?

Actually should this be extracted to a constant or file instead of rebuilding it for every response where it's needed?

psiinon: I dont thinka StringBuilder will be more efficient - the compiler will just make this into one String. The response can also be cached, so browsers should only ever ask for it once.
However it would probably be better as a resource file that gets cached in the class.

thc202: Although the response can be cached it might not be, it currently relies on heuristics. [1]
Probably better to provide an explicit expiration time?

[1] https://tools.ietf.org/html/rfc7234#section-4.2.2

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -198,10 +199,12 @@ public CoreAPI() {
 				OTHER_SEND_HAR_REQUEST,
 				new String[] { PARAM_REQUEST },
 				new String[] { PARAM_FOLLOW_REDIRECTS }));
+		this.addApiOthers(new ApiOther(OTHER_SCRIPT_JS, false));
```
thc202: Do we really want this exposed in the "normal" endpoints (which end up in the client APIs)? Why not just as API shortcut?

src/org/zaproxy/zap/extension/api/CoreAPI.java
psiinon: max-age=86400 ie one day? Although that wont help if the user updates to a new version of ZAP :/ Maybe we should have a v=1 type url param which we increment if/when the script file changes?

src/org/zaproxy/zap/extension/api/CoreAPI.java
psiinon: No, I just copy-n-pasted another one ;)

psiinon:
Updated to allow browsers to cache the script for one day, and introduced versioning for when we need to change it

kingthorin:
Looks goodLeave a comment

thc202:
Thanks!

PR #2753, opened 2016-08-08
Use PopupMenuItemAlert in core classes

thc202:
Change classes PopupMenuAlertDelete, PopupMenuAlertEdit and
PopupMenuAlertsRefresh to use PopupMenuItemAlert, instead of "manually"
accessing the Alerts tree.
Change class PopupMenuItemAlert to return just one alert selected if the
menu item does not support multiple selected alerts when a middle alert
node (one that shows the alert name) is selected, instead of returning
all child alerts or the number of all child alerts when checking if the
button of the menu item is enabled for the number of selected alerts.
The change allows menu items that do not support multiple selection to
still be used when the middle alerts are selected (for example, the
edit menu item).
Add some JavaDocs to the classes.

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/PopupMenuAlertEdit.java
+++ b/src/org/zaproxy/zap/extension/alert/PopupMenuAlertEdit.java
@@ -19,76 +19,35 @@
  */
 package org.zaproxy.zap.extension.alert;
 
-import java.awt.Component;
-
-import javax.swing.JTree;
-import javax.swing.tree.DefaultMutableTreeNode;
-
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.core.scanner.Alert;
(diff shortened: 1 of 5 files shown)


src/org/zaproxy/zap/extension/alert/PopupMenuAlertEdit.java
```diff
@@ -19,76 +19,35 @@
...
-						}
-			        }
-			    }
-        	    
-        	}
-        });
-			
 	}
-	
+
```
kingthorin: Oops :)

kingthorin:
Looks good

src/org/zaproxy/zap/extension/alert/PopupMenuAlertEdit.java
thc202: Fixed, I was checking other diff which didn't include that change ;)

PR #2759, opened 2016-08-10
Remove redundant test in SpiderHtmlParserUnitTest

thc202:
Remove redundant test (and incorrect, "parsed" should be true)
"shouldNotParseMessageIfAlreadyParsed". The behaviour is already being
asserted with "shouldNotParseHtmlResponseIfAlreadyParsed".


kingthorin:
Looks good

PR #2760, opened 2016-08-10
Add tests for SpiderSitemapXMLParser

thc202:
Add tests to assert the expected behaviour of SpiderSitemapXMLParser.
Other minor changes were done to class SpiderSitemapXMLParser:
 it's required to check if sitemap.xml files should be parsed;
 - Do not check if the value of the node (of the URL location) is non
 null, the value is guaranteed to be non null (per tests and JavaDoc).

### DIFF
--- a/src/org/zaproxy/zap/spider/parser/SpiderSitemapXMLParser.java
+++ b/src/org/zaproxy/zap/spider/parser/SpiderSitemapXMLParser.java
@@ -79,9 +79,13 @@ public class SpiderSitemapXMLParser extends SpiderParser {
 	 * Instantiates a new sitemap.xml parser.
 	 * 
 	 * @param params the params
+	 * @throws IllegalArgumentException if {@code params} is null.
 	 */
 	public SpiderSitemapXMLParser(SpiderParam params) {
 		super();
+		if (params == null) {
+			throw new IllegalArgumentException("Parameter params must not be null.");
+		}
 		this.params = params;
 	}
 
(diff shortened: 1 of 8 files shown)


kingthorin:
Looks goodLeave a comment

thc202:
Changed the name of one test, from "shouldFindUrlsInAreaElements" to "shouldFindUrlsInValidSitemapXml" :)

JordanGS:
@thc202 @psiinon @kingthorin

Since a report cannot be generated with Sites, @thc202 suggestion I have been making it work by retrieving the list of alerts with the following command.

`List<Alert> alerts = Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class).getAllAlerts();` 

To get the above call to work, the below fixed needed to be applied; creation of `public List<Alert> getAlerts(String cleanNodeName) ` in AlertTreeModel.java.

No changes needed to be made to ExtensionAlert.java which were suggested.

https://github.com/thc202/zaproxy/commit/49eeeac3869cf3f2315ec877bbdba47d163013cc

I had a 1 to 1 ratio between the GUI alerts and the report alerts. Having said that, I am no longer able to use the above call to get the alerts. Something in this merge request broke it and creates a java out of memory heap error. I'm back to square one. Any suggestions would be greatly appreciated.

thc202:
The changes to `AlertTreeModel` are not need to use `getAllAlerts()`, that's an existing core function which just relies on DB.

You sure it was this pull request? It just adds a unit test (and does minor changes to the spider parser).
Is the memory error happening with the same session?

JordanGS:
@thc202 as discussed in irc, i removed the test dir from the build path and it works now. Some changes in the test dir caused the OOME issue when making calls to `List<Alert> alerts = Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class).getAllAlerts();`

PR #2763, opened 2016-08-11
Sort statistics keys in the API views

thc202:
Change StatsAPI to return the statistics with the keys sorted as it
makes it easier to read (for humans, similar statistics are grouped
together).


kingthorin:
Looks good

psiinon:
Nice - I kept meaning to do this myself ;)

PR #2765, opened 2016-08-12
Update to Java 8 in weekly docker image

psiinon:
Planning on waiting for a couple of weeks and then changing and rereleasing the stable image to use Java 8 as well


kingthorin:
Looks good

PR #2773, opened 2016-08-15
Use UTF-8 to read/write ZAP scripts

mahoyaya:
fix: If you have a script file that contains multi-byte characters to load or save, the code is broken. It was changed to specify the utf8 when you load or save files.

### DIFF
--- a/src/org/zaproxy/zap/extension/script/ExtensionScript.java
+++ b/src/org/zaproxy/zap/extension/script/ExtensionScript.java
@@ -23,13 +23,13 @@ import java.awt.EventQueue;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileReader;
-import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Writer;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.security.InvalidParameterException;
 import java.text.MessageFormat;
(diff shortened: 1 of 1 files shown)


thc202:
Code changes look good to me, commit message could be tweaked though (subject split from the message [1]).

Worth noting that this change will/can break existing scripts that were working correctly (i.e. relying on default character encoding).
Ideally we should allow to choose the encoding, but using UTF-8 always might be a good compromise.

[1] https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md

thc202:
@mahoyaya any comment on the above? (re the commit message.)

What everyone thinks of always using the same character encoding (instead of relying on default or allow to choose).

psiinon:
I'm good with always using the same character encoding.

kingthorin:
Same here

mahoyaya:
I also best do not know. However, in the Japanese environment because the script is written and read by the character code ISO-8859-1, the multi-byte string will break. So, the script in the character encoding of UTF-8 is treated is thought to be preferred.

thc202:
Good to merge then?

kingthorin:
I think so

thc202:
@mahoyaya thanks! How would you like to be credited? [1]

[1] https://github.com/zaproxy/zap-core-help/wiki/HelpCredits

mahoyaya:
@thc202 Please display "Daiki Ichinose (@mahoyaya)" for credited. thanks.

thc202:
Done in zaproxy/zap-core-help@16431196edc82591d05e7c4f95d3b0ddbfc39b3a. Thank you!

PR #2775, opened 2016-08-15
Extract correct selected value from combined view

### LINKED ISSUE #2774: Wrong value shown in fuzz location for body text when selected through combined view
From #2766.

Steps to reproduce the issue:
2. Select a HTTP message with request body;
3. In Request tab select the combined view (request and body shown in the same text area);
4. Select a segment of the request body and open the Fuzzer dialogue;
5. Note that in the Fuzz Locations the Value has different text than the one selected (the text shown is part of the request header not body).

ZAP:
2.4.0

thc202:
Change class HttpRequestAllPanelSyntaxHighlightTextArea to return the
correct selected value, extracted directly from the request header or
the request body (instead of extracting the text from the text area,
which was using relative positions resulting in wrong text being
extracted).

Fix #2774 - Wrong value shown in fuzz location for body text when
selected through combined view


kingthorin:
Looks good

PR #2777, opened 2016-08-16
Add JavaDoc to some abstract view classes

thc202:
Document some methods, constructors and type arguments in abstract view
classes.

### DIFF
--- a/src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
+++ b/src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
@@ -29,7 +29,6 @@ import java.awt.CardLayout;
 import java.awt.Font;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
-import java.awt.HeadlessException;
 import java.awt.Insets;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
@@ -59,6 +58,8 @@ import org.zaproxy.zap.utils.ZapTextField;
 
 public class AbstractParamContainerPanel extends JSplitPane {
 
+    private static final String DEFAULT_ROOT_NODE_NAME = "Root";
+
     private static final long serialVersionUID = -5223178126156052670L;
     protected Object paramObject = null;
     
@@ -83,19 +84,20 @@ public class AbstractParamContainerPanel extends JSplitPane {
     // ZAP: Added logger
     private static Logger log = Logger.getLogger(AbstractParamContainerPanel.class);
 
+    /**
+     * Constructs an {@code AbstractParamContainerPanel} with a default root node's name ({@value #DEFAULT_ROOT_NODE_NAME}).
+     */
     public AbstractParamContainerPanel() {
         super();
         initialize();
     }
 
     /**
-     * @param parent
-     * @param modal
-     * @param title
-     * @param rootName
-     * @throws HeadlessException
+     * Constructs an {@code AbstractParamContainerPanel} with the given root node's name.
+     * 
+     * @param rootName the name of the root node
      */
-    public AbstractParamContainerPanel(String rootName) throws HeadlessException {
+    public AbstractParamContainerPanel(String rootName) {
         initialize();
         getRootNode().setUserObject(rootName);
     }
@@ -242,7 +244,11 @@ public class AbstractParamContainerPanel extends JSplitPane {
     }
 
     /**
-     * @return
+     * Gets the headline panel, that shows the name of the (selected) panel and has the help button.
+     * 
+     * @return the headline panel, never {@code null}.
+     * @see #getTxtHeadline()
+     * @see #getHelpButton()
      */
     private JPanel getPanelHeadline() {
         if (panelHeadline == null) {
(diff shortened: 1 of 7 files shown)


src/org/parosproxy/paros/view/AbstractParamDialog.java
```diff
@@ -238,21 +247,39 @@ private AbstractParamContainerPanel getJSplitPane() {
...
      *
-     * @param parentParams
-     * @param name
-     * @param panel
+     * @param parentParams the name of the parent nodes of the panel, might be {@code null}.
+     * @param name the name of the panel, must not be {@code null}.
+     * @param panel the panel, must not be {@code null}.
+     * @param sort {@code true} if the panel should be added in alphabetic order, {@code false} otherwise
      */
     // ZAP: Added sort option
```
kingthorin: Could be removed? Sort is handled numerous places here and not documented in this way....

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
```diff
@@ -420,9 +449,14 @@ public void showParamPanel(String name) {
     }
 
     /**
+     * Shows the panel with the given name.
+     * <p>
+     * The previously shown panel (if any) is notified that will be hidden.
```
kingthorin: `that it will be hidden` ?

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
```diff
@@ -402,8 +426,13 @@ public void showParamPanel(String parent, String child) {
     }
 
     /**
+     * Shows the panel with the given name.
+     * <p>
+     * Nothing happens if there's no panel with the given name (or the name is empty or {@code null}).
+     * <p>
+     * The previously shown panel (if any) is notified that will be hidden.
```
kingthorin: `that it will be hidden` ?

src/org/parosproxy/paros/view/AbstractParamDialog.java
```diff
@@ -263,37 +290,64 @@ public void showParamPanel(String parent, String child) {
     }
 
     /**
+     * Shows the panel with the given name.
+     * <p>
+     * Nothing happens if there's no panel with the given name (or the given name is empty or {@code null}).
+     * <p>
+     * The previously shown panel (if any) is notified that will be hidden.
```
kingthorin: `that it will be hidden` ?

src/org/parosproxy/paros/view/AbstractParamDialog.java
```diff
@@ -263,37 +290,64 @@ public void showParamPanel(String parent, String child) {
...
+     * @param name the name of the panel to be shown
      */
     public void showParamPanel(String name) {
     	this.getJSplitPane().showParamPanel(name);
     }
 
     /**
+     * Shows the panel with the given name.
+     * <p>
+     * The previously shown panel (if any) is notified that will be hidden.
```
kingthorin: `that it will be hidden` ?

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
```diff
@@ -457,8 +498,14 @@ public void initParam(Object obj) {
     }
 
     /**
-     * This method is to be overrided by subclass.
-     *
+     * Validates all panels, throwing an exception if there's any validation error.
+     * <p>
+     * The message of the exception can be shown in GUI components (for example, an error dialogue) as it's expected to be
```
kingthorin: If the message isn't being passed in, how is this supposed to be interpreted or addressed by the dev? (Applies to all occurrences in this PR.)

src/org/parosproxy/paros/view/AbstractParamDialog.java
thc202: You mean the ZAP comment? I'm OK with removing it.

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
thc202: Will change (ditto for other comments).

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
thc202: The message is obtained from the Exception, possibly, thrown by this method. So, callers can/should catch the exception and show its message in GUI components.

src/org/parosproxy/paros/view/AbstractParamDialog.java
kingthorin: Yes that's what I meant.

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
kingthorin: Can I suggest something like:
> The message of the exception can be shown in GUI components (for example, an error dialogue) callers can expect an internationalised message.

or something like that?

src/org/parosproxy/paros/view/AbstractParamDialog.java
thc202: Removed.

src/org/parosproxy/paros/view/AbstractParamContainerPanel.java
thc202: Sure! JavaDocs updated.

thc202:
All comments addressed.

kingthorin:
Looks good to me

PR #2779, opened 2016-08-16
Use all form submit fields when spidering

### LINKED ISSUE #2748: ZAP Spidering HTML Forms with multiple submit buttons
ZAP when encountering an HTML Form with multiple submit buttons defined

e.g.

```
<form action="test.jsp">
<input type="text" value="a" size="10" maxlength="10" name="name">
<input type="submit" value="Reload" name="submit1">
<input type="submit" value="Basic" name="submit2">
```

it spiders for the following URL
/test.jsp?name=a&submit1=Reload&submit2=Basic

instead of submitting two URLs like
/test.jsp?name=a&submit1=Reload
/test.jsp?name=a&submit2=Basic

Reference: https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java#L109

Additionally, the "value" attribute of the input type="submit" element can be used as a differentiator in the edge case where the two submit buttons have the same ID/name

e.g.
```
<form action="test.jsp">
<input type="text" value="a" size="10" maxlength="10" name="name">
<input type="submit" value="Reload" name="submit">
<input type="submit" value="Basic" name="submit">
```

the spider can generate the following requests
/test.jsp?name=a&submit=Reload
/test.jsp?name=a&submit=Basic

thc202:
Change class SpiderHtmlFormParser to use all submit fields when
spidering, using each submit field in turn with remaining form fields.
Add some tests to assert the expected behaviour of SpiderHtmlFormParser.
Other minor changes were done to "normalise" the behaviour/expectations
of class SpiderHtmlFormParser:
 it's required to check if HTML POST forms should be parsed;
 - Change the instance variable "param" to private and final.
 - Expect always a message when parsing (letting a NullPointerException
 be thrown if not).

Fix #2748 - ZAP Spidering HTML Forms with multiple submit buttons

### DIFF
--- a/src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
+++ b/src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
@@ -20,6 +20,7 @@ package org.zaproxy.zap.spider.parser;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Iterator;
@@ -56,15 +57,19 @@ public class SpiderHtmlFormParser extends SpiderParser {
 	private static final String DEFAULT_PASS_VALUE = DEFAULT_TEXT_VALUE;
 
 	/** The spider parameters. */
-	SpiderParam param;
+	private final SpiderParam param;
 
 	/**
 	 * Instantiates a new spider html form parser.
 	 * 
 	 * @param param the parameters for the spider
+	 * @throws IllegalArgumentException if {@code param} is null.
 	 */
 	public SpiderHtmlFormParser(SpiderParam param) {
 		super();
+		if (param == null) {
+			throw new IllegalArgumentException("Parameter param must not be null.");
+		}
 		this.param = param;
 	}
 
@@ -82,12 +87,7 @@ public class SpiderHtmlFormParser extends SpiderParser {
 		}
 
 		// Get the context (base url)
-		String baseURL;
-		if (message == null) {
-			baseURL = "";
-		} else {
-			baseURL = message.getRequestHeader().getURI().toString();
-		}
+		String baseURL = message.getRequestHeader().getURI().toString();
 
 		// Try to see if there's any BASE tag that could change the base URL
 		Element base = source.getFirstElement(HTMLElementName.BASE);
@@ -121,12 +121,16 @@ public class SpiderHtmlFormParser extends SpiderParser {
 				continue;
 			}
 
-			// Prepare data set
-			List<HtmlParameter> formDataSet = prepareFormDataSet(form.getFormFields());
+			FormData formData = prepareFormDataSet(form.getFormFields());
 
 			// Process the case of a POST method
 			if (method != null && method.trim().equalsIgnoreCase(METHOD_POST)) {
-				String query = "";
+				// Build the absolute canonical URL
+				String fullURL = URLCanonicalizer.getCanonicalURL(action, baseURL);
+				if (fullURL == null) {
+					return false;
+				}
+				log.debug("Canonical URL constructed using '" + action + "': " + fullURL);
 
 				/*
 				 * Ignore encoding, as we will not POST files anyway, so using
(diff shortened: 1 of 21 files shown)


src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -121,37 +121,39 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
...
 				}
 
-				log.debug("Canonical URL constructed using '" + action + "': " + fullURL);
-				notifyListenersPostResourceFound(message, depth + 1, fullURL, query);
+				for (HtmlParameter submitField : formData.getSubmitFields()) {
+					notifyPostResourceFound(
+							message,
+							depth,
+							fullURL,
+							appendEncodedUrlQueryParameter(baseRequestBody, submitField));
```
kingthorin: Does this really need to be split across 5 lines?

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+        SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser();
+        boolean parsed = true;
+        // When
+        boolean canParse = htmlParser.canParseResource(new HttpMessage(), ROOT_PATH, parsed);
+        // Then
+        assertThat(canParse, is(equalTo(false)));
+    }
+
+    @Test
+    public void shouldNotParseNonHtmlResponse() {
```
kingthorin: This name seems a bit misleading `HttpMessage()` creates an html response, though without a filename param it might be empty. So it "is" an html response (per the set content type), there's just no content....

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+        Source source = null;
+        SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser();
+        HttpMessage messageHtmlResponse = createMessageWith("NoForms.html");
+        // When
+        htmlParser.parseResource(messageHtmlResponse, source, BASE_DEPTH);
+        // Then = No exception
+    }
+
+    @Test
+    public void shouldNeverConsiderCompletelyParsed() {
```
kingthorin: `shouldNeverConsiderCompletelyParsedWhenSourceNull`? Or why should this never be considered fully parsed?

kingthorin:
Additional tests needed?
- POST test cases that mirror the included GET cases? 
- Single page with mixed form methods? (One GET one POST?)
- Form with no method specified? [_Should default to GET I guess?_]

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
thc202: Depends, that's done to improve readability, but can change that.
(I'm using the formatter suggested in the dev mailing list a long time ago.)

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
thc202: It's creating a new `HttpMessage` (but not using `createMessageWith`), which by default has no response (so, it's not HTML).

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
thc202: Whether or not it's considered parsed does not depend on the inputs, that's why it just says that. Why it has that behaviour? I guess because it's a fall back parser.

thc202:
Sure, will add those tests as well (I did test POST, though).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
kingthorin: Ok. Seems like overkill to me but it's a really minor detail.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
kingthorin: Ok

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
thc202: I don't mind using other formatter rules that don't do that (using that one for consistency).

kingthorin: Leave it for now.

thc202: OK.

psiinon:
@thc202 are you planning on adding those tests?

thc202:
Yes.

thc202:
Pull request updated with more tests. There's still some untested behaviour which will require some changes to normalise it (unrelated to this pull request, mostly about handling of HTML base element).

thc202:
And it's rebased.

kingthorin:
Looks good to me

PR #2782, opened 2016-08-17
Support the -configfile cmdline parameter

psiinon:


src/org/parosproxy/paros/CommandLine.java
```diff
@@ -328,6 +332,28 @@ private boolean parseKeywords(String[] args, int i) throws Exception {
                 this.configs.put(pair.substring(0, eqIndex), pair.substring(eqIndex + 1));
                 result = true;
             }
+        } else if (checkPair(args, CONFIG_FILE, i)) {
+            String conf = keywords.get(CONFIG_FILE);
+            File confFile = new File(conf);
+            if (! confFile.exists()) {
```
Grunny: (Not a Java developer ;)) Would `!confFile.isFile()` be a better check here (since `exists` and `canRead` would return true for a readable directory)?

src/org/parosproxy/paros/CommandLine.java
```diff
@@ -328,6 +332,28 @@ private boolean parseKeywords(String[] args, int i) throws Exception {
...
+                throw new Exception(error);
+            } else if (! confFile.canRead()) {
+            	// We cant use i18n here as the messages wont have been loaded
+            	String error = "File not readable: " + confFile.getAbsolutePath();
+            	System.out.println(error);
+                throw new Exception(error);
+            }
+            Properties prop = new Properties();
+            try (FileInputStream inStream = new FileInputStream(confFile)) {
+    			prop.load(inStream);
```
Grunny: Does ZAP have a standard for whitespace? It is currently a mix of spaces and tabs in quite a few places, which makes it a bit jarring to read. :)

kingthorin: Unfortunately no, standardizing a formatter and doing a cleanup run is on the long list of things TODO....

src/org/parosproxy/paros/CommandLine.java
psiinon: Yes, isFile (in addition to canRead) would be better :)
btw thanks for jumping in @Grunny - the more eyes we have on the code the better!

src/org/parosproxy/paros/CommandLine.java
psiinon: In the meantime new code should fit in with existing code. Which I keep on getting wrong :/
We should have done a global replace when we released 2.5.0. Guess that should be done for 2.6.0 now...

psiinon:
Fixed comments, and added ZAP comment in CommandLine

thc202:
Looks good to me.

PR #2783, opened 2016-08-17
Address some JavaDoc errors in ParameterParser

thc202:
Use HTML entities and document the parameters of method getAncestorPath.


kingthorin:
Looks good

Issue #368, opened 2015-06-04
API - report URLS just found by spider


zapbot:
```
Enhance the API to report URLs just found by the spider.
```

Original issue reported on code.google.com by `psiinon` on 2012-09-19 11:26:18

kingthorin:
Is http://zap/UI/spider/view/fullResults/ too verbose to cover this?

psiinon:
Yeah, I want to be able to see just the urls that the spider has found and not those found manually. I may well implement this fairly soon ;)

thc202:
The above API call returns just the URLs found by a specific spider scan, unless we want to exclude the seeds?

PR #2797, opened 2016-08-22
Fix Issue 368 Spider API URL List

kingthorin:
Added a Spider API end point:
- allUrls: Returns a list of all URLs found by all spider scans (based
on HistoryReferences)

Fixes zaproxy/zaproxy#368

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/SpiderAPI.java
+++ b/src/org/zaproxy/zap/extension/spider/SpiderAPI.java
@@ -18,9 +18,12 @@
 package org.zaproxy.zap.extension.spider;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.PatternSyntaxException;
 
 import net.sf.json.JSONObject;
@@ -30,8 +33,12 @@ import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.db.DatabaseException;
+import org.parosproxy.paros.db.RecordHistory;
+import org.parosproxy.paros.db.TableHistory;
+import org.parosproxy.paros.model.HistoryReference;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.zaproxy.zap.extension.api.ApiAction;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiException.Type;
@@ -86,6 +93,7 @@ public class SpiderAPI extends ApiImplementor {
 	private static final String VIEW_RESULTS = "results";
 	private static final String VIEW_FULL_RESULTS = "fullResults";
 	private static final String VIEW_SCANS = "scans";
+	private static final String VIEW_ALL_URLS = "allUrls";
 
 	/**
 	 * The Constant PARAM_URL that defines the parameter defining the url of the scan.
@@ -138,6 +146,7 @@ public class SpiderAPI extends ApiImplementor {
 		this.addApiView(new ApiView(VIEW_FULL_RESULTS, new String[] { PARAM_SCAN_ID }));
 		this.addApiView(new ApiView(VIEW_SCANS));
 		this.addApiView(new ApiView(VIEW_EXCLUDED_FROM_SCAN));
+		this.addApiView(new ApiView(VIEW_ALL_URLS));
 
 	}
 
@@ -460,6 +469,36 @@ public class SpiderAPI extends ApiImplementor {
 				resultList.addItem(new ApiResponseSet("scan", map));
 			}
 			result = resultList;
+		} else if (VIEW_ALL_URLS.equals(name)) {
+			ApiResponseList resultUrls = new ApiResponseList(name);
+			Set<String> urlSet=new HashSet<String>();
+			
+			TableHistory tableHistory = extension.getModel().getDb().getTableHistory();
+			List<Integer> ids = Collections.emptyList();
+			
+			try {
+				ids = tableHistory.getHistoryIdsOfHistType(extension.getModel().getSession().getSessionId(),
+						HistoryReference.TYPE_SPIDER, HistoryReference.TYPE_SPIDER_TASK);
+			} catch (DatabaseException e) {
+				throw new ApiException(ApiException.Type.INTERNAL_ERROR, e.getMessage());
+			}
+			
+			String url;
+			for (Integer id : ids) {
+				try {
+					RecordHistory rh = tableHistory.read(id.intValue());
+					if (rh != null) {
+						url = rh.getHttpMessage().getRequestHeader().getURI().toString();
+						if (urlSet.add(url)) {
+							resultUrls.addItem(new ApiResponseElement("url", url));
+						}
+					}
+				} catch (HttpMalformedHeaderException | DatabaseException e) {
+					throw new ApiException(ApiException.Type.INTERNAL_ERROR, e.getMessage());
+				}
+			}
+			
+			result = resultUrls;
 		} else {
 			throw new ApiException(ApiException.Type.BAD_VIEW);
 		}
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -138,6 +140,8 @@ public SpiderAPI(ExtensionSpider extension) {
 		this.addApiView(new ApiView(VIEW_FULL_RESULTS, new String[] { PARAM_SCAN_ID }));
 		this.addApiView(new ApiView(VIEW_SCANS));
 		this.addApiView(new ApiView(VIEW_EXCLUDED_FROM_SCAN));
+		this.addApiView(new ApiView(VIEW_URLS, new String[] { PARAM_SCAN_ID }));
```
thc202: This view seems to duplicate `VIEW_RESULTS` (but depends on following comments).

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +464,26 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
 				resultList.addItem(new ApiResponseSet("scan", map));
 			}
 			result = resultList;
+		} else if (VIEW_URLS.equals(name)) {
+			ApiResponseList resultUrls = new ApiResponseList(name);
+			SpiderScan scan = (SpiderScan) this.getSpiderScan(params);
+			synchronized (scan.getResourcesFound()) {
```
thc202: This can be extracted to a method and used by both views.

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +464,26 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
+				for (SpiderResource sr : scan.getResourcesFound()) {
```
thc202: Should also include URLs out of scope? (As those were also found by the spider.)
Should probably add the URLs to a Set (to prevent duplicate URLs).

thc202:
The view `allUrls` would be more useful if the scan results were kept indefinitely (they are not persisted between sessions and can be removed/discarded by the user).
Maybe it should use the persisted (spider) messages to get all URLs fetched by all spiders? That would give more consistent results.

kingthorin:
So I should base it on history references where type is spider or spider_task?

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
kingthorin: I guess it could build two sets kind of like fullResults (VIEW_FULL_RESULTS) does...

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
kingthorin: Sure if we keep 2.

thc202:
Including both types should be OK (we might need to take care of duplicates though).

kingthorin:
@psiinon #368 was originally opened by you. Any feedback? Based on what's here I guess I'll drop foundUrls (since it duplicates something that's there already). Then make allUrls based on history (using a set to avoid duplicates). Will that fill the original need?

kingthorin:
Made the executive decision to proceed as outlined :smiley: Let me know what you think.

psiinon:
I'm good with that :)

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +466,29 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
+
+			if (extHistory == null) {
+				throw new ApiException(Type.NO_IMPLEMENTOR, ExtensionHistory.NAME);
+			}
+	        
+			ApiResponseList resultUrls = new ApiResponseList(name);
+			Set<String> urlSet=new HashSet<String>();
+			
+			for (int idx = 0; idx <= extHistory.getLastHistoryId(); idx++) {
+				HistoryReference hr=extHistory.getHistoryReference(idx);
```
thc202: This requires to read (and caching almost) all HTTP messages, even if they are not of interest. It could obtain the spider HTTP messages directly, for example:
```Java
TableHistory tableHistory = extension.getModel().getDb().getTableHistory();
List<Integer> ids = tableHistory.getHistoryIdsOfHistType(extension.getModel().getSession().getSessionId(), HistoryReference.TYPE_SPIDER, HistoryReference.TYPE_SPIDER_TASK);
for (Integer id : ids) {
    String uri = tableHistory.read(id.intValue()).getHttpMessage().getRequestHeader().getURI().toString();
    // ...
```

thc202:
The view could have a description (i.e. using resource message key "spider.api.view.allUrls").

kingthorin:
I was wondering how that workshe.  Does the core just match based on the end point location/name?

psiinon:
Yeah, the i18n is just based on the api endpoints. The code generators actually tell you which ones arent present, but you should be able to work them out based on the existing ones.

kingthorin:
Done for allUrls.

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +468,32 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
+			TableHistory tableHistory = extension.getModel().getDb().getTableHistory();
+			List<Integer> ids = Collections.emptyList();
+			try {
+				ids = tableHistory.getHistoryIdsOfHistType(extension.getModel().getSession().getSessionId(),
+						HistoryReference.TYPE_SPIDER, HistoryReference.TYPE_SPIDER_TASK);
+			} catch (DatabaseException e) {
+				throw new ApiException(ApiException.Type.INTERNAL_ERROR, e.getMessage());
+			}
+			for (Integer id : ids) {
+					try {
```
thc202: The try-catch statement is not correctly indented (it has one more level than it should).

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +468,32 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
+					urlSet.add(
```
thc202: This could be tweaked to immediately add the `url` into the `resultUrls` (the `add` method returns  true if it object was added).

kingthorin: You mean wrap `resultUrls.addItem` around `urlSet.add`, because it won't do anything if urlSet.add doesn't add it (i.e.: it isn't unique)?

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
kingthorin: Thanks, will touch that up.

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
thc202: I mean something like:
```Java
if (urlSet.add(url)) {
    resultUrls.addItem(new ApiResponseElement("url", url));
}
```
If it was able to add to the Set it means it's unique and can be added to the results.

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -460,6 +468,33 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
...
+				try {
+					url = tableHistory.read(id.intValue()).getHttpMessage().getRequestHeader().getURI().toString();
+					if (urlSet.add(url)) {
+						resultUrls.addItem(new ApiResponseElement("url", url));
+					}
+				} catch (HttpMalformedHeaderException | DatabaseException e) {
+					throw new ApiException(ApiException.Type.INTERNAL_ERROR, e.getMessage());
+				}
+			}
+
```
kingthorin: Since urlSet is only used to enforce uniqueness should I be emptying it here? `urlSet=Collections.emptySet()` or will the normal GC take care of it?

thc202: GC will take care of it.

kingthorin: Great, this one done then? :smiley:

thc202: Actually, noticed a potential issue, we should also test that the `RecordHistory` is not `null` before using it, that can happen if one of the entries is deleted while we are iterating the IDs (likely to happen for `TYPE_SPIDER_TASK`, since those are deleted once no longer needed).

kingthorin: Ok, will address that.

kingthorin:
Done

thc202:
Looks good to me, thank you!

psiinon:
Looks good to me too. Might be nice to have a paging ption as well, as per the history urls, but they can be optional params added later :) Merging now...

Issue #1211, opened 2015-06-04
SQLi Scanner May Raise Seemingly Duplicate Alerts


zapbot:
```
To reproduce:
- Login to Bodgeit
- Scan the login request

Access ZAP API through browser and view the alerts (http://zap/HTML/core/view/alerts)
- Alerts contain two identical SQLi alerts


ZAP version 2.3.0.1
```

Original issue reported on code.google.com by `stephendv` on 2014-05-20 17:56:15

kingthorin:
Has this been rechecked recently?

kingthorin:
I've been told (via IRC) that this is still happening and is expected. It was suggested that perhaps we should allow/accept a Boolean for "unique" alerts.

kingthorin:
Confirmed as described. Note the "id" values are different. Will investigate further.

![1211_recheck](https://cloud.githubusercontent.com/assets/7570458/17825280/5f1dcf62-6635-11e6-9cf1-5fe998466b3f.png)

kingthorin:
Also note the "other info" of the two alerts is different.

thc202:
Right, that makes the two alerts different (not duplicates in that case). The duplicates might happen when two equal requests are scanned (passive or active), which should results in the same alerts being raised (they are the same but for different messages).

kingthorin:
The original test case above does highlight this issue. Really the only significant difference in the alerts is the other info or extrainfo (it's called the later in the code). The SQLi scanner should probably return (exit the current param being checked) once an alert is raised.

thc202:
Regarding https://github.com/zaproxy/zaproxy/issues/1211#issuecomment-241146812, the duplicates would not show up in the view "alerts" (only when accessing them by ID).

thc202:
Released in version 25 of "Active scanner rules" add-on.

PR #2810, opened 2016-08-22
Active scanners' alerts persisted twice when with GUI

thc202:
From discussion/research, in IRC channel, for #1211.

Steps to reproduce the issue:
1. Run ZAP;
2. Active scan a target website, for example, bodgeit;
3. View all alerts through the ZAP API;
4. Note that some alert IDs are missing (are shown only 0, 2, 4...);
5. View the missing alerts ("alert" view), comparing the ones shown with the missing ones they are exactly the same (e.g. 0 is equal to 1, 2 with 3...).

Remarks:
 - The issue only happens when running ZAP with GUI.
 - The duplicated alerts are not reflected in the reports or in UI components (e.g. Alerts tree).

ZAP:
Versions >= 2.4.0


kingthorin:
Looks good

PR #2811, opened 2016-08-22
Fix overlapping of text locations with same start

### LINKED ISSUE #2792: Able to overlap fuzz (HTTP) locations
Steps to reproduce the issue:
2. Open the Fuzzer dialogue for a HTTP message;
3. Select the method and add it as a location;
4. Select it again and add it as a location, not that it adds it instead of informing that the fuzz locations overlap (the issue happens when the selection starts at the first char of the header).

Add-on:
Fuzzer version 6.

thc202:
Change DefaultTextHttpMessageLocation to check if the new text location
is (or not) an insertion point, when the existing text location has the
same start and different end. If it's not an insertion point it's
overlapping the existing location.

Fix #2792 - Able to overlap fuzz (HTTP) locations


kingthorin:
Looks good

PR #2814, opened 2016-08-23
Remove redundant null check

thc202:
Remove redundant null check from StandardFieldsDialog, per a previous
check the variable can not be null.


kingthorin:
Looks good

PR #2815, opened 2016-08-23
Respect sort parameter for new intermediate panels

thc202:
Change AbstractParamContainerPanel to respect the sort parameter when
adding new intermediate parent panel nodes, so that they are inserted
in correct location.


kingthorin:
Looks good

PR #2816, opened 2016-08-23
Allow to keep standard dialogue open after saving

thc202:
Change StandardFieldsDialog to allow to configure whether or not the
dialogue should be automatically hidden after saving. The change allows,
for example, to keep the dialogue open if the saving process was not
successful to give the user a chance to try to fix the cause of the
problem (that was not possible to detect during the validation process).
 ---
Related to comments/changes in zaproxy/zap-extensions#441.

### DIFF
--- a/src/org/zaproxy/zap/view/StandardFieldsDialog.java
+++ b/src/org/zaproxy/zap/view/StandardFieldsDialog.java
@@ -96,6 +96,13 @@ public abstract class StandardFieldsDialog extends AbstractDialog {
 	private Map<String, JPanel> tabNameMap = new HashMap<String, JPanel>(); 
 
 	/**
+	 * Flag that indicates whether or not the dialogue is automatically hidden when {@link #save() saved}.
+	 * 
+	 * @see #isHideOnSave()
+	 */
+	private boolean hideOnSave;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/view/StandardFieldsDialog.java
```diff
@@ -133,6 +140,31 @@ public StandardFieldsDialog(Window owner, String titleLabel, Dimension dim, Stri
...
+		this.hideOnSave = true;
+	}
+
+	/**
+	 * Tells whether or not the dialogue is automatically hidden when {@link #save() saved}.
+	 * <p>
+	 * The default is {@code true}.
+	 * 
+	 * @return {@code true} if the dialogue should be hidden, {@code false} otherwise.
+	 * @since 2.6.0
```
kingthorin: Are we sure about this?

thc202: That's the target, changed to a TODO anyway.

thc202:
Changed the methods to protected, does not really need to be public.

kingthorin:
Looks good

PR #2817, opened 2016-08-24
Allow to add a password field to standard dialogue

thc202:
Change class StandardFieldsDialog to allow to add a JPasswordField and
obtain its value/password.


kingthorin:
Looks good to me

PR #2818, opened 2016-08-24
Reduce code duplication in standard dialogue

thc202:
Extract a method, in StandardFieldsDialog, that initialises its content
panel by adding the (main) component and default and extra buttons.


psiinon:
Looks good

Issue #2757, opened 2016-08-09
Alerts with different request method are considered the same


JordanGS:
I would like to reopen an old issue which is a blocker to all report generation.

Originally raised in: https://github.com/zaproxy/zaproxy/issues/2132
Custom Fix in: https://github.com/thc202/zaproxy/releases/tag/w2016-01-22

It was a 2.5.0 release milestone that was address but does not seem to have been applied, i have tested the custom build `w2016-01-22` and it still works as a fix but it does not seem to have been applied to the 2.5.0 release.

Could we re-open this issue and if it's an easy fix apply it to the upcoming weekly's and as part of the 2.6.0 milestone please?

EDIT: I have looked through the source and the fix has been applied to Alerts.java however this does not solve the issue in 2.5.X series at all. `w2016-01-22` is unaffected however, that would indicate that there is a change in Alerts or in another class which is causing the report generation to break.

#2132 
@thc202 
@kingthorin 
@psiinon

thc202:
OWASP ZAP Developer Group thread:
https://groups.google.com/d/topic/zaproxy-develop/GqQ9z932ziI/discussion

thc202:
Per discussion in dev thread changing the issue to include the request method when differentiating the alerts.

PR #2821, opened 2016-08-25
Initialise Alert's method to an empty string

thc202:
The change prevents NullPointerException when testing for equality,
doing comparisons or obtaining the hash code if the alert no longer has
a message (thus the method was not being set to a non-null value).
 ---
Issue reported in IRC. Related to #2757.


kingthorin:
Looks good to me, thanks

PR #2824, opened 2016-08-25
Detach Sites tree model when loading the session

thc202:
Change Session class to set a temporary SiteMap tree model into the
Sites tree just before starting to load the session to reduce the number
of (unnecessary) events that the view (EDT) has to process (which tend
to be a high number, when many alerts are added to each site node). The
SiteMap that contains the loaded data is set into the Sites tree after
the loading has finished.
The change, slightly, reduces the time it takes to load a session.


kingthorin:
Looks good

Issue #1107, opened 2015-06-04
Additional Commentary needed for Script Templates/Examples


zapbot:
```
It would be really handy if the example or template scripts make it clear what parameters
they are accepting/leveraging.

I recently turned the targeted comment finder example into a passive rule, however,
without going through the extra steps of doing print lines on ps, msg, and src along
the way there was nothing giving me any clues what those items were or where they were
coming from. 

Having done so it became obvious that I could replace hr.getHttpMessage() from the
targeted comment finder with msg which the passive template was already using.

```

Original issue reported on code.google.com by `kingthorin` on 2014-03-13 17:20:52

PR #2825, opened 2016-08-26
Additional Commentary for JS templates

kingthorin:
Add JSDoc style comments to base functions in:

Active default template.js
Passive default template.js
Proxy default template.js
Targeted default template.js

Other script types seemed to have sufficiently complete commentary.

Fixes zaproxy/zaproxy#1107

### DIFF
--- a/src/scripts/templates/active/Active default template.js
+++ b/src/scripts/templates/active/Active default template.js
@@ -1,10 +1,15 @@
-// The scanNode function will typically be called once for every page 
-// The scan function will typically be called for every parameter in every URL and Form for every page 
-
 // Note that new active scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
-
+/**
+ * Scans a "node", i.e. an individual entry in the Sites Tree.
+ * The scanNode function will typically be called once for every page. 
+ * 
+ * @param as - the ActiveScan parent object that will do all the core interface tasks 
+ *     (i.e.: sending and receiving messages, providing access to Strength and Threshold settings,
+ *     raising alerts, etc.). This is an ScriptsActiveScanner object.
+ * @param msg - the HTTP Message being scanned. This is an HttpMessage object.
+ */
 function scanNode(as, msg) {
 	// Debugging can be done using println like this
 	println('scan called for url=' + msg.getRequestHeader().getURI().toString());
@@ -36,6 +41,17 @@ function scanNode(as, msg) {
 	}
 }
 
+/**
+ * Scans a specific parameter in an HTTP message.
+ * The scan function will typically be called for every parameter in every URL and Form for every page.
+ * 
+ * @param as - the ActiveScan parent object that will do all the core interface tasks 
+ *     (i.e.: sending and receiving messages, providing access to Strength and Threshold settings,
+ *     raising alerts, etc.). This is an ScriptsActiveScanner object.
+ * @param msg - the HTTP Message being scanned. This is an HttpMessage object.
+ * @param {string} param - the name of the parameter being manipulated for this test/scan.
+ * @param {string} value - the original parameter value.
+ */
 function scan(as, msg, param, value) {
 	// Debugging can be done using println like this
 	println('scan called for url=' + msg.getRequestHeader().getURI().toString() + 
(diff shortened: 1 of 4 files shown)


src/scripts/templates/active/Active default template.js
```diff
@@ -4,7 +4,14 @@
 // Note that new active scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
-
+/**
+ * Scans a "node", i.e. an individual entry in the Site Tree.
```
thc202: "Sites" instead?

Guess the first two lines of the file could be deleted, replaced by new docs?

src/scripts/templates/active/Active default template.js
```diff
@@ -4,7 +4,14 @@
...
+ * 
+ * @param as the ActiveScan parent object that will do all the core interface tasks
```
thc202: The actual class is ScriptsActiveScanner (which is itself a scanner).

src/scripts/templates/active/Active default template.js
```diff
@@ -36,6 +43,16 @@ function scanNode(as, msg) {
 	}
 }
 
+/**
+ * Scans a specific parameter in an HTTP message.
+ * 
+ * @param as the ActiveScan parent object that will do all the core interface tasks
```
thc202: Same here.

src/scripts/templates/active/Active default template.js
```diff
@@ -36,6 +43,16 @@ function scanNode(as, msg) {
...
 }
 
+/**
+ * Scans a specific parameter in an HTTP message.
+ * 
+ * @param as the ActiveScan parent object that will do all the core interface tasks 
+ *     (i.e.: sending and receiving messages, providing access to Strength and Threshold settings,
+ *     raising alerts, etc.)
+ * @param msg the HTTP Message being scanned
+ * @param param the parameter being manipulated for this test/scan
```
thc202: Could mention that it's the name of the parameter. Also, could indicate the type (string) per JSDoc? [1]

[1] http://usejsdoc.org/tags-param.html#names-types-and-descriptions

src/scripts/templates/active/Active default template.js
```diff
@@ -36,6 +43,16 @@ function scanNode(as, msg) {
...
+ * @param value the original parameter value retrieved by the proxy or spider
```
thc202: Is the "retrieved" part really needed? That might not be the case (messages manually sent, other tools).

src/scripts/templates/passive/Passive default template.js
```diff
@@ -4,6 +4,14 @@
 // Note that new passive scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
+/**
+ * Passively scans an HTTP message.
+ * 
+ * @param ps the PassiveScan parent object that will do all the core interface tasks
```
thc202: In this case the object is ScriptsPassiveScanner, also a scanner.

src/scripts/templates/active/Active default template.js
kingthorin: Hmm ok I'll try to revise to be more specific, trying to hit a balance between dev comments and user comments :smiley:

src/scripts/templates/active/Active default template.js
kingthorin: Good call

kingthorin:
All addressed

src/scripts/templates/passive/Passive default template.js
```diff
@@ -1,9 +1,18 @@
-// The scan function will be called for request/response made via ZAP, excluding some of the automated tools
 // Passive scan rules should not make any requests 
 
 // Note that new passive scripts will initially be disabled
 // Right click the script in the Scripts tree and select "enable"  
 
+/**
+ * Passively scans an HTTP message. // The scan function will be called for
```
thc202: The // could be removed.

kingthorin:
Look good now?

thc202:
Yeah, looks good to me, thank you! (I didn't notice your update, no email notification.)

kingthorin:
No problem, I'm not worried. Things can easily get buried/missed when lots of things are moving/updating/changing... Thanks

PR #2826, opened 2016-08-26
Show which param is at fault

psiinon:
Its almost as if I'm having problems running this script on a remote system that I cant login to ;)


kingthorin:
Looks good

Issue #527, opened 2015-06-04
Content Security Policy Passive scan rule


zapbot:
```
Add a passive scan rule to check CSP for:

Including the options eval-script or inline-script

Including any http resources from an https page

Including a large number of sources

Using an 'excessive' wildcard (like *.com)
```

Original issue reported on code.google.com by `psiinon` on 2013-02-19 15:51:26

kingthorin:
https://github.com/zaproxy/zap-core-help/wiki

kingthorin:
https://pokeinthe.io/2016/04/25/state-of-csp-alexa-top-one-million-2016-04/

psiinon:
Thats @marumari's site ;)

kingthorin:
Yup seemed like a good reference point for building this.

psiinon:
Definitely!

kingthorin:
What do you guys think about using:
https://github.com/shapesecurity/salvation

Is anyone aware of documentation pertaining to browser or spec handling for multiple headers (i.e.: if more than one CSP header is specified on the response). IIRC per the base HTTP RFCs multiple headers are supposed to be merged, however I wonder if there's a CSP specific behavior/expectation. https://www.w3.org/TR/CSP2/#enforcing-multiple-policies seems to suggest that policies be merged/unified. It also seems that browsers are starting to support CSP via meta, eg: https://bugzilla.mozilla.org/show_bug.cgi?id=663570

kingthorin:
I started to try setting this up leveraging the library mentioned above. Can someone provide insight on the following?

> java.lang.UnsupportedClassVersionError: com/shapesecurity/salvation/data/Origin : Unsupported major.minor version 52.0
> 	at java.lang.ClassLoader.defineClass1(Native Method)
> 	at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
> 	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
> 	at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
> 	at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
> 	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
> 	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
> 	at java.security.AccessController.doPrivileged(Native Method)
> 	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
[...]

PR #2831, opened 2016-08-29
CSP Scanner Plugin ID Reservation

kingthorin:
Reserve plugin ID 10055 for CSP passive scanner. See zaproxy/zaproxy#527

### DIFF
--- a/src/doc/alerts.xml
+++ b/src/doc/alerts.xml
@@ -96,6 +96,7 @@ This list the alerts
 10052	X-ChromeLogger-Data Header Information Leak
 10053	Apache Range Header DoS (CVE-2011-3192)
 10054	Cookie set without SameSite attribute
+10055	Content Security Policy Scanner
 10094   Base64 Disclosure
 10095   Backup File Disclosure
 10096   Timestamp Disclosure


thc202:
Any reason to not use 10038 (i.e. change the existing scanner to do the new checks)?

kingthorin:
Mainly that's it'd require an aver haul because it's no longer just "missing". It can be done though if that's the preference.

kingthorin:
If the team agrees that overhaul is a better idea then I'm good with that. This PR can just be closed.

thc202:
It does not require many changes, just a rename of the class and a change of the messages prefix. (It would not be the first to be changed to do more checks, XFO scanner was initially just checking if the header was missing.)
IMHO it would be better to have the related CSP checks in the same scanner (the current one already has a TODO to do more checks), but I'm OK with both ways.

kingthorin:
Ok, I can agree with that. Closing this in favour of using 10038.

kingthorin:
IRC update: Reopening. Will create a separate pscan addon and use 10055 for the new rule.

See zaproxy/zaproxy#527 for further details.

thc202:
Looks good to me.

PR #2832, opened 2016-08-29
Load alerts of loaded messages

### LINKED ISSUE #2736: Bug: Can't generate reports from saved Session data
If you save a session and then restore it later, and then try to generate HTML or XML reports, now of the issues saved and restored are included in the report. If you do more scanning and new issues are found, they are included, but not the old ones.

thc202:
Change Session class to also load the alerts of the loaded messages and
include the active scanner messages (which also have alerts).

Fix #2736 - Can't generate reports from saved Session data


thc202:
Because of a bottleneck in SiteNode (and, to less extent, in HistoryReference) the session loads will take longer, working on a fix for that...

PR #2835, opened 2016-08-30
Break loops on match in HistoryFilter

thc202:
Change class HistoryFilter to immediately break the loops if the tag or
alert filter conditions match (there's no need to continue checking as
one is enough). Also, do not access the collections tags and alerts if
not really needed.


kingthorin:
Looks goodLeave a comment

PR #2839, opened 2016-08-30
Delete all Alerts at once when deleting a SiteNode

thc202:
Change method Context.deleteNode to delete all alerts at once instead of
one at a time (which generates more change events/tasks).


kingthorin:
Looks good

JordanGS:
@thc202 could this cause a memory issue (OOME) if it's all at once and not deleted in batches or one at a time?

thc202:
No, this is just removing them from the SiteNode (it's unrelated to DB).

PR #2842, opened 2016-08-31
Dont call Xvfb is its already running

psiinon:


kingthorin:
Should there be an else condition with an error message? is `/tmp/.X1-lock` xvfb specific?

psiinon:
It is xvfb specific.
Theres no need to do anything else - it happens if you try running zap-x.sh more than once. Which we do at Mozilla when we run the baseline scan against a set of services ;)

thc202:
Looks good to me.

PR #2844, opened 2016-08-31
Set ExtensionUiUtils to core and other tweaks

thc202:
Change class ExtensionUiUtils to:
 - Be a core extension, to prevent it from being disabled through the UI
 as it has UI functionality that should be always enabled;
 - Correct its description;
 - Have the logger as final static;
 - Add the session changed listener only if there's view and remove a
 (now) redundant view check.


kingthorin:
Looks good

PR #2845, opened 2016-08-31
Remove hardcoded names in script scanners

thc202:
Remove hardcoded names in classes ScriptsActiveScanner and
ScriptsPassiveScanner, the resource messages will be loaded/available by
the time the names are required.


kingthorin:
Looks good

PR #2851, opened 2016-09-06
Hook OverrideMessageProxyListener into the Proxy

thc202:
Change Control class to hook extensions' OverrideMessageProxyListener
into the Proxy during the initialisation.


psiinon:
Looks good to me

PR #2852, opened 2016-09-06
Add HTML base element tests for HTML form parser

thc202:
Change class SpiderHtmlFormParser to require a non empty value in HTML
base element to be used.
Add tests to assert the expected behaviour when handling the HTML base
element for GET and POST forms.
 ---
From https://github.com/zaproxy/zaproxy/pull/2779#issuecomment-242163412.


kingthorin:
Looks good

PR #2856, opened 2016-09-08
Initial alert overrides implementation

### LINKED ISSUE #2853: Override Alert details
Summary: Allow users to override alert details, eg to include or point to internal / company specific info.

More detail
-------------
Many companies have their own advice and guidance for avoiding security vulnerabilities.
The may have their own terminology, and can often provide very specific information such as company mandated policies for specific languages platforms.
Proposal
-----------
New ‘alert override’ configuration file which can be specified via the command line / changed via the UI and stored in the config file.
Format something like:

`<alertid>.name = New name`
`<alertid>.description = New description`
`<alertid>.otherInfo = New other info`
`<alertid>.solution = Use our ‘fixes-everything’ lib`
`<alertid>.reference = http://internet.example.com/xss.html`
`<alertid>.extra = Extra info that wont overwrite anything`

Only generic info should be overridden, alert specific info such as the URI, attack, parameter, evidence and confidence shouldn’t be changed.
The file only needs to contain what needs to be changed, eg could just be:

`# 40012 = Reflected XSS`
`40012.extra = Follow our company specific guidelines at http://internet.example.com/xss.html`

Could also support adding rather than replacing info if the relevant line starts with a ‘+’, eg

`# 40012 = Reflected XSS`
`40012.solution = + Also follow our company specific guidelines at http://internet.example.com`/xss.html

In this case the above text would be added to the standard XSS solution.
Implementation
-------------------
Introduce new ‘extraInfo’ alert field, which scan rules should not set (protected methods?). Add to HTML and XML reports if present.
If file specified load as a properties file on start or when added.
When new alerts raised check for overrides and apply directly to the alerts as required.
Will therefore automatically appear in the UI, HTML and XML reports.
Don’t apply to existing alerts? Can revisit.
Why not add to the alert filters?
--------------------------------
Alert filters are really for changing specific alerts, eg “The XSS on https://www.example.com is a False Positive”.
These overrides are for changing generic information about alerts, ie the information that is either hard coded or read from the relevant add-on properties file.
While they are both related to alerts, combining them into the same mechanism feels wrong, somehow.
Open Questions
---------------------

- Should new overrides apply to existing alerts? 
 - “No” is easier ;)
- Should it be possible to change things like risk?
 - Possibly not as that’s what alert filters are for

psiinon:
Fixes #2853 
Note that this does not introduce the extraInfo field - that can be added if still thought worthwhile.
Will submit a PR for help changes as well.


psiinon:
Pushed new version with previously missing options pane and some indentation fixes ;)

test/org/zaproxy/zap/extension/alert/ExtensionAlertUnitTest.java
```diff
@@ -0,0 +1,421 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.parosproxy.paros.core.scanner.Alert;
+
+@RunWith(MockitoJUnitRunner.class)
```
thc202: Could be removed.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -804,4 +894,5 @@ public boolean supportsDb(String type) {
     public boolean supportsLowMemory() {
     	return true;
     }
+
```
thc202: Should be reverted.

psiinon:
Hopefully fixed those

thc202:
Looks good to me, thank you!

psiinon:
I still need to submit a PR for the help ;)
Maybe we should not merge zaproxy PRs until a PR has been openned for the related help pages?

thc202:
Sounds good to me.

psiinon:
Help PR raised :) zaproxy/zap-core-help#44

kingthorin:
Thanks

PR #2864, opened 2016-09-13
Show correct error for wrong regex in ContextAPI

thc202:
Change ContextAPI to throw ILLEGAL_PARAMETER (instead of INTERNAL_ERROR)
when the regular expression (for context inclusion or exclusion) is
invalid, which better indicates what the problem is.
Change Context class to not throw Exception and document the exceptions
expected to be thrown (PatternSyntaxException if the regular expression
is not valid or IllegalArgumentException if it it's null).
Change ContextExcludePanel and ContextIncludePanel to no longer catch
the exception when setting the (already validated) regular expressions.


kingthorin:
Looks good

psiinon:
Looks good, just needs conflicts resolving

thc202:
Resolved.

PR #2865, opened 2016-09-13
Normalise Include/Exclude context panels

thc202:
Change context panels "Include in Context" and "Exclude from Context" to
use the same tables as other dialogues to manage multiple options (that
is, table shown in one side and buttons to add/modify/remove at the
other).

Change class MultipleRegexesOptionsPanel to use a modifiable list by
default, to allow to do modifications without setting a list beforehand.
Remove no longer used resource messages.


kingthorin:
Looks good

PR #2868, opened 2016-09-15
Fix tests failing in Windows (for ZAP jar 2.5.0)

thc202:
Change SessionUtilsUnitTest to assert with effective paths instead of
using hardcoded ones, which would not match correctly in all OSs.


kingthorin:
Looks.good

PR #2870, opened 2016-09-15
Remove incorrect JavaDoc tags

thc202:
Remove incorrect JavaDoc tags, return tag when the method returns void
and param tag of inexistent type parameters.
Remove commented method and corresponding JavaDoc in BrowserDialog.


thc202:
Done, commit and pull request description updated.

src/org/zaproxy/zap/session/SessionManagementMethodType.java
```diff
@@ -37,8 +37,6 @@
  * The system automatically detects and loads {@link SessionManagementMethodType} classes and,
  * through them, the corresponding session management methods.
  * </p>
```
kingthorin: Paragraph marker is now un-needed?

thc202: Will remove it (it was not needed before either).

thc202:
Change done.

PR #2872, opened 2016-09-16
Handle parsing errors in VariantAbstractRPCQuery

thc202:
Change class VariantAbstractRPCQuery to handle the exceptions thrown by
custom implementations when parsing the request body. Log the exceptions
with WARN level (instead of ERROR) as that's not caused (most likely) by
a programming error but a faulty/malformed content.
 ---
Related to discussion in OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/uVZx3UvHG3Y/discussion


kingthorin:
Looks good

PR #2874, opened 2016-09-16
Correct offset calculation in text header views

### LINKED ISSUE #2793: Wrong highlight in combined view with last part of request header
Steps to reproduce the issue:
2. Open the Fuzzer dialogue for a HTTP message with request body;
3. With the split view selected, add as fuzz location the whole request header (Ctrl+A to select all text, easier to reproduce the issue with the whole request header);
4. Switch to combined view and note that some of the request body is wrongly highlighted as fuzz location.

ZAP:
2.4.0

thc202:
Extract the calculation of offsets for view to header and header to view
into a class (HttpTextViewUtils) and changed the text views to use it,
reduces code duplication and uses the correct calculations in all cases
(some calculations were already correct).
Add tests to assert the expected behaviour of HttpTextViewUtils.

Fix #2793 - Wrong highlight in combined view with last part of request
header

### DIFF
--- a/src/org/zaproxy/zap/extension/httppanel/component/all/request/HttpRequestAllPanelTextView.java
+++ b/src/org/zaproxy/zap/extension/httppanel/component/all/request/HttpRequestAllPanelTextView.java
@@ -21,13 +21,12 @@ import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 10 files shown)


thc202:
Don't review yet, will add unit tests.

thc202:
Pull request updated with tests (and fix for #2793).

PR #2877, opened 2016-09-19
Remove fragment from form's action for all methods

thc202:
Change SpiderHtmlFormParser to remove the fragment from form's action
for all methods (POST forms should not include the fragment either).


kingthorin:
Looks good

psiinon:
What the reasoning behind this?
Does it fix and existing problems?
I dont like loosing information, and this looks like something that could be useful if & when we focus more on client side scanning...

thc202:
This change normalises the behaviour, before only the GET forms had the fragment removed, now that applies to all methods (POST).
The fragment component is for consumption of client side, so no point sending it to the server. [1]

[1] https://tools.ietf.org/html/rfc3986#section-3.5

PR #2878, opened 2016-09-20
Remove port if scheme's default in canonicalizer

thc202:
Change URLCanonicalizer to only remove the default port if it's the
default port of the scheme (currently checked for HTTP/S).


kingthorin:
Looks good

PR #2879, opened 2016-09-20
Correct attribute's name in HTML form parser

thc202:
Correct the name of an attribute (max instead of min) for number/range
fields in SpiderHtmlFormParser.
Change SpiderHtmlFormParser to allow to set the date used for default
values of the form fields (easier to test).
Add tests to assert the expected behaviour of the expected values set to
the form fields.


thc202:
Some tests are failing... timezones...

thc202:
Tests fixed.

kingthorin:
Looks good to me

PR #2880, opened 2016-09-20
Prevent (potential) NPE in HostProcess

thc202:
Reorder statements in HostProcess.scanSingleNode to prevent a potential
NullPointerException when logging info about the node being scanned.


kingthorin:
Looks good

PR #2883, opened 2016-09-21
Log exceptions' stack trace on class loading error

thc202:
Change classes AddOnLoader and AddOnLoaderUtils to also log the
exception stack trace when failed to instantiate the classes (e.g.
extensions, passive/active scanners...). The exception stack trace gives
more clues to what the problem might be since the exception's message
might be empty leading to log just:
 DEBUG org.zaproxy.zap.control.AddOnLoaderUtils  -

which is not helpful.
 ---
Related to discussion in OWASP ZAP Developer Group: https://groups.google.com/d/topic/zaproxy-develop/jEAWmD_Df98/discussion


kingthorin:
Looks good

PR #2884, opened 2016-09-21
Preserve contents of the scripts when loading

thc202:
Change ExtensionScript to read chars from the script instead of reading
each line, to preserve its line ending chars (which could end up being
replaced/removed).


thc202:
Will conflict with #2773.

kingthorin:
Looks good

PR #2891, opened 2016-09-23
Show the cause why a script was not loaded

thc202:
Change ExtensionScript to provide more details why a script was not
successfully loaded (e.g. missing script type, invalid character
sequence, other unexpected causes).


kingthorin:
Looks good

PR #2895, opened 2016-09-26
Show white space chars of matches in Search panel

thc202:
Change SearchResultTableEntry to replace the white space characters in
the string found with visible equivalent characters so that the matches
in the Search panel are like:
 Content-Length:·453¤¶Connection:·close¤¶Content-Type:

instead of:
 Content-Length: 453Connection: closeContent-Type:

which does not reflect the actual match.


kingthorin:
Looks good

PR #2905, opened 2016-09-30
Upgrade to use ubuntu:16.04

### LINKED ISSUE #2903: Ubuntu 16.04 version bump?
Hi,
I was wondering if there was any blockers for moving the Dockerfile to Ubuntu 16.04?

15.04 is EOL since February 4th and is thus heavier on transfer for those using the LTS and is without security patches, etc.

cheers!

psiinon:
Fixes #2903 
The pip upgrade is required - the pip install of the API fails otherwise


kingthorin:
Looks good

PR #2907, opened 2016-09-30
Ignore incorrect start/end positions on highlight

thc202:
Change HttpTextViewUtils to return invalid position, instead of throwing
an exception, if the start or end positions are greater than the length
of the view/header/body, otherwise it would result in an exception when
the combined view is showing a custom message in the body (when body is
too large to display per configurations), moreover it is more resilient
to future changes in the content shown.
Update the tests to assert the new behaviour.
 ---
Related to #2874 - Correct offset calculation in text header views.


kingthorin:
Looks good

PR #2909, opened 2016-10-04
Update dependencies, Fix license

bkimminich:
@khamasaki, @psiinon: As there are no automated tests for the node.js API, please review closely before merging!

I bumped the version to 0.3 because on the old(?) code location @TeamPraxis org the latest version was already 0.2 while still being 0.1 in the @zaproxy org. It seemed that the API code was generated more recently in the @zaproxy repo, so I figured this is the latest.

If everything is fine, please release to npm. And maybe @khamasaki could remove the original repo on @TeamPraxis to avoid confusion? Or at least add a note that the location was changed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ out/
 /build/dist/
 /build/results/
 /bin/
+
+# Node
+node_modules/
--- a/nodejs/api/zapv2/package.json
+++ b/nodejs/api/zapv2/package.json
@@ -1,32 +1,27 @@
 {
   "name": "zaproxy",
   "description": "ZAProxy Client API for Node.js",
-  "version": "0.1.0",
-  "homepage": "https://github.com/TeamPraxis/node-zaproxy",
+  "version": "0.3.0",
+  "homepage": "https://github.com/zaproxy/zaproxy",
   "author": {
     "name": "Keith Hamasaki",
     "email": "khamasaki@teampraxis.com"
   },
   "repository": {
     "type": "git",
-    "url": "git://github.com/TeamPraxis/node-zaproxy.git"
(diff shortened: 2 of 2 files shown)


kingthorin:
Could you fixup into a single commit?

Should some change or addition be made to the author section.

bkimminich:
> Could you fixup into a single commit?
Hm, changing history after pushing is never a good idea... I can keep in mind that you prefer squashed commits for possible future PRs, though.

bkimminich:
Oh', I forgot to mention the reason why I touched the dependencies in the first place: https://snyk.io/vuln/npm:request:20160119

thc202:
> Hm, changing history after pushing is never a good idea...

IMHO that only applies to branches that are going to be pulled/used/based by 3rd-parties (e.g. zaproxy's develop branch), which is not likely for branches of pull requests. I don't have the same immutability expectations for pull request branches as for main ones (develop/master/...).

OK, I see that your pull request branch is a develop branch, which might cause problems to someone if changed. Using another branch would avoid those problems though. Still, I think worth doing the changes given that not many users are going to be affected(?).

khamasaki:
I added a note to the readme on the old version of the Node.js API pointing here. Thanks!

psiinon:
Shouldnt we be moving the code to https://github.com/zaproxy/zap-api-nodejs?
That doesnt have to be done as part of this PR or course...

thc202:
Yeah, we should do that.

kingthorin:
@bkimminich could you fix-up or squash this into a single "atomic" commit?

bkimminich:
@kingthorin Squashed it into a single one.

thc202:
Thanks!

PR #2922, opened 2016-10-10
Do not access the EDT in daemon mode

thc202:
Change class ExtensionSearch to not access the EDT (and view classes) if
the view is not initialised when the session changes by adding a "view"
SessionChangedListener when there's a View.
 ---

### DIFF
--- a/src/org/zaproxy/zap/extension/search/ExtensionSearch.java
+++ b/src/org/zaproxy/zap/extension/search/ExtensionSearch.java
@@ -39,7 +39,7 @@ import org.parosproxy.paros.model.Session;
 import org.zaproxy.zap.extension.help.ExtensionHelp;
 import org.zaproxy.zap.view.ZapMenuItem;
 
-public class ExtensionSearch extends ExtensionAdaptor implements SessionChangedListener {
+public class ExtensionSearch extends ExtensionAdaptor {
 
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good

src/org/zaproxy/zap/extension/search/ExtensionSearch.java
```diff
@@ -71,9 +71,9 @@ public ExtensionSearch() {
 	@Override
 	public void hook(ExtensionHook extensionHook) {
 	    super.hook(extensionHook);
-	    extensionHook.addSessionListener(this);
 	    extensionHook.addOptionsParamSet(getSearchParam());
 	    if (getView() != null) {
+	        extensionHook.addSessionListener(this);
```
psiinon: I'm wondering if this could be a little confusing - you have to look at the sessionChanged method to check that its only relevant to the UI and not the daemon. What if someone changes this to do something else? OK, we should then check that its initialized in daemon mode but thats easily missed. 
2 suggestions - either add suitable comments (above and in sessionChanged?) or change sessionChangedEventHandler to only call this.getSearchPanel().resetSearchResults(); if View.isInitialized()

thc202: Change done, it's now using a "view" SessionChangedListener.

kingthorin:
Agreed

PR #2924, opened 2016-10-11
Do not use null string literal in ApiResponseSet

thc202:
Change ApiResponseSet to not use null string literal, in XML and HTML
formats, when the values are null otherwise it might seem that the null
values of the set have a value (i.e. "null"), instead of nothing.
For example, an alert with no evidence/attack (null) would shown "null"
instead of nothing.


kingthorin:
Looks good

PR #2925, opened 2016-10-11
Differentiate the source of alerts

### LINKED ISSUE #2592: Differentiate the source of alerts
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/vskRxvLh_ZQ/discussion

Allow to differentiate the source of alerts, with the following values(?):
 - Active - raised by an active scanner;
 - Manual - raised manually (by the user);
 - Passive - raised by a passive scanner;
 - Tool - raised by other tools/functionalities in ZAP (for example, fuzzer, HTTPS Info add-on, custom scripts...);
 - Unknown - for old alerts which source is not (well) known.

The source should be accessible through the ZAP API as well the GUI.

thc202:
Change the Alert class to have a source "unknown", "active", "manual",
"passive" and "tool" which is set just before the alert is raised for
active, manual and passive alerts, for remaining alerts it's considered
as raised by a tool (e.g. custom scripts). Old alerts of existing
sessions will default to "unknown" since its exact source is not
immediately known.
Change Alert tab to shown the new field of the alert, "Source:".
Change CoreAPI to return the ID of the source of the alerts.
Change databases and corresponding tables to have the source.
Change HTML, XML and MD reports to include the ID of the source of the
alerts.

Fix #2592 - Differentiate the source of alerts


kingthorin:
Looks good

PR #2927, opened 2016-10-12
Include the base URL in SpiderTextParser

thc202:
Change SpiderTextParser to include the base URL when processing the URLs
found in the response so if there's an error in the processing of those
URLs it's included the page were the problematic URL was found. Without
the change it would be logged:
 Error while Processing URL in the spidering process (on base ): Host
 could not be reliably evaluated from: http://example.com)

which does not give any information where the problematic URL was found.
The inclusion of base URL does not affect how the URLs are resolved, the
processed URLs are already absolute.

Tweak the class URLCanonicalizer to include the base URL when debug
logging URLs with no authority, for same reason.
 ---


kingthorin:
Looks good

PR #2928, opened 2016-10-13
Added filterchain to parse out new line characters

kulinacs:
Many editors automatically append new lines to the end of files.
Without this change, a trailing new line in version.txt causes the
jar created to attempt to have a newline character in the name.


thc202:
Looks good to me.

PR #2929, opened 2016-10-13
Include the "faulty" URI in exception message

thc202:
Change class HttpMethodHelper to include the "faulty" URI in the
exception message, to give more information about the problem when being
handled by caller code.
 ---
From review comments in zaproxy/zap-extensions#586 and zaproxy/zap-extensions#587.


kingthorin:
Looks good

PR #2930, opened 2016-10-13
Added setup as a dependency for dist

### LINKED ISSUE #1921: Building ant target 'dist' generates incorrectly named .jar file.
When building target 'dist', the 'setup' target, which is responsible for setting the zap.jar property, is not run. The effect of this is that you end up with a .jar file named '${zap.jar}':

...
copy-source-to-build:
      [jar] Building jar: /usr/local/zaproxy-2.4.2/build/zap/${zap.jar}
...

Adding 'setup' to the depends list for the 'dist' list appears to fix this.

kulinacs:
The dist build target requires ${zap.jar} to be set. When it is
not, the build does not actually execute successfully. Adding
setup as a dependency for dist fixes this problem, and fixes #1921

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -129,7 +129,7 @@
 
 	</target>
 
-	<target name="dist" depends="compile" description="Generates the distribution">
+	<target name="dist" depends="setup, compile" description="Generates the distribution">
 		<echo message="Creating the distribution set..." />
 
 		<!-- Create the distribution directory -->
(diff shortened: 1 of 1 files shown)


kingthorin:
Looks good to me

thc202:
- There's a typo in the commit message (in "dependency");
 - The property "zap.jar" defined in "day-stamped-release" target could be removed.

The build does not always fail if the version was not set (at least here the build does not fail, although the jar does not have the best name).

kulinacs:
Although the dist target build does not fail during the build process without the setup target, attempting to run it with zap.sh causes a bad substitution error.

thc202:
Yes, just saying that the build does not fail.

Do you plan to fix the commit message or remove the property from "day-stamped-release"?

kulinacs:
Will fix the commit message, and thought of an alternative that would also fix problem, perhaps checking the zap.jar property is set in dist, and if not, run setup, that way the existing day-stamped-release property can be kept.

thc202:
I'm fine leaving as is (maybe it's better as it ensures that the name is the expected one, in case the logic changes in "setup"?).

Looks good to me, thank you!

kingthorin:
Merge as-is then?

thc202:
Looks good to me as is, now and before :)

PR #2936, opened 2016-10-17
Always set Java mem to 1/4 available (over 512Mb)

psiinon:
This means we have no upper limit to the level set by default. The user can still override if they want.


kingthorin:
Looks ok to me.

PR #2938, opened 2016-10-18
Allow to override the host header of an HTTP request

### LINKED ISSUE #1318: Allow to override the target host of an HTTP request
```
Zap appears to use the host header to work out which IP to send the request to, so altering
it triggers an error:

For example, trying to send the following request to 192.168.1.73 via org.parosproxy.paros.network.HttpMessage
and paros.network.HttpRequestHeader:

GET /joomla/?foo=bar&cachebust=1405888831.78 HTTP/1.1
Host: ndz27s.192.168.1.73
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:30.0) Gecko/20100101 Firefox/30.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Cookie: tz_offset=3600; acopendivids=swingset,jotto,phpbb2,redmine; acgroupswithpersist=nada;
d5a4bd280a324d2ac98eb2c0fe58b9e0=ahufhedqrq6atgh9lp81ogg3p6
Referer: http://192.168.1.73/
DNT: 1
Connection: keep-alive
Cache-Control: no-cache

triggers the following error:

javax.script.ScriptException: java.net.UnknownHostException:  java.net.UnknownHostException:
ndz27s.192.168.1.73 in <script> at line number 58

The relevant Jython code looks roughly like:
headers = org.parosproxy.paros.network.HttpRequestHeader(headers_shown_above)
attack = org.parosproxy.paros.network.HttpMessage(headers, body)
as_helper.sendAndReceive(attack, True, True)


From my perspective I think a setRemoteAddr() method on org.parosproxy.paros.network.HttpMessage
would be a solid solution.
```

Original issue reported on code.google.com by `albinowax` on 2014-08-18 18:08:39

jx6f:
I found that the Host header is overwritten by `org.apache.commons.httpclient.HttpMethodBase#addHostRequestHeader`(in commons-httpclient-3.1.jar).
http://svn.apache.org/viewvc/httpcomponents/oac.hc3x/trunk/src/java/org/apache/commons/httpclient/HttpMethodBase.java?view=markup#l1361

To prevent that behavior, override it by each extended classes.

### DIFF
--- a/src/org/apache/commons/httpclient/HttpMethodBase.java
+++ b/src/org/apache/commons/httpclient/HttpMethodBase.java
@@ -65,6 +65,7 @@ import org.parosproxy.paros.network.HttpHeader;
  *  String, String, String) to preserve the intended request URI.
  *  - Change the way cookie headers are handled when using forced user mode, put all the headers in a single line see ISSUE 1874
  *  - Do not add a User-Agent header by default.
+ *  - Removed the #addHostRequestHeader method call from #addRequestHeaders to prevent the overwriting of host header.
  * 
  */
 /**
@@ -1551,7 +1552,6 @@ public abstract class HttpMethodBase implements HttpMethod {
         LOG.trace("enter HttpMethodBase.addRequestHeaders(HttpState, "
             + "HttpConnection)");
 
-        addHostRequestHeader(state, conn);
         addCookieRequestHeader(state, conn);
         addProxyConnectionHeader(state, conn);
     }


jx6f:
I missed that the `org.apache.commons.httpclient.HttpMethodBase` class has already existing in zaproxy.
https://github.com/zaproxy/zaproxy/blob/develop/src/org/apache/commons/httpclient/HttpMethodBase.java

Filling with NOP `org.apache.commons.httpclient.HttpMethodBase#addHostRequestHeader` may be better than my way.

thc202:
Probably better to just remove the statement from the method `addRequestHeaders` (as was done with the `User-Agent` header), also could you comment the change in the "Forked class..." comment?

Could you change the pull request title/commit message to not close the issue 1318? This change is not enough to close the issue, attempting to send a message with a relative request-uri will still fail (the reason why this change was not done before, still useful nonetheless).

jx6f:
Thank you for nice advices. My changes has done.

kingthorin:
Please do an interactive rebase and fix-up or squash into a single "atomic" commit: https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#what-we-zap-team-expect-from-you .

jx6f:
Oops. Thank you.

kingthorin:
Looks good

thc202:
Could you add a short note about the change in the "Forked class" comment?

May I suggest the following for pull request title / commit message: "Allow to override the host header of an HTTP request", it reflects better the change done.

jx6f:
I misunderstood. I appreciate your kindness.

thc202:
Changes look good to me (commit message could be tweaked but not a blocker IMO), thank you!

Tweaked the pull request title to not be confused with issue 1318, that's included in the release notes.

thc202:
Actually we might need other changes to keep the expected (and correct) behaviour, for example, when following redirections the Host header should be the one of the new location not the existing one. The change will also break clients relying on automatic update of the request header, for example, when same HttpMessage/HttpRequestHeader is reused to send requests to several hosts.

jx6f:
I confirmed that this PR takes terribly bad behavior when manually sended request gets a response which involves redirection.
I'm so sorry but may I close this PR? I think this PR must be rejected.

psiinon:
OK, no problem :)

jx6f:
@psiinon @thc202 Thank you so much!

PR #2942, opened 2016-10-19
Move HTML parser's test files into its own dir

thc202:
Move the files used by SpiderHtmlParserUnitTest to its own directory to
be more clear what the files are used for and by what parser.
Update SpiderHtmlParserUnitTest to use the new directory.


kingthorin:
Looks good

PR #2943, opened 2016-10-19
Reuse test files of HTML form spider parser

thc202:
Change the files to have the method as a variable so that it can be used
for both GET and POST forms.
Update SpiderHtmlFormParserUnitTest to reflect the changes.


kingthorin:
Looks good

PR #2944, opened 2016-10-19
Added ZAP_HOME variable to launch script

kulinacs:
Utilizing the ZAP_HOME variable if set will allow other Linux
distros to easily package ZAP according to their packaging
guidelines, as it will allow zap.sh to be placed somewhere away
from the Java executable.


kulinacs:
After further thought, I am of the opinion encourage arbitrary configuration change is less elegant than a symlink, and am closing this request as such.

thc202:
OK.

PR #2946, opened 2016-10-20
Set a name to spider threads

thc202:
Initialise the spider threads with custom name as it makes it easier to
identify that are threads created by ZAP, know it's purpose and to know
to which spider scan they belong. Also, correct the site/name show when
starting the spider.

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/SpiderThread.java
+++ b/src/org/zaproxy/zap/extension/spider/SpiderThread.java
@@ -108,21 +108,43 @@ public class SpiderThread extends ScanThread implements SpiderListener {
 
 	private List<ParseFilter> customParseFilters = null;
 
+	private final String id;
+
 	/**
-	 * Instantiates a new spider thread.
+	 * Constructs a {@code SpiderThread} with the given data.
 	 * 
-	 * @param extension the extension
-	 * @param site the site
+	 * @param extension the extension to obtain configurations and notify the view
+	 * @param spiderParams the spider options
+	 * @param site the name that identifies the target site
 	 * @param listenner the scan listener
+	 * @deprecated (TODO add version) Use {@link #SpiderThread(String, ExtensionSpider, SpiderParam, String, ScanListenner)}
 	 */
+	@Deprecated
 	public SpiderThread(ExtensionSpider extension, SpiderParam spiderParams, String site, ScanListenner listenner) {
+		this("?", extension, spiderParams, site, listenner);
+	}
+
+	/**
+	 * Constructs a {@code SpiderThread} with the given data.
+	 * 
+	 * @param id the ID of the spider, usually a unique integer
+	 * @param extension the extension to obtain configurations and notify the view
+	 * @param spiderParams the spider options
+	 * @param site the name that identifies the target site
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/spider/SpiderThread.java
```diff
@@ -108,21 +108,43 @@
...
-	 * @param extension the extension
-	 * @param site the site
+	 * @param extension the extension to obtain configurations and notify the view
+	 * @param spiderParams the spider options
+	 * @param site the name that identifies the target site
 	 * @param listenner the scan listener
+	 * @deprecated (TODO add version) Use {@link #SpiderThread(String, ExtensionSpider, SpiderParam, String, ScanListenner)}
 	 */
+	@Deprecated
 	public SpiderThread(ExtensionSpider extension, SpiderParam spiderParams, String site, ScanListenner listenner) {
```
kingthorin: Perhaps this is a change for another PR, but what's with `listenner` having two Ns?

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -141,11 +145,31 @@
...
+	@Deprecated
 	public Spider(ExtensionSpider extension, SpiderParam spiderParam, ConnectionParam connectionParam,
 			Model model, Context scanContext) {
+		this("?", extension, spiderParam, connectionParam, model, scanContext);
+	}
+
+	/**
+	 * Constructs a {@code Spider} with the given data.
+	 * 
+	 * @param id the ID of the spider, usually a unique integer
```
kingthorin: As above, does this need to be more clear that it's a string representation not an integer

src/org/zaproxy/zap/extension/spider/SpiderThread.java
thc202: Most likely a typo, not much that can be done to fix it (the class name, we also have ScanListenner2).

src/org/zaproxy/zap/extension/spider/SpiderThread.java
```diff
@@ -108,21 +108,43 @@
...
+		this("?", extension, spiderParams, site, listenner);
+	}
+
+	/**
+	 * Constructs a {@code SpiderThread} with the given data.
+	 * 
+	 * @param id the ID of the spider, usually a unique integer
```
thc202: I don't think it needs any remark, the type of the parameter already indicates that (moreover the compiler will inform if used int/Integer instead of String).

src/org/zaproxy/zap/spider/Spider.java
thc202: Ditto.

src/org/zaproxy/zap/extension/spider/SpiderThread.java
kingthorin: Ok

src/org/zaproxy/zap/extension/spider/SpiderThread.java
kingthorin: Fair enough

kingthorin:
Looks good

PR #2948, opened 2016-10-21
Create first Root CA certificate synchronously

thc202:
Change class ExtensionDynSSL to create the Root CA certificate
synchronously to have the certificate ready for when the Local Proxy is
started, otherwise it could fail to process immediate SSL/TLS requests.


kingthorin:
Looks good

PR #2953, opened 2016-10-24
Delay addition of the context being imported

thc202:
Changed Session to only add the context being imported if no errors
occurred while importing it, otherwise the context could be left in a
potentially inconsistent state which could cause issues in other parts
of the code (for example, if it had no name (i.e. null) it would no
longer be possible to add a new context, delete the one imported or
create a new session).


kingthorin:
Looks good

PR #2954, opened 2016-10-24
Allow to export a Context through the context menu

thc202:
Add a pop up menu item to the context menu of the contexts tree to allow
to export the selected context.


kingthorin:
Looks good

Issue #2935, opened 2016-10-16
Wrong charset used in response body if no charset set


avinas-h:
Steps to reproduce the bug.
Download the latest weekly release of ZAP (I used the 2016-10-10 one)
1- In a web browser (which is configured to the latest ZAP), visit http://genius.com/login and notice how the web page is loaded (the web page will be loaded correctly)

2- Close the browser, load any Zest proxy rule to the latest ZAP, open the web browser and visit http://genius.com/login

3- You will noticed that the web page is loaded incorrectly, in particular the styling of the page gets messed up. This can severely affect Selenium tests utilizing ZAP and Zest proxy rules.

I noticed that the issue is also there in ZAP 2.5. But I didn't find the issue in June 2016 Weekly release of ZAP.

thc202:
Is/was the browser trusting ZAP's certs? That page has styles from <https://assets.genius.com/> (note the HTTPS scheme).

avinas-h:
Yes. That is the reason why the page was loading properly when it was intercepting via ZAP without any proxy rules. Btw, the following is the configuration of my machine.
![image](https://cloud.githubusercontent.com/assets/13018530/19478928/acf803d6-9544-11e6-9d1f-d65956c73360.png)

Were you able to reproduce the issue with the latest version of ZAP and the 2.5 stable version?

avinas-h:
Update: I was also able to reproduce the issue also in a Windows 10 machine with yesterday's weekly release. Please note that you must clear the browsing history when you try to load the web page with the Zest proxy rule. Otherwise, you will not be able to see the issue because the browser uses caches stylesheets from the previous session in which the Zest proxy rule was not activated.

thc202:
I was able to reproduce the issue, it's related to the character encoding of the (CSS) response. Thanks!

avinas-h:
Great! Can't wait to have the fixed weekly release.

avinas-h:
I thought the issue was resolved when you marked it as closed. But I downloaded yesterday's weekly release (D-2016-10-24) and the issue is still there. You should be able to reproduce the issue in the same way as I explained earlier.
The following screenshot shows how the web page loading correctly when the ZAP rule is disabled.
![image](https://cloud.githubusercontent.com/assets/13018530/19684658/c3d3da5c-9ab8-11e6-8abf-9c21ef4f74ae.png)

[...]

PR #2956, opened 2016-10-25
Correct charset determination in HttpResponseBody

thc202:
Remove use of platform's default charset when determining if the charset
of the string is UTF-8, which was leading to wrong results if the
platform's default charset was not UTF-8.

Related to:
 - #2935 - Wrong charset used in response body if no charset set
 - #2941 - Attempt to determine (String) body's charset


kingthorin:
Looks good

PR #2957, opened 2016-10-26
Do not allow Contexts with same name

### LINKED ISSUE #1952: Do not allow Contexts with same name
Steps to reproduce the issue:
1. Run ZAP (2.4.2);
2. Click the "New Context..." button; 
3. Enter "Default Context" and save;
4. Note that there are now two contexts "Default Context".

Remarks:
Since context's name is used to uniquely identify/select a context throughout ZAP, it should not be possible to create contexts with the same name.

ZAP Version:
Versions >= 2.0.0

thc202:
Change GUI/API to not allow to:
- Import or create a context with no name, with an empty name or with
  name that already exists;
- Change the name of the context to be null, empty or that duplicates
  an existing name.

Fix #1952 - Do not allow Contexts with same name


thc202:
Changes done to the resource messages.

kingthorin:
Looks good d

PR #2959, opened 2016-10-26
Show correct header when selected panel is removed

thc202:
Change AbstractParamContainerPanel to (explicitly) show the first
available panel when the selected panel is removed, to show the correct
information in the panel header, title and help button. Also, do not
show the panel if already shown (skip notifications that the same panel
is hidden and then shown, header setup and re-setting the panel in the
layout).


kingthorin:
Looks good

PR #2961, opened 2016-10-27
Export context's session management data

thc202:
Change ExtensionSessionManagement to also export session management data
when exporting the context (not a problem for core implementations which
do not have any data).


kingthorin:
Looks good

PR #2962, opened 2016-10-28
Remove WAVSEP spider tests

thc202:
The WAVSEP spider tests are no longer maintained, also the (HTML) spider
parsers have now good unit test coverage and the spider is regularly
tested with WIVET (through zapbot scans).


kingthorin:
Looks good

Issue #2964, opened 2016-10-31
Allow to select the look and feel


vdun:
I'm trying to use the GTK look and feel :
```
# export _JAVA_OPTIONS='-Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel'
# zaproxy
Found Java version 1.8.0_112
Available memory: 7970 MB
Setting jvm heap size: -Xmx512m
Picked up _JAVA_OPTIONS: -Dswing.defaultlaf=com.sun.java.swing.plaf.gtk.GTKLookAndFeel
0 [main] INFO org.zaproxy.zap.GuiBootstrap  - OWASP ZAP 2.5.0 started.
...
```
but still getting the old look and feel!
The trick works on other Jar based applications except ZAP.

thc202:
The L&F used by ZAP is currently "hardcoded".

psiinon:
The relevant code is here: https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/GuiBootstrap.java#L325,L352
I'd have no problem for this to be changeable via a config option, as long as the default stays the same.

nhegde610:
Hi, 
I want to take up this issue. I have looked through the code and pretty much understood how to fix this issue. However, I don't know which files should I edit to create a config option in the options window. If possible, I would like to get some answers regarding this.

thc202:
I guess this would go under Options > Display, if so, the classes that need to be changed are [`org.parosproxy.paros.extension.option.OptionsViewPanel`](https://github.com/zaproxy/zaproxy/blob/ca3ff4cfb7d34e91ab31ccccfa7ffb03883218df/src/org/parosproxy/paros/extension/option/OptionsViewPanel.java) (this one creates the Display panel) and  `OptionsParamView` (reads and persists the corresponding options).

nhegde610:
Thank you @thc202. I have managed to create a config options in Display Options panel. While I was testing my solution, I noticed a UI bug. When I switch the look to native system look, the option for changing look  disappears. I have to resize the options window to be able to view it. This issue also arises if the user changes the font size and font name. Here are pics for better understanding.
Before resizing :
![before_resizing](https://user-images.githubusercontent.com/18752973/36634438-a8fca258-19ca-11e8-96e8-bc3b8f7f2e73.png)

After resizing: 
[...]

PR #2966, opened 2016-11-02
Use L&F specified through JVM args

thc202:
Change GuiBootstrap to use the look and feel specified through the JVM
arguments if able to find/set it, otherwise fallback to previous/current
behaviour.

Related to #2964 - Allow to select the look and feel


kingthorin:
Looks good to me

PR #2968, opened 2016-11-03
Increase page size when accessing alerts

psiinon:
It turns out that the alert API paging is not implemented very efficiently, and
choosing too small a page size can take a very long time.


thc202:
Looks good to me.

PR #2971, opened 2016-11-03
Do not initialise dev logger if there's no view

thc202:
Change ExtensionLog4j to not initialise the "logger" if there's no view,
it was only used if the view was initialised.
Change ZapOutputWriter to require the view initialised and that the scan
status label is provided (and remove view and null checks when logging,
no longer needed per previous changes). Also, remove unused constructor.


kingthorin:
Looks good

PR #2972, opened 2016-11-03
Init status label in attack scanner only with view

thc202:
Change AttackModeScanner to not initialise the scan status label if
there's no view, it's not needed in daemon mode. Also, change to use
long to track elapsed time, instead of Date, to not create the Date
objects unnecessarily.


kingthorin:
Looks good

PR #2974, opened 2016-11-04
Modifications to Enableable

kingthorin:
Within org.zaproxy.zap.utils:
* Add interface EnableableInterface (Extracted from Enableable).
* Enableable now implements EnableableInterface.

Within org.zaproxy.zap.view:
* AbstractMultipleOptionsTableModel now leverages EnableableInterface.
* AbstractMultipleOptionsTablePanel now leverages EnableableInterface.

### DIFF
--- a/src/org/zaproxy/zap/utils/EnableableInterface.java
+++ b/src/org/zaproxy/zap/utils/EnableableInterface.java
@@ -1,0 +1,29 @@
+
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2016 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/utils/EnableableInterface.java
```diff
@@ -0,0 +1,13 @@
+package org.zaproxy.zap.utils;
```
thc202: Should be added the license header.

src/org/zaproxy/zap/utils/EnableableInterface.java
```diff
@@ -0,0 +1,13 @@
+package org.zaproxy.zap.utils;
+
+public interface EnableableInterface {
+
+	boolean isEnabled();
+
+	void setEnabled(boolean enabled);
+
+	int hashCode();
```
thc202: This method (and equals) could be removed from the interface, not really needed.

src/org/zaproxy/zap/utils/EnableableInterface.java
kingthorin: Done

thc202:
Looks good to me, thank you!

PR #2975, opened 2016-11-04
Include date/time when logging that ZAP started

thc202:
The date/time allows to correlate the output logging with other logs
and events more easily.

 ---

From talks in IRC.


kingthorin:
Looks good

Issue #2970, opened 2016-11-03
Allow to configure, by script type, the enabled state of new/loaded scripts


thc202:
The change allows to set the scripts enabled by default, for example, fuzz scripts need to be enabled to be used in the Fuzzer dialogue (the scripts get automatically disabled upon errors) although given that they must be explicitly added to be used they can be already enabled without problems, which is one less step/burden to use them.

thc202:
Closing, required core changes are done, will raise new issues for changes in the add-ons.

PR #2976, opened 2016-11-04
Change ScriptType to define if enabled by default

thc202:
Change ScriptType to allow to define if the scripts of the script type
should be enabled by default (e.g. when added/loaded via GUI).

Related to #2970 - Allow to configure, by script type, the enabled state
of new/loaded scripts

### DIFF
--- a/src/org/zaproxy/zap/extension/script/ScriptType.java
+++ b/src/org/zaproxy/zap/extension/script/ScriptType.java
@@ -13,22 +13,80 @@ public class ScriptType {
 	 */
 	public static final String CAPABILITY_APPEND = "append";
 	
-	private String name;
-	private String i18nKey;
-	private ImageIcon icon;
-	private boolean isEnableable;
-	private Set<String> capabilitiesSet = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/script/ScriptType.java
```diff
@@ -13,22 +13,85 @@
...
+	 * @since TODO add version
+	 */
+	public ScriptType(
+			String name,
+			String i18nKey,
+			ImageIcon icon,
+			boolean enableable,
+			boolean enabledByDefault,
+			String[] capabilities) {
 		super();
```
kingthorin: Something seems to be up with the indentation here

thc202: That's on purpose, using an auto-formatter which likes to format that way (to improve readability). Will tweak to just two lines.

thc202: Done.

kingthorin:
Looks good, thanks

PR #2979, opened 2016-11-07
Delay init of attack mode scanner to prevent NPE

thc202:
Change ExtensionActiveScan to delay the initialisation of
AttackModeScanner to allow it to properly check if the view is
initialised, using the extension.

Caused by #2972 - Init status label in attack scanner only with view


psiinon:
Merging with just one review to fix the weekly release...

PR #2981, opened 2016-11-07
Change attack mode thread to daemon

thc202:
Change the thread used for the attack mode to be a daemon thread, to not
prevent ZAP from terminating normally. For example, if the attack mode
was enabled while starting ZAP (in daemon mode) and ZAP was not able to
bind to the address/port it would be kept running instead of
terminating.


kingthorin:
Looks good

PR #3002, opened 2016-11-14
Correctly render all nodes in checkbox tree

thc202:
Change JCheckBoxTree to correctly render the top level nodes, the
renderer will not show the checkbox if the node has no checkbox state
moreover set the node's text to the label wherever it has or not a
checkbox. Also, change to create the checkbox state of the tree nodes
before the model is set to the base class (as it might be used by base
class for painting calculations, using the custom renderer).
Update test to reflect the change in behaviour (no longer throws a
NullPointerException when setting a null model).


kingthorin:
There is a trailing n missing on NPE in the commit message

thc202:
Fixed, thank you.

PR #3007, opened 2016-11-17
Support Factory Reset

### LINKED ISSUE #2701: Enhancement Request: Factory Reset
People might get in situations where ZAP isn't working/behaving as expected, or they feel the screwed up the configuration and broke it (whether they actually broke it or not).

One of the common remediation ideas is to do a factory reset to make sure the configuration isn't screwed up. Seems like in the Options panel we could add a line as a peer of Options (but after it so its at the VERY end of the list) called Factory Reset - and this could be used to reset either EVERYTHING in ZAP, or at least ALL the options you can configure in the Options panel.

Not sure if this reset should throw away downloaded plugins, change your session properties, or anything else that can change. But it should at least reset all the Options. And it should be clear about what it does and does not reset.

psiinon:
Fixes #2701

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -1537,6 +1537,11 @@ options.cert.title                    = Certificate
 options.cert.title.cert               = Certificate
 options.dialog.title = Options
 options.dialog.rootName = Options
+options.dialog.reset.button = Reset to Factory Defaults 
+options.dialog.reset.warn = Are you sure you want to reset all of the options?\n\
+This takes effect immediately and cannot be undone!
+options.dialog.reset.error = Failed to reset the options :\n {0}
+options.dialog.save.error = Failed to save the options :\n {0}
 options.ext.button.openurl            = Open Homepage in Browser
 options.ext.label.author              = Author
 options.ext.label.core                = Core
--- a/src/org/parosproxy/paros/Constant.java
+++ b/src/org/parosproxy/paros/Constant.java
@@ -67,6 +67,7 @@
 // ZAP: 2016/06/13 Migrate config option "proxy.modifyAcceptEncoding" 
 // ZAP: 2016/07/07 Convert passive scanners options to new structure
 // ZAP: 2016/09/22 JavaDoc tweaks
+// ZAP: 2016/11/17 Issue 2701 Support Factory Reset
 
 package org.parosproxy.paros;
 
@@ -295,6 +296,8 @@ public final class Constant {
 	public static final URL BLANK_IMAGE_URL = Constant.class.getResource("/resource/icon/10/blank.png");
 	public static final URL SPIDER_IMAGE_URL = Constant.class.getResource("/resource/icon/10/spider.png");
 
+    private static Logger LOG = Logger.getLogger(Constant.class);
+
     public static String getEyeCatcher() {
         return staticEyeCatcher;
     }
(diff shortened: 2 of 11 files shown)


kingthorin:
Build fails

psiinon:
Ha! Looks like an easy one to fix...

psiinon:
Builds ok now

src/lang/Messages.properties
```diff
@@ -1537,6 +1537,11 @@ options.cert.title                    = Certificate
 options.cert.title.cert               = Certificate
 options.dialog.title = Options
 options.dialog.rootName = Options
+options.dialog.reset.button = Reset to Factory Defaults 
+options.dialog.reset.warn = Are you sure you want to reset all of the options?\n\
+This takes effect immediatelly and cannot be undone!
+options.dialog.reset.error = Failed to reset the options : {0}
```
thc202: Should have a new line before the parameter?

src/lang/Messages.properties
```diff
@@ -1537,6 +1537,11 @@ options.cert.title                    = Certificate
...
+options.dialog.save.error = Failed to save the options : {0}
```
thc202: Should have a new line before the parameter?

src/org/parosproxy/paros/Constant.java
```diff
@@ -346,12 +349,39 @@ public static String getDefaultHomeDirectory(boolean incDevOption) {
         return zapStd;
     		
     }
+    
+    public void copyDefaultConfigs(File f, boolean forceReset) throws IOException, ConfigurationException {
```
thc202: The argument `forceReset` does not seem to be used?

src/org/parosproxy/paros/view/AbstractParamDialog.java
```diff
@@ -194,10 +189,8 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
                         AbstractParamDialog.this.setVisible(false);
 
                     } catch (Exception ex) {
-                    	// The exception messages should be internationalized!
-                        View.getSingleton().showWarningDialog(ex.getMessage());
+                        Constant.messages.getString("options.dialog.save.error", ex.getMessage());
```
thc202: The message is no longer being shown.

src/org/parosproxy/paros/model/OptionsParam.java
```diff
@@ -211,11 +212,12 @@ protected void parse() {
...
 			}
 		}
-		
-//		for (int i=0; i<paramSetList.size(); i++) {
-//		    AbstractParam param = (AbstractParam) paramSetList.get(i);
-//		    param.load(getConfig());
-//		}
+    }
+    
+    public void parseParamSet() {
```
thc202: Could be `reloadConfigParamSets` instead?

src/lang/Messages.properties
```diff
@@ -1537,6 +1537,11 @@ options.cert.title                    = Certificate
 options.cert.title.cert               = Certificate
 options.dialog.title = Options
 options.dialog.rootName = Options
+options.dialog.reset.button = Reset to Factory Defaults 
+options.dialog.reset.warn = Are you sure you want to reset all of the options?\n\
+This takes effect immediatelly and cannot be undone!
```
thc202: There's a typo in immediately?

psiinon:
Hopefully fixed all of those

src/org/zaproxy/zap/spider/parser/SpiderSVNEntriesParser.java
```diff
@@ -355,4 +355,8 @@ public boolean canParseResource(HttpMessage message, String path, boolean wasAlr
 		Matcher matcher = SVN_ENTRIES_FILE_PATTERN.matcher(path);
 		return matcher.find();
 	}
+	
+	public static void main(String[] params) {
```
kingthorin: Oops?

kingthorin:
Uh now you have even more unrelated content included 😢 

I suggest feature branches 😄 Instead of working on develop and pushing directly to it.

psiinon:
Ugh :/
Yeah, I usually do feature branches, but messed up with this one :P
No idea why the tests are failing, that one works ok for me...

psiinon:
Think I've fixed the files but the build is still failing :(

psiinon:
So I think I've fixed the comments, and I dont think the test failure is related to my change - it works fine here :/

thc202:
Thanks!

thc202:
For the record all tests pass locally, with Java 7 and Java 8.

kingthorin:
So we believe the failure is some sort of false positive or one-off condition?

I don't mind merging as long as we're in agreement. Aside from the test failure it looks good to me.

kingthorin:
https://travis-ci.org/zaproxy/zaproxy/requests

Seems to show things passing...

thc202:
The build failure seems to caused by changes in the Constant class (per #3011).

src/org/parosproxy/paros/Constant.java
```diff
@@ -346,12 +349,39 @@ public static String getDefaultHomeDirectory(boolean incDevOption) {
...
+        File oldf;
+        if (isDevBuild() || isDailyBuild()) {
+            // try standard location
+            oldf = new File (getDefaultHomeDirectory(false) + FILE_SEPARATOR + FILE_CONFIG_NAME);
+        } else {
+            // try old location
+            oldf = new File (zapHome + FILE_SEPARATOR + "zap" + FILE_SEPARATOR + FILE_CONFIG_NAME);
+        }
+        
+        if (!forceReset || oldf.exists() && Paths.get(zapHome).equals(Paths.get(getDefaultHomeDirectory(true)))) {
```
thc202: This condition is not correct it should be and not or (as the old file needs to exist anyway).
The test was failing because of this, was able to reproduce locally when the default home directory didn't exist (as is the case of Travis CI builds).

Issue #2525, opened 2016-06-02
Form Handling


psiinon:
The ZAP traditional and Ajax spiders explore an application by putting basic default values in all forms. These may often not be valid values, for example using "ZAP" when an email address is required.

The enhancement would allow the user to define default values based on pattern matching against the field names and/or ids.

It would also be very useful if it could show the user all forms and their associated fields for an application, and then allow the user to update the default values.

Expected Results
- User able to specify default values for all forms used by the ZAP spiders
- Display all of the forms and fields for an application and allow the user to update the default values to be used
- Full support for defining default values via the API

Note that this enhancement has been suggested for [MWoS 2016](https://wiki.mozilla.org/Security/Automation/Winter_Of_Security_2016#ZAP:_Form_Handling)

kingthorin:
Should this also include expansion of ZAP's default entry set, and attempts to guess or intuit which might be applicable? i.e.: If the parameter name matches some dictionary of "email" address identifiers use zap@example.org, etc.

thc202:
Project page:
https://wiki.mozilla.org/Security/Automation/Winter_Of_Security_2016/ZAP_Form_Handling

rwehe:
Thanks for adding me @psiinon! 
Can we also assign to @cdlaguna?

psiinon:
Sure, I just wanted it assigned to someone before I tweeted about the 'projects' label ;)
I'll invite @cdlaguna then assign it to them :)

thc202:
Assigned :)

thc202:
A new add-on, Form Handler, is now available in the marketplace, it allows to define custom values for the form fields (requires weekly release >= 2017-03-06).

thc202:
Online help at:
https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsFormhandlerFormHandlerHelp

psiinon:
Many thanks guys!

PR #3028, opened 2016-11-23
Value Generator (previously Form Handling)

rwehe:
Core changes to prepare for the extension as detailed in #2525.
Implemented new Value Generator interface and refactored the way forms are handled by default.

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
+++ b/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
@@ -56,6 +56,8 @@ import org.zaproxy.zap.model.ScanController;
 import org.zaproxy.zap.model.StructuralNode;
 import org.zaproxy.zap.model.StructuralSiteNode;
 import org.zaproxy.zap.model.Target;
+import org.zaproxy.zap.model.ValueGenerator;
+import org.zaproxy.zap.model.DefaultValueGenerator;
 import org.zaproxy.zap.spider.SpiderParam;
 import org.zaproxy.zap.spider.filters.FetchFilter;
 import org.zaproxy.zap.spider.filters.ParseFilter;
@@ -68,6 +70,8 @@ import org.zaproxy.zap.view.ZapMenuItem;
  * The ExtensionSpider is the Extension that controls the Spider.
  */
 public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedListener, ScanController<SpiderScan> {
+
+	private ValueGenerator generator = new DefaultValueGenerator();
 
 	public static final int EXTENSION_ORDER = 30;
 	
@@ -125,6 +129,17 @@ public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedL
 		this.customFetchFilters = new LinkedList<>();
 		this.customParseFilters = new LinkedList<>();
 		this.scanController = new SpiderScanController(this);
+	}
+
+	public void setValueGenerator (ValueGenerator generator) {
+		if (generator == null){
+			throw new IllegalArgumentException("Parameter generator must not be null.");
+		}
+		this.generator = generator;
+	}
+
+	public ValueGenerator getValueGenerator() {
+		return generator;
 	}
 
 	@Override
--- a/src/org/zaproxy/zap/model/DefaultValueGenerator.java
+++ b/src/org/zaproxy/zap/model/DefaultValueGenerator.java
@@ -1,0 +1,152 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.model;
+
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.httpclient.URI;
+
+import net.htmlparser.jericho.FormControlType;
+
+/**
+ *   Default implementation of the ValueGenerator
+ *
+ *   @since 2.6.0
+ */
+
+public class DefaultValueGenerator implements ValueGenerator{
+
+	private static final String ATTR_TYPE = "type";
+	private static final String DEFAULT_NUMBER_VALUE = "1";
+	private static final String DEFAULT_TEXT_VALUE = org.parosproxy.paros.Constant.PROGRAM_NAME_SHORT;
+	private static final String DEFAULT_PASS_VALUE = DEFAULT_TEXT_VALUE;
+	private static final String DEFAULT_FILE_VALUE = "test_file.txt";
+	private static final String DEFAULT_EMPTY_VALUE = "";
+
+	private Date defaultDate;
+
+	/**
+	 * Gets the default {@code Date}, to be used for default values of date fields.
+	 *
+	 * @return the date, never {@code null}.
+	 * @see #setDefaultDate(Date)
+	 */
+	public Date getDefaultDate() {
+		if (defaultDate == null) {
+			return new Date();
+		}
+		return defaultDate;
+	}
+
+	public void setDefaultDate(Date date) {
+		this.defaultDate = date;
+	}
+
+	/**
+	 * Generates accurate field values for following types:
+	 * <ul>
+	 * <li>Text/Password/Search - DEFAULT_TEXT_VALUE</li>
+	 * <li>number/range - if min is defined, then use min. If max is defined use max,
+	 * otherwise DEFAULT_NUMBER_VALUE</li>
+	 * <li>url - http://www.example.com</li>
+	 * <li>email - foo-bar@example.com</li>
+	 * <li>color - #ffffff</li>
+	 * <li>tel - 9999999999</li>
+	 * <li>date/datetime/time/month/week/datetime-local - current date in the proper format</li>
+	 * <li>file - DEFAULT_FILE_VALUE</li>
+	 * </ul>
+	 * @return the default String value for each control type
+	 */
+
+	@Override
+	public String getValue(URI uri, String url, String fieldId, String defaultValue, List<String> definedValues, Map<String, String> envAttributes, Map<String, String> fieldAttributes) {
+
+		//If there is a default value provided, return it
+		if (!defaultValue.isEmpty()){
+			return defaultValue;
+		}
+
+		if (fieldAttributes.get("Control Type").equalsIgnoreCase(FormControlType.TEXT.toString())) {
+			// Converted FormControlType to String to allow for case insensitive comparison
+			// If the control type was reduced to a TEXT type by the Jericho library, check the
+			// HTML5 type and use proper values
+			String type = fieldAttributes.get(ATTR_TYPE);
+			if (type == null || type.equalsIgnoreCase("text")) {
+				return DEFAULT_TEXT_VALUE;
+			}
+			if (type.equalsIgnoreCase("number") || type.equalsIgnoreCase("range")) {
+				String min = fieldAttributes.get("min");
+				if (min != null) {
+					return min;
+				}
+				String max = fieldAttributes.get("max");
+				if (max != null) {
+					return max;
+				}
+				return DEFAULT_NUMBER_VALUE;
+			}
+			if (type.equalsIgnoreCase("url")) {
+				return "http://www.example.com";
+			}
+			if (type.equalsIgnoreCase("email")) {
+				return "foo-bar@example.com";
+			}
+			if (type.equalsIgnoreCase("color")) {
+				return "#ffffff";
+			}
+			if (type.equalsIgnoreCase("tel")) {
+				return "9999999999";
+			}
+			if (type.equalsIgnoreCase("datetime")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("datetime-local")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("date")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("time")) {
+				SimpleDateFormat format = new SimpleDateFormat("HH:mm:ss");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("month")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("week")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-'W'ww");
+				return format.format(getDefaultDate());
+			}
+		} else if (fieldAttributes.get("Control Type").equalsIgnoreCase(FormControlType.PASSWORD.toString())) {
+			return DEFAULT_PASS_VALUE;
+		} else if (fieldAttributes.get("Control Type").equalsIgnoreCase(FormControlType.FILE.toString())) {
+			return DEFAULT_FILE_VALUE;
+		}
+		return DEFAULT_EMPTY_VALUE;
+	}
+}
--- a/src/org/zaproxy/zap/model/ValueGenerator.java
+++ b/src/org/zaproxy/zap/model/ValueGenerator.java
@@ -1,0 +1,44 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2017 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.model;
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.httpclient.URI;
+
+public interface ValueGenerator {
+
+	/**
+	 * The interface that accepts default values for the spider.
+	 *
+	 * @param uri the uri
+	 * @param url the resolved URL
+	 * @param fieldId the name associated with the current field
+	 * @param defaultValue the value of 'value attribute' if it has one
+	 * @param definedValues the predefined values for the field, if present
+	 * @param envAttributes all attributes of the current form
+	 * @param fieldAttributes all attributes of the current field
+	 *
+	 * @since 2.6.0
+	 */
+
+	String getValue(URI uri, String url, String fieldId, String defaultValue, List<String> definedValues, Map<String, String> envAttributes, Map<String,String> fieldAttributes);
+}
--- a/src/org/zaproxy/zap/spider/Spider.java
+++ b/src/org/zaproxy/zap/spider/Spider.java
@@ -173,12 +173,12 @@ public class Spider {
 		this.spiderParam = spiderParam;
 		this.connectionParam = connectionParam;
 		this.model = model;
+		this.extension = extension;
 		this.controller = new SpiderController(this, extension.getCustomParsers());
 		this.listeners = new LinkedList<>();
 		this.seedList = new LinkedHashSet<>();
 		this.cookieManager = new CookieManager();
 		this.scanContext = scanContext;
-		this.extension = extension;
 		
 		init();
 	}
@@ -440,6 +440,15 @@ public class Spider {
 						+ isTerminated() + "].");
 			}
 		}
+	}
+
+	/**
+	 * Gets the extension.
+	 *
+	 * @return the extension
+	 */
+	protected ExtensionSpider getExtensionSpider() {
+		return this.extension;
 	}
 
 	/* SPIDER PROCESS maintenance - pause, resume, shutdown, etc. */
--- a/src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
+++ b/src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
@@ -19,25 +19,27 @@ package org.zaproxy.zap.spider.parser;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Date;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
+import net.htmlparser.jericho.Attribute;
 import net.htmlparser.jericho.Element;
-import net.htmlparser.jericho.FormControl;
-import net.htmlparser.jericho.FormControlType;
 import net.htmlparser.jericho.FormField;
 import net.htmlparser.jericho.FormFields;
 import net.htmlparser.jericho.HTMLElementName;
 import net.htmlparser.jericho.Source;
 
+import org.apache.commons.httpclient.URI;
 import org.parosproxy.paros.network.HtmlParameter;
 import org.parosproxy.paros.network.HtmlParameter.Type;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.model.DefaultValueGenerator;
+import org.zaproxy.zap.model.ValueGenerator;
 import org.zaproxy.zap.spider.SpiderParam;
 import org.zaproxy.zap.spider.URLCanonicalizer;
 
@@ -47,39 +49,47 @@ import org.zaproxy.zap.spider.URLCanonicalizer;
 public class SpiderHtmlFormParser extends SpiderParser {
 
 	private static final String ENCODING_TYPE = "UTF-8";
-	private static final String DEFAULT_NUMBER_VALUE = "1";
-	private static final String DEFAULT_FILE_VALUE = "test_file.txt";
-	private static final String DEFAULT_TEXT_VALUE = org.parosproxy.paros.Constant.PROGRAM_NAME_SHORT;
-
-	private static final String METHOD_POST = "POST";
-	private static final String ATTR_TYPE = "type";
 	private static final String DEFAULT_EMPTY_VALUE = "";
-	private static final String DEFAULT_PASS_VALUE = DEFAULT_TEXT_VALUE;
+	private static final String METHOD_POST = "POST";
+	private URI uri;
+	private String url;
+
+	/** The form attributes*/
+	private Map<String, String> envAttributes = new HashMap<String, String>();
 
 	/** The spider parameters. */
 	private final SpiderParam param;
 
-	/**
-	 * The default {@code Date} to be used for default values of date fields.
-	 * <p>
-	 * Default is {@code null}.
-	 * 
-	 * @see #setDefaultDate(Date)
-	 */
-	private Date defaultDate;
+	/**Create new Value Generator field*/
+	private final ValueGenerator valueGenerator;
 
 	/**
 	 * Instantiates a new spider html form parser.
-	 * 
+	 *
 	 * @param param the parameters for the spider
 	 * @throws IllegalArgumentException if {@code param} is null.
 	 */
 	public SpiderHtmlFormParser(SpiderParam param) {
+		this(param, new DefaultValueGenerator());
+	}
+
+	/**
+	 * Instantiates a new spider html form parser.
+	 *
+	 * @param param the parameters for the spider
+	 * @param param the parameters ValueGenerator
+	 * @throws IllegalArgumentException if {@code param} or {@code valueGenerator} is null.
+	 */
+	public SpiderHtmlFormParser(SpiderParam param, ValueGenerator valueGenerator) {
 		super();
 		if (param == null) {
 			throw new IllegalArgumentException("Parameter param must not be null.");
 		}
+		if (valueGenerator == null){
+			throw new IllegalArgumentException("Parameter valueGenerator must not be null.");
+		}
 		this.param = param;
+		this.valueGenerator = valueGenerator;
 	}
 
 	@Override
@@ -97,6 +107,7 @@ public class SpiderHtmlFormParser extends SpiderParser {
 
 		// Get the context (base url)
 		String baseURL = message.getRequestHeader().getURI().toString();
+		uri = message.getRequestHeader().getURI();
 
 		// Try to see if there's any BASE tag that could change the base URL
 		Element base = source.getFirstElement(HTMLElementName.BASE);
@@ -114,6 +125,11 @@ public class SpiderHtmlFormParser extends SpiderParser {
 		List<Element> forms = source.getAllElements(HTMLElementName.FORM);
 
 		for (Element form : forms) {
+			//Clear the attributes for each form and store their key and values
+			envAttributes.clear();
+			for (Attribute att : form.getAttributes()){
+				envAttributes.put(att.getKey(), att.getValue());
+			}
 			// Get method and action
 			String method = form.getAttributeValue("method");
 			String action = form.getAttributeValue("action");
@@ -137,6 +153,7 @@ public class SpiderHtmlFormParser extends SpiderParser {
 				action = action.substring(0, fs);
 			}
 
+			url = URLCanonicalizer.getCanonicalURL(action, baseURL);
 			FormData formData = prepareFormDataSet(form.getFormFields());
 
 			// Process the case of a POST method
@@ -241,7 +258,7 @@ public class SpiderHtmlFormParser extends SpiderParser {
 			if (field.getFormControl().getFormControlType().isSubmit()) {
 				currentList = submitFields;
 			}
-			for (String value : getValues(field)) {
+			for (String value : getDefaultTextValue(field)) {
 				currentList.add(new HtmlParameter(Type.form, field.getName(), value));
 			}
 		}
@@ -251,56 +268,83 @@ public class SpiderHtmlFormParser extends SpiderParser {
 
 	/**
 	 * Gets the values for the given {@code field}.
-	 * <p>
-	 * If the field is of submit type it returns its predefined values.
+	 * If the field is of submit type it passes the predefined values to the ValueGenerator and returns its predefined values.
+	 * Gets the default value that the input field, including HTML5 types, should have.
 	 *
 	 * @param field the field
 	 * @return a list with the values
-	 * @see #getDefaultTextValue(FormField)
 	 */
-	private List<String> getValues(FormField field) {
+	private List<String> getDefaultTextValue(FormField field) {
+
+		// Get the Id
+		String fieldId = field.getName();
+
+		// Create new HashMap 'fieldAttributes' and new list 'definedValues'
+		Map<String, String> fieldAttributes = new HashMap<String, String>();
+		List<String> definedValues = new ArrayList<String>();
+
+		//Store all values in the FormFiled field into the Map 'fieldAttributes'
+		fieldAttributes.putAll(field.getFormControl().getAttributesMap());
+
+		// Places a key, Control Type, for each FormControlType
+		fieldAttributes.put("Control Type", field.getFormControl().getFormControlType().name());
+
+
+		//Handles Submit Fields
 		if (field.getFormControl().getFormControlType().isSubmit()) {
-			return new ArrayList<>(field.getPredefinedValues());
+			List<String> submitFields = new ArrayList<String>();
+			for (String value : field.getPredefinedValues()){
+				String finalValue = this.valueGenerator.getValue(uri, url, fieldId, value, definedValues, envAttributes, fieldAttributes);
+				submitFields.add(finalValue);
+			}
+			return submitFields;
 		}
 
 		// Get its value(s)
 		List<String> values = field.getValues();
+		String defaultValue;
+
+		//If the field has a value attribute present(Predefined value)
+		//Should store the value being submitted to be passed to the ValueGenerator
+		if(field.getFormControl().getAttributesMap().containsKey("value")){
+			defaultValue = field.getFormControl().getAttributesMap().get("value");
+		}
+
 		if (log.isDebugEnabled()) {
 			log.debug("Existing values: " + values);
 		}
 
 		// If there are no values at all or only an empty value
 		if (values.isEmpty() || (values.size() == 1 && values.get(0).isEmpty())) {
-			String finalValue = DEFAULT_EMPTY_VALUE;
+			defaultValue = DEFAULT_EMPTY_VALUE;
 
 			// Check if we can use predefined values
 			Collection<String> predefValues = field.getPredefinedValues();
 			if (!predefValues.isEmpty()) {
+				//Store those predefined values in a list for the DefaultValueGenerator
+				definedValues.addAll(predefValues);
 				// Try first elements
 				Iterator<String> iterator = predefValues.iterator();
-				finalValue = iterator.next();
+				defaultValue = iterator.next();
 
 				// If there are more values, don't use the first, as it usually is a "No select"
 				// item
 				if (iterator.hasNext()) {
-					finalValue = iterator.next();
-				}
-			} else {
-				/*
-				 * In all cases, according to Jericho documentation, the only left option is for
-				 * it to be a TEXT field, without any predefined value. We check if it has only
-				 * one userValueCount, and, if so, fill it with a default value.
-				 */
-				if (field.getUserValueCount() > 0) {
-					finalValue = getDefaultTextValue(field);
+					defaultValue = iterator.next();
 				}
 			}
 
-			log.debug("No existing value for field " + field.getName() + ". Generated: " + finalValue);
-
-			values = new ArrayList<>(1);
-			values.add(finalValue);
+		} else {
+			defaultValue = values.get(0);
 		}
+
+		//Get the default value used in DefaultValueGenerator
+		String finalValue = this.valueGenerator.getValue(uri, url, fieldId, defaultValue, definedValues, envAttributes, fieldAttributes);
+
+		log.debug("Generated: " + finalValue + "For field " + field.getName());
+
+		values = new ArrayList<>(1);
+		values.add(finalValue);
 
 		return values;
 	}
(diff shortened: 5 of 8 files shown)


kingthorin:
Please fix-up or squash into a single "atomic" commit, and resolve the outstanding conflicts.

rwehe:
Hey @kingthorin, thanks for the link. @cdlaguna and I somehow  overlooked the contributing documentation but we believe we've made the changes necessary. When you get a chance please let us know if there's anything else that stands out! When we squashed the commits the date wasn't updated so it's showing the same date as our initial commit - not sure if that's an issue or not. 
Thanks!

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
```diff
@@ -119,6 +123,17 @@ private void initialize() {
 		this.customParseFilters = new LinkedList<>();
 		this.scanController = new SpiderScanController(this);
 	}
+	
+	/* */
```
kingthorin: Some actual javadoc could be included here (preferably). Or, it could be removed.

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
```diff
@@ -119,6 +123,17 @@ private void initialize() {
...
+	public void setValueGenerator (ValueGenerator generator) {
+	    this.generator = generator;
+	}
+	
+	
+
```
kingthorin: Some of these blank lines should be removed.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
+package org.zaproxy.zap.model;
```
kingthorin: The standard license header should be included.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+		if (defaultDate == null) {
+			return new Date();
+		}
+		return defaultDate;
+	}
+
+	/**
+	 * Sets the default {@code Date}, to be used for default values of date fields.
+	 * <p>
+	 * If none ({@code null}) is set it's used the current date when generating the values for the fields.
```
kingthorin: `* If none ({@code null}) is set then the current date when generating the values for the fields is used.`

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+	 *
+	 * @param date the default date, might be {@code null}.
+	 * @since TODO add version
```
kingthorin: This should probably be included on all the javadoc in this file.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+		this.defaultDate = date;
+	}
+	
+
+	/**
+	 * <p>
+	 * Generates accurate field values for following types:
+	 * <ul>
+	 * <li>Text/Password/Search - DEFAULT_TEXT_VALUE</li>
+	 * <li>number/range - if min is defined, then use min, otherwise if max is defined use max
```
kingthorin: `* <li>number/range - if min is defined, then use min. If max is defined use max,`

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+	 * otherwise DEFAULT_NUMBER_VALUE;</li>
```
kingthorin: None of the other li have semi-colons

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+	 * <li>url - http://www.example.com</li>
+	 * <li>email - contact@example.com</li>
+	 * <li>color - #000000</li>
+	 * <li>tel - 9999999999</li>
+	 * <li>date/datetime/time/month/week/datetime-local - current date in the proper format</li>
+	 * <li>file - DEFAULT_FILE_VALUE</li>
+	 * </ul> 
+	 */
+
```
kingthorin: This blank line should be removed

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+	 * <li>tel - 9999999999</li>
+	 * <li>date/datetime/time/month/week/datetime-local - current date in the proper format</li>
+	 * <li>file - DEFAULT_FILE_VALUE</li>
+	 * </ul> 
+	 */
+	
+	@Override
+	public String getValue(URI uri, Map<String, String> formAttributes, Map<String, String> fieldAttributes) {
+
+		if (fieldAttributes.get("Control Type") .equalsIgnoreCase (FormControlType.TEXT.toString())) {
```
kingthorin: Seems to be some extra spaces here.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+				}
+				String max = fieldAttributes.get("max");
+				if (max != null) {
+					return max;
+				}
+				return DEFAULT_NUMBER_VALUE;
+			}
+
+			if (type.equalsIgnoreCase("url")) {
+				return "http://www.example.com";
```
kingthorin: Use the constant defined earlier?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			if (type.equalsIgnoreCase("email")) {
+				return "foo-bar@example.com";
```
kingthorin: Use the constant defined earlier?
Different value than constant?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			if (type.equalsIgnoreCase("color")) {
```
kingthorin: Define a LIST (or array) of possible type names? (Could we encounter colour?) Or does this just assume HTML5 field types?


Is there a plan to guess type (which may just be text) based on element name?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+				return "#ffffff";
```
kingthorin: Use the constant defined earlier?
Different value than constant?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+					return max;
+				}
+				return DEFAULT_NUMBER_VALUE;
+			}
+
+			if (type.equalsIgnoreCase("url")) {
+				return "http://www.example.com";
+			}
+
+			if (type.equalsIgnoreCase("email")) {
```
kingthorin: Define a LIST (or array) of possible type names? (Could we encounter "e-mail", "email_add", "email_address", "email_addy"?) Or does this just assume HTML5 field types? 

Is there a plan to guess type (which may just be text) based on element name?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+				return "foo-bar@example.com";
+			}
+
+			if (type.equalsIgnoreCase("color")) {
+				return "#ffffff";
+			}
+
+			if (type.equalsIgnoreCase("tel")) {
+				return "9999999999";
```
kingthorin: Use the constant defined earlier?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			if (type.equalsIgnoreCase("datetime")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'");
```
kingthorin: Does java contain bundled constants for these SDFs? (Same for the bunch below.)

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+			}
+			if (type.equalsIgnoreCase("month")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-MM");
+				return format.format(getDefaultDate());
+			}
+			if (type.equalsIgnoreCase("week")) {
+				SimpleDateFormat format = new SimpleDateFormat("yyyy-'W'ww");
+				return format.format(getDefaultDate());
+			}
+		} else if (fieldAttributes.get("Control Type") .equalsIgnoreCase (FormControlType.PASSWORD.toString())) {
```
kingthorin: Extra spaces can be removed.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,136 @@
...
+			return DEFAULT_PASS_VALUE;
+		} else if (fieldAttributes.get("Control Type") .equalsIgnoreCase(FormControlType.FILE.toString())) {
```
kingthorin: Extra spaces can be removed.

src/org/zaproxy/zap/model/ValueGenerator.java
```diff
@@ -0,0 +1,8 @@
+package org.zaproxy.zap.model;
```
kingthorin: The standard license header should be included.

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -172,13 +172,13 @@ public Spider(String id, ExtensionSpider extension, SpiderParam spiderParam, Con
 		this.id = id;
 		this.spiderParam = spiderParam;
 		this.connectionParam = connectionParam;
-		this.model = model;
+		this.model = model;		
+		this.extension = extension;
```
kingthorin: Does this really need to be moved from [L181](https://github.com/zaproxy/zaproxy/pull/3028/commits/00b4690e76aab56f86329d7574382ede7fbde703#diff-1f3205859ebf47606f09a0c795150206L181) to here?

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -823,4 +823,10 @@ public Thread newThread(Runnable r) {
 			return t;
 		}
 	}
+
+	public ExtensionSpider getExtensionSpider() {
+		// TODO Auto-generated method stub
```
kingthorin: Comment could be removed?

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -19,67 +19,56 @@
...
 import java.util.LinkedList;
 import java.util.List;
-
-import net.htmlparser.jericho.Element;
-import net.htmlparser.jericho.FormControl;
-import net.htmlparser.jericho.FormControlType;
-import net.htmlparser.jericho.FormField;
-import net.htmlparser.jericho.FormFields;
-import net.htmlparser.jericho.HTMLElementName;
-import net.htmlparser.jericho.Source;
```
kingthorin: Do these really need to be moved?

It is cleaner to have them grouped nicely, however that's a formatting change and should be included as a separate PR.
https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -19,67 +19,56 @@
...
+	
 	/**
 	 * Instantiates a new spider html form parser.
 	 * 
 	 * @param param the parameters for the spider
+	 * @param param the parameters ValueGenerator
 	 * @throws IllegalArgumentException if {@code param} is null.
 	 */
-	public SpiderHtmlFormParser(SpiderParam param) {
+	public SpiderHtmlFormParser(SpiderParam param, ValueGenerator valueGenerator) {
```
kingthorin: This breaks backward compatibility, is that intentional?

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -225,14 +214,15 @@ private void processGetForm(HttpMessage message, int depth, String action, Strin
...
+
+	public FormData prepareFormDataSet(FormFields form) {
 		List<HtmlParameter> formDataSet = new LinkedList<>();
 		List<HtmlParameter> submitFields = new ArrayList<>();
 
 		// Process each form field
 		Iterator<FormField> it = form.iterator();
 		while (it.hasNext()) {
-			FormField field = it.next();
+			  FormField field = it.next();
```
kingthorin: Formatting changes should occur in a separate PR:
https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#formatting

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -321,118 +311,19 @@ private void notifyPostResourceFound(HttpMessage message, int depth, String url,
...
-	 * <li>color - #000000</li>
-	 * <li>tel - 9999999999</li>
-	 * <li>date/datetime/time/month/week/datetime-local - current date in the proper format</li>
-	 * <li>file - DEFAULT_FILE_VALUE</li>
-	 * </ul>
-	 * 
 	 * @param field the field
-	 * @return the default text value
-	 * @see #getDefaultDate()
+	 * @return hashMap fieldAttributes
```
kingthorin: It's returning a `String`

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -321,118 +311,19 @@ private void notifyPostResourceFound(HttpMessage message, int depth, String url,
...
-
-	/**
-	 * Sets the default {@code Date}, to be used for default values of date fields.
-	 * <p>
-	 * If none ({@code null}) is set it's used the current date when generating the values for the fields.
-	 *
-	 * @param date the default date, might be {@code null}.
-	 * @since TODO add version
-	 */
-	public void setDefaultDate(Date date) {
```
kingthorin: This was a public method, can it be removed without breaking any dependent code?

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -903,11 +905,12 @@ public void shouldIgnoreBaseHtmlIfActionIsAbsoluteWhenParsingPostForm() {
     @Test
     public void shouldSetValuesToFieldsWithNoValueWhenParsingGetForm() {
         // Given
-        SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser();
+    	DefaultValueGenerator valueGenerator = new DefaultValueGenerator();
+    	SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser(valueGenerator);
```
kingthorin: Seems to be an indentation inconsistency here (tabs vs spaces)?

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -939,11 +942,12 @@ public void shouldSetValuesToFieldsWithNoValueWhenParsingGetForm() {
     @Test
     public void shouldSetValuesToFieldsWithNoValueWhenParsingPostForm() {
         // Given
-        SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser();
+        DefaultValueGenerator valueGenerator = new DefaultValueGenerator();
+    	SpiderHtmlFormParser htmlParser = createSpiderHtmlFormParser(valueGenerator);
```
kingthorin: Seems to be an indentation inconsistency here (tabs vs spaces)?

kingthorin:
I don't think the date is an issue. I did a fixup on https://github.com/zaproxy/zap-extensions/pull/639 and it seems to still be dated 3 days ago....no biggy.

As for other things, I've done a quick review. Off to a great start, most of the things I pointed out are minor.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
rwehe: it appears the test files use spaces while files outside of test use tabs?

kingthorin: It's possible. We do have plans to standardize and go through an auto-formatting activity but it's on the to-do list.... 

If you need to adjust something to be consistent I suggest just copying the indentation on one of the surrounding lines and pasting it, not optimal but workable.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
thc202: Since this is a new class it's enough with one since tag, in the class' JavaDoc.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
thc202: The default generator is just an extraction of the existing behaviour. Other value generator implementations might do that (or the default in the future).

src/org/zaproxy/zap/model/DefaultValueGenerator.java
thc202: Same as previous comment regarding the current default behaviour.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
thc202: Yes, this method was not included in any main release yet.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
kingthorin: Ok

src/org/zaproxy/zap/model/DefaultValueGenerator.java
cdlaguna: We do not believe there are constants defined to return these values for each HTML type. This is all from refactored code. Should we be creating constants to store these values?

src/org/zaproxy/zap/model/DefaultValueGenerator.java
cdlaguna: @kingthorin We've reviewed the Oracle documentation concerning [SDF](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html)  and do not beleive there are any constants that correspond to the date formats. This code was also moved over from SpiderHtmlFormParser.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2015 The ZAP Development Team
```
kingthorin: :smiley: It's 2016 :smiley:

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -321,118 +318,19 @@ private void notifyPostResourceFound(HttpMessage message, int depth, String url,
...
-		this.defaultDate = date;
+		// Create an empty HashMap called 'fieldAttributes'
+		Map<String, String> fieldAttributes = new HashMap<String, String>();
+		
+		//Store all values in the FormFiled field into the Map 'fieldAttributes' 
+		fieldAttributes.putAll(field.getFormControl().getAttributesMap());
+		
+		// Places a key, Control Type, for each FormControlType
+		fieldAttributes.put("Control Type", field.getFormControl().getFormControlType().name());
+		return this.valueGenerator.getValue(null, null, fieldAttributes);
```
psiinon: Shouldnt we be supplying the url and formAttributes here?
These are core changes - we cant update this code without a full release. If we dont specify them then we probably shouldnt include them in the interface, or at least comment it to say they are not currently supported. But I'd much rather they were included, pending any i/f changes ;)

src/org/zaproxy/zap/model/ValueGenerator.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.model;
+
+import java.net.URI;
+import java.util.Map;
+
+public interface ValueGenerator {
```
psiinon: I'm having second thoughts about this interface (and I was involved in specifying it;)
We have quite a few usecases for this enhancement - HTML forms submitted by the std and ajax spiders, WSDL defns and OpenAPI defns.
The interface seems too targeted at the HTML forms, and its really not clear what keys should be used in other cases.
I also suspect that in most cases implementations will just use the field name ;)
So how about we pull out the field name as a string but also add a type string, which could be things like "htmlform", "wsdl", "openapi" etc - it shouldnt be an enum as we'd want it to be as extensible as possible. That way implementations can target specific types of fields.
And also maybe rename formAttributes to something more generic like envAttributes?
It would also be good to document (somewhere) what the keys actually are, eg for HTML form the tag forms, and give some examples. This will be more important for the WSDL / OpenAPI attributes as it will be much less clear what they will be.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
cdlaguna: We have modified the SpiderHtmlFormParser to now return a URI to the DefaultValueGenerator instead of a null. We did this by creating a new variable of type URI and adding a line of code to the parsing method that stored the URI in our new variable. Would this approach be appropriate for supplying the URI to the ValueGenerator, or is there another way we should be returning it?

cdlaguna: Would the form attributes be done in a similar manner as the fieldAttributes. Is there a list where the values are already stored, like the field attributes, or do we need to add code to obtain all attributes for the form tags and then proceed with placing them in a hash map?

psiinon: Yes, I actually do something similar here: https://github.com/psiinon/zap-extensions/blob/d65b00c54e2b8cf64e22496af7828990fdc8ac23/src/org/zaproxy/zap/extension/ascanrulesBeta/Csrftokenscan.java#L271
So you can retreive all of the form attributes using formElement.getAttributes() then you can loop through them and add them to a HashMap

cdlaguna: We have now modified the code to provide a type URI, a string URL, and HashMap EnvAttributes to the DefaultValueGenerator. We would like to know if we are taking the appropriate method to storing values inside the envAttributes.

rwehe: Is this correct to store the values? 
https://github.com/zaproxy/zaproxy/pull/3028/files#diff-e23e6cb557d6464022791cdbb73f7ddfR121

thc202:
Attached a [diff](https://github.com/zaproxy/zaproxy/files/773845/generator-form-tests.txt) with a test to verify the data provided by the form parser to the value generator. The assertions pass but not all of them are correct, there are some task comments showing the expected values (and what's not expected).

I'll add some review comments shortly.

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
```diff
@@ -69,6 +71,8 @@
  */
 public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedListener, ScanController<SpiderScan> {
 
+	public ValueGenerator generator = new DefaultValueGenerator();
```
thc202: This should be private.

src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
```diff
@@ -127,6 +131,14 @@ private void initialize() {
 		this.scanController = new SpiderScanController(this);
 	}
 
+	public void setValueGenerator (ValueGenerator generator) {
+	    this.generator = generator;
```
thc202: We should validate that's non-null (to fail sooner).

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Map;
+
+import org.apache.commons.httpclient.URI;
+
+import net.htmlparser.jericho.FormControlType;
+
+public class DefaultValueGenerator implements ValueGenerator{
```
thc202: Consider adding JavaDoc (and since tag).

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+	
+	private Date defaultDate;
```
thc202: This variable could be defined after the class variables.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+		return defaultDate;
+	}
+
+	/**
+	 * Sets the default {@code Date}, to be used for default values of date fields.
+	 * <p>
+	 * If none ({@code null}) is set then the current date when generating the values for the fields is used.
+	 *
+	 * @param date the default date, might be {@code null}.
+	 * @since TODO add version
```
thc202: This JavaDoc tag could be removed (the class is new).

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+	 */
+	public void setDefaultDate(Date date) {
+		this.defaultDate = date;
+	}
+
+	/**
+	 * <p>
```
thc202: The p tag could be removed.

src/org/zaproxy/zap/model/ValueGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.model;
+
+import java.util.Map;
+
+import org.apache.commons.httpclient.URI;
+
+public interface ValueGenerator {
```
thc202: Consider adding JavaDoc (and since tag).

src/org/zaproxy/zap/model/ValueGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+	public String getValue(URI uri, String url, String fieldId, Map<String, String> envAttributes, Map<String,String> fieldAttributes);
```
thc202: - Consider adding JavaDoc (specially to explain the meaning of the parameters).
 - The public modifier could be removed.

src/org/zaproxy/zap/spider/Spider.java
```diff
@@ -823,4 +823,8 @@ public Thread newThread(Runnable r) {
 			return t;
 		}
 	}
+
+	public ExtensionSpider getExtensionSpider() {
```
thc202: Could be protected instead (and it could be defined before the nested class).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -47,39 +49,41 @@
...
-	/**
-	 * The default {@code Date} to be used for default values of date fields.
-	 * <p>
-	 * Default is {@code null}.
-	 * 
-	 * @see #setDefaultDate(Date)
-	 */
-	private Date defaultDate;
+	/** The spider parameters. */
+	private SpiderParam param;
```
thc202: Could be left as final.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -47,39 +49,41 @@
...
 
+	/**Create new Value Generator field*/
+	private ValueGenerator valueGenerator;
```
thc202: Could be final.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -114,6 +119,9 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
 		List<Element> forms = source.getAllElements(HTMLElementName.FORM);
 
 		for (Element form : forms) {
+			for (Attribute att : form.getAttributes()){
+				envAttributes.put(att.getKey(), att.getValue());
```
thc202: The `envAttributes` should be reset before processing each form (in the attached test you can notice that the attributes of the first form are passed to the second).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -143,6 +151,7 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
 			if (method != null && method.trim().equalsIgnoreCase(METHOD_POST)) {
 				// Build the absolute canonical URL
 				String fullURL = URLCanonicalizer.getCanonicalURL(action, baseURL);
+				url = fullURL;
```
thc202: The `url` should be set before calling `prepareFormDataSet` (the attached test shows that the first `url` passed to form generator is null and for the second is the value of the first form).

thc202:
@psiinon and I talked about the fields that should be passed to the value generator, at the moment just fields that do not have a value are passed to the value generators. Ideally the form parser should pass all fields, to allow the generators to override existing values.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -114,6 +119,10 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
 		List<Element> forms = source.getAllElements(HTMLElementName.FORM);
 
 		for (Element form : forms) {
+			for (Attribute att : form.getAttributes()){
+				envAttributes.clear();
```
thc202: This should be done before this loop.

thc202:
Feel free to include the previously attached test (with the appropriate changes) in the pull request, good to have that running now.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -969,16 +981,131 @@ public void shouldSetValuesToFieldsWithNoValueWhenParsingPostForm() {
...
+        // When
+        boolean completelyParsed = htmlParser.parseResource(msg, source, BASE_DEPTH);
+        // Then
+        assertThat(valueGenerator.getFields(), hasSize(6));
+        assertThat(
+                valueGenerator.getFields().get(0),
+                is(equalTo(formField(
+                        "http://example.com/",
+                        // XXX The following should be: "http://example.org/post"
+                        null,
```
thc202: This can now be replaced with the actual value (code is now working correctly?). Same for other task comments in this test.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -969,16 +981,131 @@ public void shouldSetValuesToFieldsWithNoValueWhenParsingPostForm() {
...
+                        attributes(attribute("name", "field3"), attribute("type", "text"), attribute("Control Type", "TEXT")),
+                        attributes(
+                                attribute("action", "http://example.org/post"),
+                                attribute("method", "POST"),
+                                attribute("atta", "valueA"))))));
+        assertThat(
+                valueGenerator.getFields().get(3),
+                is(equalTo(formField(
+                        "http://example.com/",
+                        // XXX The following should be: "http://example.com/get"
```
thc202: In this one there's a typo, it should be `.org` not `.com`.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
cdlaguna: There is a single error when running the JUnit test. It's the URL field returning a target URI when it should be null.  Were the values of null supposed to be changed to match the comment above them?

thc202: Correct, that should use the value mentioned in the above comment. Sorry, for not being clear.

cdlaguna: No problem. It should be implied when it's commented in. Additionally, we should be commenting out the attribute that is  noted as well?

thc202: Yes, that needs to be changed too (as you already did). The comments can be removed, that was just as for info until the fixes were applied.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -242,6 +252,19 @@ private FormData prepareFormDataSet(FormFields form) {
 				currentList = submitFields;
 			}
 			for (String value : getValues(field)) {
```
rwehe: @psiinon @thc202  (comment for lines 255-267)
@cdlaguna and I have set it so that fields that contain the attribute 'value' are now passed to the valueGenerator. They will pass in the value associated as a parameter (defaultValue) and the DefaultValueGenerator will return the same 'defaultValue'. If a new value is provided by the valueGenerator to replace the default then it will be updated then passed to the Spider. We're open to feedback on this approach to handling predefined values.

rwehe:
After adding the defaultValue functionality to the valueGenerator we are now failing the unit test. We adjusted the unit test to include defaultValue but are still failing.

Do you have any suggestions as to what else should be changed on the unit test to accommodate for the new defaultValue parameter?

Thanks!

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,150 @@
...
+	 * <li>date/datetime/time/month/week/datetime-local - current date in the proper format</li>
+	 * <li>file - DEFAULT_FILE_VALUE</li>
+	 * </ul> 
+	 */
+	@Override
+	public String getValue(URI uri, String url, String fieldId, String defaultValue, Map<String, String> envAttributes, Map<String, String> fieldAttributes) {
+
+		//If there is a default value provided, return it
+		if (defaultValue != ""){
+			return defaultValue;
```
rwehe: This conditional should return the same default value, however our extension should be able to modify the value now that it's passed in.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
thc202: That's fine, one thing that might worth changing now is to pass all values of the fields (a new parameter `List<String> values`) as that can be used by the generators to return one of the provided values (other than default value).

Also, the method `getValues` could be removed (i.e. the logic moved to `getDefaultTextValue`), since we will be passing all fields to the value generators.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
thc202: Looks good. /the check could be changed to `!defaultValue.isEmpty()` though).

thc202:
The number of fields that are passed to the value generator is greater now (8 instead of 6), that's why it's failing. It also needs to check the new fields (also good to check the default value and the values).

You can run the test from within Eclipse (select the method `shouldProvidedCorrectFormDataToValueGenerator` and select context menu item Run As > JUnit Test), it's easier/faster to check what's failing.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -242,6 +252,21 @@ private FormData prepareFormDataSet(FormFields form) {
...
+					//Proceed if the field has a value attribute present(Predefined value)
+					if(field.getFormControl().getAttributesMap().containsKey("value") && 
+							!field.getFormControl().getAttributesMap().get("value").equalsIgnoreCase("submit")	)
```
thc202: Missed this change while reviewing yesterday (outdated content), it's fine to let the submit fields pass to the value generators.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
cdlaguna: We have moved the logic from the getValues method. Please let us know if this was done properly or needs to be revised.

At the moment defaultValues are no longer being used. It seems the logic within the for loop no longer works as we cannot call the getDefaultTextValue and return a type string. How would we go about passing the field into the generator again? Or should it already have been passed in the first time(should we restructure our loop)?

cdlaguna: Disregard last comment. All fields are now being passed to the generator after moving logic from getValues method, and are returning default values as they should be. Please review our movement of the logic and let us know if it needs revision.

test/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -903,11 +913,12 @@ public void shouldIgnoreBaseHtmlIfActionIsAbsoluteWhenParsingPostForm() {
     @Test
     public void shouldSetValuesToFieldsWithNoValueWhenParsingGetForm() {
```
cdlaguna: After moving the logic from getValues method we are now failing additional JUnit tests:
- shouldSetValuesToFieldsWithNoValueWhenParsingGetForm
- shouldSetValuesToFieldsWithNoValueWhenParsingPostForm

Not sure if this was a result of removing the method or another error on our part, working to address these issues as well as modify the new test for both new parameters.

thc202: It seems `definedValues` is not initialised before being used leading to an exception (maybe that's already fixed locally?). Once that's fixed the tests still fail, they should still pass without changing them though.
The problem seems to be that `field.getValues()` is no longer being used, before it obtained/used the first value when available and non-empty.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
thc202: The submit fields should also be passed to the value generator, other than that (and the comment added previously) it looks good.

cdlaguna:
The Value Generator works as it should now with all new parameters being passed in. JUnit Tests have been updated to have a few test cases for the two newest parameters (defaultValue & definedValues). If there are any issues with the Unit test, formatting, or functionality changes that need to be done please let us know.

Thank you.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.httpclient.URI;
+
+import net.htmlparser.jericho.FormControlType;
+
+/**
+ *   Default implementation of the DefaultValueGenerator
```
thc202: Last word should be `ValueGenerator`.

src/org/zaproxy/zap/model/DefaultValueGenerator.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ *   @since 2.6.0
+ *   @return the default String value for each control type
```
thc202: This tag can be moved to the `getValue` method (it's just for methods that return something),

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -47,39 +49,41 @@
...
-	 */
-	private Date defaultDate;
+	/**Create new Value Generator field*/
+	private final ValueGenerator valueGenerator;
 
 	/**
 	 * Instantiates a new spider html form parser.
 	 * 
 	 * @param param the parameters for the spider
+	 * @param param the parameters ValueGenerator
```
thc202: This param tag should be added to the new constructor instead.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
 
 	/**
 	 * Gets the values for the given {@code field}.
-	 * <p>
-	 * If the field is of submit type it returns its predefined values.
+	 * If the field is of submit type it passes the predefined values to the DefaultValueGenerator and returns its predefined values.
```
thc202: It should mention `ValueGenerator` instead.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
-	 * @see #getDefaultTextValue(FormField)
 	 */
-	private List<String> getValues(FormField field) {
-		if (field.getFormControl().getFormControlType().isSubmit()) {
-			return new ArrayList<>(field.getPredefinedValues());
-		}
+	private List<String> getDefaultTextValue(FormField field) {
 
 		// Get its value(s)
 		List<String> values = field.getValues();
```
thc202: Could now be moved to after the if "Handles Submit Fields", it's only needed after that if.

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
+		String fieldId = field.getName();
+		String defaultValue = "";
```
thc202: - The initialisation to empty string could be removed (it's guaranteed to be initialised when used).
 - Could be moved to after the if "Handles Submit Fields" (same reason as `values`).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
+		String finalValue;
```
thc202: This variable could be declared just in line 343 (and in/after 297, for the remaining uses it ends up being assigned to `defaultValue`, so those can be assigned directly to `defaultValue`).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
+
+		//Store all values in the FormFiled field into the Map 'fieldAttributes'
+		fieldAttributes.putAll(field.getFormControl().getAttributesMap());
+
+		// Places a key, Control Type, for each FormControlType
+		fieldAttributes.put("Control Type", field.getFormControl().getFormControlType().name());
+
+		//If the field has a value attribute present(Predefined value)
+		//Should store the value being submitted to be passed to the ValueGenerator
+		if(field.getFormControl().getAttributesMap().containsKey("value")){
```
thc202: Could be moved to after the if "Handles Submit Fields" (same reason as the move of `values`).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
-			values.add(finalValue);
+		} else {
+			defaultValue = values.get(0);
 		}
 
+		/*
+		 * In all cases, according to Jericho documentation, the only left option is for
+		 * it to be a TEXT field, without any predefined value. We check if it has only
+		 * one userValueCount, and, if so, fill it with a default value.
+		 */
```
thc202: This comment could be removed (no longer matches this code/method).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
+
+		//Get the default value used in DefaultValueGenerator
+		finalValue = this.valueGenerator.getValue(uri, url, fieldId, defaultValue, definedValues, envAttributes, fieldAttributes);
+
+		//Reset the Default Value to empty for each field
+		defaultValue = "";
```
thc202: This could be removed, it's no longer needed (it's not used after calling `getValue` above).

src/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -251,57 +262,95 @@ private FormData prepareFormDataSet(FormFields form) {
...
+
+		log.debug("No existing value for field " + field.getName() + ". Generated: " + finalValue);
```
thc202: The log message should be updated (we now have a value at this point).

src/org/zaproxy/zap/model/DefaultValueGenerator.java
rwehe: @cdlaguna and I applied the changes for all of your latest comments. 
On this comment we just moved the return tag, not the entire javadoc. Is this what you were asking? Thanks!

thc202: Yes.

thc202:
Thank you very much!

PR #3034, opened 2016-11-28
Remove alerts.xml file

thc202:
Remove alerts.xml file, its contents (i.e. alerts' data) are not used
nor maintained.
Move the registry of the scanners IDs to a new file, scanners.md, which
was previously in the alerts.xml (as XML comment).


thc202:
Comments about the format/contents of new file (and its location) are very welcome.

Issue #2898, opened 2016-09-28
Tweak spider parser to ignore/strip matched parenthesis around URLs


stephendonner:
Using https://github.com/stephendonner/docker-zap/blob/a6e591ebbdfe8be78112a5f8e41bfb396388cd20/run-docker.sh I did a scan of https://treeherder.allizom.org, and it apparently found a comment:

Bootstrap v3.3.5 (http://getbootstrap.com)

And then tried to extract the host with the parenthesis:

3390 [pool-1-thread-1] WARN org.zaproxy.zap.spider.URLCanonicalizer  - Error while Processing URL in the spidering process (on base ): Host could not be reliably evaluated from: http://getbootstrap.com)

"@thc202: we should tweak the parser to ignore the parenthesis"

Full log below:

`Started by upstream project "docker-zap" build number 44
originally caused by:
 Started by user Stephen Donner
Building in workspace /var/lib/jenkins/workspace/docker-zap/default
git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
git config remote.origin.url https://github.com/stephendonner/docker-zap.git # timeout=10
Fetching upstream changes from https://github.com/stephendonner/docker-zap.git
git --version # timeout=10
git fetch --tags --progress https://github.com/stephendonner/docker-zap.git +refs/heads/*:refs/remotes/origin/*
Checking out Revision a6e591ebbdfe8be78112a5f8e41bfb396388cd20 (refs/remotes/origin/master)
git config core.sparsecheckout # timeout=10
git checkout -f a6e591ebbdfe8be78112a5f8e41bfb396388cd20
git rev-list a6e591ebbdfe8be78112a5f8e41bfb396388cd20 # timeout=10
[default] $ /usr/bin/env bash /tmp/hudson1902304993759994340.sh
https://treeherder.allizom.org
[INFO]            ZAP is running
[INFO]            Accessing URL https://treeherder.allizom.org
[INFO]            Running spider...
[INFO]            Running an active scan...
[INFO]            Issues found: 0

=================================================================ZAP-daemon log output follows=================================================================
Found Java version 1.8.0_45-internal
Available memory: 1839 MB
Setting jvm heap size: -Xmx512m
449 [main] INFO org.zaproxy.zap.DaemonBootstrap  - OWASP ZAP D-2016-09-05 started.
550 [main] INFO org.parosproxy.paros.common.AbstractParam  - Setting config view.mode = attack was null
[...]

PR #3053, opened 2016-12-06
Tweak log message in URLCanonicalizer

thc202:
Change a log message to include the URL that is being processed, also
change to return immediately if the the URL is not valid after logging
the problem (instead of throwing an exception, which would be caught
(and logged) in the same method).

### DIFF
--- a/src/org/zaproxy/zap/spider/URLCanonicalizer.java
+++ b/src/org/zaproxy/zap/spider/URLCanonicalizer.java
@@ -21,7 +21,6 @@
 
 package org.zaproxy.zap.spider;
 
-import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


psiinon:
Are warnings the most useful thing at this level? (I know its the current behaviour;)
Might it be better to throw exceptions and then allow the caller to decide if it wants to warn / ignore etc?

thc202:
I think warning is useful, the spider should be able to resolve all URLs correctly and it's good to know when they are not (e.g. in the case it's an issue in the extraction of the URLs #2898). Maybe it should (also) be shown to the user? Spider tab?
At the moment the class is only used by spider related classes, so the behaviour of the callers would end up being the same ("warn and return"). If we plan to use it in other places we should probably extract it to other package and remove some of the spider related logic (e.g. strip of session parameters and other "cleaning" methods).

psiinon:
I agree - I think this info is useful to the user, in which case I think it should be shown in the UI rather than reported in the logs. And the spider tab sounds like the right place for it to me - we could always make this an option.

thc202:
OK, should those changes be done for 2.6.0?

psiinon:
Would be nice, but def not a blocker (imo)

thc202:
OK, thanks.

PR #3055, opened 2016-12-07
Add initiator constant for AJAX spider requests

thc202:
Add a constant to HttpSender class for requests sent by the AJAX spider.


thc202:
Yes, I didn't want to update/change other files as the AJAX Spider is still not using that value.
Let's leave the pull request open until the AJAX Spider changes are done and I'll get back to this and update the other files(?).

kingthorin:
Sure

thc202:
Core template updated, there's at least one template in an add-on that needs to be updated too, but that should be done near the 2.6.0 release (IMO). There are no help pages that need to be updated (AFAICT).

Changes for AJAX spider to use the new initiator ID are in zaproxy/zap-extensions#651.

PR #3056, opened 2016-12-07
Allow to extend ProxyThread

thc202:
Change ProxyThread to allow to be extended (from other packages) and use
a custom HttpSender, required for the AJAX Spider to use a custom
initiator ID.


src/org/parosproxy/paros/core/proxy/ProxyThread.java
```diff
@@ -141,10 +142,23 @@
...
+	protected ProxyThread(ProxyServer server, Socket socket) {
+		this(server, socket, null);
+	}
+
+	/**
+	 * Constructs a {@code ProxyThread} with the given proxy server, socket and HTTP sender.
+	 *
+	 * @param server the parent proxy server.
+	 * @param socket the connected socket to read/write the messages.
+	 * @param httpSender the object used to send the messages, might be {@code null} in which case is used a default.
```
kingthorin: `in which case a default is used.`

thc202:
Change done, thank you!

PR #3062, opened 2016-12-12
Add tests for OptionsParamApi

thc202:
Add tests for OptionsParamApi to assert the expected behaviour.
Also, do other tweaks to OptionsParamApi:
 - Remove commented code and related constant (unimplemented option);
 - Remove initialisations with default value and initialise the enabled
 instance variable as true (default value used when loading from file);
 - Properly handle malformed values in the configuration file;
 - Do not attempt to set and save the API key if the configurations was
 not set.

### DIFF
--- a/src/org/zaproxy/zap/extension/api/OptionsParamApi.java
+++ b/src/org/zaproxy/zap/extension/api/OptionsParamApi.java
@@ -18,28 +18,30 @@
 package org.zaproxy.zap.extension.api;
 
 import org.apache.commons.configuration.ConfigurationException;
+import org.apache.commons.configuration.ConversionException;
+import org.apache.log4j.Logger;
 import org.parosproxy.paros.common.AbstractParam;
 
 public class OptionsParamApi extends AbstractParam {
 
+	private static final Logger LOGGER = Logger.getLogger(OptionsParamApi.class);
+
 	public static final String ENABLED = "api.enabled";
 	public static final String SECURE_ONLY = "api.secure";
-	public static final String POST_ACTIONS = "api.postactions";
 	public static final String API_KEY = "api.key";
 	private static final String DISABLE_KEY = "api.disablekey";
 	private static final String INC_ERROR_DETAILS = "api.incerrordetails";
 	private static final String AUTOFILL_KEY = "api.autofillkey";
 	private static final String ENABLE_JSONP = "api.enablejsonp";
 	
-	private boolean enabled = false;
-	private boolean secureOnly = false;
-	private boolean disableKey = false;
-	private boolean incErrorDetails = false;
-	private boolean autofillKey = false;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/api/OptionsParamApi.java
```diff
@@ -48,16 +50,34 @@ public OptionsParamApi() {
...
-		autofillKey = getConfig().getBoolean(AUTOFILL_KEY, false);
-		enableJSONP = getConfig().getBoolean(ENABLE_JSONP, false);
-	    key = getConfig().getString(API_KEY, "");
-	    //postActions = getConfig().getBoolean(POST_ACTIONS, false);
+	    enabled = getBooleanFromConfig(ENABLED, true);
+	    secureOnly = getBooleanFromConfig(SECURE_ONLY, false);
+		disableKey = getBooleanFromConfig(DISABLE_KEY, false);
+		incErrorDetails = getBooleanFromConfig(INC_ERROR_DETAILS, false);
+		autofillKey = getBooleanFromConfig(AUTOFILL_KEY, false);
+		enableJSONP = getBooleanFromConfig(ENABLE_JSONP, false);
```
kingthorin: If these lines are changing anyway is there any reason to keep the extra indent (lines 55 to 58 etc vs 53/54)

test/org/zaproxy/zap/extension/api/OptionsParamApiUnitTest.java
```diff
@@ -0,0 +1,403 @@
...
+    public void shouldFailToSetEnabledWithoutConfig() {
+        // Given / When
+        OptionsParamApi param = new OptionsParamApi();
+        // When
+        param.setEnabled(true);
+        // Then = NullPointerException
+    }
+
+    @Test
+    public void shouldSetEnabledWithConfig() {
```
kingthorin: This title is confusing (to me) as things are set and tested disabled.

src/org/zaproxy/zap/extension/api/OptionsParamApi.java
thc202: No reason, it's just that I didn't need to change the indentation of the statements. I'll change that (worth noting that the first two statements are the ones with "wrong" indentation, using mixed chars).

test/org/zaproxy/zap/extension/api/OptionsParamApiUnitTest.java
thc202: Yeah, I initially had "EnabledState" but later changed to just "Enabled" (to match the other test names), this test is asserting that the "enabled state" is being changed/set (to false, since it's true by default).

thc202:
Changes done, thank you.

Issue #3060, opened 2016-12-12
Send Forced Browse requests through ZAP


thc202:
Forced Browse requests are sent directly by DirBuster code bypassing most of the rules/configurations set in ZAP. For example, it's not be possible to use Contexts/Users nor change its requests (e.g. HTTP Sender script).

Add-on:
Forced Browse

PR #3063, opened 2016-12-12
Add initiator constant for Forced Browse requests

thc202:
Add a constant to HttpSender class for requests sent by the Forced
Browse add-on.

Related to #3060 - Send Forced Browse requests through ZAP


thc202:
Same comment, Brute Force add-on was not yet changed to use that (worst in this case, as it will require a lot more changes...).

kingthorin:
Ok

psiinon:

thc202:
Conflict resolved.

PR #3067, opened 2016-12-14
Allow to deprecate ZAP API endpoints

### LINKED ISSUE #3061: Allow to deprecate API endpoints
Allow to deprecate API endpoints (e.g. for unused/deprecated options #3058). Deprecated API endpoints should not be shown in ZAP API UI and should be marked as deprecated in the generated ZAP API clients.

thc202:
Change the ZAP API to allow to set its endpoints as deprecated (and add
a description why they are).
Change ZAP API UI to show a note when the endpoints are deprecated.
Change JAVA API generator to annotate and add JavaDoc tag to deprecated
endpoints.

Fix #3061 - Allow to deprecate API endpoints


src/lang/Messages.properties
```diff
@@ -71,6 +71,7 @@ api.dialogue.browseApiNotEnabled.title = API Not Enabled
 api.dialogue.browseApiNotEnabled.message = The API must be enabled before you can browse it.\nDo you want to enable the API?
 api.dialogue.browseApiNotEnabled.button.confirm.label = Enable and Browse the API
 api.dialogue.browseApiNotEnabled.button.cancel.label = Cancel
+api.deprecated.option.endpoint = Option no longer in effective use.
```
kingthorin: I think we need a different word here, "effective" implies that it can be used but that the current use is partially wrong or something like that. i.e.: If it is no longer being effectively used then it was at some point, or could be at some point in the future.

src/org/zaproxy/zap/extension/api/ApiElement.java
```diff
@@ -108,4 +122,54 @@ public void setOptionalParamNames(List<String> optionalParamNames) {
...
+	 * @see #getDeprecatedDescription()
+	 */
+	public boolean isDeprecated() {
+		return deprecated;
+	}
+
+	/**
+	 * Sets whether or not the API element is deprecated.
+	 * <p>
+	 * Deprecated elements are shown in the {@link WebUI} with a note that they are deprecated and its use is discouraged. The
```
kingthorin: `their use is discouraged`

src/org/zaproxy/zap/extension/api/ApiElement.java
```diff
@@ -108,4 +122,54 @@ public void setOptionalParamNames(List<String> optionalParamNames) {
...
+	 * API generators might also marked the elements as deprecated.
```
kingthorin: `mark` (not marked)

src/lang/Messages.properties
kingthorin: I couldn't immediately think of a good alternative as I read this, I'll keep it in the back of my head and if I think of some good text add it here. If you have another idea when that's great too 👍

thc202: That's what I meant, those API endpoints can still be used but they shouldn't (either are not working properly or were replaced by better alternatives), also, the corresponding (Java) methods are no longer in use in ZAP (i.e. no ZAP code is calling those methods).

src/org/zaproxy/zap/extension/api/ApiElement.java
thc202: Will change.

src/org/zaproxy/zap/extension/api/ApiElement.java
thc202: Right :)

thc202:
Addressed some comments, the first one is still pending.

src/lang/Messages.properties
thc202: @kingthorin what do you think?

kingthorin: Ok, I'm good with that. Leave it as is.

thc202: OK, all comments should be addressed then.

PR #3089, opened 2016-12-20
Log the name of the user of the active scan

thc202:
Change HostProcess to include the name of the user (if any) when logging
the information of the scan being started. That information is useful
when reviewing what the scan was doing (or, expected to do).


src/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -251,6 +252,24 @@ public void run() {
         }
     }
 
+    /**
+     * Logs information about the scan.
+     * <p>
+     * It logs the {@link #nodeInScopeCount number of nodes} that will be scanned and the name of the {@link #user}, if any.
+     */
+    private void logScanInfo() {
+        StringBuilder strBuilder = new StringBuilder(150);
```
psiinon: nit - might be better to check it info logging is on before creating the StringBuilder?

thc202: Didn't add any check because that's the default level and it's not logged often. If you prefer I can add the check.

Issue #3095, opened 2016-12-23
ProxyThread SocketTimeoutException Overly Verbose


kingthorin:
Per IRC: 
http://logs.glob.uno/?c=mozilla%23websectools&s=20+Dec+2016&e=20+Dec+2016#c28879
http://logs.glob.uno/?c=mozilla%23websectools&s=20+Dec+2016&e=20+Dec+2016#c28894

The logging of SocketTimeoutException in ProxyThread could probably be less verbose in general use.

PR #3096, opened 2016-12-23
ProxyThread SocketTimeoutException Verbosity

kingthorin:
Only log full exception if debug is enabled.

Fixes zaproxy/zaproxy#3095

### DIFF
--- a/src/org/parosproxy/paros/core/proxy/ProxyThread.java
+++ b/src/org/parosproxy/paros/core/proxy/ProxyThread.java
@@ -64,6 +64,7 @@
 // ZAP: 2016/09/22 JavaDoc tweaks
 // ZAP: 2016/11/28 Correct proxy errors' Content-Length value.
 // ZAP: 2016/12/07 Allow to extend the ProxyThread and use a custom HttpSender
+// ZAP: 2016/12/23 Make SocketTimeoutException less verbose for general use
 
 package org.parosproxy.paros.core.proxy;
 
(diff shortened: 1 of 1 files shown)


src/org/parosproxy/paros/core/proxy/ProxyThread.java
```diff
@@ -270,7 +271,10 @@ public void run() {
 	    	if (firstHeader != null) {
 	    		log.warn("Timeout accessing " + firstHeader.getURI());
 	    	} else {
-	    		log.warn("Timeout", e);
+	    		log.warn("Socket Timeout");
```
thc202: Worth changing to "Timed out while reading first message.", this is a special case.

src/org/parosproxy/paros/core/proxy/ProxyThread.java
```diff
@@ -270,7 +271,10 @@ public void run() {
 	    	if (firstHeader != null) {
 	    		log.warn("Timeout accessing " + firstHeader.getURI());
 	    	} else {
-	    		log.warn("Timeout", e);
+	    		log.warn("Socket Timeout");
+	    		if (log.isDebugEnabled()) {
+	    			log.debug(e);
```
thc202: This should call `debug(e, e)` to include the exception's stack trace.

kingthorin:
Done & done

thc202:
Thank you! :)

PR #3106, opened 2016-12-29
Do not require status/version in add-on file name

### LINKED ISSUE #3090: Be more lenient on add-on's file name format
ZAP should not require the status nor the version in the file name of the add-on, both properties can be obtained (they already are) from the manifest bundled in the add-on (i.e. ZapAddOn.xml). This change would require less efforts to produce an add-on that is accepted by ZAP (for example, by default an add-on/artifact built using Maven/Gradle does not have one of the statuses required by ZAP).

thc202:
Change AddOn class to not require the status and/or the version in the
file name of the add-on. The add-on file name just needs to have the ID
and have a ZAP extension. Also, deprecate old constructor/methods that
require the file name to have the status/version and introduce new
constructor/methods where applicable.
Change BaseZapAddOnXmlData to read the status from the manifest file of
the add-on (ZapAddOn.xml).
Remove hardcoded manifest file name (ZapAddOn.xml) from JavaDoc and code
(by using the constant from AddOn).
Change AddOnCollection to iterate just ZAP add-on files and use the new
AddOn constructor.
Change ExtensionAutoUpdate to use the new constructors/methods and to
copy the file from manual add-on installations using a normalised file
name.
Add some tests to assert the expected behaviour of AddOn class.

Fix #3090 - Be more lenient on add-on's file name format

### DIFF
--- a/src/org/zaproxy/zap/control/AddOn.java
+++ b/src/org/zaproxy/zap/control/AddOn.java
@@ -24,12 +24,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
@@ -54,6 +57,14 @@ import org.zaproxy.zap.control.BaseZapAddOnXmlData.ExtensionWithDeps;
 import org.zaproxy.zap.extension.pscan.PluginPassiveScanner;
 
 public class AddOn  {
+
+	/**
+	 * The name of the manifest file, contained in the add-ons.
+	 * 
+	 * @since TODO add version
+	 */
+	public static final String MANIFEST_FILE_NAME = "ZapAddOn.xml";
+
 	public enum Status {unknown, example, alpha, beta, weekly, release}
 	
 	private static ZapRelease v2_4 = new ZapRelease("2.4.0");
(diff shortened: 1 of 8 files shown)


src/org/zaproxy/zap/control/AddOn.java
```diff
@@ -54,6 +57,14 @@
...
+	/**
+	 * The name of the manifest file, contained in the add-ons.
+	 * 
+	 * @since TODO add version
+	 */
+	public static final String MANIFEST_FILE_NAME = "ZapAddOn.xml";
+
 	public enum Status {unknown, example, alpha, beta, weekly, release}
 	
 	private static ZapRelease v2_4 = new ZapRelease("2.4.0");
```
kingthorin: Is this supposed to be updated per release? Aren't version numbers also maintained in a constant class somewhere? (I thought I remembered seeing them elsewhere)

src/org/zaproxy/zap/control/AddOn.java
```diff
@@ -167,8 +178,21 @@
...
+	/**
+	 * Tells whether or not the given file name matches the name of a ZAP add-on.
+	 * <p>
+	 * The file name must have the format "{@code <id>-<status>-<version>.zap}". The {@code id} is a string, the {@code status}
+	 * must be a value from {@link Status} and the {@code version} must be an integer.
+	 *
+	 * @param fileName the name of the file to check
+	 * @return {@code true} if the given file name is the name of an add-on, {@code false} otherwise.
+	 * @deprecated (TODO add version) Use {@link #isAddOnFileName(String)} instead, the checks done in this method are more
+	 *             restrict than it needs to.
```
kingthorin: Just `strict` would be fine or `restrictive`

thc202: OK, will change that.

src/org/zaproxy/zap/control/AddOn.java
thc202: No, that version was created to prevent loading (older) add-ons that could be incompatible with newer ZAP version (2.4.0).

src/org/zaproxy/zap/control/AddOn.java
thc202: Change done.

src/org/zaproxy/zap/control/AddOn.java
kingthorin: K

Issue #3113, opened 2017-01-02
Move Spider to an add-on


thc202:
Tasks:
 - [x] Migrate #7299 once merged.
     - Done in zaproxy/zap-extensions#3886
 - [x] Migrate #7356 and #7414 once merged.
     - Done in zaproxy/zap-extensions#3945 
 - [x] Ensure spider classes are all using parameterized logging (vs concatenation), and adjust class names so that acronyms aren't full caps (Uri vs URI, etc).
    - Done in zaproxy/zap-extensions#3888
 - [x] Migrate existing add-ons to use the spider from the add-on:
    - [x] Automation (Done in zaproxy/zap-extensions#3962)
    - [x] Form Handler (Superseded with zaproxy/zap-extensions#3954)
    - [x] GraqhQL (Done in https://github.com/zaproxy/zap-extensions/pull/3953 and zaproxy/zap-extensions#3955)
    - [x] OpenAPI (Done in https://github.com/zaproxy/zap-extensions/pull/3920)
    - [x] Quick Start (Done in zaproxy/zap-extensions#3972)
    - [x] SOAP (Done in https://github.com/zaproxy/zap-extensions/pull/3952)
 - [x] Use Form Handler add-on (Done in zaproxy/zap-extensions#3959)
 - [x] Merge #7335 once all above tasks done.
 - [x] Change `SpiderParam` to extend `VersionedAbstractParam`. (Done in zaproxy/zap-extensions#3976)
 - [x] Include spider add-on in weekly releases. (Done in zaproxy/zaproxy#7335)
 - [x] Add to zap-admin's allow list for website help pages. (Done in https://github.com/zaproxy/zap-admin/pull/809)
 - [x] Complete zaproxy/zaproxy#4398
     - [x] Migrate the code to the add-on. zaproxy/zap-extensions#3991
 - [x] Remove core SQLite dependency (#7473).
   - [x] Move SQLite to an add-on and change spider to use it (zaproxy/zap-extensions#4067).
 - [x] Update the HUD to use the "new" event publisher. (Done in zaproxy/zap-hud#1141)
 - [x] Update 2.12 release notes to mention the migration. (Done in https://github.com/zaproxy/zap-core-help/pull/460)
 - [x] Update add-on changelog and 2.12 release notes to mention the changes compared to 2.11 spider. (Done in https://github.com/zaproxy/zap-extensions/pull/4081)
[...]

PR #3117, opened 2017-01-04
Move Spider functionality to Spider extension

thc202:
Move (that is, create new class and deprecate old one) the spider
functionality to ExtensionSpider, also, all spider related functionality
is now done through classes in the spider extension package.
The functionality "Exclude from Spider" is shared by both spiders (AJAX
and traditional) therefore it belongs to core.

Preparation for #3113 - Move Spider to an add-on


thc202:
There's some changes that are not correct, the "Exclude from Spider" session panel should still be kept as a core functionality (it's used by all spiders, traditional and AJAX).

thc202:
Changes done and pull request description updated.

thc202:
@kingthorin please, take a look at the changes again (classes Session and SpiderThread). Restored some changes to remove the usage of ExtensionSpider by Session class.

kingthorin:
I'll try to go through this again later today.

thc202:
Thank you!

PR #3118, opened 2017-01-05
include subjectAlternativeName extension in generated certificates

mozkeeler:
When generating a certificate on the fly for a domain, zaproxy doesn't include a subjectAlternativeName extension (which is mandatory for new certificates in recent versions of Firefox). This appears to do the right thing.


mozkeeler:
Actually, to be more correct, the extension is only mandatory for certificates issued from built-in roots. The root generated by zaproxy isn't a built-in, so this normally doesn't matter. However, you can configure Firefox to treat an imported root as a built-in, whereupon this actually would matter.

psiinon:
Thanks @mozkeeler :)
https://github.com/zaproxy/zap-core-help/blob/master/src/help/zaphelp/contents/credits.html

mozkeeler:
I guess just "David Keeler (@mozkeeler)". Thanks!

thc202:
Done in zaproxy/zap-core-help@68e1f390d06f0f0b3a7fe29293c7d8f899a127ed. Thank you!

PR #3120, opened 2017-01-05
Allow to manage domains in scope with Spider API

### LINKED ISSUE #3058: Allow to configure the domains always in spider scope (Spider API)
Reported in IRC channel.

It should be possible to configure, through the ZAP API, the domains that are always in spider scope (Options > Spider > Domains that are always 'in scope').

The ZAP API should provide the same functionality as the UI, add/modify/remove and enable/disable all.

The following related auto generated API endpoints should be phased out:
 - `spider/view/optionScope`
 - `spider/view/optionScopeText`
 - `spider/action/setOptionScopeString`

they are generated from old/deprecated options (no longer shown in the GUI).

The following related auto generated API endpoints need to be changed to return correct values:
 - `spider/view/optionDomainsAlwaysInScope`
 - `spider/view/optionDomainsAlwaysInScopeEnabled`

they are generated from current "options" (although they shouldn't be being exposed through the ZAP API).

thc202:
Add views and actions that allow to manage the domains that are always
in spider scope.

Fix #3058 - Allow to configure the domains always in spider scope
(Spider API)


src/lang/Messages.properties
```diff
@@ -1876,12 +1876,20 @@ sites.spider.popup         = Spider...
 sites.showinsites.popup    = Show in Sites Tab
 
 spider.activeActionPrefix = Spidering: {0}
+spider.api.action.addDomainAlwaysInScope = Adds a new domain that's always in scope.
+spider.api.action.modifyDomainAlwaysInScope = Modifies a domain that's always in scope.
```
psiinon: What does this actually mean? Might be worth explaining a bit more + giving an overview of the params?

thc202: That API endpoint just allows to modify one of the domains that is always in scope (like in the UI, change to regex, the value or enable/disable it).
Will improve the description of the endpoint and add describe the parameters.

thc202:
Changes done and conflict resolved.

thc202:
Fixed the modification of the domain (could use wrong old values) and added description to `setOptionScopeString`.

src/lang/Messages.properties
```diff
@@ -1891,13 +1891,22 @@ spider.activeActionPrefix = Spidering: {0}
 spider.api.action.clearExcludedFromScan = Clears the regexes of URLs excluded from the spider scans.
 spider.api.action.excludeFromScan = Adds a regex of URLs that should be excluded from the spider scans.
 spider.api.action.setOptionSendRefererHeader = Sets whether or not the 'Referer' header should be sent while spidering.
+spider.api.action.addDomainAlwaysInScope = Adds a new domain that's always in scope, using the specified value and (optionally) if enabled (default, true) and if it's a regex (default, false).
```
kingthorin: The and ... and in this sentence reads a little awkwardly. How about:
> Adds a new domain that's always in scope, using the specified value. Optionally sets if the new entry is enabled (default, true) and whether or not the new value is specified as a regex (default, false).

src/lang/Messages.properties
```diff
@@ -1891,13 +1891,22 @@ spider.activeActionPrefix = Spidering: {0}
...
+spider.api.action.modifyDomainAlwaysInScope = Modifies a domain that's always in scope. Allows to modify the value, if enabled or if a regex. The domain is selected with its index.
+spider.api.action.removeDomainAlwaysInScope = Removes a domain that's always in scope, with the given index.
```
kingthorin: For this one and the one before it, should we have some kind of `See also spider.api.view.domainsAlwaysInScope`? So that users know how to get the index?

src/lang/Messages.properties
```diff
@@ -1891,13 +1891,22 @@ spider.activeActionPrefix = Spidering: {0}
...
 spider.api.action.setOptionMaxChildren = Sets the maximum number of child nodes (per node) that can be crawled, 0 means no limit.
 spider.api.action.scan = Runs the spider against the given URL (or context). Optionally, the 'maxChildren' parameter can be set to limit the number of children scanned, the 'recurse' parameter can be used to prevent the spider from seeding recursively, the parameter 'contextName' can be used to constrain the scan to a Context and the parameter 'subtreeOnly' allows to restrict the spider under a site's subtree (using the specified 'url').
 spider.api.action.scanAsUser = Runs the spider from the perspective of a User, obtained using the given Context ID and User ID. See 'scan' action for more details.
+spider.api.action.setOptionScopeString = Use actions [add|modify|remove]DomainAlwaysInScope instead.
 spider.api.view.allUrls = Returns a list of unique URLs from the history table based on HTTP messages added by the Spider.
+spider.api.view.domainsAlwaysInScope = Gets all the domains that are always in scope. For each domain is shown the index, value, if enabled and if a regex.
```
kingthorin: I'd suggest something like:
> Gets all the domains that are always in scope. For each domain the following are shown: the index, the value (domain), if enabled, and if specified as a regex.

src/lang/Messages.properties
thc202: Right, will add a comment about that.

thc202:
No worries, thanks for taking a look at it :)

thc202:
Changes applied.

PR #3136, opened 2017-01-11
Allow to manage outgoing proxy excluded domains

### LINKED ISSUE #2283: Change the return value of the optionProxyExcludedDomains view
The view [optionProxyExcludedDomains](http://zap/UI/core/view/optionProxyExcludedDomains) currently returns a list of _ProxyExcludedDomainMatcher_. This should be changed to human readable plaintext.

### LINKED ISSUE #2284: Show regex in the optionProxyChainSkipName view
The view [optionProxyChainSkipName ](http://zap/UI/core/view/optionProxyChainSkipName) does not show the regex defined in ZAP (Tools --> Options --> Connection -- > 'Add' button). The regular expressions should be added to this view.

### LINKED ISSUE #2285: Create a new action to add a domain/IP that should be skipped by the proxy
Create a new action, which will allow the user to add a new domain/IP that should be skipped by the outgoing proxy. Then it would be much easier to add more domains if required, instead of overwriting the old values with the current [setOptionProxyChainSkipName](http://zap/UI/core/action/setOptionProxyChainSkipName) action.

The action should also have parameters to specify if the newly added domain exclusion is a regex and if it's enabled (default can be true). Then the behavior would correspond much more with the ZAP GUI (Tools --> Options --> Connection -- > 'Add' button).
### Suggested action
- addOptionProxyChainSkipName(apikey`*`, domain`*`, regex`*`, enabled)

Furthermore there should be actions to remove/modify existing domain exclusions and enable/disable all exclusions at once, similar to the actions that are available in the ZAP GUI.
### Suggested actions
- removeOptionProxyChainSkipName(id`*`)
- modifyOptionProxyChainSkipName(id`*`, domain, regex, enabled)
- enableAllOptionProxyChainSkipNames()
- disableAllOptionProxyChainSkipNames()

thc202:
Add new views and actions to CoreAPI to allow to add, modify, remove and
enable/disable all the domains excluded from the outgoing proxy.
Correct existing CoreAPI (outgoing proxy related) option views to return
correct values.

Fix #2283 - Change the return value of the optionProxyExcludedDomains
view
Fix #2284 - Show regex in the optionProxyChainSkipName view
Fix #2285 - Create a new action to add a domain/IP that should be
skipped by the proxy


src/lang/Messages.properties
```diff
@@ -918,6 +918,12 @@ core.api.action.sendRequest = Sends the HTTP request, optionally following redir
 core.api.action.setMode = Sets the mode, which may be one of [safe, protect, standard, attack]
 core.api.action.setOptionDnsTtlSuccessfulQueries = Sets the TTL (in seconds) of successful DNS queries (applies after ZAP restart).
 core.api.action.shutdown = Shuts down ZAP
+core.api.action.addProxyChainExcludedDomain = Adds a domain to be excluded from the outgoing proxy, using the specified value and (optionally) if enabled (default, true) and if it's a regex (default, false).
```
kingthorin: The and ... and in this sentence reads a little awkwardly. How about:
> Adds a domain to be excluded from the outgoing proxy, using the specified value. Optionally sets if the new entry is enabled (default, true) and whether or not the new value is specified as a regex (default, false).

src/lang/Messages.properties
```diff
@@ -918,6 +918,12 @@ core.api.action.sendRequest = Sends the HTTP request, optionally following redir
...
+core.api.action.modifyProxyChainExcludedDomain = Modifies a domain excluded from the outgoing proxy. Allows to modify the value, if enabled or if a regex. The domain is selected with its index.
+core.api.action.removeProxyChainExcludedDomain = Removes a domain excluded from the outgoing proxy, with the given index.
```
kingthorin: For this one and the one before it, should we have some kind of `See also core.api.view.proxyChainExcludedDomains`? So that users know how to get the index?

src/lang/Messages.properties
```diff
@@ -936,6 +942,10 @@ core.api.view.messages = Gets the HTTP messages sent by ZAP, request and respons
 core.api.view.mode = Gets the mode
 core.api.view.numberOfMessages = Gets the number of messages, optionally filtering by URL
 core.api.view.optionDnsTtlSuccessfulQueries = Gets the TTL (in seconds) of successful DNS queries.
+core.api.view.optionProxyChainSkipName = Use view proxyChainExcludedDomains instead.
+core.api.view.optionProxyExcludedDomains = Use view proxyChainExcludedDomains instead.
+core.api.view.optionProxyExcludedDomainsEnabled = Use view proxyChainExcludedDomains instead.
+core.api.view.proxyChainExcludedDomains = Gets all the domains that are excluded from the outgoing proxy. For each domain is shown the index, value, if enabled and if a regex.
```
kingthorin: I'd suggest something like:
> Gets all the domains that are excluded from the outgoing proxy. For each domain the following are shown: the index, the value (domain), if enabled, and if specified as a regex.

thc202:
Will do the changes, thanks!

thc202:
Changes done.

src/lang/Messages.properties
thc202: Let me know if the new description looks ok (to do the same in the other pull request).

Issue #3043, opened 2016-12-01
Add helper for printing in js templates


kingthorin:
Add the helper discussed here:
https://github.com/zaproxy/community-scripts/pull/48#issuecomment-258136067

Standardize the print/println used in all templates.

PR #3138, opened 2017-01-12
Issue 3043 Add helper for printing to JS templates

kingthorin:
Add helper code that overcomes the lack of println when scripts are run
using Java 8's Nashorn JS engine, by reassigning the function when
undefined.

Fixes zaproxy/zaproxy#3043

### DIFF
--- a/src/scripts/templates/standalone/Traverse sites tree.js
+++ b/src/scripts/templates/standalone/Traverse sites tree.js
@@ -3,12 +3,14 @@
 // Standalone scripts have no template.
 // They are only evaluated when you run them. 
 
+// The following handles differences in printing between Java 7's Rhino JS engine
+// and Java 8's Nashorn JS engine
+if (typeof println == 'undefined') this.println = print;
+
 function listChildren(node, level) {
-    var i;
-    for (i=0;i<level;i++) print ("    ");
     var j;
     for (j=0;j<node.getChildCount();j++) {
-        println(node.getChildAt(j).getNodeName());
+        println(Array(level+1).join("    ") + node.getChildAt(j).getNodeName());
         listChildren(node.getChildAt(j), level+1);
     }
 }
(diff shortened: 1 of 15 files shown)


src/scripts/templates/standalone/Traverse sites tree.js
```diff
@@ -3,6 +3,10 @@
 // Standalone scripts have no template.
 // They are only evaluated when you run them. 
 
+// The following handles differences in printing between Java 7's Rhino JS engine
+// and Java 8's Nashorn JS engine
+if (typeof println == 'undefined') this.println = print;
+
```
thc202: The script could be changed to remove the usage of print (which would add the newline in Java 8, making the output look wrong) ?

src/scripts/templates/targeted/Find HTML comments.js
```diff
@@ -1,5 +1,9 @@
 // Targeted scripts can only be invoked by you, the user, eg via a right-click option on the Sites or History tabs
 
+// The following handles differences in printing between Java 7's Rhino JS engine
+// and Java 8's Nashorn JS engine
+if (typeof println == 'undefined') this.println = print;
+
```
thc202: In this script the `print`s could be replaced with `println` (and removed the newline char).

kingthorin: No problem

src/scripts/templates/standalone/Traverse sites tree.js
kingthorin: K, will figure out something for this one. Can probably do more specific reassignment of the functions in order to handle both.

src/scripts/templates/standalone/Traverse sites tree.js
```diff
@@ -3,12 +3,16 @@
...
 function listChildren(node, level) {
     var i;
-    for (i=0;i<level;i++) print ("    ");
+    for (i=0;i<level;i++)
     var j;
     for (j=0;j<node.getChildCount();j++) {
-        println(node.getChildAt(j).getNodeName());
+        println(Array(i).join("    ") + node.getChildAt(j).getNodeName());
```
thc202: It could use `level + 1` instead of `i` (and previous for loop could be removed).

This change also fixes an issue in the original script (which would print root nodes in wrong position, i.e. as if it was a sub-node of the previously printed node).

kingthorin:
Done

thc202:
Thanks!

PR #3146, opened 2017-01-16
Fix NPE when starting the spider

thc202:
Change ExtensionSpider to have the exclude list never null (which is
used by SpiderThread).

Caused by #3117 - Move Spider functionality to Spider extension


psiinon:
Merged with just one approval as its a small change which fixes the weekly release ;)

PR #3156, opened 2017-01-20
Use G1 as default garbage collector

arincon85:
After have some performance problems and freezes, I change the JVM Garbage Collector to CMS and it works for me.

I add this new configuration as default one, but you can still use .ZAP_JVM.properties to configure it.

Default configuration = -XX:+UseConcMarkSweepGC -XX:+CMSIncrementalMode -XX:+PrintGCDetails -XX:+PrintGCTimeStamps
As oracle suggest for [java 8](https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/cms.html)

-----
<sub>Edit by @kingthorin:</sub>
> Add G1 GC as default garbage collector
> Per http://openjdk.java.net/jeps/248

### DIFF
--- a/src/zap.sh
+++ b/src/zap.sh
@@ -41,12 +41,17 @@ if [ "`echo ${JAVA_OUTPUT} | grep "continuing with system-provided Java"`" ] ; t
   unset JAVA_HOME
 fi
 
+DEFAULTJAVAGC="-XX:+UseG1GC"
+
 JAVA_VERSION=$(java -version 2>&1 | awk -F\" '/version/ { print $2 }')
 JAVA_MAJOR_VERSION=${JAVA_VERSION%%[.|-]*}
 JAVA_MINOR_VERSION=$(echo $JAVA_VERSION | awk -F\. '{ print $2 }')
 
 # JEP 223, newer Java versions (>= 9) no longer use 1 as major version
-if [ $JAVA_MAJOR_VERSION -ge 9 ] || ([ $JAVA_MAJOR_VERSION -ge 1 ] && [ $JAVA_MINOR_VERSION -ge 8 ]); then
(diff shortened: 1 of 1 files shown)


thc202:
Now that ZAP is targeting Java 8 I'm fine changing the default garbage collector, although we should probably use G1 (per http://openjdk.java.net/jeps/248).

thc202:
@kingthorin, @psiinon any comments on this?

psiinon:
Not really :/ Happy to go with whatever is the recommended one :)

kingthorin:
I'm good with going to G1, per the linked JEP it's intended to provide the best overall experience and further improvements are planned for Java 9.

Want me to update this PR?

thc202:
@kingthorin yes, please. Tested the flag with IBM SDK 8, Eclipse OpenJ9, Zulu 9, OracleJRE 8/9 and all worked fine (maybe we could just set that for Java 8?).

kingthorin:
Won't it be released when the core is Java 8, so we don't really need to worry about Java 7 support?

thc202:
Right, was referring to not apply that when running with Java 9, but since that's the default no big deal?

kingthorin:
The way I read things it should be fine for Java 9

kingthorin:
Rebased current, switched to `-XX:+UseG1GC`. Let me know about the other switches, glad to remove them.

thc202:
Yeah, it's fine with Java 9 (it's just a preference, do not set/change it since we don't need to).

Yes, better to remove the other flags (that's for debug).

kingthorin:
Ok so I’ll move it inside the jep 223 conditional and only set it for Java 8?

thc202:
Sounds good.

thc202:
@arincon85, @kingthorin thanks!

PR #3164, opened 2017-01-23
Select first certificate of newly added keystore

### LINKED ISSUE #3163: Autoselect Imported Certificate
In ZAP Options, when you import a new PKCS#12 cert (and probably works the same for the other types as well), after successful import, it takes you back to the KeyStore tab (which is good).  But then to make that the Active cert the user has to:

a) Click on the cert in the left column,
b) Click on the username in the right column,
c) Click on the Set Active button

I'm requesting that when a cert is successfully imported, and they are sent back to the KeyStore tab, you automatically do a) and b) for them so they just have to do step c).

This doesn't hurt anything if they don't want to do step c) right now, but makes it MUCH easier if they do.  This should probably take like 5 minutes or less to implement this usability enhancement.

Given that #2489 is still open, I have to do this every time I start ZAP up again. If #2489 was addressed, I wouldn't care as much (but that's a much harder ticket to address).

thc202:
Change OptionsCertificatePanel to select the first certificate available
of the newly added keystore.

Fix #3163 - Autoselect Imported Certificate

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
@@ -27,6 +27,7 @@
 // ZAP: 2014/08/14 Issue 1184: Improve support for IBM JDK
 // ZAP: 2016/06/28: File chooser for PKCS#12 files now also accepts .pfx files
 // ZAP: 2017/01/09 Remove method no longer needed.
+// ZAP: 2017/01/23 Select first alias of selected keystore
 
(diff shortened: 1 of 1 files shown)


davewichers:
Not sure why you created a new issue for the same thing. Whatever. Thanks so much for knocking this out so quick!!

kingthorin:
This is a Pull Request (code submission) not an issue 😀

davewichers:
Ah. My bad! I'm a github nube. Sorry.

PR #3168, opened 2017-01-26
Create a new 'live' ZAP docker image based on the latest code in the repos

psiinon:
Will upload a version of this image to https://hub.docker.com/r/psiinon/ for people to try out once its finished building (and will link from this issue).
The plan is to auto generate this image whenever the zaproxy repo is changed as per https://docs.docker.com/docker-hub/builds/
I'm also hoping we can use it to auto generate daily builds using https://wiki.mozilla.org/TaskCluster

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -232,6 +232,7 @@
 		<replace file="${dist}/zap.bat" token="zap-dev.jar" value="${zap.jar}"/>
 		<replace file="${dist}/zap.sh" token="zap-dev.jar" value="${zap.jar}"/>
 
+		<chmod perm="755" file="${dist}/zap.sh" />
 	</target>
 
 	<target name="dist-with-source" description="Builds ZAP jar with source to be used in zap-extensions">
--- a/build/docker/Dockerfile-live
+++ b/build/docker/Dockerfile-live
@@ -1,0 +1,97 @@
+# This dockerfile builds a 'live' zap docker image using the latest files in the repos
+FROM ubuntu:16.04
+MAINTAINER Simon Bennetts "psiinon@gmail.com"
+
+RUN apt-get update && apt-get install -q -y --fix-missing \
+	make \
+	ant \
+	automake \
+	autoconf \
+	gcc g++ \
+	openjdk-8-jdk \
+	ruby \
+	wget \
+	curl \
+	xmlstarlet \
+	unzip \
+	git \
+	x11vnc \
+	xvfb \
+	openbox \
+	xterm \
+	net-tools \
+	ruby-dev \
+	python-pip \
+	firefox \
+	vim \
+	xvfb \
+	x11vnc && \
+	apt-get clean && \
+	rm -rf /var/lib/apt/lists/*  && \
+	
+	pip install --upgrade pip && \
+	gem install zapr && \
+	pip install zapcli && \
+	# Install latest dev version of the python API
+	pip install https://github.com/zaproxy/zap-api-python/releases/download/0.0.9.dev1/python-owasp-zap-v2.4-0.0.9.dev1.tar.gz && \
+	useradd -d /home/zap -m -s /bin/bash zap && \ 
+	echo zap:zap | chpasswd && \
+	mkdir /zap  && \
+	chown zap /zap && \
+	chgrp zap /zap && \
+	mkdir /zap-src  && \
+	chown zap /zap-src && \
+	chgrp zap /zap-src
+
+WORKDIR /zap-src
+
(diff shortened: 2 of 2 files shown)


build/docker/Dockerfile-develop
```diff
@@ -0,0 +1,58 @@
+# This dockerfile allows you to build and run OWASP ZAP using the latest source code
+# To build use a command line:
+#	docker build -t owasp/zap2docker-dev -f Dockerfile-develop .
+# To use use a command like
```
thc202: `To run use`?

build/docker/Dockerfile-develop
```diff
@@ -0,0 +1,58 @@
...
+# 	docker run -it owasp/zap2docker-dev bash
+#
+# /zap/src	will contain the ZAP repos with the latest source code (from when the container was built)
+# /zap/zap will contain a version of ZAP built from the source code
+#
+# If you change any of the ZAP extensions the you can rebuild them using:
```
thc202: `then you can`?

build/docker/Dockerfile-develop
```diff
@@ -0,0 +1,58 @@
...
+FROM openjdk:8-jdk
+MAINTAINER Simon Bennetts "psiinon@gmail.com"
+
+ENV ANT_VERSION 1.9.8 
+RUN curl -fsSL https://www.apache.org/dist/ant/binaries/apache-ant-$ANT_VERSION-bin.tar.gz | tar xzf - -C /usr/share \
+  && mv /usr/share/apache-ant-$ANT_VERSION /usr/share/ant \
+  && ln -s /usr/share/ant/bin/ant /usr/bin/ant
+ENV ANT_HOME /usr/share/ant
+
+# Install vi so that its easy to make quick changes
```
thc202: `Install vim`

build/docker/Dockerfile-develop
```diff
@@ -0,0 +1,58 @@
...
+RUN mkdir -p /zap/src 
+RUN chown -R zap /zap && \
+	chgrp -R zap /zap
+WORKDIR /zap/src
+
+USER zap
+
+# Pull the ZAP repos
+RUN git clone https://github.com/zaproxy/zaproxy.git && \
+	git clone https://github.com/zapbot/zap-extensions.git && \
```
thc202: Should be from zaproxy? (same comment for following clones)

psiinon:
Good spots :) Hopefully fixed now

build/docker/Dockerfile-develop
```diff
@@ -0,0 +1,52 @@
...
+RUN cd /zap/src/zaproxy && pwd && ls && \
+	ant -f build/build.xml deploy-weekly-addons && \
+	ant -f build/build.xml day-stamped-release
+	
+RUN chmod +x /zap/src/zaproxy/build/zap/zap.sh
+	
+RUN	ln -s /zap/src/zaproxy/build/zap /zap
+
+# Sanity check that we can run ZAP	
+RUN cd /zap/zap && ./zap.sh -version
```
jvehent: Every dockerfile instruction creates a new layer inside the container, so you can significantly reduce the container size by merging those into a single statement. For example: https://github.com/mozilla-services/hindsight/blob/master/Dockerfile#L20-L62

It might also be useful to delete the source code after the build and remove dev packages to help reduce the image size.

psiinon: The idea of this image was to allow people to start hacking on the ZAP source code easily, so deleting the source wouldnt be good ;)
However ... we could just have another image which was based on this and just deleted the source. And, thinking about it more, this should include all of the other sw that the weekly release includes (if its to replace the weekly release).
So the image should really be much closer to the weekly one :/
I'll do some more playing ...

jvehent: Ha yes, you'd want to keep the build tools then.

I quite like the idea of a lean container for testers and a larger one for developers. I think you could use the same dockerfile for both and alter the build using env variables: http://stackoverflow.com/questions/19537645/get-environment-variable-value-in-dockerfile#34600106

psiinon:
Here the new live one: https://hub.docker.com/r/psiinon/zap2docker-live/

psiinon:
Changed the PR to just include a new 'live' image - this one does not include the source so I've changed the title.
Example available at https://hub.docker.com/r/psiinon/zap2docker-live/
The plan is to use dockerhubs automated build to build this every day or on every change. We can keep maintaining the weekly by duplicating the Monday morning build (assuming no one has reported any issues).
Ready to review now :)
Future changes may include pushing a daily (non docker) release to github, but I've not looked into that yet.

Issue #353, opened 2015-06-04
Enhancement Request - Add timestamps to Spider progress/history frame


zapbot:
```
Would it be possible/practical to add a column to the bottom frame of ZAP so that the
Spider and Active Scan progress/history information shows a request/response timestamp.
(This would make performance problems more obvious).

Also in the header part of that frame where the progress meter itself is shown it would
be nice to include "Current Scans: <#scans> <#active_threads>" (so that you can actually
see how many threads you have configured in the options and that they're all currently
in use).

Lastly an option to reverse the sort order of that progress/history frame would be
handy so that you could choose "newest on top" and then never have to scroll to see
what's going on 'now'. (Alternately a 'auto-scroll' option might work.)
```

Original issue reported on code.google.com by `kingthorin` on 2012-08-31 17:15:59

kingthorin:
I opened this an am marking it wontfix. There seems to be lots of timing and status info available in various places.

kingthorin:
This was addressed by https://github.com/zaproxy/zaproxy/pull/2410, removing wontfix label :smile:

PR #3176, opened 2017-01-30
Allow to show more request data in History tab

thc202:
Change History tab to allow to show the request header/body sizes,
response timestamp and response header size of the HTTP messages. Also,
remove redundant functionality, already provided by HistoryTable.


kingthorin:
If this adds timestamps to history you could mark it as fixing https://github.com/zaproxy/zaproxy/issues/353

Will review the code shortly.

thc202:
The (messages) spider panel already supports that (that issue should be fixed since 2.5.0).

PR #3178, opened 2017-01-31
Coverity Fixes - High

kingthorin:
Fix 2 theoretical resource leaks identified by Coverity Scan.

DbSQL.java and SSLContextManager.java one occurrence each.


thc202:
Looks good, although in DbSQL we should also be closing `reader` (line 109), could you change that as well?

kingthorin:
Sure

thc202:
Thanks!

Issue #2618, opened 2016-06-24
Allow to choose the name of session snapshots


davewichers:
We have the ability to Persist a session, but once done, we have Snapshot Session, but not Save As, which would allow the user to pick the name of this saved session. Snapshot Session autogenerates a name for the snapshot based on the Session name and current time, but I find those names very confusing/hard to read, especially if you have made several of them. (they seem to get longer and longer).

Anyway, I propose either:

1) Replacing Snapshot Session with Save As.
2) Adding a Save As... feature for sessions (but this creates yet another option :-( )
3) If you can figure out how, enhance Snapshot Session to allow the user to use the default autogenerated snapshot name OR chose their own name.

I think 1 or 3 are the best options as ZAP already has the 'too many features' problem so adding another one should be avoided if possible.

kingthorin:
@davewichers Would you prefer Save as (newly saved session becoming active) or ability to specify snapshot name? (Where snapshot functionality stays the same as it currently is, a point in time copy of the session.)

davewichers:
Save As, because I think that is more natural/expected behavior.

Dave

psiinon:
My comment from #3186 as @davewichers probably didnt see it:

The reason we changed from Save As to Persist is the latter is more accurate.
People were getting confused - Save As implies to most people that they are saving a copy, which is not what it does. Its persisting the session in a db that keeps up to date whenever it changes.
So I dont think we should change Persist to Save as unless we also change how it works.

And to follow up...
I still think most people will expect "Save as" to save a one time copy, which it wont be. The snapshots are one time copies, but as mentioned they dont allow the snaphot to be named.

How about either 

* "Snapshot as..." which saves gives a dialog and saves the snapshop with the given name
* "Rename session..." which saves the session with the new name and makes it the active one

@davewichers (and anyone else) any preferences on those?

davewichers:
Thanks for the explanation. That is helpful. I like Snapshot As ...

Dave

kingthorin:
@psiinon @davewichers thank you both. I'll get working on it.

thc202:
Issue title updated to match the current changes.

thc202:
Fixed in #3186.

Issue #638, opened 2015-06-04
Persist and snapshot sessions instead of saving them


zapbot:
```
Change "Save Session.. " to "Persist Session..." and introduce snapshots.
'Persist' describes what actually happens more accurately.
The snapshots allow people to save instances of the session which will not get updated.
```

Original issue reported on code.google.com by `psiinon` on 2013-04-16 18:35:57

PR #3186, opened 2017-02-05
Change to Snapshot As

kingthorin:
Change Snapshot to allow the user to select the name. A datestamped name
is suggested based on the persisted session name along with a timestamp,
however the user can set it to anything they wish.

Accompanying help PR (zaproxy/zap-core-help#69)
will be updated after this change is accepted.


kingthorin:
@zaproxy/team-zaproxy how's that look?

thc202:
I still need to check the history before commenting further (i.e. "Save As" was replaced by "Persist Session").

kingthorin:
Ok

psiinon:
Whats the use case here? 
And the expected behavior?
I'm not sure this adds anything that cant be already done have persisting and snapshots.
We moved away from the 'Save as' terminology as it didnt really reflect what ZAP was actually doing and often confused people.

kingthorin:
https://github.com/zaproxy/zaproxy/issues/2618

Mainly, give the user a chance to save the session with a specified name. Once you've persisted you can't change the name, you can snapshot but then you're stuck with the snapshot naming convention.

I did come across a block of commented code the other day that basically said something along the lines of "being left in case users don't like generated snapshot names", so I guess we could also just put that code back into action.

thc202:
I didn't find the discussion in the mailing list (although I recall that this was discussed). The issue of the change is #638.

Allowing to choose the name used in the snapshots is probably better than having the two items Save As and Persist Session. (The snapshot is not an exact replacement for Save As though, as that does not change the session being used.)

kingthorin:
I see use cases for all of them, though I understand the interest in fewer options (UI clutter). Should we put it back to dave in the issue?

>Would you prefer Save as (newly saved session becoming active) or ability to specify snapshot name? (Where snapshot stays the same as it currently is, a point in time copy of the session.)

thc202:
Agreed. Yeah, worth doing that.

kingthorin:
Answer from Dave:
> Save As, because I think that is more natural/expected behavior.

kingthorin:
@psiinon @thc202 thoughts on this now that we have an answer from the original requester? 

Do we want to make visibility of Save As in the menu optional? 
Do we want to make visibility of Snapshot in the menu optional?
Do we just leave the PR as is?

thc202:
I'd prefer seeing just one menu item, either Persist Session or Save As, both menu items do the same. If we want restore the Save As functionality, we can change Persist Session to be always enabled (it could also be renamed to Save As?).

kingthorin:
I'm good with replacing Persistence with Save As. @psiinon are you ok with that? If so I'll update this PR, and include changes to the startup dialog and options screen/messages. Then update the help PR as well.

psiinon:
The reason we changed from Save As to Persist is the latter is more accurate.
People were getting confused - Save As implies to most people that they are saving a copy, which is not what it does. Its persisting the session in a db that keeps up to date whenever it changes.
So I dont think we should change Persist to Save as unless we also change how it works.
Can someone explain what we're trying to achieve here?

kingthorin:
I dunno what else to say.  I've tried to explain, Dave had tried to explain.  Just close this. PR and the issue I guess.

psiinon:
NP, I'll go through the backlog. I was just trying to explain why we changed from Save As to Persist - I think the original reason was valid and I'm wary of changing back. But if theres a need for a new option then I'm fine with that, we just need to get the terminology right.

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -413,15 +414,15 @@ public String getDescription() {
 	public void saveSnapshot() {
 	    Session session = model.getSession();
 
-	    // This code has been deliberately left here in case people dont like the auto naming of snapshots
-	    // It can be deleted at some point in the future if there are no complaints
-	    /*
 	    JFileChooser chooser = new JFileChooser(model.getOptionsParam().getUserDirectory());
 	    // ZAP: set session name as file name proposal
 	    File fileproposal = new File(session.getSessionName());
+	    String proposedFileName;
```
thc202: Could be declared in the if-then block.

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -413,15 +414,15 @@ public String getDescription() {
...
-	    /*
 	    JFileChooser chooser = new JFileChooser(model.getOptionsParam().getUserDirectory());
 	    // ZAP: set session name as file name proposal
 	    File fileproposal = new File(session.getSessionName());
+	    String proposedFileName;
 	    if (session.getFileName() != null && session.getFileName().trim().length() > 0) {
-	    	// if there is already a file name, use it
-	    	fileproposal = new File(session.getFileName());
+	    	// if there is already a file name, use it and add a timestamp
+	    	proposedFileName = session.getFileName().replace(".session","");
```
thc202: This could remove other text than the extension (although unlikely).

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -461,22 +462,6 @@ public String getDescription() {
...
-	    */
-		String fileName = session.getFileName();
-		
-		if (fileName.endsWith(".session")) {
-		    fileName = fileName.substring(0, fileName.length() - 8);
-		}
-		fileName += "-" + dateFormat.format(new Date()) + ".session";
-		
-		try {
-    	    waitMessageDialog = view.getWaitMessageDialog(Constant.messages.getString("menu.file.savingSnapshot"));	// ZAP: i18n
```
thc202: The current message is using "menu.file.savingSession" instead of "menu.file.savingSnapshot", should use the latter? (if not the message should be removed, does not seem to be used in other place)

src/org/parosproxy/paros/control/MenuFileControl.java
kingthorin: True, I'll make sure it's the last occurrence. That'll be safer.

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -453,30 +455,14 @@ public String getDescription() {
     		}
     		
     		try {
-	    	    waitMessageDialog = view.getWaitMessageDialog(Constant.messages.getString("menu.file.savingSession"));	// ZAP: i18n
+	    	    waitMessageDialog = view.getWaitMessageDialog(Constant.messages.getString("menu.file.savingSnapshot"));	// ZAP: i18n
 	    	    control.snapshotSession(fileName, this);
         	    log.info("snapshot as session file " + session.getFileName());
```
thc202: Just noticed that this should use `fileName` (instead of `session.getFileName()`). Sorry, for noticing sooner.

kingthorin: No worries, will tweak that tonight

thc202:
Thanks!

kingthorin:
Seems I commented on the wrong PR earlier. This comment should have been here, adding just so the history is intact.

Now produces a log message like:
> 18151 [AWT-EventQueue-0] INFO org.parosproxy.paros.control.MenuFileControl  - Snapshotting: C:\Users\someuser\Desktop\zap_sessions\a_session.session as C:\Users\someuser\Desktop\zap_sessions\a_session-20170227-202142.session

PR #3197, opened 2017-02-09
Fix Adv Spider Options Render

kingthorin:
A minor dimension adjustment to ensure that the Spider Advanced Options
tab isn't truncated.

I tried `pack()`ing in various places but it never seemed to impact the issue.

Before:
spider_adv_height

After:
spider_adv_height2

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
+++ b/src/org/zaproxy/zap/extension/spider/ExtensionSpider.java
@@ -801,7 +801,7 @@ public class ExtensionSpider extends ExtensionAdaptor implements SessionChangedL
 
 	public void showSpiderDialog(SiteNode node) {
 		if (spiderDialog == null) {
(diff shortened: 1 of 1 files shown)


thc202:
In the long run we should probably wrap the panels with a `JScrollPane`, it would prevent further truncation issues (e.g. if more fields are added or different font size used). Thoughts?

psiinon:
I did actually have a quick go at that a while ago but it broke some of the panels and I didnt have time to look into the problems.
But yes, I think thats the best way to go.

kingthorin:
Well that works but doesn't at the same time. In this case wouldn't that just add a scroll bar but leave the window size and component positions the same, which would still look like the before pic?

thc202:
Not sure it would be the same, but, yes, not all components might be totally visible at the same time. I was/am more worried of the possible hidden fields (e.g. if you increase the font size (say) to 18 you will not see all fields nor have any indication that they are hidden).

kingthorin:
Oh ok I see your point, I hit that in the options dialog the other day.

PR #3198, opened 2017-02-09
Use only the given context when spider filtering

### LINKED ISSUE #3192: URLs included in context are disregarded by the spider
I am trying to use ZAP via python API client. I use official release 2.5 of ZAP. I want to crawl web site at http://mysite.com, and if there are links to subdomain.mysite.com, include them into crawl.

I tried to use the following code:

```
zap.core.new_session(apikey=apikey)
zap.context.include_in_context(context_name, ".*mysite\.com.*", apikey)
zap.spider.scan(url=target, contextname=context_name, apikey=apikey)
```

However, even though spider starts with a specified context, it still ignores URLs to subdomain.mysite.com and reports them in Spider->Processed window with OUT_OF_SCOPE flag.

thc202:
Change DefaultFetchFilter to only use the context (when set) to check
if a URI is or not valid (exclusion list still applies).

Fix #3192 - URLs included in context are disregarded by the spider

### DIFF
--- a/src/org/zaproxy/zap/spider/filters/DefaultFetchFilter.java
+++ b/src/org/zaproxy/zap/spider/filters/DefaultFetchFilter.java
@@ -62,38 +62,21 @@ public class DefaultFetchFilter extends FetchFilter {
 		try {
 			
 			// Context check
-			if (this.scanContext != null)
-				if (!this.scanContext.isInContext(uri.toString()))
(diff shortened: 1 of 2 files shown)


thc202:
I was wondering if the "domains always in scope" should still be checked if a URI is not included in the context. WDYT?

PR #3217, opened 2017-02-17
Allow to resize history filter dialogue

### LINKED ISSUE #3215: History Filter dialog cant be scaled
This causes problems when a large font size is used, eg when using a high resolution screen.

thc202:
Change HistoryFilterPlugDialog to allow to be resized, also, allow its
components to freely expand.

Fix #3215 - History Filter dialog cant be scaled

### DIFF
--- a/src/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
+++ b/src/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
@@ -19,7 +19,6 @@
  */
 package org.zaproxy.zap.extension.history;
 
-import java.awt.Dimension;
 import java.awt.Frame;
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
```diff
@@ -486,8 +479,8 @@ private JScrollPane getUrlRegxExcScroller() {
 	private JScrollPane getTagScroller() {
 		if (tagScroller == null) {
 			tagList = new JList<>(getTagModel());
+			tagList.setPrototypeCellValue("Tags are short...");
```
kingthorin: Is thus actually exposed in the UI or just a placeholder? If exposed the i18n?

thc202: Just a placeholder to let the list component calculate a (more) reasonable width.

Issue #1634, opened 2015-06-04
Export URLs of specific sites


zapbot:
```
At the moment it seems that it is possible to export only spidered URLs and the only
way to achieve this is through the menu option "Export All URLs to File...".

It would be much more confortable if one could right click on contexts or sites to
export the selected URLs.

Furthermore it should be possible to export all URLs (not only the ones which are detected
throuth the "Spider..." option).
```

Original issue reported on code.google.com by `marabicudi` on 2015-05-13 09:18:21

thc202:
Changed the title of the issue per #3253.

Issue #2623, opened 2016-06-25
Enhancement: Export all URLs to File should alphabetically sort the output


davewichers:
It seems to export them in random order, maybe the same order they are in the Sites tree. Its possible that if issue #1800 I asked for a while ago is addressed, then this will happen automatically as a side affect.

thc202:
Yes, they are exported with the same order as in the Sites tree.

PR #3235, opened 2017-02-23
Issue 1634 Improve Export URLs

kingthorin:
Exporting URLs to a file now takes into consideration what is selected in the Sites Tree.

Refactored PopupMenuExportURLs so that it uses HistoryReferences instead of reading from the DB, and also made it more modular.

Added PopupMenuExportSelectedURLs which extends PopupMenuExportURLs, checks the selection (incl. multi) in the site tree and exports based on selection. (If no selection, export all).

The revised version of this functionality no longer exports the HTTP Method. It exports only URLs in order to be more true to the UI experience of the user.

Fixes zaproxy/zaproxy#1634

zap-core-help\src\help\zaphelp\contents\ui\tlmenu\report.html will be updated in a separate PR after this one is accepted.


src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -79,7 +79,19 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
         	    BufferedWriter fw = null;
                 try {
                     fw = new BufferedWriter(new FileWriter(file, isAppend));
-                    exportURLs((SiteNode) site.getModel().getRoot(), fw, html);
+                    int selectionCount=site.getSelectionCount();
+                    if (selectionCount == 0) {
+                        //No selection, export from root
+                        exportURLs((SiteNode) site.getModel().getRoot(), fw, html);
+                    } else if (selectionCount == 1) {
```
kingthorin: It would also be cool if I could check what was selected in the Context section of the pane, then walk the site tree and see which items applied to the Context. (So that if a context is selected only the URLs applicable to it are exported...which would more completely address the original issue).

Is there an easy way to check for that?

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -79,7 +79,19 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
...
+                        //No selection, export from root
+                        exportURLs((SiteNode) site.getModel().getRoot(), fw, html);
+                    } else if (selectionCount == 1) {
+                        //One selection, export starting there
+                        exportURLs((SiteNode) site.getSelectionPath()
+                            .getLastPathComponent(), fw, html);
+                    } else if (selectionCount > 1) {
+                        //Multi-selection, warn and export all
+                        extension.getView().showWarningDialog(Constant.messages.getString("exportUrls.multiselect.warning"));
+                        exportURLs((SiteNode) site.getModel().getRoot(), fw, html);
```
kingthorin: Note I did look at handling multi-select, however if one of the subsequent selections is a child of an earlier selection it ends up duplicated in the results.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
thc202: The method `SiteMapPanel.getTreeContext()` needs to be public (or, add methods to `SiteMapPanel` to query/get the selection state of the contexts tree).

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
thc202: The exported URIs can be added to a `Set` to avoid the duplicates (i.e. if the URI is already exported, skip it).

thc202: BTW, the multi-selection can break the sorting of the URLs (per #2623), so, we might need to sort them before exporting.

thc202:
Should we have another menu item? "Export (Selected) URLs to File..."? As is the user has to watch out for the selections when exporting.

kingthorin:
Ya I debated that a bit, you're probably right.

kingthorin:
To that end, do you have thoughts on refactoring? I could create a whole new class for the new menu item, but that seems like lot of duplication for little gain.

I spent a few minutes making the constructor take a string for the menu item name, and altering ExtensionHistory to create two menu items, but for some reason neither of them end up added (with no exception occurring).

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+		boolean html = file.getName().toLowerCase().endsWith(".htm") || file.getName().toLowerCase().endsWith(".html");
+        
+    	BufferedWriter fw = null;
+    	try {
+        	fw = new BufferedWriter(new FileWriter(file, false));
+
+			for (String item : aSet) {
+				item = html ? wrapHTML(item) : item;
+				fw.write(item + CRLF);
+//				fw.newLine();
```
kingthorin: Thoughts on this? I think using `newLine()` is probably better (since it considers platform).

src/lang/Messages.properties
```diff
@@ -1049,6 +1049,8 @@ enc2.tools.menu.encdec                 = Encode/Decode/Hash...
 enc2.tools.menu.encdec.mnemonic        = e
 
 exportUrls.popup = Export All URLs to File...
+exportUrls.popup.selected = Export Selected URLs to File...
+exportUrls.multiselect.warning = All URLs will be exported (ignoring the selections in the Sites Tree).
```
thc202: Can be removed?

src/org/zaproxy/zap/extension/history/PopupMenuExportSelectedURLs.java
```diff
@@ -0,0 +1,61 @@
+package org.zaproxy.zap.extension.history;
```
thc202: Missing license header.

src/org/zaproxy/zap/extension/history/PopupMenuExportSelectedURLs.java
```diff
@@ -0,0 +1,61 @@
...
+	protected void performAction() {
+	    JTree siteTree = extension.getView().getSiteTreePanel().getTreeSite();
+		writeURLs(buildOutputSet(siteTree.getSelectionPaths()));
+	}
+	
+	private SortedSet<String> buildOutputSet(TreePath[] startingPoints) {
+	    JTree siteTree = extension.getView().getSiteTreePanel().getTreeSite();
+	    ArrayList<TreePath> startingPts = new ArrayList<TreePath>();
+	    
+		if (startingPts.isEmpty()) {
```
thc202: Should check `startingPoints` instead (`ArrayUtils.isEmpty(startingPoints)`?).

src/org/zaproxy/zap/extension/history/PopupMenuExportSelectedURLs.java
```diff
@@ -0,0 +1,61 @@
...
+		for (TreePath aPath:startingPts) {
+			Enumeration<?> en = (((SiteNode) aPath.getLastPathComponent()).preorderEnumeration());
+			while (en.hasMoreElements()) {
+				SiteNode node = (SiteNode) en.nextElement();
+				if (node.isRoot()) {
+					continue;
+				}
+				HistoryReference nodeHR = node.getHistoryReference();
+				if (nodeHR != null
+						&& !HistoryReference.DEFAULT_TEMPORARY_HISTORY_TYPES.contains(nodeHR.getHistoryType())) {
```
thc202: Consider using `HistoryReference.getTemporaryTypes()` instead, as add-ons might add new types.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+		SortedSet<String> outputSet = new TreeSet<String>();
+		Enumeration<?> en = (startingPoint.preorderEnumeration());
+		while (en.hasMoreElements()) {
+			SiteNode node = (SiteNode) en.nextElement();
+			if (node.isRoot()) {
+				continue;
+			}
+			HistoryReference nodeHR = node.getHistoryReference();
+			if (nodeHR != null
+					&& !HistoryReference.DEFAULT_TEMPORARY_HISTORY_TYPES.contains(nodeHR.getHistoryType())) {
```
thc202: Consider using `HistoryReference.DgetTemporaryTypes()` instead, as add-ons might add new types.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+		return outputSet;
+	}
 
-		@SuppressWarnings("unchecked")
-		Enumeration<SiteNode> en = node.children();
-		while (en.hasMoreElements()) {
-			exportURLs(en.nextElement(), writer, html);
+	protected void writeURLs(SortedSet<String> aSet) {
+		File file = getOutputFile();
+		if (file == null) {
```
thc202: Maybe we should ask for the file before collecting the URLs? It would avoid collecting them if the user cancels the dialogue.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+    
+		boolean html = file.getName().toLowerCase().endsWith(".htm") || file.getName().toLowerCase().endsWith(".html");
+        
+    	BufferedWriter fw = null;
+    	try {
+        	fw = new BufferedWriter(new FileWriter(file, false));
+
+			for (String item : aSet) {
+				item = html ? wrapHTML(item) : item;
+				fw.write(item + CRLF);
```
thc202: It's no longer writing the method, missed this point :/ We could save the URL/method with the `Pair` class (or a specific class).

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
thc202: I'm fine with using that instead.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
kingthorin: I actually did that on purpose. I was trying to make it more true to what the user is presented in the UI. (Export URLs)

Will make sure I include this detail in the commit msg and help tweak.

thc202: OK :)

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
kingthorin: Hmmm ok i understand your efficiency concern but if I'm not passing the set in then `writeURLs` would need to call `buildOutputSet`. So from PopupMenuExportSelectedURLs can I override `buildOutputSet` with a method with a different signature? Or do I somehow have to make the signatures the same?

src/org/zaproxy/zap/extension/history/PopupMenuExportSelectedURLs.java
kingthorin: Yup I'll also make the indentation unified throughout the file.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+		sb.append("<a href=\"").append(input).append("\">");
+		sb.append(input).append("</a><br>");
+		
+		return sb.toString();
 	}
     
     private File getOutputFile() {
-
-	    JFileChooser chooser = new JFileChooser(extension.getModel().getOptionsParam().getUserDirectory());
+    	WritableFileChooser chooser = new WritableFileChooser(extension.getModel().getOptionsParam().getUserDirectory());
```
kingthorin: Will fix the indentation here to match surrounding code.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
```diff
@@ -22,131 +22,108 @@
...
+     
     private static Logger log = Logger.getLogger(PopupMenuExportURLs.class);
-
-    public PopupMenuExportURLs() {
-        super(Constant.messages.getString("exportUrls.popup"));
-
+    
+    public PopupMenuExportURLs(String menuItem) {
+        super(menuItem);
+
```
kingthorin: Will remove the whitespace additions here and above.

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
thc202: Yeah, we have to change some of the methods/signatures. How about:
```Java
private void performAction() {
    File file = getOutputFile();
    if (file == null) {
        return;
    }

    writeURLs(file, getOutputSet());
}

// Previously buildOutputSet
protected SortedSet<String> getOutputSet() {
    // ...
}

private void writeURLs(File file, SortedSet<String> aSet) {
    // ...
}
```

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
thc202: Thanks :+1:

src/org/zaproxy/zap/extension/history/PopupMenuExportURLs.java
kingthorin: Great, thanks!

thc202:
Thanks!

PR #3246, opened 2017-02-28
Specify listen port and wait for passive scanning

itsuugo:
This patch adds a new option to the zap-baseline.py script that permits
to specify the listening port of zap, thus we can know in advance the
port that we can use as a passive proxy.

Also, the patch adds a loop to wait for the time specified with the -m
option, that permits us to run external external software using the zap
proxy (i.e selenium) and obtain the analysis of the passive scanning. In
current implementation the script finish at the same time that the
spider, that makes impossible to control the time of the passive
scanning.

### DIFF
--- a/build/docker/zap-baseline.py
+++ b/build/docker/zap-baseline.py
@@ -87,6 +87,8 @@ def usage():
     print ('    -x report_xml     file to write the full ZAP XML report')
     print ('    -a                include the alpha passive scan rules as well')
     print ('    -d                show debug messages')
+    print ('    -P                specify listen port')
+    print ('    -D                delay in seconds to wait for passive scanning ')
     print ('    -i                default rules not in the config file to INFO')
     print ('    -j                use the Ajax spider in addition to the traditional one')
     print ('    -l level          minimum level to show: PASS, IGNORE, INFO, WARN or FAIL, use with -s to hide example URLs')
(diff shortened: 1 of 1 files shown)


itsuugo:
The rationale to not using other parameter was because I thought that it was the use case for the -m option. Anyway, I guess that adding a new parameter is better, will send a new PR with that approach.

thc202:
Not against a new PR but this one could be updated instead.

itsuugo:
I had to change the options to -P -D because -p -d were used for other features.

build/docker/zap-baseline.py
```diff
@@ -263,13 +269,14 @@ def main(argv):
...
         sys.exit(3)
 
-  # Choose a random 'ephemeral' port and check its available
-  while True:
-    port = randint(32768, 61000)
-    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    if not (sock.connect_ex(('127.0.0.1', port)) == 0):
-      # Its free:)
-      break
+  # Choose a random 'ephemeral' port and check its available if it wasn't specified with -p option
```
thc202: Comment should be updated to use the new arg.

build/docker/zap-baseline.py
```diff
@@ -422,6 +429,11 @@ def main(argv):
       logging.debug ('Ajax Spider complete')
 
     # Wait for passive scanning to complete
+    if (delay):
+      start_scan = datetime.now()
+      while ( (datetime.now() - start_scan).seconds < delay ):
+        time.sleep(5)
+        logging.debug ('Waiting for passive scan to finish: ' + str(delay - (datetime.now() - start_scan).seconds) + ' seconds left')
```
thc202: This might be misleading since the the passive scanner might not have finished. I'd suggest using other phrase to indicate that's delaying the passive scan check.

thc202:
Code changes looks good, the commit message could be amended to refer to the new args added (still mentioning `-m`).

thc202:
Thanks!

itsuugo:
ping @psiinon

PR #3234, opened 2017-02-22
Use TreeCellRenderers to prevent HTML injection issues

psiinon:
Fixes #3224

PR #3271, opened 2017-03-07
Prevent HTML injection in the alerts pane

psiinon:
Related to #3234 but affecting JLabels instead of the tree

### DIFF
--- a/src/org/zaproxy/zap/utils/ZapLabel.java
+++ b/src/org/zaproxy/zap/utils/ZapLabel.java
@@ -1,0 +1,55 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ *  
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0 
+ *   
+ * Unless required by applicable law or agreed to in writing, software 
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/utils/ZapLabel.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.swing.JTextArea;
+import javax.swing.UIManager;
+
+/**
+ * An alternative to JLabel which disables HTML and supports wrapping.
+ * Use this class instead of JLabel where the text could include HTML that you
+ * do not want to render.
+ * @author psiinon
+ *
```
thc202: Could you add the since tag?

src/org/zaproxy/zap/utils/ZapLabel.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public class ZapLabel extends JTextArea {
+
+	private static final long serialVersionUID = 1L;
+
+	public ZapLabel() {
+		this.setEditable(false);
```
thc202: Could be removed (already set in `setDefaults`)?

src/org/zaproxy/zap/utils/ZapLabel.java
```diff
@@ -0,0 +1,56 @@
...
+		setDefaults();
+	}
+
+	public ZapLabel(String text) {
+		super(text);
+		setDefaults();
+	}
+
+	private void setDefaults() {
+		this.putClientProperty("html.disable", Boolean.TRUE);
```
thc202: This could be removed (not needed for this UI component).

psiinon: Updated to fix those issues

thc202:
Thanks!

PR #3276, opened 2017-03-09
Require nonces on shortcuts (unless disabled)

psiinon:
Also added shortcuts to the Web UI for testing, but only if suitable options are set for automatically supplying nonces

### DIFF
--- a/src/org/zaproxy/zap/extension/api/API.java
+++ b/src/org/zaproxy/zap/extension/api/API.java
@@ -239,8 +239,14 @@ public class API {
 			JSONObject params = getParams(requestHeader.getURI().getEscapedQuery());
 
 			if (shortcutImpl != null) {
+				if (!getOptionsParamApi().isDisableKey() && !getOptionsParamApi().isNoKeyForSafeOps()) {
+					if ( ! this.hasValidKey(requestHeader, params)) {
+						throw new ApiException(ApiException.Type.BAD_API_KEY);
+					}
+				}
 				msg = shortcutImpl.handleShortcut(msg);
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/api/API.java
```diff
@@ -239,8 +239,14 @@ public boolean handleApiRequest (HttpRequestHeader requestHeader, HttpInputStrea
 			JSONObject params = getParams(requestHeader.getURI().getEscapedQuery());
 
 			if (shortcutImpl != null) {
+				if (!getOptionsParamApi().isDisableKey() && !getOptionsParamApi().isNoKeyForSafeOps()) {
+					if ( ! this.hasValidKey(requestHeader, params)) {
+						throw new ApiException(ApiException.Type.BAD_API_KEY);
```
kingthorin: Isn't a break or return needed here? It seems the shortcut will still be processed after throwing the exception unless I'm missing something...doesn't the current flow count on the exception being unhandled?

psiinon: I dont _think_ so - if there should be a key but isnt then an exception is thrown. Otherwise `msg = shortcutImpl.handleShortcut(msg);` is called. If an exception is thrown then its handled later. Or am I missing something?

kingthorin: It's 5am....I could be wrong...

kingthorin: Nope you're right, ignore me..l

Issue #2595, opened 2016-06-16
Minor Edit Alert Enhancements


davewichers:
When you right click on an alert, there are MANY items in that menu. I was looking for some kind of triage function (like set as false positive, or whatever), but apparently you have to hit edit first and then you can do that. To make it easier to find this functionality, I'd suggest:

1) Add an Edit button to the alert panel on the right. That will make this feature more prominent, and hopefully more likely a user will see it/try it/and find the change confidence functionality. Maybe in the top right corner of this panel?
2) Seems like if you did that, you could simply change the mode of the alert panel to edit mode (rather than popping up a separate Edit Alert dialog), but that's not a big deal. (If you do this, both the edit button, and right click-->edit on the alert should work the same).
3) Please left justify the CWE ID and WASC ID values in the Edit Alert dialog so they are on the left just like the data in all the other rows. When I bring up the Edit Alert dialog you can't see those values at all so they look blank. I have to expand the width of that dialog to see those values way off to the right.
4) When you save an Edit, that alert should remain selected in the Alerts tree. Now when you save, nothing is selected anymore and the alert info is not shown to the right, which is surprising/disconcerting.
5) Does it make sense to add a new 'Reported' choice for Confidence? i.e., Confirmed means 'yes its vulnerable'. Reported could be used to mean its been JIRAed or however such issues are reported to the app owner.

kingthorin:
- [x] Edit button
- [ ] Edit mode
- [x]  Left justify CWE and WASC in edit dialog
- [x] Additional Confidence choice.
  * I don't believe this last one is necessary. If user's want to tag things Reported they can append/postpend the title, edit one of the other fields, etc. Plus the "Reported" confidence would only be visible/applicable to a single saved session.

davewichers:
I'd still like to see the Reported capability be built in. @psiinon (Simon) - do you have an opinion? Also, not sure why it appears that you aren't re-selecting the edited alert. Do you plan not to? Or just haven't done it yet?

kingthorin:
Yup, that's a todo list.

kingthorin:
Keep in mind "Confirmed" confidence is user set, never code set.

davewichers:
Understand. I want to be able to do 2 things, 1) Confirm the vulnerability
is real, then 2) Set that the vulnerability has been reported somehow,
e.g., in JIRA, in a report, however.

kingthorin:
[...]

PR #3285, opened 2017-03-13
Edit Alert Enhancements

kingthorin:
Rework AlertViewPanel so that labels are in a fixed space, left align CWE and WASC in edit mode.
AlertPanel, implement an Edit button in the toolbar.
ExtensionAlert add code to updateAlertInTree so that the alert is re-selected after editing.
AlertAddDialog remove unnecessary call to ExtensionAlert.updateAlertInTree and ExtensionAlert.displayAlert.
Messages.properties add tooltip entry for new button.

This contributes to zaproxy/zaproxy#2595

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
+++ b/src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
@@ -40,6 +40,8 @@ import javax.swing.JComboBox;
 import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
+import javax.swing.JSpinner;
+import javax.swing.JTextField;
 import javax.swing.border.TitledBorder;
 
 import org.apache.log4j.Logger;
@@ -62,13 +64,15 @@ public class AlertViewPanel extends AbstractPanel {
 	private static final long serialVersionUID = 1L;
 	private static final Logger logger = Logger.getLogger(AlertViewPanel.class);
 	
+	private static final Insets DEFAULT_INSETS = new Insets(1,1,1,1);
+	
 	private JScrollPane defaultPane = null;
 	private JScrollPane alertPane = null;
 	private ZapTextArea defaultOutput = null;
 	private JPanel alertDisplay = null;
 	private CardLayout cardLayout = null;
 	
-	private ZapTextField alertUrl = null;
+	private ZapLabel alertUrl = null;
 	private ZapLabel alertName = null;
 	private JLabel alertRisk = null;
 	private JLabel alertConfidence = null;
@@ -93,6 +97,16 @@ public class AlertViewPanel extends AbstractPanel {
 	private DefaultComboBoxModel<String> paramListModel = null;
 	private ZapNumberSpinner alertEditCweId = null;
 	private ZapNumberSpinner alertEditWascId = null;
+	
+	private JLabel attackLabel;
+	private JLabel cweidLabel;
+	private JLabel evidenceLabel;
+	private JLabel otherLabel;
+	private JLabel confidenceLabel;
+	private JLabel riskLabel;
+	private JLabel sourceLabel;
+	private JLabel urlLabel;
+	private JLabel wascidLabel;
 	
 	private boolean editable = false;
 	private Alert originalAlert = null;
@@ -221,7 +235,15 @@ public class AlertViewPanel extends AbstractPanel {
 			
 			alertEditEvidence = new ZapTextField();
 			alertEditCweId = new ZapNumberSpinner();
+			if (alertEditCweId.getEditor() instanceof JSpinner.DefaultEditor) {
+				((JSpinner.DefaultEditor) alertEditCweId.getEditor()).getTextField()
+						.setHorizontalAlignment(JTextField.LEFT);
+			}
 			alertEditWascId = new ZapNumberSpinner();
+			if (alertEditWascId.getEditor() instanceof JSpinner.DefaultEditor) {
+				((JSpinner.DefaultEditor) alertEditWascId.getEditor()).getTextField()
+						.setHorizontalAlignment(JTextField.LEFT);
+			}
 
 			// Read only ones
 			alertName = new ZapLabel();
@@ -236,7 +258,7 @@ public class AlertViewPanel extends AbstractPanel {
 			alertWascId = new JLabel();
 			alertSource = new JLabel();
 
-			alertUrl = new ZapTextField();
+			alertUrl = new ZapLabel();
 			
 			alertDescription = createZapTextArea();
 			JScrollPane descSp = createJScrollPane(Constant.messages.getString("alert.label.desc"));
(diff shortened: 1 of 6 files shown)


kingthorin:
I will put together a corresponding help PR as well.

src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -291,134 +313,61 @@ public void keyPressed(java.awt.event.KeyEvent evt) {
...
-						LayoutHelper.getGBC(1, gbcRow, 1, 0.75D, new Insets(1,1,1,1)));
-				gbcRow++;
-
-				alertDisplay.add(new JLabel(Constant.messages.getString("alert.label.source")),
-						LayoutHelper.getGBC(0, gbcRow, 1, 0.25D, new Insets(1,1,1,1)));
-				alertDisplay.add(alertSource, 
-						LayoutHelper.getGBC(1, gbcRow, 1, 0.75D, new Insets(1,1,1,1)));
-				gbcRow++;
-			}
+			alertUrl.setEditable(false);
```
thc202: Since it's no longer editable (before was but the changes were not applied...) we could use a ZapLabel instead.

src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -291,134 +313,61 @@ public void keyPressed(java.awt.event.KeyEvent evt) {
...
+			gbcRow++;
+			alertDisplay.add(getCweidLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
+			alertDisplay.add(editable ? alertEditCweId : alertCweId,
+					LayoutHelper.getGBC(1, gbcRow, 1, 1, DEFAULT_INSETS));
+			gbcRow++;
+			alertDisplay.add(getWascidLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
+			alertDisplay.add(editable ? alertEditWascId : alertWascId,
+					LayoutHelper.getGBC(1, gbcRow, 1, 1, DEFAULT_INSETS));
+			gbcRow++;
+			alertDisplay.add(getSourceLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
```
thc202: This should probably be shown only when not editing (otherwise we should set the `alertSource` always to not leave the field empty).

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -376,6 +380,23 @@ private void writeAlertToDB(Alert alert, HistoryReference ref) throws HttpMalfor
 
     }
 
+    private TreePath findAlertInTree(Alert alert) {
```
thc202: We could use the methods from the alert model to find the node (which uses binary search), although that can be changed later.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -376,6 +380,23 @@ private void writeAlertToDB(Alert alert, HistoryReference ref) throws HttpMalfor
...
 
+    private TreePath findAlertInTree(Alert alert) {
+    	Enumeration<?> en = ((((AlertNode) this.getTreeModel().getRoot()).preorderEnumeration()));
+		while (en.hasMoreElements()) {
+			AlertNode node = (AlertNode) en.nextElement();
+			if (node.isRoot()) {
+				continue;
+			}
+			if (alert.compareTo(node.getUserObject())==0) {
+				if (node.getChildCount()==1) {
```
thc202: Ideally it should continue until it has no child nodes (we should only select leafs).

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -410,6 +431,13 @@ public void updateAlertInTree(Alert originalAlert, Alert alert) {
     		this.getFilteredTreeModel().updatePath(originalAlert, alert);
     	}
     	this.recalcAlerts();
+    	
+    	if (View.isInitialised()) {
+    		JTree alertTree = this.getAlertPanel().getTreeAlert();
+    		alertTree.setExpandsSelectedPaths(true);
```
thc202: This can be done once (in `getTreeAlert()`).

src/org/zaproxy/zap/extension/history/AlertAddDialog.java
```diff
@@ -205,9 +205,6 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
 
 								// Update alert display
 								extAlert.displayAlert(alert);
```
thc202: This can also be deleted, the selection already displays the alert.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: I don't mind changing it, but you'll have to be a bit more specific.

kingthorin:
That should address most of the feedback. I did noticed one issue though. When a user edits a vulnerability (branch) a particular leaf is selected, once the edit is done the appropriate leaf is selected, but the display of the branch isn't updated. Any ideas on that?

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: Something like:
```Java

    public AlertNode getAlertNode(Alert alert) {
        AlertNode parent = (AlertNode) getRoot();
        int risk = alert.getRisk();
        if (alert.getConfidence() == Alert.CONFIDENCE_FALSE_POSITIVE) {
            // Special case!
            risk = -1;
        }

        AlertNode needle = new AlertNode(risk, alert.getName());
        needle.setUserObject(alert);
        int idx = parent.findIndex(needle);
        if (idx < 0) {
            return null;
        }
        parent = parent.getChildAt(idx);
        idx = parent.findIndex(needle);
        if (idx < 0) {
            return null;
        }
        return parent.getChildAt(idx);
    }
```
in `AlertTreeModel`.

thc202:
Looks like a bug, could you provide the exact steps to reproduce that?

src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -291,134 +313,61 @@ public void keyPressed(java.awt.event.KeyEvent evt) {
...
+			alertDisplay.add(getCweidLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
+			alertDisplay.add(editable ? alertEditCweId : alertCweId,
+					LayoutHelper.getGBC(1, gbcRow, 1, 1, DEFAULT_INSETS));
+			gbcRow++;
+			alertDisplay.add(getWascidLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
+			alertDisplay.add(editable ? alertEditWascId : alertWascId,
+					LayoutHelper.getGBC(1, gbcRow, 1, 1, DEFAULT_INSETS));
+			gbcRow++;
+			//Source doesn't need to be displayed in edit mode
+			alertDisplay.add(editable ? new JLabel() : getSourceLabel(), LayoutHelper.getGBC(0, gbcRow, 1, 0, DEFAULT_INSETS));
```
thc202: Could use a if and "skip" the two statements instead?

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -410,6 +431,12 @@ public void updateAlertInTree(Alert originalAlert, Alert alert) {
     		this.getFilteredTreeModel().updatePath(originalAlert, alert);
     	}
     	this.recalcAlerts();
+    	
+    	if (View.isInitialised()) {
+    		JTree alertTree = this.getAlertPanel().getTreeAlert();
+    		alertTree.setSelectionPath(findAlertInTree(alert));
+    		alertTree.scrollPathToVisible(findAlertInTree(alert));
```
thc202: `findAlertInTree` should be called only once.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Oh I thought you meant there was some simple call already there that I was missing, I didn't expect you to write code for me....thanks!

kingthorin:
I tried to recreate in 2.5.0 this afternoon and the issue didn't happen. So either there's a problem with the saved session I was working on or it was something introduced by this change. I'll have more of a look in the morning, if I can't sort it I'll post steps.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: No problem (there are methods using the binary search but do little more than what it's needed here).

thc202:
OK, it might well be related to other changes in develop branch.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Hmmmm in the two saved sessions I tested with this always returned zero. I unfortunately need to head to work and can't do anything further with this until tonight.

I can probably just use `findLeafNodeForAlert(root, alert)` which is already in AlertTreeModel.

kingthorin:
Ok so here's the scoop on this one. Select a branch, note which URL it refers to. Goto the Alert in the tree for that URL (leaf). Do an edit, and save it. Go back to the branch and not that its display wasn't updated.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: Index 0 might be fine, if selecting the first node. How many nodes did you have? (I tested the method and it seemed to work fine, selected the expected node always).

That method is not as efficient, it traverses the whole tree (in worst case).

thc202:
Thanks, will try that.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Will test further in a few hours when I'm home. Tks @thc202

kingthorin: Hmmm ok the updated alert always becomes the 0th item in the branch, so that makes sense. However, now I'm back to the item being selected but the selection not being displayed :(

thc202:
Thanks!

PR #3308, opened 2017-03-20
Log the number of messages sent by the scanners

thc202:
Change HostProcess to also log the number of messages sent by the
scanner when it's logged that it finished. The information is useful to
know more or less how the scanner performed.

 ---
From talks in IRC channel.


psiinon:
Approved but now conflicting.

thc202:
Done.

PR #3310, opened 2017-03-21
Fix baseline for new API security checks

psiinon:
Also fixed bug in the 'delay' changes.


psiinon:
@kingthorin merging this now to expedite the weekly -  please still review and I'll fix it if you spot any problems

Issue #1644, opened 2015-06-04
Add response headers to Params tab


zapbot:
```
Is it possible to add a Headers tab (similar to Parameters tab) which contains all headers
found for a specific site/scope?
```

Original issue reported on code.google.com by `marabicudi` on 2015-05-19 12:36:09

kingthorin:
I've been thinking about this lately. I think it's ok to leave this as parameters (unless we get feedback to the contrary in the future). Headers are "parameters" of the traffic being proxied, though they technically aren't parameters unto themselves.

kingthorin:
Updated subject to more closely reflect the comments from May 2015.

kingthorin:
Oops, tks @thc202

thc202:
No problem.

PR #3317, opened 2017-03-23
Issue 1644 Add Headers To Params Tab

kingthorin:
Response header fields are now displayed in the Params tab.

ExtensionParams modified to handle response header "parameters".
HtmlParameter add "header" type to Type enum.
Messages.properties add params.type.header.
SiteParameters modified to handle response header "parameters".

Fixes zaproxy/zaproxy#1644

### DIFF
--- a/src/org/parosproxy/paros/network/HtmlParameter.java
+++ b/src/org/parosproxy/paros/network/HtmlParameter.java
@@ -26,7 +26,7 @@ import java.util.Set;
 
 public class HtmlParameter implements Comparable<HtmlParameter> {
 	public enum Type {
-		cookie, form, url
+		cookie, form, url, header
 	};
 
 	public enum Flags {
--- a/src/org/zaproxy/zap/extension/params/ExtensionParams.java
+++ b/src/org/zaproxy/zap/extension/params/ExtensionParams.java
@@ -22,6 +22,7 @@ package org.zaproxy.zap.extension.params;
 import java.awt.EventQueue;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.HashMap;
@@ -46,6 +47,8 @@ import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HtmlParameter;
+import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpHeaderField;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.extension.anticsrf.ExtensionAntiCSRF;
 import org.zaproxy.zap.extension.help.ExtensionHelp;
@@ -349,6 +352,17 @@ public class ExtensionParams extends ExtensionAdaptor
 			persist(sps.addParam(site, iter.next(), msg));
 		}
 
+		// Header "Parameters"
+		List<HttpHeaderField> headersList = msg.getResponseHeader().getHeaders();
+		List<String> setCookieHeaders = Arrays.asList(HttpHeader.SET_COOKIE.toLowerCase(), HttpHeader.SET_COOKIE2.toLowerCase());
+		for (HttpHeaderField hdrField:headersList) {
+			if (setCookieHeaders.contains(hdrField.getName().toLowerCase())) {
+				continue;
+			}
+			HtmlParameter headerParam = new HtmlParameter(HtmlParameter.Type.header, hdrField.getName(), hdrField.getValue());
+			persist(sps.addParam(site, headerParam, msg));
+		}
+		
 		// TODO Only do if response URL different to request? 
 		// URL Parameters
 		/*
@@ -385,6 +399,8 @@ public class ExtensionParams extends ExtensionAdaptor
 					extSearch.search("[?&]" + item.getName() + "=.*", ExtensionSearch.Type.URL, true, false);
 				} else if (HtmlParameter.Type.cookie.equals(item.getType())) {
 						extSearch.search(/*".*" + */item.getName() + "=.*", ExtensionSearch.Type.Header, true, false);
+				} else if (HtmlParameter.Type.header.equals(item.getType())) {
+					extSearch.search(item.getName() + ":.*", ExtensionSearch.Type.Header, true, false);
 				} else {
 					// FORM
 					extSearch.search(/*".*" + */item.getName() + "=.*", ExtensionSearch.Type.Request, true, false);
--- a/src/org/zaproxy/zap/extension/params/SiteParameters.java
+++ b/src/org/zaproxy/zap/extension/params/SiteParameters.java
@@ -39,6 +39,7 @@ public class SiteParameters {
 	private Map<String, HtmlParameterStats> cookieParams = new HashMap<>();
 	private Map<String, HtmlParameterStats> urlParams = new HashMap<>();
 	private Map<String, HtmlParameterStats> formParams = new HashMap<>();
+	private Map<String, HtmlParameterStats> headerParams = new HashMap<>();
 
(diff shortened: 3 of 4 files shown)


kingthorin:
I'll add the fixes entry after review.

kingthorin:
I'll put together the necessary help PR separately:
https://github.com/zaproxy/zap-core-help/wiki/HelpUiTabsParams

src/org/parosproxy/paros/network/HtmlParameter.java
```diff
@@ -26,7 +26,7 @@
 
 public class HtmlParameter implements Comparable<HtmlParameter> {
 	public enum Type {
-		cookie, form, url
+		cookie, form, url, header
```
thc202: Worth adding the new value all upper case.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -349,6 +350,15 @@ public boolean onHttpResponseReceive(HttpMessage msg) {
 			persist(sps.addParam(site, iter.next(), msg));
 		}
 
+		// Header "Parameters"
+		List<HttpHeaderField> headersList = msg.getResponseHeader().getHeaders();
+		Iterator<HttpHeaderField> hdrIter = headersList.iterator();
+		while (hdrIter.hasNext()) {
```
thc202: Could use an enhanced for (avoids the explicit hasNext/next calls).

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -349,6 +350,15 @@ public boolean onHttpResponseReceive(HttpMessage msg) {
 			persist(sps.addParam(site, iter.next(), msg));
 		}
 
+		// Header "Parameters"
+		List<HttpHeaderField> headersList = msg.getResponseHeader().getHeaders();
+		Iterator<HttpHeaderField> hdrIter = headersList.iterator();
+		while (hdrIter.hasNext()) {
+			HttpHeaderField hdrField = hdrIter.next();
+			HtmlParameter headerParam = new HtmlParameter(HtmlParameter.Type.header, hdrField.getName(), hdrField.getValue());
+			sps.addParam(site, headerParam, msg);
```
thc202: It's not persisting the headers (`SiteParameters` would have to be changed too, though).

thc202:
- Should Set-Cookie headers be ignored? (Cookies are already being handled.) Still useful to show separately as a header?
 - Should request headers be processed too?
 - The method `ExtensionParams.searchForSelectedParam` should be updated to handle the new type (or, the pop up menu item changed to not be shown for the new type).
 - The class `SiteParameters` was not updated to handle the new type, was that on purpose? (That means the values will not be available through the ZAP API, nor read from the session or have cumulative stats.)

There's a typo in the commit message in "parameters".

src/org/parosproxy/paros/network/HtmlParameter.java
kingthorin: Sure, was just trying to remain consistent

thc202: OK, lets leave it all lower case then (if/when we break compatibility we can change all of them).

kingthorin:
Yes set-Cookie could probably be ignored.  Will tweak for that. 
I'm by no means set on only doing response headers. However through experience that's where the majority of interesting things are. It also seemed like that's what the original ticket was after... 
Resting site params, search, and commit msg will do.

thc202:
OK, sounds good to me (we can always add that later if needed).

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -349,6 +352,17 @@ public boolean onHttpResponseReceive(HttpMessage msg) {
 			persist(sps.addParam(site, iter.next(), msg));
 		}
 
+		// Header "Parameters"
+		List<HttpHeaderField> headersList = msg.getResponseHeader().getHeaders();
+		String[] setCookieHeaders = {HttpHeader.SET_COOKIE, HttpHeader.SET_COOKIE2};
```
thc202: Could be converted to List here, also, we should lower case these headers and the one being checked (to ensure that we do a case insensitive check).

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -349,6 +352,17 @@ public boolean onHttpResponseReceive(HttpMessage msg) {
...
 
+		// Header "Parameters"
+		List<HttpHeaderField> headersList = msg.getResponseHeader().getHeaders();
+		String[] setCookieHeaders = {HttpHeader.SET_COOKIE, HttpHeader.SET_COOKIE2};
+		for (HttpHeaderField hdrField:headersList) {
+			if (Arrays.asList(setCookieHeaders).contains(hdrField.getName())) {
+				continue;
+			}
+			HtmlParameter headerParam = new HtmlParameter(HtmlParameter.Type.header, hdrField.getName(), hdrField.getValue());
+			sps.addParam(site, headerParam, msg);
```
thc202: The headers are not being persisted.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -385,6 +399,8 @@ protected void searchForSelectedParam() {
 					extSearch.search("[?&]" + item.getName() + "=.*", ExtensionSearch.Type.URL, true, false);
 				} else if (HtmlParameter.Type.cookie.equals(item.getType())) {
 						extSearch.search(/*".*" + */item.getName() + "=.*", ExtensionSearch.Type.Header, true, false);
+				} else if (HtmlParameter.Type.header.equals(item.getType())) {
+					extSearch.search(/*".*" + */item.getName() + ":.*", ExtensionSearch.Type.Header, true, false);
```
thc202: The commented code could be removed (the header name starts the line).

src/org/zaproxy/zap/extension/params/SiteParameters.java
```diff
@@ -60,7 +61,7 @@ public void setSite(String site) {
 	 * @since 2.5.0
 	 */
 	public boolean hasParams() {
-		return !cookieParams.isEmpty() || !urlParams.isEmpty() || !formParams.isEmpty();
+		return !cookieParams.isEmpty() || !urlParams.isEmpty() || !formParams.isEmpty() || !headerParams.isEmpty();
```
thc202: The JavaDoc should be updated to also mention the headers (or, remove the type enumeration?).

src/org/zaproxy/zap/extension/params/SiteParameters.java
```diff
@@ -93,21 +96,21 @@ public HtmlParameterStats getParam(HtmlParameter.Type type, String name) {
 				params.add(param);
 			}
 			break;
+		case header:
+			for (HtmlParameterStats param : this.headerParams.values()) {
+				params.add(param);
```
thc202: This could also use the `addAll` (same for above cases), as done in `getParams()`.

thc202:
Does it worth ignoring other common headers (e.g. `Content-Length`, `Date`, `Keep-Alive`)?

kingthorin:
Probably but I think if we're going to do something like that it should be declared in `HttpHeader` as a list of those common headers. So that maintenance is simple and it can be leveraged consistently elsewhere. Should I include such a change in this PR or separately and then do another PR for this to leverage it?

src/org/zaproxy/zap/extension/params/SiteParameters.java
kingthorin: Would it be better if it used `getParams().isEmpty()`?

thc202: I'd say no, that requires extra processing and a new list.

thc202:
Just accessing them, it doesn't, but if you spider them, it might be different. I'm fine leaving as is though.

thc202:
Thanks!

kingthorin:
Updated commit message with fix tag.

Issue #3261, opened 2017-03-03
Client Cert PKCS#11 - UI/Exception Handling


kingthorin:
If you enable client certs then goto the PKCS#11 tab and hit "Add to Keystore" you get an exception.

It should be handled better, or perhaps the button shouldn't be active until there's something to add.

> 31916 [AWT-EventQueue-0] WARN org.parosproxy.paros.extension.option.OptionsCertificatePanel  - Couldn't add key from ActivIdentity - Windows (x86)
> java.lang.reflect.InvocationTargetException
> 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
> 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source)
> 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)
> 	at java.lang.reflect.Constructor.newInstance(Unknown Source)
> 	at ch.csnc.extension.httpclient.SSLContextManager.createInstance(SSLContextManager.java:378)
> 	at ch.csnc.extension.httpclient.SSLContextManager.createPKCS11Provider(SSLContextManager.java:366)
> 	at ch.csnc.extension.httpclient.SSLContextManager.initPKCS11(SSLContextManager.java:352)
> 	at org.parosproxy.paros.extension.option.OptionsCertificatePanel.addPkcs11ButtonActionPerformed(OptionsCertificatePanel.java:606)
> 	at org.parosproxy.paros.extension.option.OptionsCertificatePanel.access$7(OptionsCertificatePanel.java:569)
> 	at org.parosproxy.paros.extension.option.OptionsCertificatePanel$8.actionPerformed(OptionsCertificatePanel.java:373)
> 	at javax.swing.AbstractButton.fireActionPerformed(Unknown Source)
> 	at javax.swing.AbstractButton$Handler.actionPerformed(Unknown Source)
> 	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
> 	at javax.swing.DefaultButtonModel.setPressed(Unknown Source)
> 	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(Unknown Source)
> 	at java.awt.Component.processMouseEvent(Unknown Source)
> 	at javax.swing.JComponent.processMouseEvent(Unknown Source)
> 	at java.awt.Component.processEvent(Unknown Source)
> 	at java.awt.Container.processEvent(Unknown Source)
[...]

PR #3323, opened 2017-03-24
Fix Issue 3261

kingthorin:
Refactor OptionsCertificatePanel slightly to display an error (but not
log an exception) if the PKSCS#11 PIN field is blank when the user
clicks "Add to Keystore". If a PIN is present then the existing behavior
is still observed (display error message and log exception).

Fixes zaproxy/zaproxy#3261

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
@@ -59,6 +59,7 @@ import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.model.OptionsParam;
 import org.parosproxy.paros.view.AbstractParamPanel;
+import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.utils.ZapTextField;
 
 import ch.csnc.extension.httpclient.PKCS11Configuration;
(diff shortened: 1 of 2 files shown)


src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
```diff
@@ -735,11 +730,23 @@ private void showErrorMessageSunPkcs11ProviderNotAvailable() {
 				Constant.messages.getString("options.cert.label.client.cert"), JOptionPane.ERROR_MESSAGE);
 	}
 
-	private void showGenericErrorMessagePkcs11CouldNotBeAdded() {
-		JOptionPane.showMessageDialog(null, new String[] {
-				Constant.messages.getString("options.cert.error"),
-				Constant.messages.getString("options.cert.error.password")}, 
-				Constant.messages.getString("options.cert.label.client.cert"), JOptionPane.ERROR_MESSAGE);
+	private void showGenericErrorMessagePkcs11CouldNotBeAdded(boolean isErrorLevel, String name, Exception e) {
```
thc202: The name could be changed to `logAndShow...`, to reflect the new behaviour.

thc202:
The commit was not amended.

thc202:
Thank you.

PR #3326, opened 2017-03-24
Allow add-ons to use Semantic Versioning

### LINKED ISSUE #3101: Allow add-ons to use Semantic Versioning
In line with #3090, add-ons should be able to use SemVer [1], at the moment they can only use an integer for the version which does not convey enough information for the add-ons to reliably depend on each other (there's a workaround in place but requires the add-ons to have two versions, the integer and the semantic version).
The change should not require the use of SemVer (current version scheme can still be used).

[1] http://semver.org/

thc202:
Change AddOn class to use semantic version by default, if an add-on is
still using integer version it's converted to a semantic version.
The integer/file version is also deprecated and should be used just the
semantic version (just referred as version).
The semver element used in the ZapAddOn.xml and ZapVersions.xml files is
deprecated in favour of using only the version element.
Add/update tests to assert the new behaviour.

Fix #3101 - Allow add-ons to use Semantic Versioning


psiinon:
Looks good but holding off on merging until after 2.6.0 is released, as per irc

thc202:
Right, moved the issue to 2.7.0 instead.

PR #3327, opened 2017-03-25
Ensure messages to be scanned have a response

### LINKED ISSUE #1964: ZAP should ensure response in base message before passing them to scanners
The base message in scanners might not contain a response (getBaseMsg().getResponseHeader().isEmpty() returns true), which will lead to FPs.

ZAP should ensure that the messages have a response before passing them to scanners. If not it should send a new request and use that as base message.

Noticed in pull request https://github.com/zaproxy/zap-extensions/pull/47 by @thc202 (see https://github.com/zaproxy/zap-extensions/pull/47#discussion_r41240218)

If the issue is fixed it would be recommandable to ensure that all active scanners won't do the same check (i.e. https://github.com/zaproxy/zap-extensions/blob/bbbba9bbaa5fd979f2a0b1e04cb33f5680775ac3/src/org/zaproxy/zap/extension/ascanrulesAlpha/TestUserAgent.java#L127)

thc202:
Change HostProcess to send the message if it doesn't have a response,
prior being scanned.

Fix #1964 - ZAP should ensure response in base message before passing
them to scanners

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/HostProcess.java
+++ b/src/org/parosproxy/paros/core/scanner/HostProcess.java
@@ -66,9 +66,11 @@
 // ZAP: 2016/12/20 Include the name of the user when logging the scan info
 // ZAP: 2017/03/20 Improve node enumeration in pre-scan phase.
 // ZAP: 2017/03/20 Log the number of messages sent by the scanners, when finished.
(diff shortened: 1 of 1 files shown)


psiinon:
As per irc chat - worth adding a comment to explain why the responses might be null.

thc202:
Comment added.

PR #3332, opened 2017-03-27
Replace version with 2.6.0

psiinon:


src/org/parosproxy/paros/CommandLine.java
```diff
@@ -79,7 +79,7 @@
      * Command line option to disable the default logging through standard output.
      * 
      * @see #isNoStdOutLog()
-     * @since TODO add version
+     * @since 1.6.0
```
thc202: There's a typo.

PR #3368, opened 2017-04-05
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


kingthorin:
I'm holding off on this one, I'm not sure that the user understands the point. I think they're confusing Strength and Threshold.

https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsScanpolicy

kingthorin:
Disregard, this does get things in the right order and align with the help docs.

Might be something we need to provide clarity on through crowdin so that others aren't making the same mistake(s).

PR #3370, opened 2017-04-05
Change API view "contextList" to return a list

### LINKED ISSUE #3353: ZAP API View:contextList method returns a string instead of JSON list
Issue:
ZAP API View:contextList method does not return a proper JSON list of values. It returns a string instead of a list. For example: If I have a context ZAP then list context method return value looks like this : {"contextList":"[Default Context, ZAP]"} but It should have been {"contextList":["Default Context", "ZAP"]}. Looks like instead of returning a list it is returning a string.
Note: 
Compare methods that returns scan policy names where the return type is a list instead of string.

thc202:
Change class ContextAPI to return the context names in a list, not all
names in a single string.

```JSON
{"contextList":["Context 1","Context 2"]}
```
and:
```XML
<contextList type="list">
    <contextName>Context 1</contextName>
    <contextName>Context 2</contextName>
</contextList>
```

instead of:
```JSON
{"contextList":"[Context 1, Context 2]"}
```
and:
```XML
<contextList>[Context 1, Context 2]</contextList>
```

This change breaks existing consumers that extract the names from the
string.

Fix #3353 - ZAP API View:contextList method returns a string instead of
JSON list

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ContextAPI.java
+++ b/src/org/zaproxy/zap/extension/api/ContextAPI.java
@@ -264,12 +264,11 @@ public class ContextAPI extends ApiImplementor {
         	result = new ApiResponseElement(name, getContext(params).getIncludeInContextRegexs().toString());
         	break;
         case VIEW_CONTEXT_LIST:
-        	List<String> contextNames = new ArrayList<>();
-            List<Context> contexts = Model.getSingleton().getSession().getContexts();
-            for (Context context : contexts){
-                contextNames.add(context.getName());
+            resultList = new ApiResponseList(name);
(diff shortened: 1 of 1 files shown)


psiinon:
Should we start a new 'unnumbered' release page in zap-core-help which we update with breaking changing like this?

thc202:
Yeah, will do that. (I thought on doing it but was going to be 2.7.0 ;)

psiinon:
It can be 2.7.0 and we can always change that it be decide to do an intermediate release :)

thc202:
OK, will use that version then.

thc202:
Pull request opened zaproxy/zap-core-help#88.

PR #3374, opened 2017-04-06
Adjust to more user favorable column widths

kingthorin:
Modify `ParamsPanel.setParamsTableColumnSizes()` loop through all headings and set the preferred width based on the header content. After the loop set the "value" column to be preferred much larger than it's header.

### DIFF
--- a/src/org/zaproxy/zap/extension/params/ParamsPanel.java
+++ b/src/org/zaproxy/zap/extension/params/ParamsPanel.java
@@ -20,6 +20,7 @@
 package org.zaproxy.zap.extension.params;
 
 import java.awt.CardLayout;
+import java.awt.Component;
 import java.awt.Event;
 import java.awt.GridBagConstraints;
 import java.awt.Toolkit;
@@ -32,6 +33,9 @@ import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JToolBar;
 import javax.swing.KeyStroke;
+import javax.swing.table.DefaultTableColumnModel;
+import javax.swing.table.TableCellRenderer;
+import javax.swing.table.TableColumn;
 
 import org.jdesktop.swingx.JXTable;
 import org.parosproxy.paros.Constant;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/params/ParamsPanel.java
```diff
@@ -204,23 +204,26 @@ private JScrollPane getJScrollPane() {
...
 		
-		paramsTable.getColumnModel().getColumn(2).setMinWidth(50);
-		paramsTable.getColumnModel().getColumn(2).setPreferredWidth(100);	// used
+		paramsTable.getColumnModel().getColumn(2).setMinWidth(40);
+		paramsTable.getColumnModel().getColumn(2).setPreferredWidth(40);	// used
 		
-		paramsTable.getColumnModel().getColumn(3).setMinWidth(50);
-		paramsTable.getColumnModel().getColumn(3).setPreferredWidth(100);	// numvals
+		paramsTable.getColumnModel().getColumn(3).setMinWidth(40);
+		paramsTable.getColumnModel().getColumn(3).setPreferredWidth(40);	// numvals
```
thc202: This hides part of the column name even if there's space, was that on purpose? (First time users would have to resize the column to know what values the column has.)

src/org/zaproxy/zap/extension/params/ParamsPanel.java
```diff
@@ -204,23 +204,26 @@ private JScrollPane getJScrollPane() {
...
 		
-		paramsTable.getColumnModel().getColumn(4).setMinWidth(50);
-		paramsTable.getColumnModel().getColumn(4).setPreferredWidth(100);	// % change
+		paramsTable.getColumnModel().getColumn(4).setMinWidth(40);
+		paramsTable.getColumnModel().getColumn(4).setPreferredWidth(40);	// % change
```
thc202: Ditto.

src/org/zaproxy/zap/extension/params/ParamsPanel.java
kingthorin: Tweaked

src/org/zaproxy/zap/extension/params/ParamsPanel.java
kingthorin: Tweaked

src/org/zaproxy/zap/extension/params/ParamsPanel.java
```diff
@@ -204,23 +204,26 @@ private JScrollPane getJScrollPane() {
...
+		paramsTable.getColumnModel().getColumn(0).setPreferredWidth(60);	// type
 		
 		paramsTable.getColumnModel().getColumn(1).setMinWidth(100);
-		paramsTable.getColumnModel().getColumn(1).setPreferredWidth(200);	// name
+		paramsTable.getColumnModel().getColumn(1).setPreferredWidth(100);	// name
 		
-		paramsTable.getColumnModel().getColumn(2).setMinWidth(50);
-		paramsTable.getColumnModel().getColumn(2).setPreferredWidth(100);	// used
+		paramsTable.getColumnModel().getColumn(2).setMinWidth(40);
+		paramsTable.getColumnModel().getColumn(2).setPreferredWidth(40);	// used
```
thc202: This column is now trimmed. For example, in French this column and "% Change" are trimmed as well.

May I suggest removing the hardcoded sizes and pack the columns? Didn't try it but based on other tables it should look fine in more languages, although we would have to take care to not use the whole value for the last column, given how bigger the values are.

kingthorin:
@thc202 yes the widths only really consider English, as with the originals. Can you point me to a packed example? I did try a few other variations that didn't depend on statically defining things but then the columns didn't seem to fill the width available, and it looked funny to have unused space.

kingthorin:
Hopefully that's a more i18n friendly implementation 👍

thc202:
Sorry for not answering in time, it could be used the methods of `JXTable` to pack the columns (which do a similar thing to what is being done here).

thc202:
Thank you!

kingthorin:
Should I look into changing it, or leave it?

thc202:
Leave it :)

kingthorin:
Ok @psiinon, look good to u?

psiinon:
yep :)

kingthorin:
Thanks

PR #3380, opened 2017-04-07
Fallback to default charset when loading scripts

### LINKED ISSUE #3377: ZAP can no longer load non UTF8 scripts
An example is https://github.com/mozilla/zest/blob/develop/examples/BodgeIt_Register_XSS.zst

Likely to be an unintended side effect of #2773

We should be able to cope with scripts that have been created with previous versions of ZAP :/

thc202:
Change ExtensionScript to fallback to (JVM) default charset if not able
to read with UTF-8. Also, add a method to load a script with a given
charset (e.g. Script Console add-on can prompt the user if neither UTF-8
nor default charset works).
Update the load API endpoint to allow to specify the charset.

Fix #3377 - ZAP can no longer load non UTF8 scripts


thc202:
Forgot the API changes, adding that...

thc202:
@kingthorin please, review it again, updated the API action `load` and its description.

thc202:
Thanks!

psiinon:

Issue #2615, opened 2016-06-23
Filtering ZAP Reports to Show High Risk Items


leonardwkerr:
This is a feature I wold like to see added into ZAP, filtering the alerts by the severity of the risk.

kingthorin:
This is possible in the reporting addons.

thc202:
Reopening, this is about the core view alerts. [1]

[1] https://groups.google.com/d/topic/zaproxy-users/Mxmq7wnYcx4/discussion

Issue #1620, opened 2015-06-04
Add endpoint to get number of alerts grouped by risk level


zapbot:
```
The current REST api exposes an endpoint to get the total number of alerts in the current
ZAP session.

http://zap/JSON/core/view/numberOfAlerts/?zapapiformat=JSON&baseurl=

 It would be great to modify the existing endpoint(for all formats) to return the number
of alerts grouped by risk level. A sample output could be

{
  "totalNumberOfAlerts": "170",
  "highRisk": "100",
  "mediumRisk": "20",
  "lowRisk": "20",
  "informational": "30"
}

```

Original issue reported on code.google.com by `diveshpremdeep` on 2015-05-01 00:37:39

PR #3386, opened 2017-04-09
Core View Alerts Risk Filter

kingthorin:
Added an optional parameter to the core view alerts API so that the results (alerts) returned can be limited to a specific risk.

Fixes zaproxy/zaproxy#2615

Added endpoint alertsSummary to fix zaproxy/zaproxy#1620 which outputs an overall summary, such as `{"High":0,"Low":132,"Medium":39,"Informational":153}`


src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -167,6 +167,7 @@
 	private static final String PARAM_IDX = "idx";
 	private static final String PARAM_IS_REGEX = "isRegex";
 	private static final String PARAM_IS_ENABLED = "isEnabled";
+	private static final String PARAM_RISK = "risk";
```
thc202: `riskID`? Otherwise we should mention in the description that's expected the ID (not the name).

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -822,14 +823,20 @@ public ApiResponse handleApiView(String name, JSONObject params)
...
+			
 			processAlerts(
 					this.getParam(params, PARAM_BASE_URL, (String) null), 
 					this.getParam(params, PARAM_START, -1), 
 					this.getParam(params, PARAM_COUNT, -1), new Processor<Alert>() {
 
 						@Override
 						public void process(Alert alert) {
-							resultList.addItem(alertToSet(alert));
+							if (risk == -1) {
```
thc202: This would be better if done in `processAlerts`, `numberOfAlerts` should also have the risk parameter.

kingthorin: It is in `processAlerts`...

thc202: This is being done in the `Processor` instantiated not `processAlerts` (which is around 1298/1305).

kingthorin: Ok get ya now...tks

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -838,9 +843,21 @@ public void process(Alert alert) {
...
+					this.getParam(params, PARAM_RISK, -1), counter);
 			
 			result = new ApiResponseElement(name, Integer.toString(counter.getCount()));
+		} else if (VIEW_ALERTS_SUMMARY.equals(name)) {
+			final ApiResponseList resultList = new ApiResponseList(name);
+			Map<String, Object> alertData = new HashMap<>();
+			int[] riskIdArray = {Alert.RISK_HIGH, Alert.RISK_MEDIUM, Alert.RISK_LOW, Alert.RISK_INFO};
+			CounterProcessor<Alert> counter = new CounterProcessor<>();
+			for (int riskId:riskIdArray) {
+				processAlerts(this.getParam(params, PARAM_BASE_URL, (String) null), -1, -1, riskId, counter);
```
thc202: The counter needs to be reset before each count, otherwise the following risks include the counts of the previous ones. But, probably better to have a new counter that counts all the risks at the same time, that would allow to just read the alerts once, instead of now that it reads all alerts four times (once for each risk).

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -838,9 +843,21 @@ public void process(Alert alert) {
...
+				alertData.put(Alert.MSG_RISK[riskId], String.valueOf(counter.getCount()));
```
thc202: The value could be left as integer (helpful for JSON format).

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -838,9 +843,21 @@ public void process(Alert alert) {
...
+			}
+			resultList.addItem(new ApiResponseSet<Object>("risk", alertData));
```
thc202: Wrapping in a list does not seem right (we have just one set). I'd suggest removing the list (this means that the JSON object needs some tweaks, there are other endpoints doing the same).

src/org/zaproxy/zap/extension/api/CoreAPI.java
kingthorin: Fair enough, I did kind of debate that while implementing.

src/org/zaproxy/zap/extension/api/CoreAPI.java
kingthorin: So `{"0":"145","1":"126","2":"38","3":"0"}` with no name?

src/org/zaproxy/zap/extension/api/CoreAPI.java
kingthorin: Ok resetting makes sense. I don't see how to pass multiple counters back and forth (in the current model).

thc202: You would have to create a new implementation of `Processor<Alert>`, that counts the risks individually based on the alerts passed in. Then just expose the risk counts (could return a `Map` ready for the `ApiResponseSet`).

src/org/zaproxy/zap/extension/api/CoreAPI.java
thc202: Same as is now but without the list:
```JSON
{"alertsSummary":{"High":0,"Low":4,"Medium":3,"Informational":0}}
```

That can be done, for example, with:
```Java
new ApiResponseSet<Integer>(name, alertData) {
    @Override
    public net.sf.json.JSON toJSON() {
        JSONObject jo = new JSONObject();
        jo.put(getName(), super.toJSON());
        return jo;
    };
};
```
(having this assign to the `result`.)

src/org/zaproxy/zap/extension/api/CoreAPI.java
kingthorin: That seems overly complicated for such a simple task. If it's only going to be a specific case of summarizing risk counts, let's just stick with resetting, less code same result.

kingthorin: But it would be less processing... will dig into it tonight I guess.

thc202: Right, "4 times" less processing (less GC pressure, less I/O ops...) and it's not more complicated, it just moves the logic around into the processor. For example (if you don't want to have the processor create the map for you) it could be:
```Java
final int[] risks = { 0, 0, 0, 0 };
Processor<Alert> counter = new Processor<Alert>() {

    @Override
    public void process(Alert alert) {
        risks[alert.getRisk()]++;
    }
};
processAlerts(this.getParam(params, PARAM_BASE_URL, (String) null), -1, -1, -1, counter);

Map<String, Integer> alertData = new HashMap<>();
for (int i = 0; i < risks.length; i++) {
    alertData.put(Alert.MSG_RISK[i], risks[i]);
}

// ...
```

kingthorin: Awesome @thc202!

src/org/parosproxy/paros/view/AbstractParamDialog.java
```diff
@@ -227,7 +227,7 @@ public void actionPerformed(ActionEvent e) {
      *
      * @return javax.swing.JSplitPane
      */
-    private AbstractParamContainerPanel getJSplitPane() {
+    public AbstractParamContainerPanel getJSplitPane() {
```
thc202: This slipped through.

src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -838,9 +843,26 @@ public void process(Alert alert) {
...
+				@Override
+				public void process(Alert alert) {
+					riskSummary[alert.getRisk()]++;
+				}
+			};
+			processAlerts(this.getParam(params, PARAM_BASE_URL, (String) null), -1, -1, -1, counter);
+
+			Map<String, Object> alertData = new HashMap<>();
+			for (int i = 0; i < riskSummary.length; i++) {
+				alertData.put(String.valueOf(i), riskSummary[i]);
```
thc202: Using the numbers as key breaks the XML format.

src/org/parosproxy/paros/view/AbstractParamDialog.java
kingthorin: Oh, thanks for catching that.

kingthorin:
Fixed

thc202:
Thank you!

PR #3387, opened 2017-04-10
Allow esc to close AbstractFormDialog

kingthorin:
Add handling to AbstractFormDialog so that they can be closed with the escape key.

(For example when adding Global Exclude URLs)

### DIFF
--- a/src/org/zaproxy/zap/view/AbstractFormDialog.java
+++ b/src/org/zaproxy/zap/view/AbstractFormDialog.java
@@ -27,13 +27,17 @@ import java.awt.Rectangle;
 import java.awt.Window;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.KeyEvent;
 
+import javax.swing.AbstractAction;
 import javax.swing.BorderFactory;
 import javax.swing.Box;
 import javax.swing.BoxLayout;
 import javax.swing.JButton;
+import javax.swing.JComponent;
 import javax.swing.JDialog;
 import javax.swing.JPanel;
+import javax.swing.KeyStroke;
 
 import org.parosproxy.paros.Constant;
 import org.zaproxy.zap.extension.help.ExtensionHelp;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
```diff
@@ -117,6 +123,18 @@ protected void init() {
         getEnabledCheckBox().setSelected(true);
         getDescTextField().setText("");
         token = null;
+        
+		// Handle escape key to close the dialog
```
thc202: Mixed indentation chars (class is using spaces).

src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
```diff
@@ -117,6 +123,18 @@ protected void init() {
...
+		KeyStroke escape = KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0, false);
```
thc202: This should be added in `getFieldsPane` (or the constructors), this method (`init`) is called each time the dialogue is shown. Unless it was just to apply to add dialogue? (In that case it should be added a flag to just do this once.)

src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
```diff
@@ -117,6 +123,18 @@ protected void init() {
...
+		AbstractAction escapeAction = new AbstractAction() {
+
+			@Override
+			public void actionPerformed(ActionEvent e) {
+				dispatchEvent(new WindowEvent(DialogAddToken.this, WindowEvent.WINDOW_CLOSING));
```
thc202: It should call `clearAndHide` (or `clearFields` and `dispose`) instead (same behaviour as Cancel button).

kingthorin:
If I move it to `AbstractFormDialog` where would I add it? I guess `getFieldsPanel()` is always overridden? (I'm confused by that declaration)

thc202:
If moved to base class it should be added to `initialise()`.

kingthorin:
Thanks!

src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
thc202: The `getFieldsPane` suggestion was only if this was going to be applied to (just) `DialogAddToken` and `DialogModifyToken`. Sorry for not being clear.

src/org/zaproxy/zap/view/AbstractFormDialog.java
```diff
@@ -84,6 +88,19 @@ private void initialise(boolean initView) {
...
 			initView();
 		}
+		
+		// Handle escape key to close the dialog
+		KeyStroke escape = KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0, false);
+		AbstractAction escapeAction = new AbstractAction() {
+
+			private static final long serialVersionUID = -8088438488574461587L;
+
+			public void actionPerformed(ActionEvent e) {
```
thc202: It's missing the override annotation.

kingthorin: Done

thc202:
Thank you.

PR #3398, opened 2017-04-12
Increase limit of global script variables' value

comibat:
It is unreasonable limitation of 1K for global script variable value. Let's make it 1M

### DIFF
--- a/src/org/zaproxy/zap/extension/script/ScriptVars.java
+++ b/src/org/zaproxy/zap/extension/script/ScriptVars.java
@@ -28,7 +28,7 @@ import javax.script.ScriptContext;
 public class ScriptVars {
 	
 	private static int MAX_KEY_SIZE = 30;
-	private static int MAX_VALUE_SIZE = 1024;
+	private static int MAX_VALUE_SIZE = 1024*1024;
 	private static int MAX_SCRIPT_VARS = 20;
 	private static int MAX_GLOBAL_VARS = 50;
 


kingthorin:
Unreasonable based on what criteria?

The potential memory footprint with this change goes 70KB to 70MB.

Also why multiply when you can just properly define the value?

psiinon:
Yeah, I'd be very interested to know what you (or anyone else) are doing that would require this change.
Its a very unexpected use of global script variables.

comibat:
@kingthorin 
Q: Unreasonable based on what criteria?
A: If you want to share some data between scripts, it's the only way (that i know of). In case that you want to share more than 70k, having in mind that everything else eats a lot more memory, how would you call this limitation?

Q: The potential memory footprint with this change goes 70KB to 70MB.
A: Yes, but that's small, comparing to memory consumption of everything else.

Q: Also why multiply when you can just properly define the value?
A: To make it obvious, but this i can change.

@psiinon : check the first answer.

psiinon:
I was just wondering what sort of data you'd want to share that would be of this size :)
We never hear enough about what people are using ZAP for, so I'd love to hear more about what the scripts are doing and what sort of data they need to pass. We might even be able to suggest better ways of passing such data, or alternatively plan better options for the future.

kingthorin:
@comibat Sorry if my first response came across short or something. I'm with @psiinon just trying to get a better feel for the use case(s), etc.

comibat:
Sorry guys, i was off for some time. Let me give a usecase example.
Active rules has two methods - scanNode and scan; the first one will be called for every request made, the second one for every param.
HTTP Sender has also two methods - sendingRequest and responseReceived. The second one will be called for >> all requests/responses sent/received by ZAP.
So, what if i want to put a payload into a param at one point in time, and later on detect it somewhere in response? I need to use the scan method for placing the payload and responseReceived for detecting it.
For that purpose, i need to share the information about the payload, and since those two don't have commonly shared variables (as far as i know), i have to use the global vars. If you have a lot of payloads that you want to track, then you need more space for that. 
This can be very useful for checking stored XSSs, for example.

thc202:
This change does not impact currently working scripts, so, existing use cases would still work as before. For new scripts I would expect the user to still take care to not create more variables or store more data than really needed (as before). IMHO it's better to give some margin and allow more use cases than require rebuild ZAP to support them.

comibat:
Thanks. Keep up the good work.

psiinon:
Thanks for the explanation. Maybe we should look at providing an interface for storing arbitrary data in the db - that would be more scalable.

comibat:
@psiinon That would be exactly what i'd love to see in future releases.

thc202:
@comibat how would you like to be credited?

And feel free to raise an issue for that one ;)

comibat:
@thc202 re credits - sure, why not. How that goes? Do you need something a name or ...? I'm not familiar with this. 
I'll post the issue.

thc202:
Yeah, a name and/or a handle.

Thanks.

comibat:
ok then, the name is Milic Vuletic

thc202:
Done in zaproxy/zap-core-help@be8a3e7a9d54f59d8ef28fcd7eda7bfb86d6c398. Thank you!

comibat:
np. Thank you :)

PR #3411, opened 2017-04-14
Clear cached certs when setting Root CA cert

thc202:
Change CachedSslCertifificateServiceImpl to clear the certificates cache
when a new Root CA certificate is set, to ensure that it's used the new
Root CA cert for issuing the (new) certs. Also, sync when the Root CA
cert is set to avoid concurrency issues.
Change ExtensionDynSSL to set the cert through the previously mentioned
class, to ensure the cache is cleared.


thc202:
This might explain the difficulties faced by the users when importing the Root CA cert to the browsers (i.e. even after importing the cert it still says that the issuer is unknown; usually, the issue was solved by restarting ZAP).

psiinon:

thc202:
OK, will address that now.

thc202:
Conflict resolved.

PR #3429, opened 2017-04-19
Allow to configure the parse response limit

### LINKED ISSUE #1313: Spider - Allow to configure the size limit of parseable responses
```
1. Access url to get large response body
2. ZAP cannot spider/scan if we got message 'Very large response body - switch views
to display...'




```

Original issue reported on code.google.com by `vincent.xuewei` on 2014-08-14 09:05:05

thc202:
Change Spider to allow to configure maximum size in bytes for a response
to be parsed.
Add tests to assert the expected behaviour of DefaultParseFilter. Also,
normalise its behaviour (null/empty handling of the HTTP message),
tweak the case of constants and other minor tweaks.

Fix #1313 - Spider - Allow to configure the size limit of parseable
responses


src/org/zaproxy/zap/spider/filters/DefaultParseFilter.java
```diff
@@ -39,57 +38,85 @@
...
+		if (responseMessage.getResponseBody().length() > params.getMaxParseSizeBytes()) {
 			if (log.isDebugEnabled()) {
 				log.debug("Resource too large: " + responseMessage.getRequestHeader().getURI());
 			}
 			return true;
 		}
 
 		// If it's a redirection, accept it, as the SpiderRedirectParser will process it
-		if (HttpStatusCode.isRedirection(responseMessage.getResponseHeader().getStatusCode()))
+		if (HttpStatusCode.isRedirection(responseMessage.getResponseHeader().getStatusCode())) {
```
kingthorin: Remove the comment as it no longer seems true?

thc202: It's still true, I didn't change the logic (just added the braces).

kingthorin: Ok

PR #3431, opened 2017-04-20
Removed unused variables and code which was commented out

rain0r:


rain0r:
@kingthorin  Done

src/org/parosproxy/paros/network/HttpMethodHelper.java
```diff
@@ -59,19 +59,13 @@
...
 	
 	private static final String CRLF	=	"\r\n";
 	private static final String LF		=	"\n";
 	private static final Pattern patternCRLF	= Pattern.compile("\\r\\n", Pattern.MULTILINE);
 	private static final Pattern patternLF		= Pattern.compile("\\n", Pattern.MULTILINE);
 	
-	private String mUserAgent = "";
-	
 	public void setUserAgent(String userAgent) {
-		mUserAgent = userAgent;
```
kingthorin: If the variable is being dropped and the assignment, why keep the method? (Just in case there's some 3rd party code we don't know about?)

If it's being left, add appropriate deprecation details?

rain0r:
@kingthorin 
> If the variable is being dropped and the assignment, why keep the method? (Just in case there's some 3rd party code we don't know about?)
> 
> If it's being left, add appropriate deprecation details?

You are right. Actually I threw that out in the first place. It did go missing when I reverted all the formatting changes. It's out now and the method `setUserAgent` was nowhere used.

kingthorin:
DOH, we can't win with this one. Looks like something happened with your interactive rebase, the PR now only contains that one change.....

src/org/parosproxy/paros/network/HttpMethodHelper.java
```diff
@@ -68,12 +68,6 @@
 	private static final Pattern patternCRLF	= Pattern.compile("\\r\\n", Pattern.MULTILINE);
 	private static final Pattern patternLF		= Pattern.compile("\\n", Pattern.MULTILINE);
 	
-	private String mUserAgent = "";
```
thc202: When restoring the other changes, please, add the "ZAP comment". [1]

rain0r:
Restored old changes and added "ZAP" comment.

thc202:
Thanks!

PR #3432, opened 2017-04-20
Fix Context Export Resource Messages

kingthorin:
Tweak resource keys in Messages.properties and resource usage in ContextExportDialog to actually refer to export not import as appropriate.


src/lang/Messages.properties
```diff
@@ -894,7 +895,6 @@ context.import.label.context	   = Context:
 context.import.label.dir		   = Directory:
 context.import.label.file		   = File:
 context.import.label.overwrite	   = Overwrite:
```
thc202: Worth fixing these as well? (Note that in the export dialogue the other resource keys, labels and error messages, are also using import.)

kingthorin: Sure, I'll take a look at it.

thc202: Thank you.

kingthorin:
:zap:

thc202:
Thank you!

PR #3438, opened 2017-04-24
Allow to set a tab scrollable in std dialogue

thc202:
Change StandardFieldsDialog to allow to set a tab/panel scrollable, to
properly show all the fields (that is, by giving an indication with the
scrollbar that a field is hidden, partially or totally).
Change SpiderDialog to set the Advanced tab as scrollable (needed
because the tab is showing more options).

---
Related to #3197.

### DIFF
--- a/src/org/zaproxy/zap/view/StandardFieldsDialog.java
+++ b/src/org/zaproxy/zap/view/StandardFieldsDialog.java
@@ -86,7 +86,23 @@ public abstract class StandardFieldsDialog extends AbstractDialog {
 	private JPanel mainPanel = null;
 	private List<JPanel> tabPanels = null;
 	private List<Integer> tabOffsets = null;
+
+	/**
+	 * The component used when showing the panels in tabs.
+	 * 
+	 * @see #getTabComponent(JPanel)
+	 * @see #panelToScrollPaneMap
+	 */
 	private JTabbedPane tabbedPane = null;
+
+	/**
+	 * The map that contains the {@code JScrollPane}s of the tabs ({@code JPanel}s) that were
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/view/StandardFieldsDialog.java
```diff
@@ -1656,18 +1691,122 @@ public void setTabsVisible(String[] tabLabels, boolean visible) {
...
+	 * <p>
+	 * By default this method returns a {@code JScrollPane} that has the vertical scrollbar shown as needed and the horizontal
+	 * scrollbar never shown.
+	 *
+	 * @param tabPanel the panel of the tab that should be scrollable, never {@code null}.
+	 * @return the JScrollPane
+	 * @since TODO add version
+	 */
+	protected JScrollPane createTabScrollable(JPanel tabPanel) {
+		return new JScrollPane(tabPanel, JScrollPane.VERTICAL_SCROLLBAR_AS_NEEDED, JScrollPane.HORIZONTAL_SCROLLBAR_NEVER);
```
kingthorin: Does it make sense to make them both "as needed" just in case something is added and not packed or sized appropriately?

thc202: Good question, it looked fine as is in the tabs I tested, I'll take a look again. I will also change the method to pass the label of the panel (extending classes can easily change the behaviour for each scrollable tab).

thc202: Changes done (both scrollbars are shown as needed and the label of the panel is provided when creating the JScrollPane).

PR #3472, opened 2017-05-02
Move alert related functionality to alert package

thc202:
Move AlertAddDialog and PopupMenuAlert to alert package and change
ExtensionAlert to use/add them. The ExtensionAlert is required for
adding and editing the alerts (while the dialogues would be shown if the
extension was disabled the changes done to the alerts would be lost
since they were not being persisted to the session).
Update related classes to use the new/moved methods/classes.
Rename a resource message key to be under alert keys.


psiinon:
LGTM but now has conflicts

PR #3487, opened 2017-05-04
Update Python API generator for Python 3

thc202:
Change PythonAPIGenerator to use the six library, for compatibility with
Python 2 and 3. Also, tweak it to be more compliant with PEP8.

For/from zaproxy/zap-api-python#4 - Fix for Python 3


thc202:
Added more tweaks so the generated code is more compliant with PEP8.

PR #3499, opened 2017-05-08
Allow to show more message data in Search tab

### LINKED ISSUE #3498: Minor Enhancement: Support Same Columns in Search Results as History Tab
I'd like to be able to see the same columns in the Search results tab as the History tab (except with the added Match column in the Search tab). I think the default visible columns are fine, but can you add the other columns as well, in the same order as the History tab and have them off by default? Then I can click on the little config widget on the far right and enable whichever columns I want.

I would think this would be really easy to implement. Thx

thc202:
Change Search tab to allow to show the same columns (hidden by default)
as in the History tab.

Fix #3498 - Support Same Columns in Search Results as History Tab


davewichers:
Wow that was fast! Thanks!

Issue #3481, opened 2017-05-04
Export Tables via UI


rain0r:
Make the table data from the tab Spider → Messages available via UI.
This can / should also be done for tables in general, e.g. for the table under Active Scan.

Currently, it is possible to select entries and copy / export them via CTRL + C or to export them via API endpoint `http://zap:8080/UI/core/view/messages/`

PR #3502, opened 2017-05-09
Issue 3481 Export Tables via GUI

kingthorin:
- org.zaproxy.zap.utils.TableExportButton > New component that extends JButton and accepts a JTable (JXTable) as a parameter.
- Messages.properties > Moved param specific export button resource keys and values to be used generically.
- ParamsPanel > Revised to use the new generic component.
- SpiderPanel > Add StateChange listener to tabbedpane, add functionality to show/enable (or hide/disable) the appropriate export button.
- ActiveScanPanel > Add export button.

Fixes zaproxy/zaproxy#3481

### DIFF
--- a/src/org/zaproxy/zap/extension/params/ParamsPanel.java
+++ b/src/org/zaproxy/zap/extension/params/ParamsPanel.java
@@ -24,21 +24,10 @@ import java.awt.Component;
 import java.awt.Event;
 import java.awt.GridBagConstraints;
 import java.awt.Toolkit;
-import java.awt.event.ActionEvent;
 import java.awt.event.KeyEvent;
-import java.io.File;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.List;
-
-import javax.swing.AbstractAction;
-import javax.swing.Box;
 import javax.swing.ImageIcon;
-import javax.swing.JButton;
 import javax.swing.JComboBox;
 import javax.swing.JLabel;
-import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JToolBar;
@@ -47,9 +36,6 @@ import javax.swing.table.DefaultTableColumnModel;
 import javax.swing.table.TableCellRenderer;
 import javax.swing.table.TableColumn;
 
-import org.apache.commons.csv.CSVFormat;
-import org.apache.commons.csv.CSVPrinter;
-import org.apache.log4j.Logger;
 import org.jdesktop.swingx.JXTable;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.extension.AbstractPanel;
@@ -57,15 +43,14 @@ import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HtmlParameter;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.utils.SortedComboBoxModel;
+import org.zaproxy.zap.utils.TableExportButton;
 import org.zaproxy.zap.view.ScanPanel;
-import org.zaproxy.zap.view.widgets.WritableFileChooser;
 
 public class ParamsPanel extends AbstractPanel{
 	
 	private static final long serialVersionUID = 1L;
 
 	public static final String PANEL_NAME = "params";
-    private Logger LOGGER = Logger.getLogger(ParamsPanel.class);
 	
 	private ExtensionParams extension = null;
 	private JPanel panelCommand = null;
@@ -79,6 +64,7 @@ public class ParamsPanel extends AbstractPanel{
 
 	private JXTable paramsTable = null;
 	private ParamsTableModel paramsModel = new ParamsTableModel();
+	private TableExportButton exportButton = null;
 	
     //private static Log log = LogFactory.getLog(ParamsPanel.class);
     
@@ -154,6 +140,7 @@ public class ParamsPanel extends AbstractPanel{
 			GridBagConstraints gridBagConstraints0 = new GridBagConstraints();
 			GridBagConstraints gridBagConstraints1 = new GridBagConstraints();
 			GridBagConstraints gridBagConstraints2 = new GridBagConstraints();
+			GridBagConstraints gridBagConstraints3 = new GridBagConstraints();
 			GridBagConstraints gridBagConstraintsx = new GridBagConstraints();
 
 			gridBagConstraints0.gridx = 0;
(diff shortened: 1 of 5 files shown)


src/org/zaproxy/zap/extension/params/ParamsPanel.java
```diff
@@ -154,7 +141,8 @@ private  void initialize() {
 			GridBagConstraints gridBagConstraints0 = new GridBagConstraints();
 			GridBagConstraints gridBagConstraints1 = new GridBagConstraints();
 			GridBagConstraints gridBagConstraints2 = new GridBagConstraints();
-			GridBagConstraints gridBagConstraintsx = new GridBagConstraints();
+			GridBagConstraints gridBagConstraints3 = new GridBagConstraints();
+			GridBagConstraints gridBagConstraintsx = new GridBagConstraints();
```
thc202: Some tabs were added at the end of the line.

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -160,6 +163,9 @@
 	/** The found count value label. */
 	private JLabel foundCountValueLabel;
 	
+	private static ZapTableExportButton exportUrlsButton = null;
```
thc202: Could be instance variable (same for the following variable).

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -160,6 +163,9 @@
 	/** The found count value label. */
 	private JLabel foundCountValueLabel;
 	
+	private static ZapTableExportButton exportUrlsButton = null;
+	private static ZapTableExportButton exportMessagesButton = null;
```
thc202: I'd suggest changing `ZapTableExportButton` to allow to set a table and use just one button.

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -172,6 +178,25 @@ public SpiderPanel(ExtensionSpider extension, SpiderParam spiderScanParam) {
 		super("spider", new ImageIcon(SpiderPanel.class.getResource("/resource/icon/16/spider.png")), extension);
 
 		tabbedPane = new JTabbedPane();
+		tabbedPane.addChangeListener(new ChangeListener() {
+			@Override
+			public void stateChanged(ChangeEvent e) {
+				switch(tabbedPane.getSelectedIndex()) {
+				case 0:
+					getUrlsExportButton().setEnabled(true);
```
thc202: This button should also be "re-enabled" when the Messages tab is hidden, otherwise, if you have the Messages tab selected and then hide it, the export button shown/enabled is the Messages one instead of URLs.

src/org/zaproxy/zap/utils/ZapTableExportButton.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.swing.JTable;
+
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.view.widgets.WritableFileChooser;
+
+public class ZapTableExportButton extends JButton {
```
thc202: - `Zap` prefix could be removed, it seems to imply that it applies only to `ZapTable`s.
 - It should be added the since tag.

src/org/zaproxy/zap/utils/ZapTableExportButton.java
```diff
@@ -0,0 +1,97 @@
...
+					try (CSVPrinter pw = new CSVPrinter(
+							Files.newBufferedWriter(chooser.getSelectedFile().toPath(), StandardCharsets.UTF_8),
+							CSVFormat.DEFAULT)) {
+						pw.printRecord(getColumnNames(table));
+						int rowCount = table.getRowCount();
+						int colCount = table.getColumnCount();
+						for (int row = 0; row < rowCount; row++) {
+							List<Object> valueOfRow = new ArrayList<Object>();
+							for (int col = 0; col < colCount; col++) {
+								valueOfRow.add(table.getValueAt(row, col).toString());
```
thc202: This needs to be changed to handle null values (it happens with Spider panel, when a entry has no flags).

thc202:
Any reason why the button was not added to tabs History, Search, and HTTP Sessions? ;)

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
kingthorin: I had it that way originally but the enablement and visibility were not properly consistent

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
kingthorin: I did start that way but had no luck, however that was before I switched away from instanced buttons.....will fight with both a bit more I guess

src/org/zaproxy/zap/utils/ZapTableExportButton.java
kingthorin: Ok

kingthorin:
Nope not specifically, I just tackled those mentioned in the ticket and the one I'd done a few weeks ago (Params). I guess I'll tackle this feedback and set of changes, then the others separately once this is good.

thc202:
OK, sounds good to me.

kingthorin:
Ok I think I've addressed all the feedback.

src/org/zaproxy/zap/utils/TableExportButton.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.swing.JTable;
+
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.view.widgets.WritableFileChooser;
+/**
+ * A {@code JButton} class to facilitate exporting tables to a file (such as CSV).
```
thc202: It could mention that the data is exported as shown in the table (e.g. filters, sorting, column order, column visibility affect the data exported).

src/org/zaproxy/zap/utils/TableExportButton.java
```diff
@@ -0,0 +1,134 @@
...
+					try (CSVPrinter pw = new CSVPrinter(
+							Files.newBufferedWriter(chooser.getSelectedFile().toPath(), StandardCharsets.UTF_8),
+							CSVFormat.DEFAULT)) {
+						pw.printRecord(getColumnNames(getTable()));
+						int rowCount = getTable().getRowCount();
+						int colCount = getTable().getColumnCount();
+						for (int row = 0; row < rowCount; row++) {
+							List<Object> valueOfRow = new ArrayList<Object>();
+							for (int col = 0; col < colCount; col++) {
+								String value = getTable().getValueAt(row, col) == null ? "" : getTable().getValueAt(row, col).toString();
```
thc202: Worth tweaking this to just call `getValueAt` once. Maybe delay the null check for the following statement?

src/org/zaproxy/zap/utils/TableExportButton.java
```diff
@@ -0,0 +1,134 @@
...
+										+ ex.getMessage());
+						LOGGER.error("Export Failed: " + ex.getMessage(), ex);
+					}
+				}
+			}
+		}));
+	}
+
+	/**
+	 * Gets a {@code List} of column names for the given table.
```
thc202: `of (visible) column names`?

src/org/zaproxy/zap/utils/TableExportButton.java
```diff
@@ -0,0 +1,134 @@
...
+	 * 
+	 * @param table the Table for which the column names should be returned
+	 * @return the {@code List} of column names
+	 */
+	private List<String> getColumnNames(JTable table) {
```
thc202: The parameter could be removed and be used the instance variable (or the getter).

thc202:
Thanks!

PR #3505, opened 2017-05-09
Correct ZAP home dir used in stable Docker image

thc202:
Change Dockerfile-stable to use the home dir of the main ZAP versions
(.ZAP) instead of the ones used in dev/weekly versions (.ZAP_D).


psiinon:
Should we regenerate the stable docker image?

thc202:
Yeah, worth doing that. (@zapbot scans might need it?)

PR #3523, opened 2017-05-11
Hide "Starting new HTTP connection" messages

psiinon:
Just a minor annoyance ;)


thc202:
@kingthorin yeah, that suppresses the messages.

PR #3529, opened 2017-05-12
Allow to add tags with Passive Rules

denniskniep:


src/org/zaproxy/zap/extension/pscan/scanner/ScriptsPassiveScanner.java
```diff
@@ -131,7 +131,11 @@ public void raiseAlert(int risk, int confidence, String name, String description
 
 		this.parent.raiseAlert(currentHRefId, alert);
 	}
-
+
```
thc202: This indentation change could be reverted.

thc202:
Thanks!

denniskniep:
No problem. But there is unfortunately another indentation change in the new file `src/scripts/templates/passive/Passive default template.js`
I will fix this quickly.

kingthorin:
Thanks @denniskniep

PR #3533, opened 2017-05-13
Additional Table Export Buttons

kingthorin:
As per https://github.com/zaproxy/zaproxy/pull/3502#issuecomment-300445402 add table export buttons to History, Search, and HTTP Sessions.


thc202:
Thanks!

PR #3540, opened 2017-05-15
Include extension when validating if file exists

thc202:
Change PopupMenuExportURLs and TableExportButton to set the extension
(if needed) before checking if the file exists, otherwise the overwrite
confirmation could be bypassed.


psiinon:

PR #3545, opened 2017-05-16
Delay success message when exporting

thc202:
Change TableExportButton to show the message after flushing/closing the
file, to ensure that all the data was already written when the message
is shown.


kingthorin:
So extensions can't use the generic component until a core jar update happens. Are we concerned about back porting the latest changes (like this) to the generic component to places like the fuzzer export?

thc202:
Yeah, worth doing that.

PR #3547, opened 2017-05-16
Add core views to obtain messages by ID

### LINKED ISSUE #3514: Allow to obtain multiple messages by ID
Hi,

I asked a question in the zap proxy google group and was instructed to raise an issue. The question was posted here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/zaproxy-users/BRUJPXWp6Z8/QJ8D_h-CAwAJ

The messages view of the core component enables you to retrieve a list of request-response pairs, but rather than giving you the ability to start at a certain id, it starts searching at the n-th request.

**Example:**

Assume I have a request-response history with the following identifiers: 1,2,5,6,10,20,21,22,23

I can inspect the details of the request-response pair with id=5 using /core/view/message/?id=5&theOtherParameters

I **cannot** get the details of request-response pairs 5,6 and 10 by doing this: /core/view/messages/?start=5&count=3&theOtherParameters. This will return the details of request-response pair 10,20 and 21. Where the request-response pair with id=10 is the 5th one in my history.

It doesn't make a whole lot of sense to me. Sure, if you only have a 100 requests you can kinda count where you want to start. But when you have several 1000 of requests it seems impossible to extract the set of requests you want.

Would you consider adding an endpoint that offers the functionality I'm looking for?

thc202:
Add core views to obtain multiple messages by ID (supports the standard
formats and HAR).

Fix #3514 - Allow to obtain multiple messages by ID


src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -1276,6 +1302,25 @@ public void process(HttpMessage msg) {
 		}
 	}
 
+	private static List<Integer> getIds(JSONObject params) throws ApiException {
+		List<Integer> listIds = new ArrayList<>();
+		for (String id : params.getString(PARAM_IDS).split(",")) {
+			String trimmedId = id.trim();
```
kingthorin: Any reason not to do the split and trim at the same time, which would avoid the extra var?

thc202: Unfortunately that's not possible, the split returns an array. I could remove the empty check and just call trim when invoking the parse?

Or, we could use streams:
```Java
listIds = Arrays.stream(params.getString(PARAM_IDS).split(","))
        .map(String::trim)
        .filter(e -> !e.isEmpty())
        .map(Integer::valueOf)
        .collect(Collectors.toList());
```

kingthorin: It's a pretty minor detail, it can be left as is. 

IMHO streams are less readable, maybe because I'm just not used to them yet...

thc202: Removed the empty check.

PR #3561, opened 2017-05-19
Package API scan script

psiinon:
First version, still various things I'd like to add to the script but I think its useful enough to release as is in order to get more feedback.
There is definitely common code with the baseline script, at some point I think the common code should be split out into a common file.


psiinon:
Oh, worth mentioning that I've excluded the Path Traversal scan rule for the moment - I think this should be re-added when the regex fix has been implemented and merged.

build/docker/policies/API-Minimal.policy
```diff
@@ -0,0 +1,238 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<configuration>
+<policy>API Minimal</policy>
+<scanner>
+<level>MEDIUM</level>
+<strength>MEDIUM</strength>
+</scanner>
+<plugins>
+<p6>
+<enabled>false</enabled>
```
thc202: Could be re-enabled per zaproxy/zap-extensions#898? (Seems to be ready for merge.)

EDIT: Moved the comment to the pull request.

build/docker/zap-api-scan.py
```diff
@@ -0,0 +1,685 @@
...
+      logging.debug ('Docker CID: ' + cid)
+      insp_output = subprocess.check_output(['docker', 'inspect', cid])
+      #logging.debug ('Docker Inspect: ' + insp_output)
+      insp_json = json.loads(insp_output)
+      zap_ip = str(insp_json[0]['NetworkSettings']['IPAddress'])
+      logging.debug ('Docker ZAP IP Addr: ' + zap_ip)
+
+      # Copy across the files that may not be in all of the docker images
+      try:
+        cp_to_docker(cid, 'Alert_on_HTTP_Response_Code_Errors.js', '/home/zap/.ZAP_D/scripts/scripts/httpsender/')
```
thc202: Getting some errors here `zaproxy/build/docker/Alert_on_HTTP_Response_Code_Errors.js: no such file or directory`.

build/docker/zap-api-scan.py
```diff
@@ -0,0 +1,685 @@
...
+        logging.debug ('Import SOAP File ' + target_file)
+        res = zap._request(zap.base + 'soap/action/importFile/', {'file': base_dir + target_file})
+        urls = zap.core.urls
+        if len(urls) > 0:
+          # Choose the first one - will be striping off the path below
+          target = urls[0]
+        else:
+          logging.error ('Failed to import any URLs')
+    
+    logging.info ('Number of Imported  URLs: ' + str(len(urls)))
```
thc202: There's an extra scape before URLs.

build/docker/zap-api-scan.py
```diff
@@ -0,0 +1,685 @@
...
+
+# Scan rules that are being addressed
+in_progress_issues = {}
+
+logging.basicConfig(level=logging.INFO, format='%(asctime)s %(message)s')
+# Hide "Starting new HTTP connection" messages
+logging.getLogger("requests").setLevel(logging.WARNING)
+
+def usage():
+    print ('Usage: zap-baseline.py -t <target> [options]')
```
Grunny: Might it be worth using [`argparse`](https://docs.python.org/2/library/argparse.html) instead of parsing arguments with getopt which makes it easy to generate help docs as well?

psiinon: Probably yes, but this is based on zap-baseline.py so that should be changed as well. My plan is to refactor these scripts to reduce the duplication, so that would be good to look at then.

build/docker/zap-api-scan.py
psiinon: Fixed those issues.

thc202:
Thanks!

kingthorin:
Thanks

build/docker/zap-api-scan.py
Grunny: :+1:

I think this should be changed to `'Usage: zap-api-scan.py` rather than still mentioning `zap-baseline.py`. Also, even with the manual help text, it might be easier to maintain if it was one multiline string, i.e.
```python
print("""Usage: zap-api-scan.py -t <target> [options]
    -t target         target API definition, currently only an OpenAPI URL, eg https://www.example.com/openapi.json
    -f format         either openapi or soap
...etc....
""")
```

Which could also be done in both.

psiinon: Oh yeah, thats a bug :) I'll submit a new PR for it...

PR #3580, opened 2017-05-25
Do not change extensions' enabled state

thc202:
Change OptionsExtensionPanel and OptionsExtensionTableModel to not
change the enabled state of the loaded extensions, the extensions'
enabled state should be changed only by bootstrap classes (during ZAP
initialisation or installation of an add-on).
Change Extension to document the methods isEnabled and setEnabled.


kingthorin:
So this shouldn't be user controlable?  Should the checkbox control in the options panel be disabled so that users don't think they can change it?

thc202:
It is, the user can still change the state, but it's not directly changing the state of the extensions themselves (i.e. the instances).

kingthorin:
Oh ok, gotcha

PR #3582, opened 2017-05-26
Extract table export action to a top level class

thc202:
Extract the action from TableExportButton so that it can be reused in
other UI components (e.g. pop up menus).


thc202:
Yeah, that's possible already, one could use the methods `JButton.setIcon`/`setText` if the defaults are not OK. I'll add some constructors to allow to specify the text/icon (avoids creating/getting the default icon/text unnecessarily).

thc202:
Constructors added.

Issue #3593, opened 2017-05-30
ZAP active scanner in daemon mode skips all scripts


ani2796:
I'm currently running ZAP in daemon mode by building it in IntelliJ. However, whenever I make calls to the ZAP API for the active scanner, it gives me the following errors:

121531 [ZAP-ProxyThread-51] INFO org.parosproxy.paros.core.scanner.Scanner  - scanner started
121604 [Thread-7] INFO org.parosproxy.paros.core.scanner.HostProcess  - Scanning 62 node(s) from http://localhost:8080
121615 [Thread-7] INFO org.parosproxy.paros.core.scanner.HostProcess  - start host http://localhost:8080 | ScriptsActiveScanner strength MEDIUM threshold MEDIUM
121637 [Thread-7] INFO org.parosproxy.paros.core.scanner.HostProcess  - skipped plugin [no scripts enabled] http://localhost:8080 | ScriptsActiveScanner in 0.022s with 0 message(s) sent
121637 [Thread-7] INFO org.parosproxy.paros.core.scanner.HostProcess  - completed host http://localhost:8080 in 0.057s
121637 [Thread-6] INFO org.parosproxy.paros.core.scanner.Scanner  - scanner completed in 0.106s

The spider, however, is working just fine. I need to run the spider before I get the active scanner to even start. Otherwise the active scanner tells me that the URL scan tree is not found.

psiinon:
The Active Scanner can only act on urls that it knows about. If you just run it without exploring your site then there will be no urls and therefore it will exit.
You either need to run one of the spiders or explore you site in another way, eg by proxying regression tests or manual requests.

ani2796:
Thanks for the prompt reply! But the active scanner didn't work even after I had run the spider. That was the trace I had copied in my original post.

Also, is there a possible way to start ZAProxy server using java code without using system commands?

thc202:
When built from source ZAP does not include any add-ons by default, you either need to install the "Active scanner rules" add-ons (e.g. [1]) or build/copy the add-ons to src/plugin directory (and build ZAP again).

It's possible to do that as long ZAP jar and its dependencies are on the classpath (you can call ZAP main method, which starts ZAP), although it's probably better to use a JVM just for ZAP independently of your program.

[1] https://github.com/zaproxy/zap-core-help/wiki/HelpCmdline

ani2796:
[...]

PR #3595, opened 2017-05-30
Print args and error msg when failed to parse args

thc202:
Change ZAP class to print the arguments and the exception's message when
failed to parse the arguments.
Change CommandLine to accept null arguments, remove print statements
(already done by ZAP class, when printing the exception's message) and
add a message to an exception thrown.
Add tests to CommandLineUnitTest to assert expected behaviour.

Per comments in #3593.

### DIFF
--- a/src/org/zaproxy/zap/ZAP.java
+++ b/src/org/zaproxy/zap/ZAP.java
@@ -21,6 +21,7 @@ package org.zaproxy.zap;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Enumeration;
 import java.util.Locale;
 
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/ZAP.java
```diff
@@ -87,7 +88,9 @@ public static void main(String[] args) throws Exception {
         } catch (final Exception e) {
         	// Cant use the CommandLine help here as the 
         	// i18n messages wont have been loaded
-            System.out.println("Failed due to invalid parameters. Use '-h' for more details.");
+            System.out.println("Failed due to invalid parameters: " + Arrays.toString(args));
+            e.printStackTrace();
```
psiinon: Does the printStackTrace() actually provide anything useful? I cant see how it will help users (or us) and I think it might end up just looking really messy. Stack traces are for working out where an error is, this is just the user supplying an invalid cmdline option..

thc202: Probably not given that Exception is being thrown for parsing errors. I'll change that to just print the exception message.

thc202: Pull request updated.

PR #3606, opened 2017-06-01
Add more constructors to custom pop up menus

thc202:
Change ExtensionPopupMenuItem and ExtensionPopupMenuItemMessageContainer
to expose more constructors (e.g. easier to customise the pop up menus
when used from scripts).


src/org/parosproxy/paros/extension/ExtensionPopupMenuItem.java
```diff
@@ -54,12 +57,43 @@
 	private boolean precedeWithSeparator = false;
 	private boolean succeedWithSeparator = false;
 
+    /**
+     * Constructs a {@code ExtensionPopupMenuItem} with no text nor icon.
```
kingthorin: It's a minor detail but these should all be "an" since when rendered the next word will start with an E (vowel).

thc202: Right, fixed. Thank you.

Issue #3555, opened 2017-05-18
Changing Session Name, doesn't Update the Window Name


egerod:
When updating the session name after opening the Session Properties view through a context, the window name is not updated to reflect the new session name.

Steps to reproduce: Open ZAP, double click default context to open Session Properties, navigate to general tab, change the session name, click ok.
Expected Outcome: Window name contains the new session name.
Outcome: The window still show the old session name.
This was done with ZAP 2.6 and on windows 10.

Finally, if the session properties are open through the File drop down menu or the key bind, the window name is updated as intended.

kingthorin:
I agree that the behavior should be consistent but is it misleading if the window title is updated and the session is still not persisted (saved)?

thc202:
I don't think it would be misleading, we show both, the session name and file name (for the latter when persisted).

kingthorin:
Ok, perfect.

kingthorin:
http://logs.glob.uno/?c=mozilla%23websectools&s=19+May+2017&e=19+May+2017

kingthorin:
Ok so just to make sure I'm taking the right direction on this:

1. `setTitle()` will be removed from MenuFileControl.
2. The functionality from MenuFileControl will be added to `MainFrame.setTitle()`.
3. `ExtensionUiUtils.sessionChangedEventHandler()` should call `View.getSingleton().getMainFrame().setTitle("")` [or something like that].
4. Remove title handling from `Control`.

That's what I took from the chat log in the end. I'm left unclear as to whether or not changes were still necessary in `SessionGeneralPanel`  (in saveParam perhaps?)

PR #3607, opened 2017-06-02
Issue 3555 Session Name Window Title Consistency

kingthorin:
Control - Removed setTitle calls
MenuFileControl - Remove local setTitle functionality
MainFrame - Deprecate old setTitle(String). Add new setTitle(Session).
SessionGeneralPanel - saveParam now calls MainFrame's setTitle().
ExtensionUiUtils - sessionChangedEventHandler(Session) now calls MainFrame's setTitle().

Fixes zaproxy/zaproxy#3555

### DIFF
--- a/src/org/parosproxy/paros/control/MenuFileControl.java
+++ b/src/org/parosproxy/paros/control/MenuFileControl.java
@@ -44,6 +44,7 @@
 // ZAP: 2015/12/14 Log exception and internationalise error message
 // ZAP: 2016/10/26 Issue 1952: Do not allow Contexts with same name
 // ZAP: 2017/02/25 Issue 2618: Let the user select the name for snapshots
+// ZAP: 2017/06/01 Issue 3555: setTitle() functionality moved in order to ensure consistent application
 
 package org.parosproxy.paros.control;
  
@@ -189,7 +190,6 @@ public class MenuFileControl implements SessionListener {
             public void sessionSaved(final Exception e) {
                 if (EventQueue.isDispatchThread()) {
                     if (e == null) {
-                        setTitle();
                         view.getSiteTreePanel().getTreeSite().setModel(model.getSession().getSiteTree());
                     } else {
                         view.showWarningDialog(Constant.messages.getString("menu.file.newSession.error"));
@@ -246,7 +246,6 @@ public class MenuFileControl implements SessionListener {
                     }
 
                     view.getSiteTreePanel().getTreeSite().setModel(model.getSession().getSiteTree());
-                    setTitle();
 
                     if (waitMessageDialog != null) {
                         waitMessageDialog.setVisible(false);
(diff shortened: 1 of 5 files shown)


src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -511,12 +493,7 @@ public void sessionOpened(File file, Exception e) {
 
     @Override
     public void sessionSaved(Exception e) {
-        if (e == null) {
-            // ZAP: Set the title consistently
-            setTitle();
-            //File file = new File(model.getSession().getFileName());
-            //view.getMainFrame().setTitle(file.getName().replaceAll(".session\\z","") + " - " + Constant.PROGRAM_NAME);
-        } else {
+        if (e != null) {
```
kingthorin: Can someone validate that this works out the same? From my PoV is does, but a second opinion is warranted.

src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -644,12 +646,34 @@ public void setShowTabNames(boolean showTabNames) {
...
+	 * Sets the title of the main window.
+	 * <p>
 	 * The actual title set is the given {@code title} followed by the program name and version.
 	 * 
 	 * @see Constant#PROGRAM_NAME
 	 * @see Constant#PROGRAM_VERSION
+	 * @deprecated as of TODO add version
 	 */
-	@Override
+	@Override @Deprecated
```
kingthorin: Should these be on separate lines?

Should I make the body of this function simply be a call to the new function?

kingthorin:
I know I asked a bunch of questions in the issue last night, but I just decided to jump in with both feet and move ahead...hopefully it's good or close 🤞

src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -644,12 +646,34 @@ public void setShowTabNames(boolean showTabNames) {
...
 	 * Sets the title of the main window.
 	 * <p>
+	 * The actual title set is the current session name, then the filename of the session (if persisted), 
+	 * followed by the program name and version.
+	 * 
+	 * @see Constant#PROGRAM_NAME
+	 * @see Constant#PROGRAM_VERSION
+	 * @since TODO add version
+	 */
+	public void setTitle() {
```
thc202: This could accept a Session instead of getting it from the Model?

src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -644,12 +646,34 @@ public void setShowTabNames(boolean showTabNames) {
...
+	}
+	
+	/**
+	 * Sets the title of the main window.
+	 * <p>
 	 * The actual title set is the given {@code title} followed by the program name and version.
 	 * 
 	 * @see Constant#PROGRAM_NAME
 	 * @see Constant#PROGRAM_VERSION
+	 * @deprecated as of TODO add version
```
thc202: It should mention what's the replacement.

src/org/parosproxy/paros/view/MainFrame.java
thc202: They usually are.

Yes, that's better.

src/org/parosproxy/paros/control/MenuFileControl.java
thc202: It does.

kingthorin:
I think that addresses all the feedback.

thc202:
Thank you!

src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -644,12 +647,37 @@ public void setShowTabNames(boolean showTabNames) {
...
+	 * 
+	 * @param session the {@code Session} from which the window title is being built
+	 * 
+	 * @see Constant#PROGRAM_NAME
+	 * @see Constant#PROGRAM_VERSION
+	 * @since TODO add version
+	 */
+	public void setTitle(Session session) {
+		StringBuilder strBuilder = new StringBuilder();
+		strBuilder.append(session.getSessionName()).append(" - ");
```
thc202: Missed this one, sorry! We need to check that the session is not null (it might be null when discarding a session).

kingthorin:
Ok, will submit another PR for that.

thc202:
Thank you.

kingthorin:
https://github.com/zaproxy/zaproxy/pull/3622

PR #3622, opened 2017-06-06
Ensure Session Not null When Setting Window Title

kingthorin:
See
https://github.com/zaproxy/zaproxy/pull/3607#pullrequestreview-42257730

### DIFF
--- a/src/org/parosproxy/paros/view/MainFrame.java
+++ b/src/org/parosproxy/paros/view/MainFrame.java
@@ -658,10 +658,12 @@ public class MainFrame extends AbstractFrame {
 	 */
 	public void setTitle(Session session) {
 		StringBuilder strBuilder = new StringBuilder();
-		strBuilder.append(session.getSessionName()).append(" - ");
-		if (!session.isNewState()) {
-			File file = new File(session.getFileName());
-			strBuilder.append(file.getName().replaceAll(".session\\z", "")).append(" - ");
+		if (session != null) {
+			strBuilder.append(session.getSessionName()).append(" - ");
+			if (!session.isNewState()) {
+				File file = new File(session.getFileName());
+				strBuilder.append(file.getName().replaceAll(".session\\z", "")).append(" - ");
+			}
 		}
 		strBuilder.append(Constant.PROGRAM_NAME).append(' ').append(Constant.PROGRAM_VERSION);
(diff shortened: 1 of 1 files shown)


src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -658,7 +658,8 @@ public void setShowTabNames(boolean showTabNames) {
 	 */
 	public void setTitle(Session session) {
 		StringBuilder strBuilder = new StringBuilder();
-		strBuilder.append(session.getSessionName()).append(" - ");
+		String sessionName = session == null ? "" : session.getSessionName();
+		strBuilder.append(sessionName).append(" - ");
 		if (!session.isNewState()) {
```
thc202: It also requires the null check here (could have an if wrapping all "session statements").

kingthorin: I should trust myself more. I did it this way, then I tweaked it to account for null on this too. Then I read the doc on isNewState and thought 'otherwise false' likely already handles null session and undid my revised change...

Will update this PR when I get into the office....On my way now.

thc202: OK.

kingthorin:
I had to push this blindly, I'm unable to run things this morning for some reason (environmental I'm sure). Logically I think it's fine, but know that I haven't been able to run it.

src/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -658,10 +658,13 @@ public void setShowTabNames(boolean showTabNames) {
 	 */
 	public void setTitle(Session session) {
 		StringBuilder strBuilder = new StringBuilder();
-		strBuilder.append(session.getSessionName()).append(" - ");
-		if (!session.isNewState()) {
-			File file = new File(session.getFileName());
-			strBuilder.append(file.getName().replaceAll(".session\\z", "")).append(" - ");
+		System.out.println("fail");
```
thc202: Other than this, looks good to me :)

kingthorin: My morning just keeps getting better and better.....

kingthorin:
There we go 👍

thc202:
Thank you!

PR #3632, opened 2017-06-07
Do not close session when changing its properties

thc202:
Change how the changes to the session's properties (e.g. name,
description) are persisted/notified to other classes, now it just
persist those properties and notifies accordingly instead of saving
(closing and re-opening) the session, which stopped all the active
actions (e.g. scans) without warning.

### DIFF
--- a/src/org/parosproxy/paros/view/SessionGeneralPanel.java
+++ b/src/org/parosproxy/paros/view/SessionGeneralPanel.java
@@ -27,6 +27,7 @@
 // ZAP: 2015/02/10 Issue 1528: Support user defined font size
 // ZAP: 2017/01/09 Remove method no longer needed.
 // ZAP: 2017/06/01 Issue 3555: setTitle() functionality moved in order to ensure consistent application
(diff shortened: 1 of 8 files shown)


src/org/parosproxy/paros/view/SessionGeneralPanel.java
```diff
@@ -49,6 +51,8 @@
 public class SessionGeneralPanel extends AbstractParamPanel {
 
 	private static final long serialVersionUID = -8337361808959321380L;
+	
+	private static final Logger logger = Logger.getLogger(SessionGeneralPanel.class);
```
kingthorin: CONSTant?

kingthorin: CONSTant?

thc202: Probably, I'll change that.

thc202: Changed.

PR #3637, opened 2017-06-08
Do not traverse Sites tree while active scanning

### LINKED ISSUE #3539: Enhancement: Collect messages to scan before active scanning
Prior to starting an active scan collect the message(s) to be scanned. So that any changes to those messages during the scan (such as an analyst or pen tester working with the same messages via the proxy etc) do not impact the messages being active scanned.

thc202:
Change HostProcess to collect all the messages that should be scanned
before starting the actual scan, instead of traversing the Sites tree
each time a node/message needs to be scanned, to prevent changes done to
the Sites tree from affecting the active scan (e.g. by new messages
added to the Sites tree).
Change HistoryReference to allow to keep the HttpMessage in memory when
constructing it (it's now immediately used when scanning).

Fix #3539 - Collect messages to scan before active scanning


thc202:
Updated based on review comments received directly from @psiinon. Changed HostProcess to use the new constructor :/ and add a method to `HistoryReference` to clear the `HttpMessage`.

Issue #9, opened 2015-06-04
No Way to Specify Custom 404


zapbot:
```
What steps will reproduce the problem?
1. Scan something that returns custom a 302 to a custom 404 error message.

What is the expected output? What do you see instead?
Lots of False positives, especially when it comes to default files.

What version of the product are you using? On what operating system?
1.0.0. OS doesn't matter.

Please provide any additional information below.

```

> Original issue reported on code.google.com by `logicalgambit` on 2010-10-06 16:13:56
---

Part 1
======
- [x] Initial implementation - https://github.com/zaproxy/zaproxy/pull/3638
- [x] Analyzer fallback - https://github.com/zaproxy/zaproxy/pull/6194
- [x] UnitTests - https://github.com/zaproxy/zaproxy/pull/6232, https://github.com/zaproxy/zaproxy/pull/6245, https://github.com/zaproxy/zaproxy/pull/6310
- [x] Update User Guide / zap-core-help - https://github.com/zaproxy/zap-core-help/pull/335

Part 2
======
- [x] pscan rules adapted (When targeting 2.10.0) - https://github.com/zaproxy/zap-extensions/pull/2616, https://github.com/zaproxy/zap-extensions/pull/2627 https://github.com/zaproxy/zap-extensions/pull/2630
- [x] ascan rules adapted (When targeting 2.10.0) - https://github.com/zaproxy/zap-extensions/pull/2639 https://github.com/zaproxy/zap-extensions/pull/2642

yhawke:
Dear all, I suffered the same trouble yesterday in a pentest when a nice programmer decided to use 200 and a page sying "this not exists" instead of 404 error... all the forced, svn, git, backup file plugins went crazy and the db became so big that everything has crashed!!!!
I don't know if someone is working on this, and I want to give my 2 cents.
Probably a good approach is using context, people could define custom 404 and 500 or other error messages using http response codes and regexes as you said before, then crawlers, active and passive plugins should use specific methods to check for a specific Response code (I don't agree for the isError, I think we can be more generic with isResponseCode(int Response code) then the implementation should apply the specific custom control or apply it to the Response Code directly).
But I think that a big improvement could be also to alert the user that a custom 404 exists. I saw that lot of scanners perform a check for an unexisting file and then, if the answer is 200, they alert the user that a custom error exists and the scanner hasn't been configured for this.
[...]

PR #3638, opened 2017-06-09
Initial CustomPages Implementation

kingthorin:
All the "gutts" of the solution for other components to take advantage of.

Add support for custom error page definitions associated with a Context. Expose those details and functionality to Passive and Active scanners (rules). Assume that if a context with custom pages is setup the user wants those details to be leveraged. Default behavior is to check custom pages and fallback to status code checking (based on the applicable custom page definition type and related status codes).

See zaproxy/zaproxy#9

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -66,6 +66,7 @@
 // ZAP: 2019/10/21 Use and expose Alert builder.
 // ZAP: 2020/01/27 Extracted code from sendAndReceive method into regenerateAntiCsrfToken method in
 // ExtensionAntiCSRF.
+// ZAP: 2020/09/23 Add functionality for custom error pages handling (Issue 9).
 package org.parosproxy.paros.core.scanner;
 
 import java.io.IOException;
@@ -87,6 +88,7 @@ import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.control.AddOn;
 import org.zaproxy.zap.extension.anticsrf.ExtensionAntiCSRF;
+import org.zaproxy.zap.extension.custompages.CustomPage;
 import org.zaproxy.zap.model.Tech;
 import org.zaproxy.zap.model.TechSet;
 
@@ -585,6 +587,62 @@ public abstract class AbstractPlugin implements Plugin, Comparable<Object> {
      */
     protected boolean isFileExist(HttpMessage msg) {
         return parent.getAnalyser().isFileExist(msg);
+    }
+
+    /**
+     * Tells whether or not the message matches the specific {@code CustomPageType}
+     *
+     * @param msg the message that will be checked
+     * @param cpType the custom page type to be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    private boolean isCustomPage(HttpMessage msg, CustomPage.Type cpType) {
+        return parent.isCustomPage(msg, cpType);
+    }
+
+    /**
+     * Tells whether or not the message matches {@code CustomPageType.OK_200} definitions.
+     *
+     * @param msg the message that will be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isPage200(HttpMessage msg) {
+        return isCustomPage(msg, CustomPage.Type.OK_200);
+    }
+
+    /**
+     * Tells whether or not the message matches {@code CustomPageType.ERROR_500} definitions.
+     *
+     * @param msg the message that will be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isPage500(HttpMessage msg) {
+        return isCustomPage(msg, CustomPage.Type.ERROR_500);
+    }
+
+    /**
+     * Tells whether or not the message matches {@code CustomPageType.NOTFOUND_404} definitions.
+     *
+     * @param msg the message that will be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isPage404(HttpMessage msg) {
+        return isCustomPage(msg, CustomPage.Type.NOTFOUND_404);
+    }
+
+    /**
+     * Tells whether or not the message matches {@code CustomPageType.OTHER} definitions.
+     *
+     * @param msg the message that will be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isPageOther(HttpMessage msg) {
+        return isCustomPage(msg, CustomPage.Type.OTHER);
     }
 
     /**
--- a/zap/src/main/java/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
@@ -1,0 +1,276 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2019 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.custompages;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.regex.Pattern;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.utils.Enableable;
+
+public class DefaultCustomPage extends Enableable implements CustomPage {
+
+    /**
+     * The Constant FIELD_SEPARATOR used for separating DefaultCustomPage's fields during
+     * encode/decode operations.
+     */
+    private static final String FIELD_SEPARATOR = ";";
+
+    private static final Logger LOGGER = Logger.getLogger(DefaultCustomPage.class);
+
+    private int contextId;
+    private String pageMatcher;
+    private CustomPageMatcherLocation pageMatcherLocation;
+    private boolean regex;
+    private Type type;
+    private Pattern pattern;
+
+    /**
+     * Constructs a {@code DefaultCustomPage} with the given details.
+     *
+     * @param contextId the context ID for which the {@code DefaultCustomPage} is being created
+     * @param pageMatcher the page matcher of the {@code DefaultCustomPage}
+     * @param pageMatcherLocation the {@link CustomPageMatcherLocation} of the {@code
+     *     DefaultCustomPage}
+     * @param regex a boolean specifying whether or not the {@code DefaultCustomPage} page matcher
+     *     is represented by a Regex pattern
+     * @param type the {@link CustomPage.Type} of the {@code DefaultCustomPage}
+     * @param enabled a boolean specifying whether or not the {@code DefaultCustomPage} is enabled
+     *     or not
+     */
+    public DefaultCustomPage(
+            int contextId,
+            String pageMatcher,
+            CustomPageMatcherLocation pageMatcherLocation,
+            boolean regex,
+            Type type,
+            boolean enabled) {
+        super();
+        this.contextId = contextId;
+        this.pageMatcher = pageMatcher;
+        this.pageMatcherLocation = pageMatcherLocation;
+        this.regex = regex;
+        this.type = type;
+        this.setEnabled(enabled);
+    }
+
+    /**
+     * Constructs a {@code DefaultCustomPage} with the given details. {@link CustomPage.Type} and
+     * {@link CustomPageMatcherLocation} are specified by {@code int} ID instead of {@code enum}
+     * literal.
+     *
+     * @param contextId the context ID for which the {@code DefaultCustomPage} is being created
+     * @param pageMatcher the page matcher of the {@code DefaultCustomPage}
+     * @param pageMatcherLocationID the ID of the {@link CustomPageMatcherLocation} of the {@code
+     *     DefaultCustomPage}
+     * @param regex a boolean specifying whether or not the {@code DefaultCustomPage} page matcher
+     *     is represented by a Regex pattern
+     * @param type the {@code CustomPage.Type} of the {@code DefaultCustomPage}
+     * @param enabled a boolean specifying whether or not the {@code DefaultCustomPage} is enabled
+     *     or not
+     */
+    public DefaultCustomPage(
+            int contextId,
+            String pageMatcher,
+            int pageMatcherLocationID,
+            boolean regex,
+            Type type,
+            boolean enabled) {
+        super();
+        this.contextId = contextId;
+        this.pageMatcher = pageMatcher;
+        this.pageMatcherLocation =
+                CustomPageMatcherLocation.getCustomPagePageMatcherLocationWithId(
+                        pageMatcherLocationID);
+        this.regex = regex;
+        this.type = type;
+        this.setEnabled(enabled);
+    }
+
+    @Override
+    public int getContextId() {
+        return contextId;
+    }
+
+    @Override
+    public void setContextId(int contextId) {
+        this.contextId = contextId;
+    }
+
+    @Override
+    public String getPageMatcher() {
+        return pageMatcher;
+    }
+
+    @Override
+    public void setPageMatcher(String pageMatcher) {
+        this.pageMatcher = pageMatcher;
+    }
+
+    @Override
+    public CustomPageMatcherLocation getPageMatcherLocation() {
+        return pageMatcherLocation;
+    }
+
+    @Override
+    public void setPageMatcherLocation(CustomPageMatcherLocation cppmt) {
+        this.pageMatcherLocation = cppmt;
+    }
+
+    @Override
+    public boolean isRegex() {
+        return regex;
+    }
+
+    @Override
+    public void setRegex(boolean regex) {
+        this.regex = regex;
+    }
+
+    @Override
+    public Type getType() {
+        return type;
+    }
+
+    @Override
+    public void setType(Type cpt) {
+        this.type = cpt;
+    }
+
+    /**
+     * Determines if a {@code HttpMessage} is a Custom Page of a particular {@code CustomPage.Type}.
+     *
+     * @param msg the HTTP message to be evaluated
+     * @param cpt the CustomPage.Type of the Custom Pages against which the HTTP message should be
+     *     evaluated
+     * @return {@code true} if the HTTP message is a Custom Page of the type in question, {@code
+     *     false} otherwise
+     */
+    public boolean isCustomPage(HttpMessage msg, Type cpt) {
+        if (isEnabled() && getType() == cpt) {
+            String value = getPageMatcherByType(msg);
+            return matchByLocation(value);
+        }
+        return false; // Default
+    }
+
+    private String getPageMatcherByType(HttpMessage msg) {
+        if (CustomPageMatcherLocation.URL.equals(getPageMatcherLocation())) {
+            return msg.getRequestHeader().getURI().toString();
+        } else if (CustomPageMatcherLocation.RESPONSE_CONTENT.equals(getPageMatcherLocation())) {
+            return getHttpMessageAsString(msg);
+        }
+        LOGGER.error(
+                "Could not get page matcher for the given message, with: "
+                        + getPageMatcherLocation());
+        return "";
+    }
+
+    private boolean matchByLocation(String value) {
+        if (isRegex()) {
+            return isRegexMatch(pageMatcher, value);
+        }
+        return CustomPageMatcherLocation.URL.equals(getPageMatcherLocation())
+                ? value.equals(getPageMatcher())
+                : value.contains(getPageMatcher());
+    }
+
+    private static String getHttpMessageAsString(HttpMessage msg) {
+        return msg.getResponseHeader().toString() + msg.getResponseBody().toString();
+    }
+
+    private boolean isRegexMatch(String pageMatcher, String toMatch) {
+        if (pattern == null) {
+            pattern = Pattern.compile(pageMatcher);
+        }
+
+        return pattern.matcher(toMatch).find();
+    }
+
+    public String toString() {
+        StringBuilder cp = new StringBuilder();
+        cp.append("ContextId: ").append(this.getContextId());
+        cp.append(", Content: ").append(this.getPageMatcher()).append(".");
+        cp.append(", Content Type: ").append(this.getPageMatcherLocation().getName());
+        cp.append(", Is RegEx: ").append(this.isRegex());
+        cp.append(", Type: ").append(this.getType().getName());
+        cp.append(", IsEnabled: ").append(this.isEnabled());
+        return cp.toString();
+    }
+
+    /**
+     * Encodes the DefaultCustomPage in a String. Fields that contain strings are Base64 encoded.
+     *
+     * @param cp the custom page to be encoded
+     * @return the encoded string
+     */
+    public static String encode(CustomPage cp) {
+        StringBuilder encodedCP = new StringBuilder();
+        String matcherComponent = cp.getPageMatcher() != null ? cp.getPageMatcher() : "";
+
+        encodedCP
+                .append(
+                        new String(
+                                Base64.getEncoder()
+                                        .encode(matcherComponent.getBytes(StandardCharsets.UTF_8)),
+                                StandardCharsets.US_ASCII))
+                .append(FIELD_SEPARATOR);
+        encodedCP.append(cp.getPageMatcherLocation().getId()).append(FIELD_SEPARATOR);
+        encodedCP.append(cp.isRegex()).append(FIELD_SEPARATOR);
+        encodedCP.append(cp.getType().getId()).append(FIELD_SEPARATOR);
+        encodedCP.append(cp.isEnabled()).append(FIELD_SEPARATOR);
+
+        return encodedCP.toString();
+    }
+
+    /**
+     * Decodes a DefaultCustomPage from an encoded string. The string provided as input should have
+     * been obtained through calls to {@link #encode(CustomPage)}.
+     *
+     * @param contextId the ID of the context for which the encoded ({@code DefaultCustomPage})
+     *     string is being decoded
+     * @param encodedString the encoded string
+     * @return the DefaultCustomPage
+     */
+    protected static DefaultCustomPage decode(int contextId, String encodedString) {
+        String[] pieces = encodedString.split(FIELD_SEPARATOR, -1);
+        DefaultCustomPage defaultCustomPage = null;
+        try {
+            defaultCustomPage =
+                    new DefaultCustomPage(
+                            contextId, // ContextID
+                            new String(
+                                    Base64.getDecoder().decode(pieces[0]),
+                                    StandardCharsets.UTF_8), // Content
+                            CustomPageMatcherLocation.getCustomPagePageMatcherLocationWithId(
+                                    Integer.parseInt(pieces[1])), // Content
+                            // Type
+                            Boolean.parseBoolean(pieces[2]), // IsRegex
+                            Type.getCustomPageTypeWithId(Integer.parseInt(pieces[3])), // Type
+                            Boolean.parseBoolean(pieces[4])); // Enabled
+        } catch (Exception ex) {
+            LOGGER.error(
+                    "An error occured while decoding DefaultCustomPage from: " + encodedString, ex);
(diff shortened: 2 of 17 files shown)


thc202:
Best way to push this forward, thanks for raising the PR!

kingthorin:
If it's too much of a push let me know. I had a day off and wasn't sure what to work on so I tiddied up and pushed it....

thc202:
It should be fine. I noticed a ZapAddOn.xml file, was that on purpose?

kingthorin:
Was just an oversight, will remove it.

kingthorin:
Added types "Not Found" and "Other" to make the implementation more generic.

src/org/zaproxy/zap/extension/custompages/CustomPage.java
```diff
@@ -0,0 +1,120 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+ * See the License for the specific language governing permissions and 
+ * limitations under the License. 
+ */
+package org.zaproxy.zap.extension.custompages;
+
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.utils.EnableableInterface;
+
+public interface CustomPage extends EnableableInterface{
```
thc202: Maybe change the interface to just have the method `isCustomPage`? (And let `EnableableInterface` and the other methods be part of the default implementation?)
This would allow to have simpler/immutable default implementations (for 200, 404, 500).

src/org/zaproxy/zap/extension/custompages/CustomPageType.java
```diff
@@ -0,0 +1,145 @@
...
+package org.zaproxy.zap.extension.custompages;
+
+import org.apache.commons.lang.Validate;
+import org.parosproxy.paros.Constant;
+
+/**
+ * Defines the types supported by the Custom Page package.
+ */
+public enum CustomPageType {
+	ERROR(1, Constant.messages.getString("custompages.type.error")),
```
thc202: This one would be for 500 and the following for 404, the "other" would be a 200?

thc202:
Thanks. I'd suggest plug this into `Plugin`/`AbstractPlugin` and add an example implementation showing how/when it should be used.

src/org/zaproxy/zap/extension/custompages/CustomPageType.java
kingthorin: Basically. I didn’t want to link them strongly to a particular response code. For “other” I didn’t have a planned mapping or much of an idea at all aside from giving othe devs options for using the module :)

src/org/zaproxy/zap/extension/custompages/CustomPage.java
thc202: Per comments in IRC, not worth changing.

src/org/zaproxy/zap/extension/custompages/CustomPageType.java
thc202: I'd prefer have them mapped to specific status codes (either by docs or name) as that's what the scanners check/need(?).

kingthorin:
> Thanks. I'd suggest plug this into `Plugin`/`AbstractPlugin` and add an example implementation showing how/when it should be used.

Basically scanners could do something like this:
```Java
private boolean isCustomErrorPage(HttpMessage msg) {
	for (Context context : Model.getSingleton().getSession().getContexts()) {
		for (CustomPage cp : context.getCustomPages()) {
			if (cp.isCustomPage(msg, CustomPageType.ERROR)) {
				return true;
			}
		}
	}
	return false;
}
```

In circumstances like: 
https://github.com/zaproxy/zap-extensions/blob/36b290f29eb4f5ae51e49b88439b230e63854f63/src/org/zaproxy/zap/extension/pscanrules/CacheControlScanner.java#L66-L68

src/org/zaproxy/zap/extension/custompages/CustomPageType.java
kingthorin: So like `ERROR_500`, `NOTFOUND_404`, and `OTHER`. The point of including "other" was to give scripters etc a way to define something that doesn't fit a preconceived idea, that probably wouldn't be used by any of our packaged scanners (rules), then they can decide how to handle it under whatever given circumstance they might have. 

I'm ok with including a status code (number) in the naming if that helps/makes sense.

kingthorin:
That latest commit (dc15045 on 20171213) just changes the enum equality checks in the isCustomPage function (from `==` to `equals()`)

kingthorin:
Ref: https://mozilla.logbot.info/websectools/20171213

src/org/zaproxy/zap/extension/custompages/CustomPage.java
```diff
@@ -0,0 +1,120 @@
...
+	 * @param contextId the ID of the context
+	 */
+	void setContextId(int contextId);
+
+	/**
+	 * Returns the content for a Custom Page.
+	 * 
+	 * @return the content {@code String}
+	 */
+	String getContent();
```
denniskniep: I think the terms `Content` and  `ContentType` are a bit confusing.
These terms are strongly related to the http wording. 

How about renaming them into `Matcher` and `MatcherType` or  `PageMatcher` and `PageMatcherType` ? This renaming is also related to the class `CustomPageContentType`

src/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
```diff
@@ -0,0 +1,304 @@
...
+	}
+	
+	/**
+	 * Determines if a {@code HttpMessage} is a Custom Page of a particular {@code CustomPageType}.
+	 * 
+	 * @param msg the HTTP message to be evaluated
+	 * @param cpt the CustomPageType of the Custom Pages against which the HTTP message should be evaluated
+	 * @return {@code true} if the HTTP message is a Custom Page of the type in question, {@code false} otherwise
+	 */
+	public boolean isCustomPage(HttpMessage msg, CustomPageType cpt) {
```
denniskniep: Maybe we could make this a bit simpler (fewer nestings)

```
       public boolean isCustomPage(HttpMessage msg, CustomPageType cpt) {
		if (isEnabled() && getType() == cpt) {
			String value = getContentByType(msg);
			return matchByType(value);
		}
		return false; // Default
	}

	private String getContentByType(HttpMessage msg){
		if(CustomPageContentType.URL.equals(getContentType())){
			return msg.getRequestHeader().getURI().toString();
		} else if(CustomPageContentType.RESPONSE_CONTENT.equals(getContentType())){
			return getHttpMessageAsString(msg);
		}
		throw new RuntimeException("Not Supported");
	}

	private boolean matchByType(String value){
		if (!isRegex() && CustomPageContentType.URL.equals(getContentType())) {
			return value.equals(getContent());
		}
		if (!isRegex()){
			return value.contains(getContent());
		}		
		return isRegexMatch(content, value);
	}

```

src/org/zaproxy/zap/extension/custompages/resources/Messages.properties
```diff
@@ -0,0 +1,41 @@
+# Custom Pages 
+#
+# This file defines the default (English) variants of all of the internationalised messages
+
+custompages.content.type.url=URL
+custompages.content.type.responsebody=Response Body
```
denniskniep: We should rename this. Actually it is Header + Body 

```
      private String getHttpMessageAsString(HttpMessage msg) {
		return msg.getResponseHeader().toString() + msg.getResponseBody().toString();
	}
```

src/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
```diff
@@ -0,0 +1,232 @@
...
+			contextManagers.put(contextId, manager);
+		}
+		return manager;
+	}
+
+	@Override
+	public void loadContextData(Session session, Context context) {
+		try {
+			List<String> encodedCustomPages = session.getContextDataStrings(context.getIndex(), TYPE_CUSTOM_PAGE);
+			for (String e : encodedCustomPages) {
```
denniskniep: Don`t know, but `persistContextData()` has here a null check

kingthorin:
Thanks @denniskniep

src/org/zaproxy/zap/extension/custompages/CustomPageType.java
```diff
@@ -0,0 +1,145 @@
...
+	 * 
+	 * @param name
+	 *            the name of the custom page type
+	 * @return the custom page type that matches the given {@code name}.
+	 * @throws IllegalArgumentException
+	 *             if the given {@code name} is {@code null} or empty.
+	 * @see #getCustomPageTypeWithId()
+	 * @see #getDefaultType()
+	 */
+	public static CustomPageType getCustomPageTypeWithName(String name) {
```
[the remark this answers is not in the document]
kingthorin: The latter two types were a quick addition at some point when I knew there was still further work to do, I probably just missed touching this up. Thanks for catching it.

denniskniep:
Maybe we should add 2 default entries, when a context is created:
Check if StatusCode 500 -> CustomPageType.ERROR
Check if StatusCode 404 -> CustomPageType.NOTFOUND

Than Scanners etc. can simply query everytime the `isCustomErrorPage()` API. And do not need to additionally query the http statuscode with `OR`

denniskniep:
- [x] Test2
- [ ] Test1

src/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
kingthorin: Looking at other extensions it seems this is a pattern. I guess if loadContextData is called then the Context isn't (can't be) null?

thc202: Correct.

denniskniep: right, thats fine

src/org/zaproxy/zap/extension/custompages/CustomPage.java
kingthorin: I've left the UI as Content and Content Type for the time being, however, I've changed all the code and javadoc.

src/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
```diff
@@ -0,0 +1,232 @@
...
+import org.zaproxy.zap.model.ContextDataFactory;
+import org.zaproxy.zap.view.AbstractContextPropertiesPanel;
+import org.zaproxy.zap.view.ContextPanelFactory;
+
+/**
+ * A ZAP extension which implements Custom Pages, which
+ * allows you to define custom pages which may be used in various
+ * ZAP components.
+ */
+public class ExtensionCustomPages extends ExtensionAdaptor
```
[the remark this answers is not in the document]
kingthorin: Done

src/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
kingthorin: Done

kingthorin:
I see what you're getting at, however there are probably legit cases where someone would want to check one or the other but not necessarily default to both.

Could include convenient methods to cut down on useage verbosity. For Example:
`isCustomErrorOrStatus500` - isCustomPage(msg, CustomPageType.ERROR_500) or msg.getResponseHeader().getStatusCode() == 500
`isCustomErrorOrServerError` - isCustomPage(msg, CustomPageType.ERROR_500) or HttpStatusCode.isServerError(msg.getResponseHeader().getStatusCode())

Then something similar for 404s and isClientError...

psiinon:
This pull request **introduces 9 alerts** when merging f064a9d27d5ea6996a9f7da7f4446f019e0760c0 into c7174026252318de5f1867be6f4ca2aacc2b2f22 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-a4dca2464139262e97ad2fa1dbe8db8db59a51b6)

**new alerts:**

* 9 for Spurious Javadoc @param tags

---

*Comment posted by [LGTM.com](https://lgtm.com)*

kingthorin:
I rebased it, will address those the next time I'm working on it.

psiinon:
This pull request **introduces 9 alerts** when merging f81432b76868bf6b078b5e2c42cda486599cc3eb into dec36d2d7e7995d75813fc050158b327cbc0d885 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-ddecba0b422e7db01503bcc44e6b651ebd1710b2)

**new alerts:**

* 9 for Spurious Javadoc @param tags

---

*Comment posted by [LGTM.com](https://lgtm.com)*

psiinon:
This pull request **introduces 1 alert** when merging 69be3992d0c78bf7e4eeaca9a3e73c6a36b3b9a9 into dec36d2d7e7995d75813fc050158b327cbc0d885 - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-90a28ed5551bbcd7bb4df0d4b8f5221ba12234be)

**new alerts:**

* 1 for Spurious Javadoc @param tags

---

*Comment posted by [LGTM.com](https://lgtm.com)*

zap/src/main/java/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
```diff
@@ -229,6 +229,13 @@ public void sessionModeChanged(Mode mode) {
...
     }
 
+    /**
+     * Tells whether or not the msg matches the specific {@code CustomPageType}
+     *
+     * @param msg the message that will be checked
+     * @param cpType the custom page type to be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     */
     public boolean isCustomPage(HttpMessage msg, CustomPageType cpt) {
```
kingthorin: - Should we check contexts whether the user requests it or not?
- Does this while implementation actually need to be an "Option" panel instead of context based? (More like Global Excludes?)
- How do we make this pan out for pscan (only if the user has select "Scan only in scope"?)
- Is there a way (I didn't see it) in HostProcess (or whatever) to check isScanningContext? For which if we were only applying CustomPages to context scanning we could bail early here and return false? (isInScope?)

kingthorin:
Minimal testable pscan script:
```js
function scan(ps, msg, src) {
     var CustomPageType = Java.type("org.zaproxy.zap.extension.custompages.CustomPageType");
	print(ps.getHelper().isCustomPage(msg, CustomPageType.ERROR_500) + ' : ' + msg.getRequestHeader().getURI().toString());
}
```

kingthorin:
Here's a minimal but testable ascan script:
```js
function scan(as, msg, param, value) {
	// Copy requests before reusing them
	msg = msg.cloneRequest();

	// setParam (message, parameterName, newValue)
	as.setParam(msg, param, "foobar");
		
	// sendAndReceive(msg, followRedirect, handleAntiCSRFtoken)
	as.sendAndReceive(msg, false, false);

	var CustomPageType = Java.type("org.zaproxy.zap.extension.custompages.CustomPageType");
	print(as.getParent().isCustomPage(msg, CustomPageType.ERROR_500) + ' : ' + msg.getRequestHeader().getURI().toString());

	// Check if the scan was stopped before performing lengthy tasks
	if (as.isStop()) {
		return
	}
}
```

psiinon:
Some questions and suggestions:
What does the 'Response' option match against? Is it the body or the header as well?
I think it would be good to be able to match against the header :)
Also, how about adding a 'match tag' option - that would allow more complicated rules to be created if necessary, esp as I think we can tag things from scripts.
What does 'Other' imply?
How will the scan rules use this? Will they just be able to use `isPage404(msg)` and not worry about whether custom pages have been set up or not? If so it might make it a bit clearer if we used `isErrorPage(..)`, `isOkPage(...)`, `isNotFoundPage(..)` etc?

psiinon:
Another thought :)
How about a right click option when selecting text in a response - `Flag as Context -> Custom Page`?
Just a nice to have at some point.

kingthorin:
> What does the 'Response' option match against? Is it the body or the header as well?
I think it would be good to be able to match against the header :)

IIRC it's both, I'd have to double check the code :)

> Also, how about adding a 'match tag' option - that would allow more complicated rules to be created if necessary, esp as I think we can tag things from scripts.

That could be added, the problem is used within Active Scan, messages in active scan aren't passively scanned and therefore aren't tagged.

> What does 'Other' imply?

"Other" was included to provide greater flexibility for scripts etc. or between core releases if a dev decides they want to do something special or not yet handled.

> How will the scan rules use this? Will they just be able to use `isPage404(msg)` and not worry about whether custom pages have been set up or not? If so it might make it a bit clearer if we used `isErrorPage(..)`, `isOkPage(...)`, `isNotFoundPage(..)` etc?

Yes that's the plan, as per a recent suggestion fro @thc202 they now fall back to checking status so that dev (rules) don't have to check both customPages and status code or things like `isServerError` etc.

> How about a right click option when selecting text in a response - `Flag as Context -> Custom Page`?

Default to setting it as a literal URL match?

psiinon:
> > What does the 'Response' option match against? Is it the body or the header as well?
> > I think it would be good to be able to match against the header :)
> 
> IIRC it's both, I'd have to double check the code :)

Cool.

> 
> > Also, how about adding a 'match tag' option - that would allow more complicated rules to be created if necessary, esp as I think we can tag things from scripts.
> 
> That could be added, the problem is used within Active Scan, messages in active scan aren't passively scanned and therefore aren't tagged.

Arg, forgot about that. I wonder if theres another way for us to be able to define custom pages via scripts, without adding yet another script type? Definitely not essential, was just thinking of maximum flexibility.

> 
> > What does 'Other' imply?
> 
> "Other" was included to provide greater flexibility for scripts etc. or between core releases if a dev decides they want to do something special or not yet handled.
 
Fair enough.

> > How will the scan rules use this? Will they just be able to use `isPage404(msg)` and not worry about whether custom pages have been set up or not? If so it might make it a bit clearer if we used `isErrorPage(..)`, `isOkPage(...)`, `isNotFoundPage(..)` etc?
> 
> Yes that's the plan, as per a recent suggestion fro @thc202 they now fall back to checking status so that dev (rules) don't have to check both customPages and status code or things like `isServerError` etc.

+1 

> > How about a right click option when selecting text in a response - `Flag as Context -> Custom Page`?
> 
> Default to setting it as a literal URL match?

I was thinking this would be when they right clicked some text in the response that they had selected, in which case the default should be an exact match.

zap/src/main/java/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
```diff
@@ -0,0 +1,367 @@
...
+            Context.CONTEXT_CONFIG + ".custompages";
+    public static final String CONTEXT_CONFIG_CUSTOM_PAGE = CONTEXT_CONFIG_CUSTOM_PAGES + ".page";
+
+    private static final int TYPE_CUSTOM_PAGE = 600;
+
+    /** The Custom Page panels, mapped to each context. */
+    private Map<Integer, ContextCustomPagePanel> customPagePanelsMap = new HashMap<>();
+
+    /** The context managers, mapped to each context. */
+    private Map<Integer, ContextCustomPageManager> contextManagers = new HashMap<>();
```
thc202, line 73: The custom pages are being managed here and in the `Context`.

zap/src/main/java/org/zaproxy/zap/extension/custompages/CustomPageType.java
```diff
@@ -0,0 +1,142 @@
...
+
+import org.apache.commons.lang.Validate;
+import org.parosproxy.paros.Constant;
+
+/** Defines the types supported by the Custom Page package. */
+public enum CustomPageType {
+    ERROR_500(1, Constant.messages.getString("custompages.type.error")),
+    NOTFOUND_404(2, Constant.messages.getString("custompages.type.notfound")),
+    OK_200(3, Constant.messages.getString("custompages.type.ok")),
+    OTHER(4, Constant.messages.getString("custompages.type.other"));
```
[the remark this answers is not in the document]
kingthorin: That was meant to provide flexibility for scripts for use cases I hadn’t predicted.

kingthorin:
This is great! Thanks for the review. I’ll likely have questions as I go through. I’ll hop into IRC in a bit.

zap/src/main/java/org/zaproxy/zap/extension/custompages/CustomPageType.java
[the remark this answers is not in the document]
thc202: Makes sense, worth adding JavaDoc mentioning that case.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -587,6 +588,62 @@ protected boolean isFileExist(HttpMessage msg) {
...
 
+    /**
+     * Tells whether or not the message matches the specific {@code CustomPageType}
+     *
+     * @param msg the message that will be checked
+     * @param cpType the custom page type to be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isCustomPage(HttpMessage msg, CustomPageType cpType) {
```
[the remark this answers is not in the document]
kingthorin: Changed to private. Will deal with `isFileExist` in a future set of changes that involved analyzer.

zap/src/main/java/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
```diff
@@ -0,0 +1,367 @@
...
+import org.zaproxy.zap.view.ContextPanelFactory;
+import org.zaproxy.zap.view.popup.PopupMenuItemContext;
+import org.zaproxy.zap.view.popup.PopupMenuItemSiteNodeContextMenuFactory;
+
+/**
+ * A ZAP extension which implements Custom Pages, which allows you to define custom pages which may
+ * be used in various ZAP components.
+ */
+public class ExtensionCustomPages extends ExtensionAdaptor
+        implements ContextPanelFactory, ContextDataFactory, SessionChangedListener {
```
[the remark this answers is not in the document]
kingthorin: Do you mean `ExtensionCustomPages` should have inner class that implements these 3?

zap/src/main/java/org/zaproxy/zap/extension/custompages/CustomPageType.java
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * <p>Default return is {@code getDefaultType()}.
+     *
+     * @param name the name of the custom page type
+     * @return the custom page type that matches the given {@code name}.
+     * @throws IllegalArgumentException if the given {@code name} is {@code null} or empty.
+     * @see #getCustomPageTypeWithId()
+     * @see #getDefaultType()
+     */
+    public static CustomPageType getCustomPageTypeWithName(String name) {
```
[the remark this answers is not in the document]
kingthorin: It was done as a custom enum so that names shown in the combobox could be i18n but if you see a simpler way I'm up for that too.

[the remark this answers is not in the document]
thc202: Right. Actually the wrapper is unnecessary you can have the combo box with the enum type, you just need to override `toString()` to return the `name`.

zap/src/main/java/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java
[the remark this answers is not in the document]
thc202: Yes.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1528,6 +1528,49 @@ core.api.view.zapHomePath = Gets the path to ZAP's home directory.
...
+
+custompages.desc=Custom Pages Definition
+
+custompages.dialog.add.title=Add Custom Page
+custompages.dialog.add.button.confirm=Confirm
+custompages.dialog.add.field.label.type=Type:
+custompages.dialog.add.field.label.enabled=Enabled:
+custompages.dialog.add.field.label.regex=Is Regex?
+custompages.dialog.add.field.label.content=Content:
+custompages.dialog.add.field.label.contenttype=Content Type:
```
thc202, line 1542: This could also be changed to "location" (there are some variables that still need to be renamed accordingly).

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1528,6 +1528,49 @@ core.api.view.zapHomePath = Gets the path to ZAP's home directory.
...
+custompages.dialog.modify.button.confirm=Confirm
+
+custompages.dialog.remove.button.cancel=Cancel
+custompages.dialog.remove.button.confirm=Remove
+custompages.dialog.remove.checkbox.label=Do not show this message again
+custompages.dialog.remove.text=Are you sure you want to remove the selected Custom Page?
+custompages.dialog.remove.title=Remove Custom Page
+
+custompages.error.nofile=File not found : {0}
+custompages.panel.description=Define custom pages leveraged by ZAP components.
```
thc202, line 1555: The `leveraged by ZAP components` is implied?

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -587,6 +588,62 @@ protected boolean isFileExist(HttpMessage msg) {
...
+    }
+
+    /**
+     * Tells whether or not the message matches {@code CustomPageType.OK_200} definitions.
+     *
+     * @param msg the message that will be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     * @since TODO Add version
+     */
+    protected boolean isPage200(HttpMessage msg) {
```
thc202, line 610: The `ScriptsActiveScanner` needs to override these methods and change to public to allow the scripts to use them.

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanData.java
```diff
@@ -136,4 +140,63 @@ public Context getContext() {
...
     }
+
+    /**
+     * Tells whether or not the message matches the specific {@code CustomPageType}
+     *
+     * @param msg the message that will be checked
+     * @param cpType the custom page type to be checked
+     * @return {@code true} if the message matches, {@code false} otherwise
+     */
+    private boolean isCustomPage(HttpMessage msg, CustomPage.Type cpType) {
```
thc202, line 151: Needs to check (and return) if the `context` is `null`.

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanData.java
```diff
@@ -136,4 +140,63 @@ public Context getContext() {
...
+        if (customPageMap == null) {
+            customPageMap = new HashMap<CustomPage.Type, Boolean>();
+        }
+        return customPageMap.computeIfAbsent(
+                cpType, k -> customPageMap.put(cpType, context.isCustomPage(msg, cpType)));
```
thc202, line 156: Could just do `context.isCustomPage(msg, k)`, the put is done internally.

zap/src/main/java/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
```diff
@@ -0,0 +1,279 @@
...
+     * The Constant FIELD_SEPARATOR used for separating DefaultCustomPage's fields during
+     * encode/decode operations.
+     */
+    private static final String FIELD_SEPARATOR = ";";
+
+    private static final Logger LOGGER = Logger.getLogger(DefaultCustomPage.class);
+
+    private int contextId;
+    private String pageMatcher;
+    private CustomPageMatcherLocation pageMatcherType;
```
thc202, line 41: The variable needs rename.

zap/src/main/java/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
```diff
@@ -0,0 +1,279 @@
...
+package org.zaproxy.zap.extension.custompages;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.regex.Pattern;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.utils.Enableable;
+
+class DefaultCustomPage extends Enableable implements CustomPage {
```
thc202, line 29: This one might have to be public if other code is to be allowed to add them.

zap/src/main/java/org/zaproxy/zap/extension/custompages/DefaultCustomPage.java
```diff
@@ -0,0 +1,279 @@
...
+        return false; // Default
+    }
+
+    private String getPageMatcherByType(HttpMessage msg) {
+        if (CustomPageMatcherLocation.URL.equals(getPageMatcherType())) {
+            return msg.getRequestHeader().getURI().toString();
+        } else if (CustomPageMatcherLocation.RESPONSE_CONTENT.equals(getPageMatcherType())) {
+            return getHttpMessageAsString(msg);
+        }
+        LOGGER.error("Could not get page matcher for the given message.");
```
thc202, line 179: `... message, with: " + getPageMatcherType()` ? (The getter and variable should be renamed to use "location".)

zap/src/main/java/org/zaproxy/zap/extension/custompages/CustomPageType.java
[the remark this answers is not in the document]
kingthorin: Great, thanks!

thc202 (approved):

thc202:
Thanks!

psiinon (approved):

kingthorin:
OMG it’s merged!!!!! :horray:

psiinon:
Worth having a tracker for subsequent related changes?

kingthorin:
I'll add a bunch of TODO bullets to issue 9, it's been left open for now.

PR #3642, opened 2017-06-09
Allow to customise the export action of ZapTable

thc202:
Add a method to ZapTable that allows extending classes to create a
custom export action.


thc202:
Tweaked to handle null action.

PR #3656, opened 2017-06-14
added healthchek to docker images

### LINKED ISSUE #3649: Add docker health check
Dockerfile allow you to declare [HEALTHCHECK](https://docs.docker.com/engine/reference/builder/#healthcheck) command. It would be great to add this command to the docker image, and report Zap as healthy when it complete loading. This could be useful when using Zap in CI environment (using `docker-compose`), and waiting to Zap to complete loading before starting the tests. The only problem is how we detect that Zap complete loading - I think the API is responsive before Zap complete loading.

omerlh:
close #3649

### DIFF
--- a/build/docker/Dockerfile-live
+++ b/build/docker/Dockerfile-live
@@ -101,3 +101,4 @@ RUN chown zap:zap /zap/zap-x.sh && \
 WORKDIR /zap
 	
 USER zap
+HEALTHCHECK --retries=5 --interval=5s CMD zap-cli status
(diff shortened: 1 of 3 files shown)


thc202:
Thanks!

thc202:
In the pull request description the issue should be 3649 ;)

omerlh:
Fixed :)

thc202:
@omerlh how would you like to be credited?

thc202:
OK, which name/handle should we use?

omerlh:
Omer Levi Hevroni

PR #3664, opened 2017-06-15
Do not start the next plugin if scan paused

thc202:
Change HostProcess to not start the following plugin while the scanner
is paused, also, check if the scan is paused before starting the scan of
new message by the current plugin.


psiinon:

PR #3666, opened 2017-06-16
Reduce code duplication in Scanner

mqliang:


src/org/parosproxy/paros/core/scanner/Scanner.java
```diff
@@ -46,15 +46,8 @@
 
 import java.security.InvalidParameterException;
 import java.text.DecimalFormat;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
thc202: Wildcard imports are not usually used, the preference is to just import the required types.

src/org/parosproxy/paros/core/scanner/Scanner.java
```diff
@@ -256,56 +249,38 @@ public void scan(Target target) {
...
-	                thread = pool.getFreeThreadAndRun(hostProcess);
-	                if (thread == null) Util.sleep(500);
-	            } while (thread == null && !isStop());
-	            if (thread != null) {
-		            notifyHostNewScan(hostAndPort, hostProcess);
-	            }
-	    	}
+	    } else {
+            // TODO need to change for lowmem
+            if (Constant.isLowMemoryOptionSet()) {
```
thc202: This changes the previous behaviour, now it will always through the exception even if not scanning in scope nor a context, but I guess it's fine.

src/org/parosproxy/paros/core/scanner/Scanner.java
```diff
@@ -256,56 +249,38 @@ public void scan(Target target) {
...
+                throw new InvalidParameterException("Not yet supported for the low memory option :(");
+            }
+
+            List<SiteNode> nodes = null;
```
thc202: Should be initialised to an empty list, otherwise it could lead to a `NullPointerException` when iterating (if neither in scope nor scanning a context).

kingthorin:
@mqliang are you able to address the feedback?

mqliang:
@thc202  @kingthorin comment addressed.

src/org/parosproxy/paros/core/scanner/Scanner.java
```diff
@@ -256,56 +257,38 @@ public void scan(Target target) {
...
-		            notifyHostNewScan(hostAndPort, hostProcess);
-	            }
-	    	}
+	    } else if (target.getContext() != null || target.isInScopeOnly()){
+            // TODO need to change for lowmem
+            if (Constant.isLowMemoryOptionSet()) {
+                throw new InvalidParameterException("Not yet supported for the low memory option :(");
+            }
+
+            List<SiteNode> nodes = new LinkedList<>();
```
thc202: Could use `Collections.emptyList()` instead.

thc202:
The "ZAP comment" was not added to the class.

mqliang:
Done

thc202:
Thank you!

PR #3681, opened 2017-06-19
Use number spinner for connection timeout

thc202:
Change OptionsConnectionPanel to use a number spinner for the connection
timeout, which already disallows invalid values. Remove validations no
longer needed.
Change ConnectionParam to expose the default value and ensure the
connection timeout is never negative.


src/org/parosproxy/paros/network/ConnectionParam.java
```diff
@@ -126,7 +134,7 @@
 	
 	// ZAP: Added prompt option and timeout
 	private boolean proxyChainPrompt = false;
-	private int timeoutInSecs = 120;
+	private int timeoutInSecs = DEFAULT_TIMEOUT;
```
thc202: It should be used 20, right? (That was the value used when reading from configuration file.)

kingthorin: Yup

PR #3685, opened 2017-06-20
Inform of active actions when changing the session

### LINKED ISSUE #3456: Bug: Paused Scans Wiped Out when Persisting a Session
If you pause a scan, and then Save a session with 'Persist Session' when the save is done, the paused scan is 'gone'. Seems like it shouldn't have any affect on an active scan. If there is no easy way to fix this, please at least warn the user that the currently active scan is going to get wiped out so they can make the correct choice for their situation.

thc202:
Change MenuFileControl to inform of active actions (for example, spider,
active scans) before creating, opening, or saving the session.

Fix #3456 - Paused Scans Wiped Out when Persisting a Session


src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -164,6 +169,32 @@ public void newSession(boolean isPromptNewSession) throws ClassNotFoundException
 				break;
 		}
 	}
+
+	private boolean informStopActiveActions() {
+		String activeActions = wrapEntriesInLiTags(control.getExtensionLoader().getActiveActions());
+		if (!activeActions.isEmpty()) {
+			String message = MessageFormat
```
kingthorin: Couldn't this just be: `Constant.messages.getString("menu.file.session.activeactions", activeActions)` ?

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -164,6 +169,32 @@ public void newSession(boolean isPromptNewSession) throws ClassNotFoundException
 				break;
 		}
 	}
+
+	private boolean informStopActiveActions() {
+		String activeActions = wrapEntriesInLiTags(control.getExtensionLoader().getActiveActions());
+		if (!activeActions.isEmpty()) {
+			String message = MessageFormat
+					.format(Constant.messages.getString("menu.file.session.activeactions"), activeActions);
```
kingthorin: Couldn't this just be: `Constant.messages.getString("menu.file.session.activeactions", activeActions)` ? (Without the MessageFormat)

src/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -164,6 +169,32 @@ public void newSession(boolean isPromptNewSession) throws ClassNotFoundException
...
+			if (view.showConfirmDialog(message) != JOptionPane.OK_OPTION) {
+				return false;
+			}
+		}
+		return true;
+	}
+
+	private static String wrapEntriesInLiTags(List<String> entries) {
+		if (entries.isEmpty()) {
+			return "";
```
kingthorin: Should we log something here? If the "actions" are empty wouldn't that be something strange?

src/org/parosproxy/paros/control/MenuFileControl.java
thc202: Sure.

src/org/parosproxy/paros/control/MenuFileControl.java
thc202: No, that's expected to happen, when nothing is running/active.

src/org/parosproxy/paros/control/MenuFileControl.java
thc202: Change done.

src/org/parosproxy/paros/control/MenuFileControl.java
kingthorin: Ok I was thinking more of some sort of weird race or time of check/time of use issue where we check something is running but then by the time it's added to the list it's done....

All good.

thc202: That case can happen, but no very problematic IMO. Warrants further changes?

kingthorin: No, probably fine. If it comes up we've thought about it and can tackle it if seen.

thc202: OK.

PR #3687, opened 2017-06-20
Log number of alerts raised by the active scanners

thc202:
Change HostProcess to log the number of alerts raised by each scanner.


psiinon:
Is this count also available via the API? I think that would be really useful :)

thc202:
Yeah, we can expose that too, but that change might break existing (JSON) clients, the data is returned in an array instead of an object (we could add another endpoint to fix the structure?).

thc202:
Pull request raised #3733, it does not seem to break the JSON clients (the new value is returned at the of the end array).

PR #3694, opened 2017-06-22
Correct path handling

### LINKED ISSUE #3692: zap-api-scan.py correct path handling
As per https://groups.google.com/d/msg/zaproxy-users/tZbTqNMKtVQ/Z92fn_TxAAAJ

psiinon:
Fixes #3692

### DIFF
--- a/build/docker/zap-api-scan.py
+++ b/build/docker/zap-api-scan.py
@@ -197,17 +197,6 @@ def main(argv):
         usage()
         sys.exit(3)
 
-    if target.startswith('http://') or target.startswith('https://'):
-        target_url = target
-    else:
-        # assume its a file
-        if not os.path.exists(target):
-            logging.warning('Target must either start with \'http://\' or \'https://\' or be a local file')
-            usage()
-            sys.exit(3)
-        else:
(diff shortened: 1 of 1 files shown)


build/docker/zap-api-scan.py
```diff
@@ -197,26 +197,27 @@ def main(argv):
...
+                usage()
+                sys.exit(3)
+
     if target.startswith('http://') or target.startswith('https://'):
         target_url = target
     else:
         # assume its a file
-        if not os.path.exists(target):
+        if not os.path.exists(base_dir + target):
             logging.warning('Target must either start with \'http://\' or \'https://\' or be a local file')
```
kingthorin: This line should be dropped in favor of the new L215?

If only looking for `base_dir + target` then it isn't going to start with an http scheme?

psiinon: Thats because the user might have incorrectly missed off the https:// eg:
zap-api-scan.py -t www.example.com/openapi.json

kingthorin: Ok that's fair.

PR #3717, opened 2017-06-29
Report nodes added by the spider

### LINKED ISSUE #3714: Spider - report # of new endpoints discovered
From #2583:
> ... could it report the number of new endpoints it discovered during the Spider. That way the user could know if doing it was useful or not. If it found ZERO new endpoints, it didn't accomplish anything.

psiinon:
Fixes #3714 
The 'sub' tabs are now always shown in the Spider tab, including a new one which shows all of the nodes added. A count of these is also shown in the spider toolbar, and the nodes are also accesible via the API.
A right click option to "Show in Sites tree" would be useful, but I dont think its essemtial ;)


src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -310,40 +383,50 @@ private JLabel getFoundCountValueLabel() {
...
 	@Override
 	protected int addToolBarElements(JToolBar toolBar, Location location, int gridX) {
 		if (ScanPanel2.Location.afterProgressBar == location) {
 			toolBar.add(new JToolBar.Separator(), getGBC(gridX++, 0));
 			toolBar.add(getFoundCountNameLabel(), getGBC(gridX++, 0));
 			toolBar.add(getFoundCountValueLabel(), getGBC(gridX++, 0));
+
 			toolBar.add(new JToolBar.Separator(), getGBC(gridX++, 0));
-			toolBar.add(getShowMessagesToggleButton(), getGBC(gridX++, 0));
+            toolBar.add(getAddedCountNameLabel(), getGBC(gridX++, 0));
```
kingthorin: I'm going to trust that these stick out like a sore thumb in the GitHub diff, and that you're going to catch them. If you want me to comment them all then say so and I'll nitpick it later :grin:

src/org/zaproxy/zap/extension/spider/SpiderThread.java
```diff
@@ -484,7 +509,16 @@ public void run() {
...
 
-		SessionStructure.addPath(Model.getSingleton().getSession(), historyReference, message);
+        StructuralNode node = SessionStructure.addPath(
+                Model.getSingleton().getSession(), historyReference, message, true);
+		if (node != null) {
+		    try {
+                addUriToAddedNodesModel(SessionStructure.getNodeName(message), 
+                        message.getRequestHeader().getMethod(), "");
+            } catch (URIException e) {
+                log.error("Error while recording added node: " + e.getMessage(), e);
```
kingthorin: Without searching the code base for this message text is "recording" going to be meaningful to a user or dev?

Perhaps this should be about adding the node to the tree?

src/org/zaproxy/zap/extension/spider/SpiderThread.java
```diff
@@ -589,4 +630,8 @@ public void setCustomParseFilters(List<ParseFilter> customParseFilters) {
 		this.customParseFilters  = customParseFilters;
 	}
 
+    public int getNumberOfNodesAdded() {
+        return getAddedNodesTableModel().getRowCount();
```
kingthorin: Maybe there's nothing wrong with this but it just seems strange to me. Shouldn't the spider actually keep track of this vs the table model row count?

Since this is done in a thread class can we end up with a sync or time of check vs time of use issue?

src/org/zaproxy/zap/model/SessionStructure.java
```diff
@@ -269,11 +286,14 @@ private static RecordStructure addStructure (Session session, String host, HttpM
 					tmpMsg = getTempHttpMessage(session, parentName, msg);
 					parentHistoryId = tmpMsg.getHistoryRef().getHistoryId();
 				}
-				RecordStructure parentRs = addStructure(session, host, tmpMsg, paths, size-1, parentHistoryId); 
+				RecordStructure parentRs = addStructure(session, host, tmpMsg, paths, size-1, parentHistoryId, false); 
 				parentId = parentRs.getStructureId();
 			}
 			msgRs = Model.getSingleton().getDb().getTableStructure().insert(
 					session.getSessionId(), parentId, historyId, nodeName, url, method);
+		} else if (newOnly) {
```
kingthorin: May as well do thIs check first

src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -363,6 +364,20 @@ public synchronized SiteNode addPath(HistoryReference ref) {
...
+    }
+
+    /**
+     * Add the HistoryReference with the corresponding HttpMessage into the SiteMap.
+     * This method saves the msg to be read from the reference table.  Use 
+     * this method if the HttpMessage is known.
+     * Note that this method must only be called on the EventDispatchThread
+     * @param msg
+     * @param newOnly   Only return a SiteNode if one was newly created
+     * @return
```
kingthorin: Param msg should be doc's too

src/lang/Messages.properties
```diff
@@ -2037,6 +2037,7 @@ spider.api.action.setOptionMaxParseSizeBytes = Sets the maximum size, in bytes,
 spider.api.action.scan = Runs the spider against the given URL (or context). Optionally, the 'maxChildren' parameter can be set to limit the number of children scanned, the 'recurse' parameter can be used to prevent the spider from seeding recursively, the parameter 'contextName' can be used to constrain the scan to a Context and the parameter 'subtreeOnly' allows to restrict the spider under a site's subtree (using the specified 'url').
 spider.api.action.scanAsUser = Runs the spider from the perspective of a User, obtained using the given Context ID and User ID. See 'scan' action for more details.
 spider.api.action.setOptionScopeString = Use actions [add|modify|remove]DomainAlwaysInScope instead.
+spider.api.view.addedNodes = Returns a list of the names of the nodes added to the Sites tree by the specified scan.
```
thc202: `URLs` instead of `names`?

src/lang/Messages.properties
```diff
@@ -2147,7 +2149,8 @@ spider.table.messages.column.processed.successfully = Successfully
 spider.toolbar.button.showmessages.label = Show Messages
 spider.toolbar.button.showmessages.label.selected = Hide Messages
 spider.toolbar.button.showmessages.tooltip = Shows the HTTP messages sent while spidering
```
thc202: This and two above should be removed (no longer used).

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -396,6 +446,18 @@ protected void updateFoundCount() {
 		}
 	}
 
+    /**
+     * Update the count of added nodes.
+     */
+    protected void updateAddedCount() {
```
thc202: Should be called in `switchView`.

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -267,6 +301,45 @@ public void show(Component invoker, int x, int y) {
...
+                    new DefaultTableRenderer(new MappedValue(StringValues.EMPTY, IconValues.NONE), JLabel.CENTER));
+            addedNodesTable.getColumnExt(0).setHighlighters(new ProcessedCellItemIconHighlighter(0));
+
+            addedNodesTable.getColumnModel().getColumn(0).setMinWidth(80);
+            addedNodesTable.getColumnModel().getColumn(0).setPreferredWidth(90); // processed
+
+            addedNodesTable.getColumnModel().getColumn(1).setMinWidth(60);
+            addedNodesTable.getColumnModel().getColumn(1).setPreferredWidth(70); // method
+
+            addedNodesTable.getColumnModel().getColumn(2).setMinWidth(400); // name
```
thc202: `// URI`? (and remove the comment from the following line)

src/org/zaproxy/zap/extension/spider/SpiderThread.java
psiinon: The table model is the only place we currently record this. Is it really worth having another variable which should just be the same as the row count? I dont think there should be any sync issues, but if they are they'll just be transitory wont they?

src/org/zaproxy/zap/model/SessionStructure.java
psiinon: That would break the functionality - we should only check 'newOnly' if msgRs is not null

src/lang/Messages.properties
psiinon: strictly speaking they arent URLs, thay are node names :)

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
psiinon: Its actually the node name, not the url :)

src/lang/Messages.properties
thc202: But it's returning the URL (not just the node name)?

src/org/zaproxy/zap/extension/spider/SpiderThread.java
kingthorin: I'm good with this explanation. When I read it it just seemed strange.

src/lang/Messages.properties
psiinon: Well, its actually returning the 'fully qualified' node name, eg:

- http://localhost:8080/bodgeit/basket.jsp (quantity_18,update)

So it includes the full URL + the names of the post parameters

psiinon:
Hopefully addressed all of the comments (and indentation issues) but keep the comments coming :)

src/org/zaproxy/zap/model/SessionStructure.java
kingthorin: Ok.

Was just thinking bailing earlier means fewer conditionals to be checked, however, as you've pointed out it is dependent on other conditions being checked/met

src/lang/Messages.properties
thc202: Right, didn't notice that.

src/org/zaproxy/zap/extension/spider/SpiderAPI.java
```diff
@@ -623,6 +625,14 @@ public ApiResponse handleApiView(String name, JSONObject params) throws ApiExcep
 			}
 			
 			result = resultUrls;
+		} else if (VIEW_ADDED_NODES.equals(name)) {
+			result = new ApiResponseList(name);
+			SpiderScan scan = (SpiderScan) this.getSpiderScan(params);
+			if (scan != null) {
+				for (String s : scan.getAddedNodesTableModel().getAddedNodes()) {
```
thc202: This will lead to an exception when running in daemon mode (`getAddedNodesTableModel()` returns null).

psiinon: Good spot :)

psiinon: And now hopefully fixed :)

thc202:
Thank you!

PR #3733, opened 2017-07-06
Ascan API - Return alert count for each scanner

thc202:
Change ActiveScanAPI to return the count of alerts raised by each
scanner.
Change HostProcess to expose the PluginStats (allow consumers to obtain
all the data, instead of adding more methods to HostProcess).
Move PluginStats to a top level class.

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/PluginStats.java
+++ b/src/org/parosproxy/paros/core/scanner/PluginStats.java
@@ -1,0 +1,165 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


psiinon:
@kingthorin ? :)

thc202:
Conflict resolved.

src/org/parosproxy/paros/core/scanner/PluginStats.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @see #isSkipped()
+     * @see #setSkippedReason(String)
+     */
+    void skipped() {
+        this.skipped = true;
+    }
+
+    /**
+     * Gets the reason why the plugin was skipped.
```
kingthorin: Sets?

src/org/parosproxy/paros/core/scanner/PluginStats.java
```diff
@@ -0,0 +1,158 @@
...
+        return skipped;
+    }
+
+    /**
+     * Skips the plugin.
+     *
+     * @see #isSkipped()
+     * @see #setSkippedReason(String)
+     */
+    void skipped() {
```
kingthorin: Can we make this present tense? `skip`

thc202: Sure, will change that.

src/org/parosproxy/paros/core/scanner/PluginStats.java
thc202: Yeah, c&p error.

thc202:
Changed done.

Issue #3727, opened 2017-07-05
Sites Tree Alpha Sort should ignore HTTP Method


davewichers:
A while ago, you addressed my request to alphabetically sort the Sites Tree with: #1800. However, I just now noticed that the sort includes the HTTP Method in its analysis. This causes all URLs accessed with a GET to be listed before all the URLs accessed with a POST inside the same folder, which is non-intuitive/confusing. Please change the default behavior to ignore the HTTP method when sorting the tree (or offer multiple sorting options).

<sub>Edit: Enhancement request(s) moved to https://github.com/zaproxy/zaproxy/issues/3738</sub>

PR #3737, opened 2017-07-09
Issue 3727 SiteTree Sorting Should Exclude Method

kingthorin:
SiteMap - Tweak embedded class SiteNodeStringComparator compare function so that if the node names match (without the method being included), they are then sorted based on method.
SiteNode - Added a getName() method which returns the node's name without the method (verb) prefixing the return string.

Fixes zaproxy/zaproxy#3727

Before:
before

After:
after

### DIFF
--- a/src/org/parosproxy/paros/model/SiteMap.java
+++ b/src/org/parosproxy/paros/model/SiteMap.java
@@ -52,6 +52,7 @@
 // ZAP: 2016/07/07 Do not add the message to past history if it already belongs to the node
 // ZAP: 2017/01/23: Issue 1800: Alpha sort the site tree
 // ZAP: 2017/06/29: Issue 3714: Added newOnly option to addPath
+// ZAP: 2017/07/09: Issue 3727: Sorting of SiteMap should not include HTTP method (verb) in the node's name
 
 package org.parosproxy.paros.model;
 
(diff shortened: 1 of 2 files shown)


src/org/parosproxy/paros/model/SiteMap.java
```diff
@@ -878,8 +878,17 @@ private int findIndexFor(SiteNode child, SiteNode parent, int idx1, int idx2) {
...
+		String s1 = sn1.getNodeName(false);
+		String s2 = sn2.getNodeName(false);
+		int initialComparison = s1.compareToIgnoreCase(s2);
+
+		if (initialComparison == 0) {
+			s1 = sn1.getNodeName(true);
+			s2 = sn2.getNodeName(true);
+			
+			return s1.compareToIgnoreCase(s2);
+		} else {
```
thc202: The else statement can be removed (the if-then returns).

src/org/parosproxy/paros/model/SiteNode.java
```diff
@@ -230,6 +230,33 @@ public boolean isParentOf (String nodeName) {
...
+	 * @param withMethodPrefix
+	 *            a {@code boolean} specifying whether or not the returned
+	 *            string should include the HTTP Method as a prefix.
+	 * @return the name of the site node
+	 * @see #getNodeName()
+	 * @see #getCleanNodeName()
+	 * @see #getCleanNodeName(boolean)
+	 * @since TODO add version
+	 */
+    public String getNodeName(boolean withMethodPrefix) {
```
thc202: I'd suggest dropping the boolean parameter and return always without the HTTP method (no need to duplicate the existing behaviour). Regarding the name, it could well be `getName()` or if we want to keep the "Node" `getNodeNameWithoutMethod()`.

kingthorin:
Done, done, and done  ⚡️

thc202:
Thanks!

Issue #765, opened 2015-06-04
Ctrl-F key doesnt work on Resend / Manual Request Editor dialogs


zapbot:
```
Ctrl-F should bring up the find dialog in the Resend / Manual Request dialogs
This works in the Request and Response panes.
The right click works, and the menu indicated the ctrl-F key should work .. but it
doesnt.
```

Original issue reported on code.google.com by `psiinon` on 2013-08-14 10:04:08

kingthorin:
Right-click (context) does display and allow "Find...."

kingthorin:
I thought this was going to be as simple as adding the accelerator to `getPopupMenuFind()` https://github.com/zaproxy/zaproxy/blob/d76e69ace9e7d20346070f0d457c11f60054caea/src/org/parosproxy/paros/extension/edit/ExtensionEdit.java#L109-L122 but I'm not having any luck. 

The other hotkeys in the text fields seem to be related to default swing functionality (text selection/editing), so there's not something that's an obvious example.

kingthorin:
Since there is no menu and hotkeys in popups only seem to work when the popup is displayed, I had the idea of adding a button the the toolbar and assigning the hot key to it.

So I've got this working. By adding the following in `org.parosproxy.paros.extension.manualrequest.http.impl.ManualHttpRequestEditorDialog` `getManualSendPanel()`:

```
KeyStroke findKeyStroke = KeyStroke.getKeyStroke(KeyEvent.VK_F, Toolkit.getDefaultToolkit().getMenuShortcutKeyMask()); 
Action findAction = new AbstractAction("findAction") {  
	private static final long serialVersionUID = 8365172573382425660L;
	@Override
	public void actionPerformed(ActionEvent e) {     
         System.out.println("Find");
         FindDialog findDialog = new FindDialog();
         if (getFocusOwner() instanceof JTextComponent) {
        	 findDialog.setLastInvoker((JTextComponent) getFocusOwner());
        	 findDialog.setVisible(true);
         }
    }
}; 

JButton findButton = new JButton("Find");
findButton.getActionMap().put("findAction", findAction);
findButton.getInputMap(JComponent.WHEN_IN_FOCUSED_WINDOW).put(findKeyStroke, "findAction");
requestResponsePanel.addToolbarButton(findButton);
```

[...]

PR #3752, opened 2017-07-13
Issue 765 - Ctrl-F doesn't work in Resend

kingthorin:
Added a Find action to the HttpPanel toolbar (activated by Ctrl-F). 
FindDialog added a constructor that accepts a parent window.

Fixes zaproxy/zaproxy#765

### DIFF
--- a/src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
+++ b/src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
@@ -21,8 +21,11 @@ import java.awt.BorderLayout;
 import java.awt.CardLayout;
 import java.awt.Component;
 import java.awt.FlowLayout;
+import java.awt.KeyboardFocusManager;
+import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
```diff
@@ -91,6 +101,7 @@
     private JPanel moreOptionsComponent;
     private JToolBar toolBarComponents;
     private JToolBar toolBarMoreOptions;
+    private static Action findAction = null;
```
thc202: Could be moved to the same place as the other static variables (null initialisation could be removed).

src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
```diff
@@ -187,6 +200,24 @@ private JPanel getPanelHeader() {
         return panelHeader;
     }
     
+    private Action getFindAction() {
```
thc202: Should be added the static modifier.

kingthorin:
Will do, thanks

kingthorin:
Done & done

thc202:
Thank you.

PR #3762, opened 2017-07-18
Prevent multiple find dialogs per parent

kingthorin:
FindDialog add functionality to track find dialogs & parents, only allowing one open per parent. Deprecate no params constructor (searching code base [incl. extensions] it does not seem used anyway). Set dialog
visible in initialize method.
HttpPanel remove setting visibility.
ExtensionEdit align handling of FindDialog, add FIND_DEFAULT_KEYSTROKE to be used in various locations.
PopupFindMenu updated to leverage new KeyStroke constant.

### DIFF
--- a/src/org/parosproxy/paros/view/FindDialog.java
+++ b/src/org/parosproxy/paros/view/FindDialog.java
@@ -23,6 +23,7 @@
 // ZAP: 2012/05/03 Changed the method find to check if txtComp is null.
 // ZAP: 2014/01/30 Issue 996: Ensure all dialogs close when the escape key is pressed (copy tidy up)
 // ZAP: 2017/07/12 Issue 765: Add constructor with window parent, to facilitate ctrl-F in various HttpPanels
+// ZAP: 2017/07/17: Prevent opening multiple dialogs per parent.
 
 package org.parosproxy.paros.view;
 
@@ -33,6 +34,9 @@ import java.awt.GridBagLayout;
 import java.awt.HeadlessException;
 import java.awt.Toolkit;
 import java.awt.Window;
+import java.awt.event.WindowAdapter;
+import java.awt.event.WindowEvent;
+import java.util.Map;
 
 import javax.swing.JButton;
 import javax.swing.JFrame;
@@ -40,6 +44,7 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 import javax.swing.text.JTextComponent;
 
+import org.apache.commons.collections.map.ReferenceMap;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.extension.AbstractDialog;
 import org.parosproxy.paros.model.Model;
@@ -48,6 +53,9 @@ import org.zaproxy.zap.utils.ZapTextField;
 public class FindDialog extends AbstractDialog {
 
 	private static final long serialVersionUID = -3223449799557586758L;
+	
+	@SuppressWarnings("unchecked")
+	private static Map<Object, FindDialog> parentsMap = new ReferenceMap(ReferenceMap.WEAK, ReferenceMap.HARD);
 
 	private JPanel jPanel = null;
 	private JButton btnFind = null;
@@ -65,7 +73,10 @@ public class FindDialog extends AbstractDialog {
 
     /**
      * @throws HeadlessException
+     * @deprecated TODO add version, use #FindDialog(Window, boolean) instead
+     * @see #getDialog(Window, boolean)
      */
+    @Deprecated
     public FindDialog() throws HeadlessException {
         super();
  		initialize();
@@ -75,7 +86,10 @@ public class FindDialog extends AbstractDialog {
      * @param arg0
      * @param arg1
      * @throws HeadlessException
+     * @deprecated TODO add version, use #FindDialog(Window, boolean) instead 
+     * @see #getDialog(Window, boolean)
      */
+    @Deprecated
     public FindDialog(Frame arg0, boolean arg1) throws HeadlessException {
         super(arg0, arg1);
         initialize();
@@ -87,6 +101,8 @@ public class FindDialog extends AbstractDialog {
      * @param parent the parent window of the FindDialog.
      * @param modal whether or not this FindDialog should be modal
      * @throws HeadlessException
+     * @see #getDialog(Window, boolean)
+     * @since TODO add version
      */
     public FindDialog(Window parent, boolean modal) throws HeadlessException {
     	super(parent, modal);
@@ -108,6 +124,42 @@ public class FindDialog extends AbstractDialog {
         txtFind.requestFocus();
         this.getRootPane().setDefaultButton(btnFind);
         pack();
+        this.setVisible(true);
+	}
+	
+	private static Map<Object, FindDialog> getParentsMap() {
+		return parentsMap;
+	}
+	
+	/**
+	 * Get the FindDialog for the parent if there is one or creates and returns a new one.
+	 * 
+	 * @param parent the parent Window (or Frame) for this FindDialog
+	 * @param modal a boolean indicating whether the FindDialog should ({@code true}), 
+	 * or shouldn't ({@code false}) be modal.
+	 * @return The existing FindDialog for the parent (if there is one), or a new FindDialog.
+	 * @since TODO add version
+	 */
+	public static FindDialog getDialog(Window parent, boolean modal) {
+		FindDialog activeDialog = getParentsMap().get(parent);
+		if (activeDialog != null) {
+			activeDialog.getTxtFind().requestFocus();
+			return activeDialog;
+		}
+		FindDialog newDialog = new FindDialog(parent, modal); 
+		getParentsMap().put(parent, newDialog);
+		newDialog.addWindowListener(new WindowAdapter() {
+			@Override
+			public void windowClosed(WindowEvent e) {
+				getParentsMap().remove(parent);
+			}
+		});
+		return newDialog;
+	}
+
+	private void discard() {
+		this.setVisible(false);
+		this.dispose();
 	}
 	
 	private void find() {
@@ -224,7 +276,8 @@ public class FindDialog extends AbstractDialog {
 				@Override
 				public void actionPerformed(java.awt.event.ActionEvent e) {
 
-				    FindDialog.this.setVisible(false);
+					FindDialog.this.discard();
+					FindDialog.this.dispatchEvent(new WindowEvent(FindDialog.this, WindowEvent.WINDOW_CLOSING));
 				}
 			});
 
--- a/src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
+++ b/src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
@@ -22,10 +22,8 @@ import java.awt.CardLayout;
 import java.awt.Component;
 import java.awt.FlowLayout;
 import java.awt.KeyboardFocusManager;
-import java.awt.Toolkit;
 import java.awt.event.ActionEvent;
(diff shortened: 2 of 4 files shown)


src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -108,6 +128,34 @@ private void initialize() {
         txtFind.requestFocus();
         this.getRootPane().setDefaultButton(btnFind);
         pack();
+        this.setVisible(true);
+        
+        this.addWindowListener(new WindowAdapter() {
+            @Override
+            public void windowClosed(WindowEvent e) {
+                getParentsList().remove(getParentName());
```
kingthorin: Should I call `super.windowClosed(e)`?

thc202: It's not necessary (base method does not do anything).

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -65,10 +73,14 @@ public void setLastInvoker(JTextComponent lastInvoker) {
 
     /**
      * @throws HeadlessException
+     * @deprecated TODO add version
+     * @see {@link #FindDialog(Frame, boolean)}
```
thc202: The `@link` tag can be dropped (same for following line).

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -87,16 +99,24 @@ public FindDialog(Frame arg0, boolean arg1) throws HeadlessException {
      * @param parent the parent window of the FindDialog.
      * @param modal whether or not this FindDialog should be modal
      * @throws HeadlessException
+     * @since TODO add version
      */
     public FindDialog(Window parent, boolean modal) throws HeadlessException {
     	super(parent, modal);
-    	initialize();
+
+    	initialize(parent.toString());
```
thc202: `toString()` is not reliable, for example, if the parent window is resized it will be a different "name" (thus opening another find dialogue).

src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
```diff
@@ -210,7 +210,6 @@ public void actionPerformed(ActionEvent e) {
                     if (focusOwner instanceof JTextComponent) {
                         FindDialog findDialog = new FindDialog(SwingUtilities.getWindowAncestor(focusOwner), false);
```
thc202: Instead of creating a new `FindDialog` it should be requested one (e.g. `FindDialog.getDialog(Window)`, that returns the existing one or creates one if none exists), otherwise the following statement has no effect if the dialogue is already open.

src/org/parosproxy/paros/view/FindDialog.java
kingthorin: Well I can use `parent.getName()` but that seemed to return generic things like "frame0" and I was concerned about a collision.

src/org/parosproxy/paros/view/FindDialog.java
kingthorin: I can change it no problem, but was trusting auto-complete in Eclipse

src/org/zaproxy/zap/extension/httppanel/HttpPanel.java
kingthorin: Change usage in `ExtensionEdit` as well?

thc202: Yeah, that could be changed as well.

src/org/parosproxy/paros/view/FindDialog.java
thc202: Why not use the parent itself with a `ReferenceMap` (with weak keys/values)? That would ensure there's no collisions.

src/org/parosproxy/paros/view/FindDialog.java
thc202: Seems to be a wrong auto-complete suggestion then, I get warnings with JavaDoc and Eclipse:
```
FindDialog.java:77: error: unexpected content
     * @see {@link #FindDialog(Frame, boolean)}
       ^
FindDialog.java:78: error: unexpected content
     * @see {@link #FindDialog(Window, boolean)}
       ^
FindDialog.java:81: warning - Tag @see:illegal character: "123" in "{@link #FindDialog(Frame, boolean)}"
FindDialog.java:81: warning - Tag @see:illegal character: "64" in "{@link #FindDialog(Frame, boolean)}"
FindDialog.java:81: warning - Tag @see:illegal character: "123" in "{@link #FindDialog(Window, boolean)}"
FindDialog.java:81: warning - Tag @see:illegal character: "64" in "{@link #FindDialog(Window, boolean)}"

```

kingthorin: It must not have been considering what was already on the line, fixed.

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -87,16 +101,25 @@ public FindDialog(Frame arg0, boolean arg1) throws HeadlessException {
...
     }
 
 	/**
 	 * This method initializes this
 	 */
-	private void initialize() {
+	private void initialize(Object parent) {
+		if (getParentsList().containsKey(parent)) {
+			this.discard();
+			getParentsList().get(parent).requestFocus();
```
kingthorin: This seems to have stopped working :(

thc202: Yes, that's expected because the `FindDialog` instances are now being reused. This statement needs to be moved to the `getDialog` methods.

thc202: This if statement is no longer needed, right?

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -48,13 +53,17 @@
...
+	@SuppressWarnings("unchecked")
+	private static Map<Object, FindDialog> parentsList = new ReferenceMap(ReferenceMap.WEAK, ReferenceMap.WEAK);
 
 	private JPanel jPanel = null;
 	private JButton btnFind = null;
 	private JButton btnCancel = null;
 	private ZapTextField txtFind = null;
 	private JPanel jPanel1 = null;
     private JTextComponent lastInvoker = null;
+    private Object parentObject;
```
thc202: This could be replaced with calls to `getParent()`?

src/org/parosproxy/paros/view/FindDialog.java
thc202: Might worth changing the request focus to `txtFind` instead of the dialogue.

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -108,6 +125,47 @@ private void initialize() {
...
+	}
+	
+	private static Map<Object, FindDialog> getParentsList() {
+		return parentsList;
+	}
+	
+	public static FindDialog getDialog(Window parent, boolean modal) {
+		if (getParentsList().containsKey(parent)) {
+			FindDialog activeDialog = getParentsList().get(parent);
+			activeDialog.getTxtFind().requestFocusInWindow();
```
kingthorin: Re: https://github.com/zaproxy/zaproxy/pull/3762#discussion_r128081747

Still doesn't work (on Win7/java8_131 anyway).

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -108,6 +124,47 @@ private void initialize() {
...
```
kingthorin: Re: https://github.com/zaproxy/zaproxy/pull/3762#discussion_r128081747

Still doesn't work (on Win7/java8_131 anyway).

kingthorin: Re: https://github.com/zaproxy/zaproxy/pull/3762#discussion_r128081747

Still doesn't work (on Win7/java8_131 anyway).

kingthorin:
More thoughts:

- Instead of statically assigning ctrl-F should I get getting whatever is set in the keyboard options?
- The context menu should display the same key combination (whether that's ctrl-F or the keyboard option setting). So that the user sees it and might use it in the future.
- When a text component is selected and the user goes to the Edit > Find... menu item, nothing happens because the focus is wrong (menu gets focus not the text component) [Do we just ignore that? Should I open another issue? Have thoughts on a simple solution?].

src/org/parosproxy/paros/view/FindDialog.java
thc202: Could you try with `requestFocus()`, that's the method used before.

kingthorin: Will do

thc202:
1. We could allow to register "core" actions, that are later configured with the custom keyboard shortcut (similar to `ZapMenuItem`s).
2. Sounds good :)
3. I think that depends on the component, it works with the Output tab (and some other text fields/areas). Don't you see the same? (`FindDialog` attempts to get the previously focused component of the frame.)

src/org/parosproxy/paros/view/FindDialog.java
kingthorin: That fixed it 👍

kingthorin:
No. 2 done. A constant has been added so that by default there is consistency. A user can still customize away the consistency, but for now that's acceptable.
No. 3 fixed.

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -48,6 +53,9 @@
 public class FindDialog extends AbstractDialog {
 
 	private static final long serialVersionUID = -3223449799557586758L;
+	
+	@SuppressWarnings("unchecked")
+	private static Map<Object, FindDialog> parentsList = new ReferenceMap(ReferenceMap.WEAK, ReferenceMap.HARD);
```
thc202: This should be renamed, the name implies a list but it's a map (same for the getter).

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -108,6 +124,48 @@ private void initialize() {
         txtFind.requestFocus();
         this.getRootPane().setDefaultButton(btnFind);
         pack();
+        this.setVisible(true);
+	}
+	
+	private void registerParent(Object parent) {
```
thc202: This could be inlined, it's just used in one place (also, it could just `put` without needing to check if it's not contained, we already know it isn't).

src/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -108,6 +124,48 @@ private void initialize() {
...
+	 * 
+	 * @param parent the parent Window (or Frame) for this FindDialog
+	 * @param modal a boolean indicating whether the FindDialog should ({@code true}), 
+	 * or shouldn't ({@code false}) be modal.
+	 * @return The existing FindDialog for the parent (if there is one), or a new FindDialog.
+	 * @since TODO add version
+	 */
+	public static FindDialog getDialog(Window parent, boolean modal) {
+		if (getParentsList().containsKey(parent)) {
+			FindDialog activeDialog = getParentsList().get(parent);
```
thc202: Might worth moving this statement to before the if (and check that's non-`null` in the if statement), that ensures the object does not get garbage collected in the meantime.

kingthorin:
Thanks @thc202 I'll take care of that tonight

thc202:
There are still two comments to address, missing ZAP comments (`ExtensionEdit` and `PopupFindMenu`) and name of parents getter in `FindDialog`.

kingthorin:
Done, done & done :zap:

thc202:
Thank you!

Issue #3763, opened 2017-07-18
ZAP baseline scan Docker randomly fails with `ValueError: need more than 1 value to unpack`


BookGin:
When running [the baseline scan](https://github.com/zaproxy/zaproxy/wiki/ZAP-Baseline-Scan#usage), a Python error happens **randomly**.
```
$ docker run -t owasp/zap2docker-stable zap-baseline.py -t https://www.example.com

Usage of loopback devices is strongly discouraged for production use. Either use `--storage-opt dm.thinpooldev` or use `--storage-opt dm.no_warn_on_loop_devices=true` to suppress this warning.
_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created.
Jul 18, 2017 3:42:00 AM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
Traceback (most recent call last):
  File "/zap//zap-baseline.py", line 617, in <module>
    main(sys.argv[1:])
  File "/zap//zap-baseline.py", line 580, in main
    except IOError as (errno, strerror):
ValueError: need more than 1 value to unpack
```

Here is the information of my system:
```
Linux 3.10 x86_64 GNU/Linux
Red Hat Enterprise Linux Server 7.1 (Maipo)
Docker version 1.7.1, build 446ad9b/1.7.1
```

thc202:
Most likely fixed in #3474. Could you try the weekly image?

BookGin:
@thc202 Thanks for your quick response.

However, another error occurred when trying the weekly image:
```
_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created.
Jul 19, 2017 2:35:04 AM java.util.prefs.FileSystemPreferences$1 run
INFO: Created user preferences directory.
2017-07-19 02:36:59,364 I/O error(5): Failed to connect to ZAP after 120 seconds
ERROR Failed to connect to ZAP after 120 seconds
```
I cannot verify if it's fixed.

psiinon:
@BookGin does this happen consistently?
Can you post the command you used to start the baseline scan?
We run the weekly baseline against over 100 sites every day at Mozilla and its working fine right now :)

BookGin:
@psiinon 
For the weekly and live docker images, the timeout error happens consistently:
```
$ docker run --rm -t owasp/zap2docker-weekly zap-baseline.py -t http://mywebsite.com:8000/test.html
[...]

PR #3766, opened 2017-07-19
Restore log dump on all IOError in Docker scripts

thc202:
Change Docker scripts zap-api-scan.py, zap-baseline.py, and
zap-full-scan.py to dump the log on all IOError caught (a previous
change inadvertently moved the statement to the else branch).


thc202:
Hopefully this should help with #3763, by giving more details about ZAP's loading state.

PR #3746, opened 2017-07-11
docker: Check local zap python api client is up to date

g-k:
Fixes: #3356 

The baseline script should work offline, so these checks should suppress errors that exit. 

Note: this adds setuptools as a dependency but it's usually installed already.

Functional Tests:

- [x] without a local zap client: `ImportError` (it might be nice to make this a more helpful warning)

```
~/zaproxy - [3356-baseline-check-python-api-version●] » pip freeze | grep -i zap
~/zaproxy - [3356-baseline-check-python-api-version●] » python build/docker/zap-baseline.py
Traceback (most recent call last):
  File "build/docker/zap-baseline.py", line 51, in <module>
    import zapv2
ImportError: No module named zapv2
```

- [x] with outdated local zap client: update warning

```
~/zaproxy - [3356-baseline-check-python-api-version●] » pip freeze | grep -i zap
python-owasp-zap-v2.4==0.0.9
~/zaproxy - [3356-baseline-check-python-api-version●] » python build/docker/zap-baseline.py
2017-07-11 17:47:08,158 A newer version of python_owasp_zap_v2.4 is available. Please run 'pip install -U python_owasp_zap_v2.4' to update to the latest version.
...
```

- [x] with updated local zap client: no warning

```
~/zaproxy - [3356-baseline-check-python-api-version●] » pip freeze | grep -i zap
python-owasp-zap-v2.4==0.0.11
~/zaproxy - [3356-baseline-check-python-api-version●] » python build/docker/zap-baseline.py
Usage: zap-baseline.py -t <target> [options]
```

- [x] with develop/HEAD local zap client version: no warning

```
~/zaproxy - [3356-baseline-check-python-api-version●] » pip freeze | grep -i zap
python-owasp-zap-v2.4==0.0.12
~/zaproxy - [3356-baseline-check-python-api-version●] » python build/docker/zap-baseline.py
Usage: zap-baseline.py -t <target> [options]
    -t target         target URL including the protocol, eg https://www.example.com
```

- [x] when offline: error loading current version

```
~/zaproxy - [3356-baseline-check-python-api-version●] » python build/docker/zap-baseline.py
2017-07-11 17:52:29,129 Error fetching latest zap python API client version: <urlopen error [Errno 8] nodename nor servname provided, or not known>
Usage: zap-baseline.py -t <target> [options]
...
```

PR #3767, opened 2017-07-19
Tox smoketest for scan scripts

g-k:
refs: https://github.com/zaproxy/zaproxy/pull/3746#discussion_r128315445

Initial/WIP PR for smoketesting the docker scan scripts blocked on how to exit with code 0.

Changes:
* zap-api-scan python3 compatibility fix
* tox file for making sure the scripts run (run `tox` from `build/docker`)

Questions:

* currently this fails with exit code 3. Possible solutions:
  * test against an actual API or app (probably slow but good to test)
  * have tox ignore the exit code (makes the tests useless)
  * add an option (like `--version`, `--usage`, `--help` or a shorthand version)  to the scan scripts that exits with 0
* should we run this in CI?


build/docker/tox.ini
```diff
@@ -0,0 +1,9 @@
+[tox]
+skipsdist = True
+envlist = py26,py27,py34,py35,py36
+[testenv]
+deps = -rrequirements.txt
```
thc202: It's missing the `requirements.txt` file.

thc202:
I guess we could start with the 3rd solution?

Worth including that if it doesn't add much time to the build(?).

g-k:
Updated to add a `-h` option for help and a requirements file.

The whitespace changes could go in a separate PR if that'd be preferable.

This still isn't running from CI. How do we want to do that?

Should we:

1. run tox from ant
1. run it from travis CI
1. run it from Circle CI?
1. something else

I'm leaning toward 1. but haven't looked into how to do that. Should that be a separate PR?

kingthorin:
@g-k can you address the conflicts?

g-k:
@kingthorin yep, rebased on develop.

build/docker/zap-api-scan.py
```diff
@@ -244,13 +247,10 @@ def main(argv):
     elif config_url:
         # load config file from url
         try:
-            load_config(urllib2.urlopen(config_url), config_dict, config_msg, out_of_scope_dict)
+            load_config(urlopen(config_url), config_dict, config_msg, out_of_scope_dict)
         except ValueError as e:
             logging.warning(e)
             sys.exit(3)
-        except:
```
thc202: Was this removed on purpose?

thc202:
I think we can leave the CI changes to other PR (in the meantime we can run it manually, the scripts don't change that often anyway). Regarding the option, I guess we can go with 2? (We can run tox directly without Ant, any advantages for the latter?)

build/docker/zap-api-scan.py
g-k: No, I got confused resolving a merge conflict.

g-k: Fixed.

g-k:
> I think we can leave the CI changes to other PR

👍 

> I guess we can go with 2? (We can run tox directly without Ant, any advantages for the latter (i.e. run with Ant)?)

I guess it's slightly simpler, but would make tests slower.

2. sounds good.

thc202:
Thanks @g-k and @kingthorin for bumping :)

PR #3770, opened 2017-07-20
Clear dockerfiles 🐋

flibustier:
List of changes : 
- Remove `MAINTAINER` because it's deprecated
- 2 dependencies were duplicated (`x11vnc` and `xvfb`)
- `pip install python-owasp-zap-v2.4` was already satisfied by `zapcli`
- add `-na` flag to `wget` for a better output
- remove `export JAVA_HOME` from script because it's already defined in Dockerfile
- remove whitespaces at end of lines

### DIFF
--- a/build/docker/Dockerfile-stable
+++ b/build/docker/Dockerfile-stable
@@ -1,6 +1,6 @@
 # This dockerfile builds the zap stable release
 FROM ubuntu:16.04
-MAINTAINER Simon Bennetts "psiinon@gmail.com"
+LABEL maintainer="psiinon@gmail.com"
 
 RUN apt-get update && apt-get install -q -y --fix-missing \
 	make \
@@ -14,8 +14,6 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	xmlstarlet \
 	unzip \
 	git \
-	x11vnc \
-	xvfb \
 	openbox \
 	xterm \
 	net-tools \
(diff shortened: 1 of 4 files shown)


build/docker/Dockerfile-stable
```diff
@@ -30,25 +27,20 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 RUN pip install --upgrade pip
 RUN gem install zapr
 RUN pip install zapcli
-# Install latest dev version of the python API
-RUN pip install python-owasp-zap-v2.4
```
psiinon: Removing this makes me a bit nervous - zapcli is not maintained by us, so I dont like the idea of us relying on its dependencies...

flibustier:
Oh I see… In this case, we should run this command before zapcli, to avoid a useless command

build/docker/Dockerfile-stable
flibustier: Oh I see… In this case, we should run this command before zapcli, to avoid a useless command

Grunny: One advantage of leaving this after zapcli is that it makes sure the latest version of `python-owasp-zap-v2.4` is installed even if `zapcli` is a version behind for example. I try to keep the version of the dependency used by zapcli constantly up to date since there aren't usually breaking changes, but I sometimes am a bit slow, and this makes sure people can use the latest API client library in their own scripts. :)

flibustier: Seems legit, I will put it back!

build/docker/Dockerfile-stable
```diff
@@ -27,28 +24,22 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/*
 
-RUN pip install --upgrade pip
 RUN gem install zapr
-RUN pip install zapcli
-# Install latest dev version of the python API
-RUN pip install python-owasp-zap-v2.4
+RUN pip install --upgrade pip zapcli python-owasp-zap-v2.4
```
thc202: Same change could be applied to other files?

build/docker/Dockerfile-stable
```diff
@@ -65,15 +56,11 @@ ENV ZAP_PATH /zap/zap.sh
 ENV ZAP_PORT 8080
 ENV HOME /home/zap/
 
-
-COPY zap-x.sh /zap/ 
-COPY zap-* /zap/ 
-COPY zap_* /zap/ 
+COPY zap* /zap/
```
thc202: Would you mind applying this change to `Dockerfile-live`?

build/docker/Dockerfile-weekly
```diff
@@ -36,16 +33,13 @@ RUN useradd -d /home/zap -m -s /bin/bash zap
 RUN echo zap:zap | chpasswd
 RUN mkdir /zap 
 WORKDIR /zap
-RUN chown zap /zap && \
-	chgrp zap /zap
+RUN chown zap:zap /zap
```
thc202: Same here (in relation to `Dockerfile-stable`).

build/docker/Dockerfile-stable
flibustier: Right!

build/docker/Dockerfile-stable
flibustier: My bad, I will fix it!

build/docker/Dockerfile-stable
```diff
@@ -27,28 +24,22 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
...
-RUN pip install zapcli
-# Install latest dev version of the python API
-RUN pip install python-owasp-zap-v2.4
+RUN pip install --upgrade pip zapcli python-owasp-zap-v2.4
 
 RUN useradd -d /home/zap -m -s /bin/bash zap 
 RUN echo zap:zap | chpasswd
-RUN mkdir /zap 
+RUN mkdir /zap && chown zap:zap /zap
+
```
thc202: It has a space at the end of the line.

thc202:
Resolved the conflicts, addressed the last comments and added the maintainer label (same as in #3808).

thc202:
@flibustier thanks! How would you like to be credited?

flibustier:
Hi ! Sorry I wasn't available this last few days… Thanks for merging! 👍 
For the credit, it's at your convenience :)

kingthorin:
Check the page he linked, let us know what name/handle/whatever you'd like listed.

flibustier:
Okay I get it! "Jon (@flibustier)" will be fine, thank you very much :)

thc202:
Thank you!

PR #3781, opened 2017-07-25
Allow to hook HttpSenderListener

thc202:
Change class ExtensionHook to allow to add HttpSenderListener, so it can
be automatically added/removed by core code.
Change ExtensionLoader to add/remove the HttpSenderListener when the
extensions are initialised/unloaded.
Change core extensions to use the new method, less changes once moved
to an add-on (e.g. ExtensionForcedUser).

### DIFF
--- a/src/org/parosproxy/paros/extension/ExtensionHook.java
+++ b/src/org/parosproxy/paros/extension/ExtensionHook.java
@@ -29,6 +29,7 @@
 // ZAP: 2016/05/30 Allow to add AddOnInstallationStatusListener
 // ZAP: 2016/05/30 Issue 2494: ZAP Proxy is not showing the HTTP CONNECT Request in history tab
 // ZAP: 2016/08/18 Allow to add ApiImplementor
+// ZAP: 2017/07/25 Allow to add HttpSenderListener.
 
 package org.parosproxy.paros.extension;
 
@@ -48,6 +49,7 @@ import org.zaproxy.zap.extension.AddonFilesChangedListener;
 import org.zaproxy.zap.extension.api.ApiImplementor;
 import org.zaproxy.zap.extension.AddOnInstallationStatusListener;
 import org.zaproxy.zap.model.ContextDataFactory;
+import org.zaproxy.zap.network.HttpSenderListener;
 import org.zaproxy.zap.view.SiteMapListener;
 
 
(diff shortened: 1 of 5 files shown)


src/org/parosproxy/paros/extension/ExtensionHook.java
```diff
@@ -368,4 +380,39 @@ public void addApiImplementor(ApiImplementor apiImplementor) {
...
+     * @throws IllegalArgumentException if the given {@code httpSenderListener} is {@code null}.
+     * @since TODO add version
+     */
+    public void addHttpSenderListener(HttpSenderListener httpSenderListener) {
+        if (httpSenderListener == null) {
+            throw new IllegalArgumentException("Parameter httpSenderListener must not be null.");
+        }
+
+        if (httpSenderListeners == null) {
+            httpSenderListeners = new ArrayList<>();
```
kingthorin: Should this be parameterized as `ArrayList<HttpSenderListener>()`?

src/org/parosproxy/paros/extension/ExtensionHook.java
```diff
@@ -368,4 +380,39 @@ public void addApiImplementor(ApiImplementor apiImplementor) {
...
+        httpSenderListeners.add(httpSenderListener);
+    }
+
+    /**
+     * Gets the {@link HttpSenderListener}s added to this hook.
+     *
+     * @return an unmodifiable {@code List} containing the added {@code HttpSenderListener}s, never {@code null}.
+     * @since TODO add version
+     */
+    List<HttpSenderListener> getHttpSenderListeners() {
```
kingthorin: `public`?

src/org/parosproxy/paros/extension/ExtensionHook.java
thc202: It is, the type can be omitted when inferred by the compiler.

src/org/parosproxy/paros/extension/ExtensionHook.java
thc202: This method is not intended to be used by "external" classes.

src/org/parosproxy/paros/extension/ExtensionHook.java
kingthorin: Ok

src/org/parosproxy/paros/extension/ExtensionHook.java
kingthorin: K

PR #3786, opened 2017-07-26
Fix unicode encoding errors when writing reports

### LINKED ISSUE #3780: zap-baseline.py errors with UnicodeEncodeError: 'ascii' codec can't encode characters in position 243223-243224: ordinal not in range(128)
When running a scan of our application using zap-baseline.py within the owasp/zap2docker-stable:latest docker container, I get the following error:
```
Traceback (most recent call last):
  File "/zap/zap-baseline.py", line 561, in main
    f.write (zap.core.htmlreport())
UnicodeEncodeError: 'ascii' codec can't encode characters in position 243223-243224: ordinal not in range(128)
...
ERROR <type 'exceptions.UnicodeEncodeError'>
```
I get this error when requesting an html report (-r) or a simple error when requesting a markdown report (-w). XML seems to be ok, although I suspect the dev's won't be happy with that.

The full command line is:
```
/zap/zap-baseline.py -a -j -m 30 -t ${url} -r report.html
```

Grunny:
Fix unicode encoding errors when writing reports in a way that supports
both Python 2.7 and Python 3.5. The string in Python 2.7 is a byte string,
and the default system encoding is ascii, which is why it breaks with just
mode w as it tries to encode a utf-8 string as ascii. Changing the mode to
wb fixes this as then it doesn't try to encode it further when writing. But,
in Python 3, the string is unicode so needs to be encoded as bytes before
being written with mode b.

Fixes #3780

/cc @omerlh

### DIFF
--- a/build/docker/zap-api-scan.py
+++ b/build/docker/zap-api-scan.py
@@ -465,18 +465,15 @@ def main(argv):
 
             if report_html:
                 # Save the report
-                with open(base_dir + report_html, 'w') as f:
-                    f.write(zap.core.htmlreport())
+                write_report(base_dir + report_html, zap.core.htmlreport())
 
             if report_md:
                 # Save the report
-                with open(base_dir + report_md, 'w') as f:
-                    f.write(zap.core.mdreport())
+                write_report(base_dir + report_md, zap.core.mdreport())
 
             if report_xml:
                 # Save the report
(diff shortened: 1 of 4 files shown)


Grunny:
See https://github.com/Grunny/zap-cli/pull/31 and expand the comments for background on the fix.

thc202:
Thanks!

Issue #3531, opened 2017-05-13
Unreleased zap-api-scan script gets stuck on "Records to passive scan" seemingly indefinitely


stephendonner:
Using the following revision of https://hub.docker.com/r/psiinon/zap2docker-api/ - which is built from https://github.com/psiinon/zaproxy/blob/418d4273716fb7f62bafc5cd5ebd6d40531eeadf/build/docker/zap-api-scan.py

Steps (for me) to Reproduce:

1. docker pull psiinon/zap2docker-api
2. docker run -t psiinon/zap2docker-api zap-api-scan.py -t http://petstore.swagger.io/v2/swagger.json -d
3. After you see "Active Scan complete", followed by "Records to Scan...", I see the following seemingly endless loop:

[timestamp] Records to passive scan: 6893
<snip>
[timestamp + 1 hour, etc.] Records to passive scan: 6893
...

I've had to CTRL-C these runs when this occurs

Attached are both a screenshot of the repeating/stuck console output, as well as the more-relevant 
<img width="1158" alt="screen shot 2017-05-12 at 10 12 41 am" src="https://cloud.githubusercontent.com/assets/387249/26020731/3108b5be-3737-11e7-8535-78f8d95fedd8.png">
zap.out log when doing a ```kill -3 [java PID]```

[zap.out.txt](https://github.com/zaproxy/zaproxy/files/998222/zap.out.txt)

thc202:
Thanks for raising the issue.

Worth noting that this is not caused by the new functionality/script. It seems ZAP is stuck passive scanning a message (probably one with a big response), relevant thread/stack trace:
```
"ZAP-PassiveScanner" #20 daemon prio=5 os_prio=0 tid=0x00007f7d50920800 nid=0x56 runnable [0x00007f7d4ed69000]
   java.lang.Thread.State: RUNNABLE
	at java.util.regex.Pattern$CharProperty.match(Pattern.java:3776)
	at java.util.regex.Pattern$Curly.match0(Pattern.java:4260)
	at java.util.regex.Pattern$Curly.match(Pattern.java:4234)
	at java.util.regex.Pattern$BmpCharProperty.match(Pattern.java:3798)
	at java.util.regex.Pattern$Start.match(Pattern.java:3461)
	at java.util.regex.Matcher.search(Matcher.java:1248)
	at java.util.regex.Matcher.find(Matcher.java:637)
[...]

PR #3787, opened 2017-07-26
Added passive scan timeout

psiinon:
Very occationally the passive scan seems to get locked on specific response, and then the baseline just hangs forever.

### DIFF
--- a/build/docker/zap-baseline.py
+++ b/build/docker/zap-baseline.py
@@ -54,7 +54,6 @@ from zapv2 import ZAPv2
 from zap_common import *
 
 
-timeout = 120
 config_dict = {}
 config_msg = {}
 out_of_scope_dict = {}
@@ -93,6 +92,7 @@ def usage():
     print ('    -n context_file   context file which will be loaded prior to spidering the target')
     print ('    -p progress_file  progress file which specifies issues that are being addressed')
     print ('    -s                short output format - dont show PASSes or example URLs')
+    print ('    -T                max time in minutes to wait for ZAP to start and the passive scan to run')
     print ('    -z zap_options    ZAP command line options e.g. -z "-config aaa=bbb -config ccc=ddd"')
     print ('')
     print ('For more details see https://github.com/zaproxy/zaproxy/wiki/ZAP-Baseline-Scan')
@@ -121,6 +121,7 @@ def main(argv):
     zap_ip = 'localhost'
     zap_options = ''
     delay = 0
+    timeout = 0
 
     pass_count = 0
     warn_count = 0
@@ -133,7 +134,7 @@ def main(argv):
     check_zap_client_version()
 
     try:
-        opts, args = getopt.getopt(argv, "t:c:u:g:m:n:r:w:x:l:daijp:sz:P:D:")
+        opts, args = getopt.getopt(argv, "t:c:u:g:m:n:r:w:x:l:daijp:sz:P:D:T:")
     except getopt.GetoptError as exc:
         logging.warning('Invalid option ' + exc.opt + ' : ' + exc.msg)
         usage()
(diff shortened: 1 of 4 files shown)


build/docker/zap-baseline.py
```diff
@@ -314,7 +314,8 @@ def main(argv):
                 time.sleep(5)
                 logging.debug('Delay passive scan check ' + str(delay - (datetime.now() - start_scan).seconds) + ' seconds')
 
-        zap_wait_for_passive_scan(zap)
+        # Allow twice as long as the time taken to spider for the passive scanning
+        zap_wait_for_passive_scan(zap, mins * 60 * 2)
```
psiinon: Maybe this is too short - I've checked on our daily scans and for some sites is can regularly take 10-20 mins for the passive scan to complete. Note that the box is being used for other things. Suggestions for a suitable multiplyer or should this be another param to the baseline script?

kingthorin: I think it's a fine place to start.

I wonder if we could/should integrate something like this into the core? If we have a way to abandon or terminate a particular plugin (or passive scanning of a request in general) that could be a way to handle things like the passive scanners that use Hirshberg. 
(Like: _Passive scanning a message x took longer than the configured timeout of 120sec, abandoning assessment of passive plugin y._)

build/docker/zap_common.py
```diff
@@ -298,13 +298,20 @@ def zap_active_scan(zap, target, policy):
...
     logging.debug('Records to scan...')
+    time_taken = 0
     while (int(zap.pscan.records_to_scan) > 0):
         logging.debug('Records to passive scan : ' + zap.pscan.records_to_scan)
         time.sleep(2)
-    logging.debug('Passive scanning complete')
+        time_taken += 2
+        if timeout and time_taken > timeout:
+          break
+    if timeout and time_taken > timeout:
```
kingthorin: This is the same as the previous `if`, can the logging be added to the same block as the break?

build/docker/zap-baseline.py
psiinon: +1 we should also log an error with the plugin class so we can see if there any that are consistently failing

kingthorin: Should it be specifiable by the commandline (in case someone determines it is too short for their use)?

build/docker/zap_common.py
psiinon: thats was how I originally wrote it, but then you get the 'Passive scanning complete' message even when its timed out, which didnt seem right :/

build/docker/zap-baseline.py
psiinon: I'm a bit concerned that we could ed up with too many cmdline options, but in this case I think it might well be the best option

build/docker/zap_common.py
kingthorin: Set a boolean "timedout = false" or something like that and check it for the completed message?

build/docker/zap-baseline.py
thc202: Other scripts would benefit from this check (e.g. #3531, but if this is going to be added to core then it might not be necessary).

The timeout could also be dependent on the number of messages to scan?

kingthorin:
The earlier conversations weren't meant to halt this one. I think it's probably fine to go as is for the time being...

psiinon:
I'm still planning on reworking this - I think a new cmdline option would be better, one that also applies to starting ZAP - I've had a couple of cases where thats hung too.

psiinon:
Reworked - added a '-T' option to all of the packaged scripts.

build/docker/zap-full-scan.py
```diff
@@ -285,7 +288,7 @@ def main(argv):
     try:
         zap = ZAPv2(proxies={'http': 'http://' + zap_ip + ':' + str(port), 'https': 'http://' + zap_ip + ':' + str(port)})
 
-        wait_for_zap_start(zap, timeout)
+        wait_for_zap_start(zap, , timeout * 60)
```
thc202: It has an extra comma before `timeout`.

psiinon: Arg, cnp ;)
Fixed

thc202:
Thank you!

PR #3791, opened 2017-07-27
Allow to set fields to DynamicFieldsPanel

thc202:
Change DynamicFieldsPanel to allow to set the fields after creating the
panel allowing to dynamically change the fields shown (instead of
require removing/creating/adding the panel each time the fields change).

### DIFF
--- a/src/org/zaproxy/zap/view/DynamicFieldsPanel.java
+++ b/src/org/zaproxy/zap/view/DynamicFieldsPanel.java
@@ -13,6 +13,8 @@ import org.zaproxy.zap.utils.ZapTextField;
 
 public class DynamicFieldsPanel extends JPanel {
 
+	private static final String[] NO_FIELDS = new String[0];
+
 	/** The Constant serialVersionUID. */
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/view/DynamicFieldsPanel.java
```diff
@@ -25,19 +27,53 @@
...
+		setFields(requiredFields, optionalFields);
+	}
+
+	/**
+	 * Sets the (required) fields that should be shown in the panel.
+	 * <p>
+	 * Any fields previously set are removed.
+	 *
+	 * @param requiredFields the required fields.
+	 * @throws IllegalArgumentException if the given argument is {@code null}.
```
kingthorin: Should `see` the other method? 

At first I started to comment that the exception wasn't declared or thrown, but then I actually read the code :)

thc202: Change done.

PR #3805, opened 2017-08-08
Issue 3599 - Always Scan DDN Params

kingthorin:
* Add VariantDdnPath.
* Modify AbstractAppParamPlugin to handle/leverage VariantDdnPath.
* Modify Messages.properties to combine the DDN scanning option with one
of the default options (URL Query String).

Core help update will be submitted after this is accepted.

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
+++ b/src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
@@ -1,0 +1,123 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
+ *   http://www.apache.org/licenses/LICENSE-2.0 
+ *   
+ * Unless required by applicable law or agreed to in writing, software 
+ * distributed under the License is distributed on an "AS IS" BASIS, 
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. 
+ * See the License for the specific language governing permissions and 
+ * limitations under the License. 
+ */
+package org.parosproxy.paros.core.scanner;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.model.SessionStructure;
+
+/**
+ * Variant class used for URL path elements that are defined as Data Driven Nodes.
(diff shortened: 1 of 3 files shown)


src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
```diff
@@ -0,0 +1,109 @@
...
+public class VariantDdnPath implements Variant {
+
+	private static final Logger LOGGER = Logger.getLogger(VariantDdnPath.class);
+	private final List<NameValuePair> stringParam = new ArrayList<>();
+
+	@Override
+	public void setMessage(HttpMessage msg) {
+		try {
+			List<String> treePaths = Model.getSingleton().getSession().getTreePath(msg);
+			int i = 0;
```
thc202: Could start in 1 instead of adding one when creating the `NameValuePair`, in any case a comment explaining the reason why adding one (or starting at 1) would be better.

src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+		return "";
+	}
+
+	private String setParameter(HttpMessage msg, NameValuePair originalPair, String name, String value,
+			boolean escaped) {
+		try {
+			URI uri = msg.getRequestHeader().getURI();
+			String[] paths = msg.getRequestHeader().getURI().getPath().toString().split("/");
```
thc202: - Can use the variable `uri`;
 - Should use `getEscapedPath()`;
 - `toString()` can be removed (it's already returning a String).

src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
```diff
@@ -0,0 +1,109 @@
...
+			if (originalPair.getPosition() < paths.length) {
+				String encodedValue = (escaped) ? value : getEscapedValue(value);
+				paths[originalPair.getPosition()] = encodedValue;
+				String path = StringUtils.join(paths, "/");
```
thc202: Could use `String.join`.

src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
```diff
@@ -0,0 +1,109 @@
...
+				try {
+					uri.setEscapedPath(path);
+				} catch (URIException e) {
+					// Looks like it wasn't escaped after all
```
thc202: Worth logging as warn or even error, if this happens it's because of a scanner misusing the variant (i.e. calling `setEscapedParameter` with a non-escaped value).

src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
```diff
@@ -0,0 +1,109 @@
...
+					uri.setPath(path);
+				}
+			}
+		} catch (URIException e) {
```
thc202: This catch can be moved to above catch block (no longer needed outside per use of `getEscapedPath` also the level can be changed to debug, not much that can be done if the path is already wrong).

src/org/parosproxy/paros/core/scanner/VariantDdnPath.java
kingthorin: Hahaha excellent point, I totally hadn't thought about it from anyone else's perspective, just made it work.

Will tweak.

kingthorin:
Fixed-up.

I also named the position variable more specifically and added a class lead-in JavaDoc block that explains the variant and position numbering.

thc202:
Thank you!

psiinon:
Thanks :) Worth updating the help too? (unless you've already submitted a PR for that;)

PR #3806, opened 2017-08-08
Make the .deb depend on java8-runtime not openjdk-7-jre

dominics:
This way, the sysadmin can decide which Java runtime to use.

The `java8-runtime` virtual package is provided by the OpenJDK packages, so this retains BC. This also allows sysadmins to attempt to run with newer versions of Java (because e.g. `openjdk-9-jre` still `Provides: java8-runtime`)

### DIFF
--- a/build/debian/control/control
+++ b/build/debian/control/control
@@ -5,6 +5,6 @@ Architecture: all
 Maintainer: Simon Bennetts <psiinon@gmail.com>
 Section: net
 Priority: extra
-Depends: openjdk-7-jre
+Depends: java8-runtime
 Homepage: https://github.com/zaproxy/zaproxy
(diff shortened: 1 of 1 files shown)


build/debian/control/control
```diff
@@ -5,6 +5,6 @@ Architecture: all
 Maintainer: Simon Bennetts <psiinon@gmail.com>
 Section: net
 Priority: extra
-Depends: openjdk-7-jre
+Depends: java7-runtime
```
kingthorin: Should probably actually be `java8-runtime` since 2.6.0 targets Java 8.

dominics: Even better!

thc202:
@dominics thanks! How would you like to be credited?

dominics:
@thc202 

> Dominic Scheirlinck (@dominics)

would be great, thanks.

PR #3808, opened 2017-08-08
added bare docker image file

omerlh:
Added a new docker file, that is a lot smaller (569 MB instead of 1.3 GB). I removed a lot of things and used Alpine instead of Ubuntu. The idea is to have a very small image for CI, with only the required stuff - at least for us, this is the minimum required to run Zap. As I think this could help others too, I create a PR. If you think otherwise - I can push this image to our repo :)

### DIFF
--- a/build/docker/Dockerfile-bare
+++ b/build/docker/Dockerfile-bare
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
     rm -rf /var/lib/apt/lists/*
 
 # Download and expand the latest stable release 
-RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions-dev.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
+RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
     cp -R ZAP*/* . &&  \
     rm -R ZAP* 
 
--- a/build/docker/Dockerfile-bare
+++ b/build/docker/Dockerfile-bare
@@ -1,24 +1,23 @@
 # This dockerfile builds the zap stable release
 FROM ubuntu:16.04 as builder
-MAINTAINER Simon Bennetts "psiinon@gmail.com"
 
 WORKDIR /zap
 
 RUN apt-get update && apt-get install -q -y --fix-missing \
-	wget \
-	curl \
-	xmlstarlet \
-	unzip && \
-	apt-get clean && \
-	rm -rf /var/lib/apt/lists/*
+    wget \
+    curl \
+    xmlstarlet \
+    unzip && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
 
 # Download and expand the latest stable release 
 RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions-dev.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
-	cp -R ZAP*/* . &&  \
-	rm -R ZAP* 
+    cp -R ZAP*/* . &&  \
+    rm -R ZAP* 
 
 FROM java:openjdk-8-jdk-alpine
-MAINTAINER Simon Bennetts "psiinon@gmail.com"
+LABEL maintainer="psiinon@gmail.com"
 
 WORKDIR /zap
 COPY --from=builder /zap .
@@ -27,7 +26,7 @@ RUN echo "http://dl-3.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
     adduser -h /home/zap -s /bin/bash zap -D zap && \
     rm -rf /var/cache/apk/* && \
     chown zap /zap && \
-	chgrp zap /zap && \
+    chgrp zap /zap && \
     chown -R zap:zap /zap
 
 #Change to the zap user so things get done as the right person (apart from copy)
--- a/build/docker/Dockerfile-bare
+++ b/build/docker/Dockerfile-bare
@@ -1,0 +1,41 @@
+# This dockerfile builds the zap stable release
+FROM ubuntu:16.04 as builder
+MAINTAINER Simon Bennetts "psiinon@gmail.com"
+
+WORKDIR /zap
+
+RUN apt-get update && apt-get install -q -y --fix-missing \
+	wget \
+	curl \
+	xmlstarlet \
+	unzip && \
+	apt-get clean && \
+	rm -rf /var/lib/apt/lists/*
+
+# Download and expand the latest stable release 
+RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions-dev.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
+	cp -R ZAP*/* . &&  \
+	rm -R ZAP* 
+
+FROM java:openjdk-8-jdk-alpine
+MAINTAINER Simon Bennetts "psiinon@gmail.com"
+
+WORKDIR /zap
+COPY --from=builder /zap .
+RUN echo "http://dl-3.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories &&\
+    apk add --update --no-cache bash netcat-openbsd && \
+    adduser -h /home/zap -s /bin/bash zap -D zap && \
+    rm -rf /var/cache/apk/* && \
+    chown zap /zap && \
+	chgrp zap /zap && \
+    chown -R zap:zap /zap
+
+#Change to the zap user so things get done as the right person (apart from copy)
+USER zap
+
+ENV PATH $JAVA_HOME/bin:/zap/:$PATH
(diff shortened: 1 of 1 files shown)


build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,43 @@
...
+	chgrp zap /zap && \
+    chown -R zap:zap /zap
+
+#Change to the zap user so things get done as the right person (apart from copy)
+USER zap
+
+ENV PATH $JAVA_HOME/bin:/zap/:$PATH
+ENV ZAP_PATH /zap/zap.sh
+
+# Default port for use with zapcli
```
psiinon: Are these still needed? Doesnt look like zapcli is included

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,43 @@
+# This dockerfile builds the zap stable release
+FROM ubuntu:16.04 as builder
```
kingthorin: Alpine?

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,43 @@
...
+USER zap
+
+ENV PATH $JAVA_HOME/bin:/zap/:$PATH
+ENV ZAP_PATH /zap/zap.sh
+
+# Default port for use with zapcli
+ENV ZAP_PORT 8080
+ENV HOME /home/zap/
+
+HEALTHCHECK --retries=5 --interval=5s CMD zap-cli status
```
psiinon: Ditto

build/docker/Dockerfile-bare
omerlh: I am using the new [multi stage build](https://docs.docker.com/engine/userguide/eng-image/multistage-build/) feature. So, this is only the build image - to download Zap and extract it. This container is removed after this part, and the real one (See from line 20) is based on alpine.

kingthorin: Ok fair enough....tks.

psiinon:
Another thought - have you looked at using the ZAP 'core' release, which is even smaller - does that provide everything you need?
Its not currently in the versions file, but we can add it if it would be useful

build/docker/Dockerfile-bare
omerlh: Those env variables are needed, I'll remove the port variable. Also noticed that I need to update the `healthcheck` as it depends on `zap-cli`.

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,43 @@
...
+RUN apt-get update && apt-get install -q -y --fix-missing \
+	wget \
+	curl \
+	xmlstarlet \
+	unzip && \
+	apt-get clean && \
+	rm -rf /var/lib/apt/lists/*
+
+# Download and expand the latest stable release 
+RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions-dev.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
```
omerlh: It was also weird to me, but I noticed that the stable image is using the dev from some reason - so I assumed this is something I am missing.

build/docker/Dockerfile-bare
omerlh: Fixed, now using netcat for that.

omerlh:
No, I was not aware of that - currently, there is no such dockerfile, correct? If so, I can rename it to `dockerfile-core` instead. Also, maybe I should not add all the python scripts - as I do not install python in this dockerfile. Are they part of the core release?

omerlh:
I had an issue - for some reason, it took for Zap more than 3 minutes to start - and because of the health check is not working. I put the logs [here](https://gist.github.com/omerlh/70dbf8a2d23e29861b0cc536ef9ac880), and I'm running Zap using the following command:
```
zap.sh -daemon -host 0.0.0.0 -port 8090 -config api.disablekey=true -config database.recoverylog=false -config connection.timeoutInSecs=120 -addonupdate -addoninstall pscanrulesAlpha -addoninstall pscanrulesBeta -config api.addrs.addr.name=.* -config api.addrs.addr.regex=true
```
So far I did not have such issues with Zap, so I'm suspecting this is somehow related to my changes.

thc202:
Does not seem to be related to your changes, it took more than usual because of the downloads/updates of the add-ons(?), ZAP would be ready after 40 seconds if not instructed to update/install the add-ons.

thc202:
About the health check, did you change the env variable `ZAP_PORT` to 8090?

omerlh:
Yep, tested that, and usually, it works, but it still fails on my system. Maybe I should increase again the number of retries - or change the interval. Perosnally, I think that we should keep the interval on 5 seconds.

omerlh:
@psiinon @thc202 what's missing in this PR? look to me like I fixed everything we discussed :)

build/docker/Dockerfile-bare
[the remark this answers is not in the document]
thc202: It's just missing this change. I've opened a pull request (#3914) to change that in the other images, with `ZapVersions-dev.xml` it's also downloading add-ons (e.g. `webdriverlinux-beta-3.zap`, `jxbrowserlinux32-alpha-3.zap`).

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,41 @@
...
+	cp -R ZAP*/* . &&  \
+	rm -R ZAP* 
+
+FROM java:openjdk-8-jdk-alpine
+MAINTAINER Simon Bennetts "psiinon@gmail.com"
+
+WORKDIR /zap
+COPY --from=builder /zap .
+RUN echo "http://dl-3.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories &&\
+    apk add --update --no-cache bash netcat-openbsd && \
```
thc202: This spacing could use tabs (to normalise with other lines in this and other Docker files).

omerlh: just checked, and it is tab instead of space - unless I am missing something

thc202: It's using spaces here, in a following line is using a tab (which stands out from the surrounding lines).

omerlh: ok, I tried to clean it up a bit, hope that solved it

build/docker/Dockerfile-bare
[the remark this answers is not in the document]
omerlh: fixed

build/docker/Dockerfile-bare
thc202: Looks good. Could you squash the commits?

omerlh: I can, but usually, this is done through GitHub - if this is not an option let me know and I'll squash them.

thc202: We usually squash before merge, but I'm OK with squash while merging the pull request as well.

thc202:
Thank you!

omerlh:
BTW maybe worth updating the [wiki](https://github.com/zaproxy/zaproxy/wiki/Docker), I can do it after the merge - what do you think?

build/docker/Dockerfile-bare
tanelikaivola: Even when running multi-stage builds instead of having ubuntu, consider alpine:
```
FROM alpine as builder
RUN apk add --no-cache curl wget xmlstarlet unzip
```

This covers the build stage run commands used here just as well as ubuntu and pulls in 100MB less.

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,40 @@
...
+    wget \
+    curl \
+    xmlstarlet \
+    unzip && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Download and expand the latest stable release 
+RUN curl -s https://raw.githubusercontent.com/zaproxy/zap-admin/master/ZapVersions.xml | xmlstarlet sel -t -v //url |grep -i Linux | wget --content-disposition -i - -O - | tar zxv && \
+    cp -R ZAP*/* . &&  \
```
tanelikaivola: Instead of copying just move? `mv ZAP*/* .`

build/docker/Dockerfile-bare
```diff
@@ -0,0 +1,40 @@
...
+    rm -R ZAP* 
+
+FROM java:openjdk-8-jdk-alpine
```
tanelikaivola: java:* is deprecated, use openjdk:* (openjdk:8u121-jdk-alpine seems to work for now)

build/docker/Dockerfile-bare
omerlh: You are totally right, I just assumed that it does not happen that much so I can just live with ubuntu but I can change it.

build/docker/Dockerfile-bare
omerlh: That's a good one, I considered whether or not to choose openjdk, but I did not want to change too much in this commit - the original image used java.

thc202:
@omerlh might worth wait for #3918 and the push of the image(?).

omerlh:
Sure.

thc202:
@omerlh the image is now available under `owasp/zap2docker-bare`, would you mind updating the wiki?

omerlh:
@thc202 Done.

thc202:
Thanks! Right, we will be adding that shortly and announce the new image :)

omerlh:
Great :)

thc202:
For the record, the docker hub page was already updated and the image announced :)  Thanks!

omerlh:
I noticed and retweeted :) thanks for mention me!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #3559, opened 2017-05-19
ZAP API option to output report in JSON format


claudijd:
I'd like to request a feature for adding JSON output to the ZAP API so that scripts can more easily present parsable data outputs for other automation purposes.

psiinon:
Makes sense - we support XML, HTML and MD reports, not supporting JSON is an omission.

PR #3810, opened 2017-08-08
Add JSON output support to zap-full-scan.py

claudijd:
@psiinon I'm pretty sure this is all that's needed, but I really didn't run it independently because I don't have a full ZAP dev environment setup.  I just happened to notice support was added, but not necessarily to this script.

### DIFF
--- a/build/docker/zap-api-scan.py
+++ b/build/docker/zap-api-scan.py
@@ -89,6 +89,7 @@ def usage():
     print('    -r report_html    file to write the full ZAP HTML report')
     print('    -w report_md      file to write the full ZAP Wiki(Markdown) report')
     print('    -x report_xml     file to write the full ZAP XML report')
+    print('    -J report_json    file to write the full ZAP JSON document')
     print('    -a                include the alpha passive scan rules as well')
     print('    -d                show debug messages')
     print('    -P                specify listen port')
@@ -119,6 +120,7 @@ def main(argv):
     report_html = ''
     report_md = ''
     report_xml = ''
+    report_json = ''
     target = ''
     target_file = ''
     target_url = ''
@@ -142,7 +144,7 @@ def main(argv):
     check_zap_client_version()
 
     try:
-        opts, args = getopt.getopt(argv, "t:f:c:u:g:m:n:r:w:x:l:daijp:sz:P:D:T:")
+        opts, args = getopt.getopt(argv, "t:f:c:u:g:m:n:r:J:w:x:l:daijp:sz:P:D:T:")
     except getopt.GetoptError as exc:
         logging.warning('Invalid option ' + exc.opt + ' : ' + exc.msg)
         usage()
@@ -172,6 +174,8 @@ def main(argv):
             progress_file = arg
         elif opt == '-r':
             report_html = arg
+        elif opt == '-J':
+            report_json = arg    
         elif opt == '-w':
             report_md = arg
         elif opt == '-x':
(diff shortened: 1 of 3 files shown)


claudijd:
Related to #3559

build/docker/zap-full-scan.py
```diff
@@ -85,6 +85,7 @@ def usage():
     print('    -r report_html    file to write the full ZAP HTML report')
     print('    -w report_md      file to write the full ZAP Wiki(Markdown) report')
     print('    -x report_xml     file to write the full ZAP XML report')
+    print('    -J report_json    file to write the full ZAP JSON document')
```
psiinon: This will need to be added to the "getopt.getopt(..)" call as well.
Ideally we'd also support this option in the baseline and api scan too :)

build/docker/zap-full-scan.py
```diff
@@ -420,6 +424,10 @@ def main(argv):
                 # Save the report
                 write_report(base_dir + report_html, zap.core.htmlreport())
 
+            if report_json:
+                # Save the report
+                write_report(base_dir + report_json, zap.core.jsonreport())
```
thc202: This was not yet included in the Python client, we need to release a new version with that.

psiinon: Ah! Its a post 2.6.0 change isnt it :/
It will still work on the weekly (which is what this script uses), so we could use the underlying method to call the endpoint.

thc202: Yeah, sounds good.

build/docker/zap-full-scan.py
claudijd: Fixed in zap-full-scan.py, can probably patch the other scripts as well if it's about as trivial of an update

thc202: Yes, it's the same changes for the other scripts.

claudijd: Ok, just pushed the changes for zap-api-scan.py and zap-full-scan.py

build/docker/zap-baseline.py
```diff
@@ -379,6 +383,10 @@ def main(argv):
                 # Save the report
                 write_report(base_dir + report_html, zap.core.htmlreport())
 
+            if report_json:
+                # Save the report
+                write_report(base_dir + report_json, zap.core.htmlreport())
```
claudijd: I forgot to make this json report

build/docker/zap-api-scan.py
```diff
@@ -467,6 +471,10 @@ def main(argv):
                 # Save the report
                 write_report(base_dir + report_html, zap.core.htmlreport())
 
+            if report_json:
+                # Save the report
+                write_report(base_dir + report_json, zap.core.htmlreport())
```
claudijd: Same here, needs to be json

kingthorin:
Hey @claudijd once you're comfortable that this is done/ready could you do an interactive rebase and fix-up or squash into a single "atomic" commit?

- https://www.instructables.com/id/How-to-Rebase-Interactive-With-Eclipse-eGit/

build/docker/zap-full-scan.py
thc202: @claudijd the call `zap.core.jsonreport()` needs to be replaced with `zap._request_other(zap.base_other + 'core/other/jsonreport/')`, for the time being.

thc202: @claudijd are you available to apply this change? (I think it's the only outstanding comment to merge the changes.)

thc202:
Rebased and addressed the remaining comment.

thc202:
@claudijd thanks! How would you like to be credited?

claudijd:
@thc202 "Jonathan Claudius (@claudijd)" is good by me, thanks!

PR #3814, opened 2017-08-10
Issue 2965 multipart form post file details variant

kingthorin:
* VariantMultipartFormQuery - Rewritten & renamed (VariantMultipartFormParameters). Added the ability to manipulate/inject the "filename" and "Content-Type" portions of a multipart form post.
* VariantMultipartFormParametesUnitTest - Added unit tests to cover the new code.


kingthorin:
There is one minor "catch" with this implementation, if multiple files are included and happen to have the same field `name`  & `filename` then there is a submission by the active scanners for each one, though the matching `filename`s are all replaced by the injection/manipulation.

thc202:
Looks good, still need to try it though :)

Yeah, I don't think that's a big problem (we can tweak that later on, if needed).

kingthorin:
VariantMultipartFormQuery - I'd originally implemented this as a second class and run them both at the same time for head-to-head comparison, but then swapped it out. Unfortunately that means all of the formatting changed too....

kingthorin:
Would this be easier to review and deal with if I split it into two PRs?

thc202:
One pull request is fine to me.

kingthorin:
Should content type of the file parameters be a variant (or part of a variant)?
Although we don't have scanners that target file upload content should I add handling for file content? Should it be separate or part of the normal multi-part form injection?

(Side question: Can the active scan plugins tell which type of variant they're being called for? i.e.: Can devs make active scanners that only work for a particular variant?)

kingthorin:
Just noticed we do have some unit tests for these. I'll look into adding some coverage for these.

https://github.com/zaproxy/zaproxy/tree/develop/test/org/parosproxy/paros/core/scanner

thc202:
Thanks!

kingthorin:
`VariantMultipartFormFileDetailsUnitTest` added (coverage is VERY basic, so I"m open to ideas. I dislike how I've had to handle building the request bodies. Can we come up with a good way to read them from an xml or json file?). Which lead to a few minor changes in `VariantMultipartFormFileDetails`.

Will tackle `VariantMultipartFormQuery` unit tests next.

kingthorin:
In writing the unit tests for VariantMultipartFormQuery it has become clear that handling potential differences in line endings is not a simple task. Is it acceptable to repalceAll on the request and rebuild it in a friendly way? Is that too much modification to the original? Does StringUtils (or something) have a convenient function for such conversions?

thc202:
> Should content type of the file parameters be a variant (or part of a variant)?

I'd make it part of the existing variant and rename the variant (to drop the "Query").

> Although we don't have scanners that target file upload content should I add handling for file content?

Worth adding that.

> Should it be separate or part of the normal multi-part form injection?

Maybe separate?

> Can devs make active scanners that only work for a particular variant?

Yes, scanners can check the type of the `NameValuePair` being scanned.

> Can we come up with a good way to read them from an xml or json file?

We have some tests in the zap-extensions repo that read from files, but in this case I don't see a problem with how that's being done. We could use HttpComponents library to build multipart content.

> Is it acceptable to replaceAll on the request and rebuild it in a normalized way?

Do we need to? From the multipart RFCs (2046, 7578) it should be used CRLF (same as "previous" parser). We could also just extract the positions (start/end) of each injection point and just replace that when "rebuilding" the body.

kingthorin:
Thanks for the feedback @thc202, will dig into it more tonight.

kingthorin:
Ok so here's the update, as things stand in 2cbb957:
* I did rename the *Query variant to *Parameters.
* I did make things use CRLF instead of just LF, and it all seems to be behaving.

Outstanding:
* Add handling for manipulating the content-type of the file parameters.
* Add handling for manipulating the content of the file parameters.

kingthorin:
How do I restart a travis build? It seems to be failing due to the rename of the one file. It's complaining about *Query when *Query has been moved (renamed)?

https://travis-ci.org/zaproxy/zaproxy/builds/281064341#L452-L464

kingthorin:
Just so you're aware I haven't forgotten about this and am still progressing as time permits.

kingthorin:
I've revised this based on the discussions from IRC last [Friday](https://mozilla.logbot.info/websectools/20171201). I haven't implemented the content-type manipulation part yet, but I'd be glad of some feedback at this juncture. 
[In it's current form all params are handled as normal params (value/content manipulation, that includes both normal params and file params), and for file params the filename part can be manipulated.]

kingthorin:
Now handles the "Conent-Type" element for file params.

kingthorin:
(d1c121a 20171215) Added further unit tests.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+					// http://fiddle.re/23cudd (Click Java, hit "test")
+				}
+
+				String name = nameMatcher.group("name");
+				// Value doesn't include boundary, headerline, or double CRLF
+				String value = part.replaceAll(boundary + partHeaderLine + HttpHeader.CRLF + HttpHeader.CRLF, "");
+				value = value.replaceAll(getBoundary(contentType) + "--" + HttpHeader.CRLF, "");// Strip final boundary
+				extractedParameters.add(new NameValuePair(NameValuePair.TYPE_POST_DATA, name, value, position));
+				int start = offset + part.indexOf(HttpHeader.CRLF + HttpHeader.CRLF) + 4; // 4 for two CRLFs
+				int end = offset + part.length();
```
thc202: `start + value.length() - HttpHeader.CRLF.length()`, otherwise the general `MultipartFormParameter` includes more than it should.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+					LOGGER.debug("Group name: " + name + " value: " + valueMatcher.group("value"));
+				}
+				if (partHeaderLine.contains("filename=")) {
+					// Extract the filename
+					Matcher fnValueMatcher = FILENAME_PART_PATTERN.matcher(part);
+					fnValueMatcher.find();
+					String fnValue = fnValueMatcher.group("filename");
+					extractedParameters.add(new NameValuePair(NameValuePair.TYPE_POST_DATA, name, fnValue, ++position));
+					int fnStart = offset + part.indexOf(fnValue);
+					int fnEnd = offset + (fnStart + fnValue.length());
```
thc202: Should be just `fnStart + fnValue.length()` (`fnStart` already has the `offset`).

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+					}
+					multiPartParams.add(new MultipartFormParameter(name, fnValue, fnStart, fnEnd, position,
+							MultipartFormParameter.Type.FILE_NAME));
+					// Extract the content-type
+					Matcher ctValueMatcher = CONTENTTYPE_PART_PATTERN.matcher(part);
+					ctValueMatcher.find();
+					String ctValue = ctValueMatcher.group("contenttype");
+					extractedParameters.add(new NameValuePair(NameValuePair.TYPE_POST_DATA, name, ctValue, ++position));
+					int ctStart = offset + part.indexOf(ctValue);
+					int ctEnd = offset + (ctStart + ctValue.length());
```
thc202: Same here (regarding the extra `offset`).

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+		return setParameter(msg, originalPair, name, value, true);
+	}
+
+	private String setParameter(HttpMessage msg, NameValuePair originalPair, String name, String value,
+			boolean escaped) {
+		String contentType = msg.getRequestHeader().getHeader(HttpHeader.CONTENT_TYPE);
+		String boundary = getBoundary(contentType);
+
+		StringBuilder newBodyBuilder = new StringBuilder(msg.getRequestBody().toString());
+		for (int idx = 0; idx < params.size(); idx++) {
```
thc202: The loop can be removed now (and just replace).

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+			if (LOGGER.isDebugEnabled()) {
+				LOGGER.debug("i: " + idx + " pos: " + originalPair.getPosition() + " S: " + mpPart.getStart() + " E: "
+						+ mpPart.getEnd());
+			}
+			if (idx == originalPair.getPosition()-1) {// Positions start at 1
+				if (MultipartFormParameter.Type.GENERAL.equals(mpPart.getType())) {
+					value = value + HttpHeader.CRLF;
+				}
+				newBodyBuilder.replace(mpPart.getStart(), mpPart.getEnd(), value);
+				if (idx == (params.size()-1)) {
```
thc202: Same here

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+	public String setParameter(HttpMessage msg, NameValuePair originalPair, String name, String value) {
+		return setParameter(msg, originalPair, name, value, false);
+	}
+
+	@Override
+	public String setEscapedParameter(HttpMessage msg, NameValuePair originalPair, String name, String value) {
+		return setParameter(msg, originalPair, name, value, true);
+	}
+
+	private String setParameter(HttpMessage msg, NameValuePair originalPair, String name, String value,
```
thc202: Since we are not using the `name` nor `escaped` they could be removed.

test/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -0,0 +1,205 @@
...
+		String paramName = "person";
+		String newValue = "injected";
+		String origValue = "anonymous" + CRLF + "anonymous2" + CRLF;
+		variant.setMessage(message);
+		// When
+		variant.setParameter(message,
+				new NameValuePair(NameValuePair.TYPE_POST_DATA, paramName, origValue, 1),
+				paramName, newValue);
+		// Then
+		assertFalse(message.getRequestBody().toString().contains(origValue));
```
thc202: Better with exact match, that ensures we have the proper contents/format. Same comment for the following tests.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,195 @@
...
+
+		StringBuilder newBodyBuilder = new StringBuilder(msg.getRequestBody().toString());
+		for (int idx = 0; idx < params.size(); idx++) {
+			MultipartFormParameter mpPart = multiPartParams.get(idx);
+			if (LOGGER.isDebugEnabled()) {
+				LOGGER.debug("i: " + idx + " pos: " + originalPair.getPosition() + " S: " + mpPart.getStart() + " E: "
+						+ mpPart.getEnd());
+			}
+			if (idx == originalPair.getPosition()-1) {// Positions start at 1
+				if (MultipartFormParameter.Type.GENERAL.equals(mpPart.getType())) {
```
thc202: This does not seem to be necessary or shouldn't be, with the replacements.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
kingthorin: The bus ness is being used to get the param from the list, I guess I could change it to a map using param name as the key?

thc202: Right, but the list has the same order (less one) as the parameters so we can just access it (`int idx = originalPair.getPosition() -1; MultipartFormParameter mpPart = multiPartParams.get(idx);`).

kingthorin: Sorry been working with the same code too long to see an alternative (some sort of “functional fixed ness”). Thanks @thc202 !

kingthorin:
Review comments addressed.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,189 @@
...
+					// Need to skip one find for some reason...
+					// https://regex101.com/r/4ig6Wk/1
+					// http://fiddle.re/23cudd (Click Java, hit "test")
+				}
+
+				String name = nameMatcher.group("name");
+				// Value doesn't include boundary, headerline, or double CRLF
+				String value = part.replaceAll(boundary + partHeaderLine + HttpHeader.CRLF + HttpHeader.CRLF, "");
+				value = value.replaceAll(getBoundary(contentType) + "--" + HttpHeader.CRLF, "");// Strip final boundary
+				value = value.trim();// Trailing CRLF
```
thc202: This might remove more than needed (leading and trailing), not be a big issue though.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,189 @@
...
+		return setParameter(msg, originalPair, value);
+	}
+
+	@Override
+	public String setEscapedParameter(HttpMessage msg, NameValuePair originalPair, String name, String value) {
+		return setParameter(msg, originalPair, value);
+	}
+
+	private String setParameter(HttpMessage msg, NameValuePair originalPair, String value) {
+		String contentType = msg.getRequestHeader().getHeader(HttpHeader.CONTENT_TYPE);
```
thc202: Can be removed (no longer needed), same for following statement.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -0,0 +1,189 @@
...
+		String boundary = getBoundary(contentType);
+
+		StringBuilder newBodyBuilder = new StringBuilder(msg.getRequestBody().toString());
+		int idx = originalPair.getPosition() -1; 
+		MultipartFormParameter mpPart = multiPartParams.get(idx);
+		if (LOGGER.isDebugEnabled()) {
+			LOGGER.debug("i: " + idx + " pos: " + originalPair.getPosition() + " S: " + mpPart.getStart() + " E: "
+					+ mpPart.getEnd());
+		}
+		if (idx == originalPair.getPosition()-1) {// Positions start at 1
```
thc202: The if statement can be removed (now redundant).

test/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+		HttpMessage message = createMessage();
+		String paramName = "person";
+		String newValue = "injected";
+		variant.setMessage(message);
+		// When
+		variant.setParameter(message,
+				new NameValuePair(NameValuePair.TYPE_POST_DATA, paramName, DEFAULT_PARAM_CONTENT, 1),
+				paramName, newValue);
+		StringBuilder testBodyBuilder = new StringBuilder(320);
+		testBodyBuilder.append("--------------------------d74496d66958873e").append(CRLF);
```
thc202: How about changing `createMessage` to allow to pass the 4 fields and call that with the corresponding values? It would reduce duplication and still check the correct structure.

test/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+		testBodyBuilder.append(CRLF);
+		testBodyBuilder.append(newValue).append(CRLF);
+		testBodyBuilder.append("--------------------------d74496d66958873e").append(CRLF);
+		testBodyBuilder.append("Content-Disposition: form-data; name=\"somefile\"; filename=\"file.ext\"").append(CRLF);
+		testBodyBuilder.append("Content-Type: ").append(DEFAULT_CONTENT_TYPE).append(CRLF);
+		testBodyBuilder.append(CRLF);
+		testBodyBuilder.append(DEFAULT_FILE_PARAM_CONTENT).append(CRLF);
+		testBodyBuilder.append("--------------------------d74496d66958873e--").append(CRLF);
+		// Then
+		assertTrue(message.getRequestBody().toString().equals(testBodyBuilder.toString()));
```
thc202: These could use `assertThat`.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
kingthorin: Hmmmm it proved more reliable than reducing the end mark by `HttpHeader.CRLF.length()`, guess I’ll have to debug the issue(s) with that further or see if StringUtils has a trimTrailing or similar.

test/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
kingthorin: Yup I like that.

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: Right, we should be changing the `value` instead of removing the chars from the end. Using:
```Java
value = value.replaceAll(HttpHeader.CRLF + "(" + Pattern.quote(getBoundary(contentType)) + "--" + HttpHeader.CRLF + ")?$", ""); // Strip final boundary
```
seems to work in all cases.

kingthorin: Thanks @thc202

thc202:
Thank you!

Issue #3798, opened 2017-08-03
java.awt.Toolkit initialised when running without view


nkpathrut:
I have tried to implement the following youtube demo but not performed well. 

https://youtu.be/swDRxxqFmjo

I have tried to perform security test on the sample App as show in video.

The following logs I got in Jenkins : 

Started by user admin
Building in workspace /var/lib/jenkins/workspace/zap_demo1
 > git rev-parse --is-inside-work-tree # timeout=10
Fetching changes from the remote Git repository
 > git config remote.origin.url https://github.com/Eficode/security_testing_demo_app # timeout=10
Fetching upstream changes from https://github.com/Eficode/security_testing_demo_app
 > git --version # timeout=10
 > git fetch --tags --progress https://github.com/Eficode/security_testing_demo_app +refs/heads/*:refs/remotes/origin/*
 > git rev-parse refs/remotes/origin/master^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/origin/master^{commit} # timeout=10
Checking out Revision 7086f9056675491013a2eb80ba5980cf8636ed4f (refs/remotes/origin/master)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 7086f9056675491013a2eb80ba5980cf8636ed4f
 > git rev-list 7086f9056675491013a2eb80ba5980cf8636ed4f # timeout=10

<details>
<summary>Log output (click to expand)</summary>

```console
[ZAP Jenkins Plugin] START PRE-BUILD ENVIRONMENT VARIABLE REPLACEMENT
	HOST = [ localhost ]
	PORT = [ 9999 ]

	SESSION FILENAME = [  ]
	INTERNAL SITES = [  ]

	CONTEXT NAME = [ testsession ]

	INCLUDE IN CONTEXT = [ http://127.0.0.1:7272 ]

	EXCLUDE FROM CONTEXT = [  ]

	STARTING POINT (URL) = [ http://127.0.0.1:7272 ]
	REPORT FILENAME = [ JENKINS_ZAP_VULNERABILITY_REPORT_new_reports ]
	REPORT TITLE = [  ]

	COMMAND LINE = 

[ZAP Jenkins Plugin] END PRE-BUILD ENVIRONMENT VARIABLE REPLACEMENT

[ZAP Jenkins Plugin] CLEAR LOGS IN SETTINGS...
	ZAP HOME DIRECTORY [ ~/.ZAP ]
[...]

PR #3815, opened 2017-08-10
Address Toolkit initialisations in daemon mode

thc202:
Change core extensions/classes to address Toolkit initialisations when
in daemon mode (e.g. caused by creation of icons or keyboard shortcuts).

Part of #3798 - java.awt.Toolkit initialised in daemon mode


psiinon:
Is there any way we can automatically test for these sort of issues?

thc202:
I think so, we can force the headless mode and use a custom ToolKit that errors when used. I'll investigate that.

Where would we run those tests? Travis CI? @zapbot? (Ideally we would exercise all code paths.)

PR #3816, opened 2017-08-10
Do not try to read plugin dir if it does not exist

thc202:
Change AddOnCollection to return if the provided plugin directory does
not exist or if it's not a directory, instead of continuing and try to
read it (leading to an exception). Also, change the log level to WARN,
it's not a ZAP error/issue if it does not exist or it's not a directory.

### DIFF
--- a/src/org/zaproxy/zap/control/AddOnCollection.java
+++ b/src/org/zaproxy/zap/control/AddOnCollection.java
@@ -161,10 +161,12 @@ public class AddOnCollection {
     		return;
     	}
     	if (! dir.exists()) {
-    		logger.error("No such directory: " + dir.getAbsolutePath());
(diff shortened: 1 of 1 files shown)


psiinon:
It might not strictly be a ZAP issue, but it could imply an install issue.
When have you seen this happen?
It its the main ZAP plugin dir then ZAP will have very little functionality, and if its the user one then would they be able to download add-ons?
I'm happy with returning, just wondering about the change to warn instead of error :)

thc202:
After renaming the main plugin directory for checking/testing other things (didn't want the add-ons interfering). I'd prefer leaving it as warn instead of error (ZAP can still work fine without the plugin directories, specially if using custom ones), but if error works for catching installation issues I can restore it.

psiinon:
np:)

Issue #3818, opened 2017-08-11
Change right-click "Resend..." to "Send to Manual Request Editor"


claudijd:
One of the more common user actions I think a lot of testers do after they've built a site map is the ability to right-click and select a meaningful action.  One of the most common actions for me is to take a pre-existing request from the site map or individual alerts and send those requests to either the Manual Request Editor (repeater in Burp terminology) or the Fuzzer (intruder in Burp terminology).

It's possible I'm just doing it wrong, but I think a lot of other Burp heavy users may have this as an expectation when trying out or playing with ZAP for the first time.  If they can't find a quick way to do something useful, this may stunt adoption in the tool.

Hope this helps!

kingthorin:
Resend...

![001](https://user-images.githubusercontent.com/7570458/29224333-db65babe-7e97-11e7-9aac-13defe3635b8.png)
![002](https://user-images.githubusercontent.com/7570458/29224335-db664ba0-7e97-11e7-8b5a-a91ba43af02c.png)
![003](https://user-images.githubusercontent.com/7570458/29224334-db65b212-7e97-11e7-9142-a53e8f04184e.png)

claudijd:
@kingthorin what is the difference between "Resend..." and "Manual Request Editor"?  Are they basically the same?

kingthorin:
Yes essentially the same.

Should be more consistent in naming.

claudijd:
What about "Send to Manual Request Editor" in the drop down, but then it drops you into the "Manual Request Editor"?

kingthorin:
Seems reasonable to me.

thc202:
In almost all tabs that show a (HTTP) message the Fuzzer can be accessed with the context menu "Attack" > "Fuzz...". In the Request/Response tabs it can be accessed with "Fuzz..." menu item, adding the selection as a fuzz position.

psiinon:
I dont like 'Send to Manual Request Editor' I'm afraid - sounds too Burp like :/

psiinon:
"Open in Manual Editor..." I could live with maybe, or perhaps "Resend / Open in Manual Editor..." ?

kingthorin:
Ok that's fair.

PR #3820, opened 2017-08-14
Issue 3818 - Rename resend

kingthorin:
Rename context entries from "Resend..." to "Open/Resend with Request Editor..."

I chose "Request Editor" vs "Manual Editor" as mentioned in the ticket as I didn't think that 'Manual Editor' was specific enough for inexperienced/newer users.

Also standardize on the dialog title.
`manReq.resend.popup` resource key can't be removed because the websocket addon depends on it.

Fixes zaproxy/zaproxy#3818

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -1253,7 +1253,7 @@ history.panel.mnemonic		           = h
 history.panel.title                    = History
 history.purge.popup                    = Delete
 history.purge.warning                  = Are you sure you want to delete the record(s)?
-history.resend.popup                   = Resend...
+history.resend.popup                   = Open/Resend with Request Editor...
 history.scan.warning                   = Error getting History.
 history.scope.button.selected          = Show all URLs
 history.scope.button.unselected        = Show only URLs in Scope
@@ -1429,7 +1429,8 @@ manReq.checkBox.followRedirect = Follow redirect
 manReq.checkBox.useSession     = Use current tracking session
 manReq.dialog.title            = Manual Request Editor
 manReq.pullDown.method         = Method
-manReq.resend.popup            = Resend
+# TODO remove manReq.resend.popup once websockets addon is no longer using it
+manReq.resend.popup            = Manual Request Editor
 manReq.tab.request             = Request
 manReq.tab.response            = Response
 manReq.display.tabs            = Separate tabs for Request and Response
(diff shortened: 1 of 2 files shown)


kingthorin:
Would "with" be better than "in"?

thc202:
"with" sounds better.

Should the dialogue be renamed to "Request Editor"? (Currently it is "Resend".)
How about the "Manual Request Editor" dialogue, should that be renamed as well? Should we have just one dialogue?

kingthorin:
Ok I'll tweak it to "with".

As for the dialogues we could standardize on "Manual Request Editor" as the name. Are there actually multiple dialogue implementations?

thc202:
Both are using the same class just changing the name/title of the dialogue. Would it be confusing having two "Manual Request Editor" dialogues? (They can be opened at the same time.)

kingthorin:
I don't think that's a big deal. Can always revise further if there are questions later...

src/lang/Messages.properties
```diff
@@ -1429,7 +1429,7 @@ manReq.checkBox.followRedirect = Follow redirect
 manReq.checkBox.useSession     = Use current tracking session
 manReq.dialog.title            = Manual Request Editor
 manReq.pullDown.method         = Method
-manReq.resend.popup            = Resend
+manReq.resend.popup            = Manual Request Editor
```
thc202: This would no longer be needed per change in `ExtensionHistory`?

thc202:
OK.

src/lang/Messages.properties
kingthorin: >`manReq.resend.popup` resource key can't be removed because the websocket addon depends on it.

kingthorin: https://github.com/zaproxy/zap-extensions/blob/master/src/org/zaproxy/zap/extension/websocket/ui/ResendWebSocketMessageMenuItem.java#L53

thc202: OK, just searched the zaproxy repo, worth adding a TODO comment mentioning that should be deleted once WebSockets is no longer using it.

kingthorin: Comment added, will submit a websocket PR as well.

thc202:
Thanks!

src/lang/Messages.properties
thc202: Great :)

PR #3837, opened 2017-08-21
Add anoncsrf as anticsrf token name

g-k:
used by django-session-csrf: https://github.com/mozilla/django-session-csrf#differences-from-django


kingthorin:
Should probably be updated here too:
https://github.com/zaproxy/zaproxy/blob/c93ef3376b23d71bc2ab1249d15ef4cdef8e51f0/src/org/zaproxy/zap/extension/anticsrf/AntiCsrfParam.java#L45-L46

thc202:
Thanks!

Issue #3350, opened 2017-03-31
wrong comparison of HTTP Messages in queryEquals() method of HttpMessage class


raghuveerbathini:
The method queryEquals(HttpMessage), of org.parosproxy.paros.network.HttpMessage class, compares url and body parameter names of two HTTP Messages. 

The comparison is happening in wrong way, that it is comparing the first HTTP Message to itself, hence returns 'true' always.

        **set1 = getParamNameSet(HtmlParameter.Type.url);
	set2 = getParamNameSet(HtmlParameter.Type.url);**
	if (!set1.equals(set2)) {
	        return false;
	}
	if (getRequestHeader().getMethod().equalsIgnoreCase(HttpRequestHeader.POST)) {
	        
	        **set1 = getParamNameSet(HtmlParameter.Type.form);
		set2 = getParamNameSet(HtmlParameter.Type.form);**	        

		if (!set1.equals(set2)) {
		        return false;
		}
	    }

The following is the correct way to compare the two Messages:

        **set1 = getParamNameSet(HtmlParameter.Type.url);
	set2 = msg.getParamNameSet(HtmlParameter.Type.url);**
	if (!set1.equals(set2)) {
	        return false;
	}
	if (getRequestHeader().getMethod().equalsIgnoreCase(HttpRequestHeader.POST)) {
	        
	        **set1 = getParamNameSet(HtmlParameter.Type.form);
		set2 = msg.getParamNameSet(HtmlParameter.Type.form);**	        

		if (!set1.equals(set2)) {
		        return false;
		}
	    }

thc202:
Indeed, I'm wondering if it does worth maintaining those methods, that's not working properly since 2.3.0 and ZAP nor the (zap-extensions) add-ons are using it.

kingthorin:
I'd it is un-used let's drop it.

psiinon:
+1

kingthorin:
So do we want to drop the `queryEquals` function and it's use altogether? Since `queryEquals` is used within `equalType` which does not appear to be used anywhere do we want to drop both?

https://github.com/zaproxy/zaproxy/blob/a1097f98f327d89966777e471c73bcc55d56fb91/src/org/parosproxy/paros/network/HttpMessage.java#L482

https://github.com/zaproxy/zaproxy/blob/a1097f98f327d89966777e471c73bcc55d56fb91/src/org/parosproxy/paros/network/HttpMessage.java#L431

[...]

PR #3843, opened 2017-08-23
Issue 3350 HttpMessage correct queryEquals method

kingthorin:
`queryEquals` correct comparison and add JavaDoc.

Addresses zaproxy/zaproxy#3350


thc202:
Lets hold this, per comments in #3350.

src/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -479,15 +480,27 @@ public boolean equalType(HttpMessage msg) {
 	    return result;
 	}
 	
+	/**
+	 * Compares the parameter names used in GET and POST messages. For POST
```
thc202: Could you add this explanation to `equalType` too? That's the method the users of the class will use/see (without checking the source).

kingthorin: Yup

thc202:
Thanks!

PR #3851, opened 2017-08-28
Spider New Scan UI Consistency

kingthorin:
Minor changes to SpiderDialog so that when the "New Scan" button is clicked the target (and associated context(s)/user(s)) combo fields are populated based on site tree selection (if any) instead of blank.
Cancelling the dialog now resets the target to nothing.

### DIFF
--- a/src/org/zaproxy/zap/extension/spider/SpiderDialog.java
+++ b/src/org/zaproxy/zap/extension/spider/SpiderDialog.java
@@ -114,6 +114,7 @@ public class SpiderDialog extends StandardFieldsDialog {
         this.addTargetSelectField(0, FIELD_START, this.target, true, false);
         this.addComboField(0, FIELD_CONTEXT, new String[] {}, "");
         this.addComboField(0, FIELD_USER, new String[] {}, "");
+        setUsers();
         this.addCheckBoxField(0, FIELD_RECURSE, true);
         this.addCheckBoxField(0, FIELD_SUBTREE_ONLY, subtreeOnlyPreviousCheckedState);
         // This option is always read from the 'global' options
@@ -423,5 +424,11 @@ public class SpiderDialog extends StandardFieldsDialog {
     void reset() {
         target = null;
         reset(true);
+    }
+
+    @Override
+    public void cancelPressed() {
+        this.target=null;
+        super.cancelPressed();
     }
 }
--- a/src/org/zaproxy/zap/extension/spider/SpiderPanel.java
+++ b/src/org/zaproxy/zap/extension/spider/SpiderPanel.java
@@ -49,6 +49,7 @@ import org.jdesktop.swingx.renderer.IconValues;
 import org.jdesktop.swingx.renderer.MappedValue;
 import org.jdesktop.swingx.renderer.StringValues;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.model.ScanController;
 import org.zaproxy.zap.model.ScanListenner2;
@@ -495,7 +496,7 @@ public class SpiderPanel extends ScanPanel2<SpiderScan, ScanController<SpiderSca
 			scanButton.addActionListener(new ActionListener () {
 				@Override
 				public void actionPerformed(ActionEvent e) {
-					extension.showSpiderDialog(null);
+					extension.showSpiderDialog(getSiteTreeTarget());
 				}
 			});
 		}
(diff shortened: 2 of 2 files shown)


src/org/zaproxy/zap/extension/spider/SpiderDialog.java
```diff
@@ -112,8 +112,9 @@ public void init(Target target) {
         this.removeAllFields();
 
         this.addTargetSelectField(0, FIELD_START, this.target, true, false);
-        this.addComboField(0, FIELD_CONTEXT, new String[] {}, "");
+        this.addComboField(0, FIELD_CONTEXT, getContextNames(), "");
```
thc202: If we do this always (instead of just when there's no `target`?) we should validate that the starting point is not in conflict with selected context (e.g. as is now it's possible to selected "Context X" while having as starting point a "Context Y").

src/org/zaproxy/zap/extension/spider/SpiderDialog.java
```diff
@@ -424,4 +425,18 @@ void reset() {
         target = null;
         reset(true);
     }
+    /**
+     * Get a current list of the context names. (To populate the contexts combo field.)
+     * 
+     * @return the list of context names
+     * @since TODO add version
```
thc202: This could be removed, this method is not part of the "public API".

src/org/zaproxy/zap/extension/spider/SpiderDialog.java
```diff
@@ -424,4 +425,18 @@ void reset() {
...
+     */
+    private List<String> getContextNames() {
```
thc202: Can be declared static.

src/org/zaproxy/zap/extension/spider/SpiderDialog.java
kingthorin: I'll have a look at it, maybe I've over simplified something that was done intentionally.

thc202: Maybe, but this change makes the selection of the context/user more straightforward. If we don't pre-populate the context combo box when there's already a `target` we don't need to validate.

kingthorin: Ok maybe I just need to re-work this slightly. There is still a consistency issue with the original behavior. Though not the consistency issue I was initially aiming to correct (I was trying to make the button functionality closer to the context menu functionality.).

1. Open a session that has a context defined.
2. Don't select anything in the sites/context tree.
3. Click "New Scan" in the spider tab, "Select" a starting point within your context.
4. Hit cancel.
5. Hit "New Scan" again. 
6. Note: The starting point is still set, but the context selection is empty.

So I guess when a `SpiderDialog` is opened we should check if there is a target (starting point) and use that to set the context and user combo fields?
(Or is it the cancel/clear functionality that should be resetting the target?)

thc202: Both? We should use the existing target to set up the combo boxes and cancelling the dialogue should restore the previous settings (iirc, we talked about that already).

kingthorin: I think we talked about it for active scan policies, and elsewhere probably.... or I may have forgotten (I apologize if it's the later).

kingthorin:
This doesn't necessarily cover everything discussed previously but should be a step in the right direction (I hope).

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -49,9 +49,11 @@
 import org.jdesktop.swingx.renderer.MappedValue;
 import org.jdesktop.swingx.renderer.StringValues;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.model.ScanController;
 import org.zaproxy.zap.model.ScanListenner2;
+import org.zaproxy.zap.model.Target;
```
thc202: This unused import should be removed.

src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -541,4 +543,13 @@ protected boolean isHighlighted(final Object cellItem) {
 			return true;
 		}
 	}
+	
+	private SiteNode getSiteTreeTarget() {
+		if (!extension.getView().getSiteTreePanel().getTreeSite().isSelectionEmpty()) {
+			SiteNode targetNode = (SiteNode) extension.getView().getSiteTreePanel().getTreeSite().getSelectionPath()
```
thc202: Could return the value immediately.

thc202:
Thanks!

Issue #1984, opened 2015-10-16
Alerts API should always return 'evidence' (and all other keys)


tbach:
First noticed when upgrading automation from ZAP 2.3.1 to 2.4.x.

In the past (and in most cases in 2.4.x) when ZAP's API returns alerts, if keys like 'evidence' do not have any data they are returned as an empty string. In 2.4.x, certain alerts seem to not include the 'evidence' key if there is no data. This may affect other keys as well, but this is the one I've noticed.

In example below, note that the first several alerts include an empty value for the evidence key whereas the final output does not include the evidence key at all. This seems to affect at least alert 10015 (Incomplete or no cache-control and pragma HTTPHeader set).

Expected behavior: All alert keys are returned in each alert dict with empty string values if they are not present in the alert. This allows automation to assume dict structure.

Ex:

> > > for alert in zap.core.alerts():
> > > ...     print alert
> > > ...     print alert['evidence']

{u'description': u"The Anti-MIME-Sniffing header X-Content-Type-Options was not set to 'nosniff'. This allows older versions of Internet Explorer and Chrome to perform MIME-sniffing on the response body, potentially causing the response body to be interpreted and displayed as a content type other than the declared content type. Current (early 2014) and legacy versions of Firefox will use the declared content type (if one is set), rather than performing MIME-sniffing.", u'url': u'https://demo.testfire.net', u'other': u'This issue still applies to error type pages (401, 403, 500, etc) as those pages are often still affected by injection issues, in which case there is still concern for browsers sniffing pages away from their actual content type.\nAt "High" threshold this scanner will not alert on client or server error responses.', u'reference': u'http://msdn.microsoft.com/en-us/library/ie/gg622941%28v=vs.85%29.aspx\nhttps://www.owasp.org/index.php/List_of_useful_HTTP_headers', u'solution': u"Ensure that the application/web server sets the Content-Type header appropriately, and that it sets the X-Content-Type-Options header to 'nosniff' for all web pages.\nIf possible, ensure that the end user uses a standards-compliant and modern web browser that does not perform MIME-sniffing at all, or that can be directed by the web application/web server to not perform MIME-sniffing.", u'param': u'', u'confidence': u'Medium', u'alert': u'X-Content-Type-Options Header Missing', u'evidence': u'', u'messageId': u'1', u'wascid': u'15', u'id': u'5', u'cweid': u'0', u'attack': u'', u'risk': u'Low'}

[...]

PR #3858, opened 2017-08-31
Issue API should always return all Alert elements

kingthorin:
* `CoreAPI` ensure messageId is always included in API output.
* `Alert` ensure element setters set empty string if passed null. (So that all elements ultimately end up in API responses [even if empty strings]). Also cleaned up unnecessary `//ZAP` comments.

Fixes zaproxy/zaproxy#1984


src/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -1480,6 +1480,8 @@ private void getURLs(String baseUrl, SiteNode parent, ApiResponseList list) {
 		map.put("solution", alert.getSolution());
 		if (alert.getHistoryRef() != null) {
 			map.put("messageId", String.valueOf(alert.getHistoryRef().getHistoryId()));
+		} else {
+			map.put("messageId", "");
```
thc202: Could use the ternary operator like in the setters (it would avoid the duplication of the name).

thc202:
Thank you!

Issue #3832, opened 2017-08-18
Errors in Input Vector Templates


kingthorin:
[Input Vector default template L44](https://github.com/zaproxy/zaproxy/blob/c93ef3376b23d71bc2ab1249d15ef4cdef8e51f0/src/scripts/templates/variant/Input%20Vector%20default%20template.js#L44)
Should use `>=` not just `>` 
(I'm not sure about the other one because it might be based on some specific use case I'm unaware of ... what's the google implication? [Input Vector sharp query separator.js L13](https://github.com/zaproxy/zaproxy/blob/c93ef3376b23d71bc2ab1249d15ef4cdef8e51f0/src/scripts/templates/variant/Input%20Vector%20sharp%20query%20separator.js#L13))

Calls to `getParamName` and `getParamValue` should use the `helper.` prefix.
* https://github.com/zaproxy/zaproxy/blob/c93ef3376b23d71bc2ab1249d15ef4cdef8e51f0/src/scripts/templates/variant/Input%20Vector%20default%20template.js#L48-L49
* https://github.com/zaproxy/zaproxy/blob/develop/src/scripts/templates/variant/Input%20Vector%20sharp%20query%20separator.js#L38-L39

thc202:
Seems to me that's just to do something with the number of extracted parameters. If the original intention was to check for `>=` then the check is not needed (the function `setParameter` would not be called at all if there are no parameters to attack)?

Regarding the second script, that's because of the use of parameters in the fragment, but since the fragment is not sent to the server it does not have much effect.

kingthorin:
Thanks @thc202, if these are templates should we remove or comment pieces that might mislead or confuse users? For example if building a new script variant it isn't much of a leap for the test request to have a single param....

thc202:
Yeah, the simpler the better. The introduction to other useful methods can be done with comments? (We do that in other templates.)

kingthorin:
For context the reason this came up is:
1) In Pixi it's possible to conduct a NoSQLi attack on the login page, however it actually requires manipulation of the param name. (https://hub.docker.com/r/deadrobots/pixi/, https://github.com/thedeadrobots/pixi)
2) Because I was looking into the above I was reminded of https://github.com/zaproxy/zaproxy/issues/343 ...

When I began using the template(s) I ran into the gotchas above... :zap:

thc202:
Did you implement already the input vector that attacks the names? Which "parameters" does it attack?

[...]

PR #3867, opened 2017-09-06
Issue 3832 Fix Input Vector Variant Templates

kingthorin:
* Add a comment outlining where to find more details on the helper parameter.
* Modify setParameters method to apply if there are any params (vs. previous >1 application).
* Ensure the `helper.` prefix is used where necessary to avoid errors when using the templates.
* Remove "Input Vector sharp query separator", since URI fragments aren't sent to the server.
* Overhaul "Input Vector default template".

Fixes zaproxy/zaproxy#3832


kingthorin:
In removing an `if` condition I also had to adjust whitespace to account for indentation.

src/scripts/templates/variant/Input Vector default template.js
```diff
@@ -41,19 +45,17 @@ function parseParameters(helper, msg) {
...
-        
-        for (var i = 0; i < size; i++) {
-            pname = getParamName(i);
-            pvalue = getParamValue(i);                
-            if (pname == param) {      
-                    pvalue = helper.encodeBase64(value);
-            }
-                    
-            query = pname + "=" + pvalue + "&" + query;
+    query = helper.getParamValue(size);
```
thc202: Just realised that this should be `size - 1`, to obtain the `notb64` parameters.
Also, in the statement:
```JavaScript
nob64 = nob64 + "&" + vars;
```
it should be:
```JavaScript
nob64 = nob64 + "&" + vars[i];
```
(to just add the current parameter, not all of them)

If you are OK changing the other semantic issues I have other suggestions (like use a local variable, instead of adding "helper" parameter).

thc202:
Feel free to remove the whitespaces and do the other tweaks in this PR, the statements are already being (re)indented/changed so it will not make much difference.

I'd suggest completely removing `Input Vector sharp query separator.js`, that will not work, the fragment is not used by ZAP (that's actually encoded when parsed).

src/scripts/templates/variant/Input Vector default template.js
kingthorin: Sure, just let me know.

kingthorin:
Done & done. Let me know what other changes you had in mind, happy to implement them.

src/scripts/templates/variant/Input Vector default template.js
```diff
@@ -41,19 +45,17 @@ function parseParameters(helper, msg) {
...
-            }
-                    
-            query = pname + "=" + pvalue + "&" + query;
+    query = helper.getParamValue(size-1);
+
+    for (var i = 0; i < size; i++) {
+        pname = helper.getParamName(i);
+        pvalue = helper.getParamValue(i);
+        if (pname == param) {
+                pvalue = helper.encodeBase64(value);
```
thc202: The return value should be URL encoded (it's calling `setEscapedQuery`).

thc202:
Adding some comments here since I can't comment this line in the Files tab:
https://github.com/kingthorin/zaproxy/blob/c97f5881744e5e49be58d83dc98f6f4a020676a9/src/scripts/templates/variant/Input%20Vector%20default%20template.js#L21
should call `getEscapedQuery()` and `pair[1]` needs to be URL decoded (prior the Base64 checks).
The variable `nob64` could be declared globally and used in the two functions (no longer needs to add/handle the helper parameter).

src/scripts/templates/variant/Input Vector default template.js
```diff
@@ -41,19 +45,17 @@ function parseParameters(helper, msg) {
...
         }
-            
-        msg.getRequestHeader().getURI().setEscapedQuery(query);
+
+    query = pname + "=" + pvalue + "&" + query;
```
thc202: This should be a little to the right?

kingthorin:
This template is doomed.
`pair = vars[i].split("=");` can't work simply on base64 values since they're potentially right padded with `=`.

src/scripts/templates/variant/Input Vector default template.js
kingthorin: Can change to `pvalue = helper.encodeBase64(encodeURI(value));`

kingthorin:
This "template" is proving challenging. There has to be a better way to dissect and reassemble the query parameters. I tried testing with a URL like: `http://example.org/?foo=bar+bar&thing=Zm9vYmFyZg==` it's really ugly....

thc202:
`=` are supposed to be escaped in the query component, the use of `getEscapedQuery` should prevent the wrong splits (since `=` would still be encoded).

kingthorin:
`=` isn't encoded.
```
console.log(encodeURI("fred=="));
fred==
```
So you have to split it and re-pad it. Though I've got that dealt with.

In this latest iteration I tested just this variant with an active scan script that just injected `B` as it's attack. I tested with a base URL of:
http://example.org/?foo=bar+bar&thing=Zm9vYmFyZg==
Which resulted in 3 active scan requests (as expected):
http://example.org/2074679664341891143 - Analyzer
http://example.org/?thing=Zm9vYmFyZg==&foo=B - Attacked the `foo` param
http://example.org/?thing=Qg==&foo=bar+bar - Attacked the `thing` param

thc202:
In this case we should use `encodeURIComponent`/`decodeURIComponent` as we are handing the query component (or parts of it), those should encode/decode the `=`.

Instead of tagging we could use an array with the "state" of the parameters?

src/scripts/templates/variant/Input Vector default template.js
```diff
@@ -4,56 +4,75 @@
...
+            if (pname == param) {
+                //Encode the injected value
+                pvalue = encodeURI(helper.encodeBase64(value));
+            } else {
+                //Use the original value; strip the tag
+                pvalue = encodeURI(helper.encodeBase64(pvalue.slice(B64_TAG.length)));
+            }
+        } else { //Handle regular values
+            if (pname == param) {
+                pvalue = value;
```
thc202: Should check `escaped` to encode or not the value?

thc202:
Thank you!

Issue #3854, opened 2017-08-30
Error whilst using Api - ERROR ExtensionUserManagement - Unable to persist Users


FlickerBean:
Getting this error whilst trying to use Form Based Authentication through the python api.

```
2017-08-30 09:18:11,542 [ZAP-ProxyThread-5] ERROR ExtensionUserManagement - Unable to persist Users.
java.lang.NullPointerException
        at org.zaproxy.zap.authentication.ManualAuthenticationMethodType$ManualAuthenticationCredentials.encode(ManualAuthenticationMethodType.java:203)
        at org.zaproxy.zap.users.User.encode(User.java:290)
        at org.zaproxy.zap.extension.users.ExtensionUserManagement.persistContextData(ExtensionUserManagement.java:271)
        at org.parosproxy.paros.model.Model.saveContext(Model.java:505)
        at org.parosproxy.paros.model.Session.saveContext(Session.java:1166)
        at org.zaproxy.zap.model.Context.save(Context.java:450)
        at org.zaproxy.zap.extension.users.UsersAPI.handleApiAction(UsersAPI.java:174)
        at org.zaproxy.zap.extension.api.API.handleApiRequest(API.java:431)
        at org.parosproxy.paros.core.proxy.ProxyThread.processHttp(ProxyThread.java:436)
        at org.parosproxy.paros.core.proxy.ProxyThread.run(ProxyThread.java:297)
        at java.lang.Thread.run(Thread.java:748)
2017-08-30 09:18:11,553 [ZAP-ProxyThread-6] WARN  API - ApiException while handling API request:
Missing Parameter (missing_parameter) : sessionName
        at org.zaproxy.zap.utils.ApiUtils.getNonEmptyStringParam(ApiUtils.java:83)
        at org.zaproxy.zap.authentication.ManualAuthenticationMethodType$2.handleAction(ManualAuthenticationMethodType.java:423)
        at org.zaproxy.zap.extension.users.UsersAPI.handleApiAction(UsersAPI.java:217)
        at org.zaproxy.zap.extension.api.API.handleApiRequest(API.java:431)
        at org.parosproxy.paros.core.proxy.ProxyThread.processHttp(ProxyThread.java:436)
        at org.parosproxy.paros.core.proxy.ProxyThread.run(ProxyThread.java:297)
        at java.lang.Thread.run(Thread.java:748)
```

I note a similar bug was fixed some time back: https://github.com/zaproxy/zaproxy/issues/1770

[...]

PR #3907, opened 2017-09-26
Normalise form-based login URL validation/usage

thc202:
Change FormBasedAuthenticationMethodType to validate and use the login
URL always in encoded form, instead of encoded in some places and
decoded in others. Attempting to set a decoded/improperly encoded login
URL could lead to a URIException preventing the URL from being set.
While the URL had to be encoded to be set it would later be used as if
it was not encoded (leading to double encoding).
This change might break existing API users that are setting a login URL
partially/improperly encoded.

Related to #3854 - Error whilst using Api - ERROR
ExtensionUserManagement - Unable to persist Users


thc202:
Edited the pull request and commit message to correct the exception thrown (a `URIException` not `NullPointerException`).

PR #3910, opened 2017-09-27
Allow to skip scanners through the API

thc202:
Change ActiveScanAPI to allow to skip the scanners.
Change HostProcess to allow to skip scanners by ID and don't allow to
skip scanners already finished/skipped or if the scan was already
stopped.

### DIFF
--- a/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
+++ b/src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
@@ -107,6 +107,8 @@ public class ActiveScanAPI extends ApiImplementor {
 	private static final String ACTION_MODIFY_EXCLUDED_PARAM = "modifyExcludedParam";
 	private static final String ACTION_REMOVE_EXCLUDED_PARAM = "removeExcludedParam";
 
+	private static final String ACTION_SKIP_SCANNER = "skipScanner";
+
 	private static final String VIEW_STATUS = "status";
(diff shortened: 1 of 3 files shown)


src/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -537,6 +542,15 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
 				}
 				break;
+			case ACTION_SKIP_SCANNER:
+				int pluginId = getParam(params, PARAM_SCANNER_ID, -1);
+				if (pluginId == -1) {
+					throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, PARAM_SCANNER_ID);
+				}
+
+				String reason = Constant.messages.getString("ascan.progress.label.skipped.reason.user");
+				getActiveScan(params).getHostProcesses().forEach(hp -> hp.pluginSkipped(pluginId, reason));
```
kingthorin: I'm sure this is right but I'd like to understand this `->` syntax a bit better. I assume this works because `getHostProcesses()` returns an iterable. So `forEach` `hp` (host process) in the list (or whatever type of iterable) execute `pluginSkipped(pluginId, reason)`. Is that the correct interpretation?

thc202: Yeah, that's it.

Issue #3912, opened 2017-09-28
VariantDirectWebRemotingQuery might hang in infinite loop


vmolotkov:
[diff.txt](https://github.com/zaproxy/zaproxy/files/1340287/diff.txt)
There is obvious bug in method 

`org.parosproxy.paros.core.scanner.VariantDirectWebRemotingQuery.parseContent(String).`

The following two conditions, if true, cause infinite loop:
```
...
           if ( patternIgnoreArray.matcher(paramDetails[1]).matches()) continue; //to the next iteration
           if ( patternIgnoreObject.matcher(paramDetails[1]).matches()) continue; //to the next iteration
...
```
Active scanner threads get stuck and scan process freezes eventually.
Proposed fix in the attachment.

thc202:
Would you mind opening a pull request instead? [1]

[1] https://help.github.com/articles/creating-a-pull-request-from-a-fork/

vmolotkov:
Done

thc202:
Thanks!

kingthorin:
Fixed in https://github.com/zaproxy/zaproxy/pull/3913

PR #3913, opened 2017-09-28
VariantDirectWebRemotingQuery might hang in infinite loop #3912

vmolotkov:
Fix for VariantDirectWebRemotingQuery might hang in infinite loop #3912


thc202:
Thank you!

vmolotkov:
Vladimir Molotkov (BrandMaker GmbH)

Issue #3322, opened 2017-03-24
Switch JSON variant parser to use a library instead of 'maual' parsing


kingthorin:
The `VariantJSONQuery` parser 'manually' parses JSON content. It works but there are a few issues. It's manual parsing should be replaced leveraging an established library.

juhakivekas:
https://github.com/zaproxy/zaproxy/issues/3895#issuecomment-330473425
what library should be used here? net.sf.json?

thc202:
Yes, that should suit our needs.

juhakivekas:
I explored the possibilities for this and did not find a solution that is easy enough to be quickly implemented. Some of the json parsing libraries I examined did not support getting the offset of parsed tokens (json-lib, gson) and while jackson would have been good for the case, it seemed that it's a little on the big side for such a small component as the variant. The parser is still homebrew, but at least some bugs are fixed.

thc202:
That's fine, just note that we really don't need the offsets (the variant can be changed to directly implement `Variant`), we could use a Java class representation of the JSON payload (e.g. `JSONObject`, `JSONArray`), similar to what you did in the custom JSON variant in the community-scripts :)

thc202:
See #7798.

PR #3915, opened 2017-09-29
quick fix for json variant parser

### LINKED ISSUE #3895: JSON Input Vector fails if string has quoted chars
There seems to be a very fundamental issue with the JSON variant parser. The following HTTP post body causes an exception when an active scan is started with the following request (only POST parameter chosen):

	POST http://example.com/ HTTP/1.1
	Connection: keep-alive
	Content-Type: application/json
	Content-Length: 12
	Proxy-Connection: Keep-Alive
	Host: example.com
	
	{"foo":"\""}

ZAP stacktrace when a scanner is run on the request:

	3109792 [ZAP-ActiveScanner-0] WARN org.parosproxy.paros.core.scanner.VariantJSONQuery  - Failed to parse the request body: Object not ended with '}' or ']' at position 11
	java.lang.IllegalArgumentException: Object not ended with '}' or ']' at position 11
		at org.parosproxy.paros.core.scanner.VariantJSONQuery.parseObject(Unknown Source)
		at org.parosproxy.paros.core.scanner.VariantJSONQuery.parseContent(Unknown Source)
		at org.parosproxy.paros.core.scanner.VariantAbstractRPCQuery.setRequestContent(Unknown Source)
		at org.parosproxy.paros.core.scanner.VariantAbstractRPCQuery.setMessage(Unknown Source)
		at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scan(Unknown Source)
		at org.parosproxy.paros.core.scanner.AbstractPlugin.run(Unknown Source)

My versions:

	ZAP 2.6.0
	Active scanner release quality rules version 26
	Java 1.8.0_77

juhakivekas:
Fixes #3895, but not #3322 
Also fixes minor bug that causes the filed name to include the last quotation mark. It's the added `-1` and it came up when writing tests.

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
+++ b/src/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
@@ -37,6 +37,7 @@ public class VariantJSONQuery extends VariantAbstractRPCQuery {
     public static final int BEGIN_OBJECT = '{';
     public static final int END_OBJECT = '}';
     public static final int END_ARRAY = ']';
+    public static final int BACKSLASH = '\\';
 
     private SimpleStringReader sr;
 
@@ -99,7 +100,7 @@ public class VariantJSONQuery extends VariantAbstractRPCQuery {
                     chr = sr.skipWhitespaceRead();
                     if (chr == BEGIN_OBJECT) {
                         objectRead = true;
-                                
+
                         chr = sr.skipWhitespaceRead();
                         if (chr == END_OBJECT) {    // empty object
                             return;
@@ -123,12 +124,9 @@ public class VariantJSONQuery extends VariantAbstractRPCQuery {
                     if (chr == QUOTATION_MARK) {
                         
                         beginToken = sr.getPosition();
-                        while ((chr = sr.read()) != QUOTATION_MARK) {
-                            if (chr == -1)
-                                throw new IllegalArgumentException("EOF reached while reading JSON field name");
-                        }
+                        readEscapedString();
                         
-                        endToken = sr.getPosition();                        
+                        endToken = sr.getPosition() - 1;
                         // Now we have the string object name
                         // we can do something here for value filtering...
                         field = getToken(beginToken, endToken);
@@ -189,11 +187,7 @@ public class VariantJSONQuery extends VariantAbstractRPCQuery {
         // Check if the value is a string        
         if (chr == QUOTATION_MARK) {
             int beginToken = sr.getPosition();
-            while ((chr = sr.read()) != QUOTATION_MARK) {
-                if (chr == -1) {
-                    throw new IllegalArgumentException("EOF reached while reading JSON field name");
-                }
-            }
+            readEscapedString();
             
             // Now we have the string object value
             // Put everything inside the parameter array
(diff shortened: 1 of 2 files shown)


test/org/parosproxy/paros/core/scanner/VariantJSONQueryTest.java
```diff
@@ -0,0 +1,75 @@
+package org.parosproxy.paros.core.scanner;
```
kingthorin: Should probably have the standard license header

src/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
```diff
@@ -189,8 +190,11 @@ private void parseValue(String fieldName) {
         // Check if the value is a string        
         if (chr == QUOTATION_MARK) {
             int beginToken = sr.getPosition();
-            while ((chr = sr.read()) != QUOTATION_MARK) {
-                if (chr == -1) {
+            while (true) {
```
thc202: Worth extracting a method and change the read of the property name to also use it (to fixing that one as well).

src/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
```diff
@@ -189,8 +190,11 @@ private void parseValue(String fieldName) {
         // Check if the value is a string        
         if (chr == QUOTATION_MARK) {
             int beginToken = sr.getPosition();
-            while ((chr = sr.read()) != QUOTATION_MARK) {
-                if (chr == -1) {
+            while (true) {
+                chr = sr.read();
+                if (chr == BACKSLASH) { sr.read(); }
```
thc202: The statements should be in a new line (just a matter of consistency with the rest of the code).

test/org/parosproxy/paros/core/scanner/VariantJSONQueryTest.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Unit test for {@link VariantJSONQuery}.
+ */
+public class VariantJSONQueryTest {
```
thc202: It should end with `UnitTest` to be automatically picked by Ant test target.

test/org/parosproxy/paros/core/scanner/VariantJSONQueryTest.java
```diff
@@ -0,0 +1,75 @@
...
+        // Given
+        VariantJSONQuery variantJSONQuery = new VariantJSONQuery();
+        // When
+        List<NameValuePair> parameters = variantJSONQuery.getParamList();
+        // Then
+        assertThat(parameters, is(empty()));
+    }
+
+    @Test
+    public void shouldFindStringValueInContent(){
```
thc202: I'd be more explicit and say "Object" (or "JsonObject") instead of "Content" (for when the tests check an array).

test/org/parosproxy/paros/core/scanner/VariantJSONQueryTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertThat(parameters.get(0).getName(), is("foo"));
+        assertThat(parameters.get(0).getValue(), is("bar\""));
+    }
+
+    private HttpMessage getMessageWithBody(String body){
+        try {
+            return new HttpMessage(
+                    new HttpRequestHeader("POST / HTTP/1.1\nHost: www.example.com\nContent-Type: application/json"),
+                    new HttpRequestBody(body));
+        } catch(HttpMalformedHeaderException e) {
```
thc202: It's preferable to re-thrown the exception (or don't catch at all). This might start to fail if the parsing is changed (e.g. be stricter).

src/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
```diff
@@ -294,6 +288,23 @@ private void parseToken(String token) {
         }
     }
 
+    private void readEscapedString(){
+        int chr;
+        while (true) {
+            chr = sr.read();
+            if (chr == BACKSLASH) {
+                sr.read();
+                continue;
```
thc202: This statement could be removed (does not add much, it already goes to next cycle).

thc202:
Thank you!

kingthorin:
Thanks @juhakivekas

PR #3919, opened 2017-10-02
Add Java 9 to Travis CI builds

thc202:
Change tests to not use PowerMock which has issues with Java 9, moreover
most of the tests didn't actually require it (e.g. Constant.messages is
accessible).
Change PluginFactory to allow to initialise it without using add-ons (to
ease the test setup).


thc202:
Was just to verify that ZAP does not compile with Java 9...

kingthorin:
Are issues going to be raised based on these results? Do we want some of this worked on?

thc202:
Already working on fixing some of the issues (some of them might have to wait because of BC...). We already have a Java 9 issue (#2602), not sure if we need to raise another one (compilation specific?).

kingthorin:
Probably not. I don't mind digging up guidance on some of the depredations and tackling them. Let me know if there's something I can tackle.

thc202:
Opened two PRs that address the errors (#3924) and some deprecations (#3923), there's still "a lot" of deprecations, specially with `Observable`/`Observer`. Didn't take a look at them in detail but I guess some of them can be suppressed, in others use the new Flow API and in others just deprecate the classes...

thc202:
We also need to "fix" the tests, some of them are failing.

thc202:
Rebased and updated (the build is now green).

PR #3928, opened 2017-10-04
Fixes #3927: zap-full-scan on docker tries to reach out to external network

### LINKED ISSUE #3927: zap-full-scan on docker tries to reach out to external network
When running zap-full-scan.py in the owasp/zap2docker-weekly image it fails when there is no access to the outside internet. It's trying to reach out to the external network to check the client API version number (https://pypi.python.org/pypi/python-owasp-zap-v2.4/json). I'm using a self-contained docker container with ZAP and the Python API already installed and they should be matching versions. Why does it need to verify this? And why does it need to try to hit an external URL to do it?

I'm running this inside of an isolated network that can't reach outside and the system under test is another docker container on the same host, so it should not be required to enable external network access to use it.

mpeters:


build/docker/zap_common.py
```diff
@@ -374,6 +374,10 @@ def get_latest_zap_client_version():
 
 
 def check_zap_client_version():
+    if running_in_docker():
+        logging.info('No need to check ZAP Python API client while running in Docker')
```
thc202: I'd not log at all, if you prefer log something I'd suggest debug instead of info, it does not add much for the usage of the script in Docker.

mpeters: Done

thc202:
Thank you!

PR #3933, opened 2017-10-05
Include body size in Large Response HTTP views

### LINKED ISSUE #3931: Minor Enhancement to Very large response body message
When ZAP decides its not a good idea to show the response body, it displays the above message. However, the user really has no clue how BIG it really is unless they show it.

To help with that, please enhance this message to say:

Very large response body (##### bytes) - switch views ...

Where ##### is the # of bytes in the response body.

thc202:
Change Large Response HTTP views to include the size of the body in the
information message shown.

Fix #3931 - Minor Enhancement to Very large response body message


davewichers:
Wow guys. That was fast. Thx.

Issue #3959, opened 2017-10-17
Properly generate cert for ZAP API and Callback when accessed with IP address


denniskniep:
When I use https within the API or Callback Url I get an error:

Firefox: Secure Connection Failed (SSL_ERROR_INTERNAL_ERROR_ALERT)
Chrome: This site can’t provide a secure connection (ERR_SSL_PROTOCOL_ERROR)

thc202:
With 2.6.0 (or dev/weekly)? While proxying or accessing directly? Java version?

denniskniep:
Currently working with a dev build. I am accessing the Url without  proxying. (If I am proxying there is no problem with the API).  I think Callback is never through proxy right? 
My Java Version is 1.8

thc202:
Which revision/commit did you use to build? If it's working while proxying it might be a problem with SNI. (Yeah, for the callback it does not need to proxy just wanted to know if it was working for the API.)

No errors in the log when that happens?

denniskniep:
I tried it with the latest commit 8017bc48ea0b5ba5278ca1a2a871d9ded3a13028

The Log output was:
`[ZAP-ProxyThread-34] WARN  SSLConnector - No domain extracted from SSL/TLS handshake session.`

thc202:
Yeah, that's a problem with SNI. Which version of Firefox are you using? And what's the "update" Java version?

denniskniep:
Firefox 56 
Java 1.8.0_131

denniskniep:
The `ExtendedSSLSession` contains no `SNIServerName`
I accessed the API and the CallbackEndpoint with an Ip-Address, not with an DNS hostname.

The ["Server Name Indication"](https://de.wikipedia.org/wiki/Server_Name_Indication) is defined in the  [RFC6066](https://tools.ietf.org/html/rfc6066#page-6).

According to the RFC there are no Ip-Addresses allowed:

> Currently, the only server names supported are DNS hostnames;
> Literal IPv4 and IPv6 addresses are not permitted in "HostName".

Maybe a workaround is to issue the certificate with all in Zap used Ip-Addresses and HostNames into the SubjectAlternativeName.
https://en.wikipedia.org/wiki/Subject_Alternative_Name

thc202:
Right, I assumed you were using the hostname not the IP address.

PR #3975, opened 2017-10-23
Fix for #3959, #4016;. Create valid cert for an SSLEndpoint accessed in the browser by an IpAddress.

denniskniep:
Added the IpAddress to the cert as SubjectAlternativeName.


denniskniep:
I implemented an other approach in contrast to the approach proposed in the issue. The cert is issued only with the needed IpAddress. Not with all possibles. I think fewer is generally better.

denniskniep:
Currently there is a "ERR_CERT_COMMON_NAME_INVALID" problem when proxying through zap and calling the API or the CallbackEndpoint with an IpAddress. The IpAddress is in the hostname. But I think this is a seperate issue? Should I raise one or should it be fixed inside this PR?

src/org/parosproxy/paros/security/SslCertificateService.java
```diff
@@ -63,7 +65,7 @@
 	 * @throws UnrecoverableKeyException
 	 * @throws MissingRootCertificateException when it wasn't initialized.
 	 */
-	KeyStore createCertForHost(String hostname)
+	KeyStore createCertForHost(String hostname, InetAddress listeningAddress)
```
thc202: I'd suggest using a default method instead (to keep binary compatibility).

thc202: How about `KeyStore createCert(CertData)`? (With `CertData` containing the desired common name and alternative (IPs/domain) names.) That would be more flexible for the caller and allow for future changes without breaking binary compatibility.

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -530,7 +531,8 @@ public Socket createSocket(Socket socket, String host, int port,
 	 */
 	public Socket createTunnelServerSocket(String targethost, Socket socket) throws IOException {
 		// ZAP: added host name parameter
-		SSLSocket s = (SSLSocket) getTunnelSSLSocketFactory(targethost).createSocket(socket, socket
+		InetAddress listeningAddress = socket.getInetAddress();
```
thc202: This should call `getLocalAddress()`, otherwise we are generating the cert with remote address.

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -539,7 +541,7 @@ public Socket createTunnelServerSocket(String targethost, Socket socket) throws
 	}
 
 	// ZAP: added new ServerSocketFaktory with support of dynamic SSL certificates
-	public SSLSocketFactory getTunnelSSLSocketFactory(String hostname) {
+	public SSLSocketFactory getTunnelSSLSocketFactory(String hostname, InetAddress listeningAddress) {
```
thc202: In this case we should also introduce a new method (and deprecate the "old" one?).

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -673,11 +677,11 @@ private void createX509KeyManager(Socket socket) {
 			String hostname = extractHostname(sslSocket.getHandshakeSession());
 
 			if (hostname == null) {
-				logAndThrow("No domain extracted from SSL/TLS handshake session.");
+				logger.info("No domain extracted from SSL/TLS handshake session.");
```
thc202: This could be debug no longer critical (we will validate/fail later with a more relevant message).

src/org/parosproxy/paros/security/CachedSslCertifificateServiceImpl.java
```diff
@@ -51,18 +52,24 @@ private CachedSslCertifificateServiceImpl() {
...
+			return this.cache.get(key);
 		}
-		final KeyStore ks = delegate.createCertForHost(hostname);
-		this.cache.put(hostname, ks);
+		final KeyStore ks = delegate.createCertForHost(hostname, listeningAddress);
+		this.cache.put(key, ks);
 		return ks;
 	}
 
+	private String getKey(String hostname, InetAddress listeningAddress) {
```
thc202: This could be static.

src/org/parosproxy/paros/security/CachedSslCertifificateServiceImpl.java
```diff
@@ -51,18 +52,24 @@ private CachedSslCertifificateServiceImpl() {
...
+		String keyForCache = String.format("Host:%s;InetAddress:%s", hostname, listeningAddress.getHostAddress());
```
thc202: It could return the value immediately (the method name already indicates what this is).

src/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -142,7 +142,24 @@ public KeyStore createCertForHost(String hostname) throws NoSuchAlgorithmExcepti
...
-		certGen.addExtension(Extension.subjectAlternativeName, false, new GeneralNames(new GeneralName(GeneralName.dNSName, hostname)));
+
+		ArrayList<GeneralName> subjectAlternativeNames = new ArrayList<>();
+		boolean subjectAlternativeNamesIsCritical = true;
+		if (hostname != null) {
+			subjectAlternativeNames.add(new GeneralName(GeneralName.dNSName, hostname));
+			subjectAlternativeNamesIsCritical = false;
+		}
+
+		if(!listeningAddress.isAnyLocalAddress()){
```
thc202: - Does this (not) happen? Even if the socket is bound to all addresses we get a specific IP address.
 - We could keep the old behaviour and just add the IP address if we don't have the hostname.

src/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -142,7 +142,24 @@ public KeyStore createCertForHost(String hostname) throws NoSuchAlgorithmExcepti
...
+			subjectAlternativeNames.add(new GeneralName(GeneralName.iPAddress, listeningAddress.getHostAddress()));
+		}
+
+		if (hostname == null && listeningAddress.isAnyLocalAddress()) {
```
thc202: This validation could be done earlier (although see previous comment about wildcard address).

thc202:
> But I think this is a separate issue? Should I raise one or should it be fixed inside this PR?

I'm fine this being handled in another PR/issue (it was already failing), but could also be fixed in this PR (does not require much changes), as you prefer.

denniskniep:
Created an Issue for that: #4013

thc202:
Thank you.

thc202:
Now in #4016 :)

src/org/parosproxy/paros/security/CertData.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public GeneralName[] getSubjectAlternativeNames() {
+        GeneralName[] subjectAlternativeNamesArray = new GeneralName[subjectAlternativeNames.size()];
+        return subjectAlternativeNames.toArray(subjectAlternativeNamesArray);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = commonName != null ? commonName.hashCode() : 0;
+        result = 31 * result + (subjectAlternativeNames != null ? subjectAlternativeNames.hashCode() : 0);
```
juhakivekas: This looks like magic, but it's unlikely to collide anyway

VANSMOOKEYMAWA:
Leave a comment GOOD

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -529,17 +534,28 @@ public Socket createSocket(Socket socket, String host, int port,
...
 		return s;
 	}
 
-	// ZAP: added new ServerSocketFaktory with support of dynamic SSL certificates
+
+	/**
+	 * @deprecated (TODO add version) No longer used/needed.
+	 */
+	@Deprecated
+	// ZAP: added new ServerSocketFactory with support of dynamic SSL certificates
```
kingthorin: Add the ZAP comment into the javadoc so that the annotation appears right before the applicable code?

denniskniep: I added the // ZAP comment into the javadoc

psiinon:
@thc202 ok to merge this?

thc202:
Not as is, will add the comments today.

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -32,6 +32,8 @@
 // ZAP: 2015/10/13 Issue 1975: Allow use of default disabled cipher suites (such as RC4-SHA)
 // ZAP: 2017/04/14 Validate that SSLv2Hello is set in conjunction with at least one SSL/TLS version.
 // ZAP: 2017/09/22 Rely on SNI if the domain is no known when creating the SSL/TLS tunnel.
+// ZAP: 2017/10/23 Pass the listeningIpAddress when creating the SSL/TLS tunnel.
+// ZAP: 2017/11/07 Check if hostname is an ipAddress when creating the SSL/TLS tunnel.
```
thc202: Both comments could be merged.

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -529,17 +534,29 @@ public Socket createSocket(Socket socket, String host, int port,
...
 		s.startHandshake();
 		return s;
 	}
 
-	// ZAP: added new ServerSocketFaktory with support of dynamic SSL certificates
+
+	/**
+	 * @deprecated (TODO add version) No longer used/needed.
+	 *
+	 * ZAP: added new ServerSocketFactory with support of dynamic SSL certificates
```
thc202: Better just remove it (it should not be in the JavaDoc, also there's already related ZAP comments in the license header).

src/org/parosproxy/paros/network/SSLConnector.java
```diff
@@ -529,17 +534,29 @@ public Socket createSocket(Socket socket, String host, int port,
...
+	 */
+	@Deprecated
 	public SSLSocketFactory getTunnelSSLSocketFactory(String hostname) {
+		return getTunnelSSLSocketFactory(hostname, null);
+	}
+
+	// ZAP: added new ServerSocketFactory with support of dynamic SSL certificates
```
thc202: Can be removed (as previous comment).

src/org/parosproxy/paros/security/CachedSslCertifificateServiceImpl.java
```diff
@@ -48,18 +48,20 @@ private CachedSslCertifificateServiceImpl() {
 		delegate = SslCertificateServiceImpl.getService();
 	}
 
-	private Map<String, KeyStore> cache = new HashMap<>();
+	private Map<Integer, KeyStore> cache = new HashMap<>();
```
thc202: It should use `CertData` as key (the hash code is not guaranteed to be unique).

src/org/parosproxy/paros/security/CertData.java
```diff
@@ -0,0 +1,65 @@
...
+        if(commonName != null){
+            addSubjectAlternativeName(new GeneralName(GeneralName.dNSName, commonName));
+        }
+    }
+
+    public String getCommonName() {
+        return commonName;
+    }
+
+    public void addSubjectAlternativeName(GeneralName subjectAlternativeName){
```
thc202: I was suggesting allow to add the hostname and IP to avoid having this class depend on implementation details. Be good if it didn't depend on BC classes (like the interface doesn't).

src/org/parosproxy/paros/security/CertData.java
```diff
@@ -0,0 +1,65 @@
...
+        return commonName == null;
+    }
+
+    public GeneralName[] getSubjectAlternativeNames() {
+        GeneralName[] subjectAlternativeNamesArray = new GeneralName[subjectAlternativeNames.size()];
+        return subjectAlternativeNames.toArray(subjectAlternativeNamesArray);
+    }
+
+    @Override
+    public int hashCode() {
```
thc202: Should override equals as well.

src/org/parosproxy/paros/security/SslCertificateService.java
```diff
@@ -63,7 +64,32 @@
 	 * @throws UnrecoverableKeyException
 	 * @throws MissingRootCertificateException when it wasn't initialized.
 	 */
-	KeyStore createCertForHost(String hostname)
+	default KeyStore createCertForHost(String hostname)
```
thc202: This should be the other way around, the default should be the new method which would call the old method (older implementations wouldn't have the new method).

src/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -109,23 +110,25 @@ public synchronized void initializeRootCA(KeyStore keystore) throws KeyStoreExce
...
-    		throw new IllegalArgumentException("Error, 'hostname' is not allowed to be null!");
-    	}
+	@Override
+	public KeyStore createCertForHost(CertData certData) throws NoSuchAlgorithmException, InvalidKeyException, CertificateException, NoSuchProviderException, SignatureException, KeyStoreException, IOException, UnrecoverableKeyException {
 
     	if (this.caCert == null || this.caPrivKey == null || this.caPubKey == null) {
     		throw new MissingRootCertificateException(this.getClass() + " wasn't initialized! Got to options 'Dynamic SSL Certs' and create one.");
     	}
 
+		if (certData.getCommonName() == null && certData.getSubjectAlternativeNames().length == 0) {
```
thc202: Worth keeping the array (to avoid creating it twice).

src/org/parosproxy/paros/security/CertData.java
thc202: Note that the hash code is not unique so collisions will happen sooner or later.

src/org/parosproxy/paros/network/SSLConnector.java
kingthorin: Ok that's partially my fault, I think I suggested moving it here earlier. I didn't check the header comments ...

src/org/parosproxy/paros/security/CertData.java
denniskniep: So I should change the `getSubjectAlternativeNames` as well?

thc202: Yes, that would have to be changed too.

thc202:
This looks good, will tackle the conflict and update the code to stop using removed libraries (#4794).

thc202:
Conflict addressed and now using other library (Commons Validator) to check if a IP address (also removed unused imports and fixed a typo).

thc202:
@denniskniep thanks!

PR #3977, opened 2017-10-23
Include spider messages when comparing sessions

thc202:
Change ExtensionCompare to also include (normal and AJAX) spider
messages when creating the map for session comparisons.


thc202:
Yeah, that happened with two sessions that had just spider requests, while both sessions had a lot of nodes/messages in Sites tree the report was always empty.

kingthorin:
Ok

PR #3989, opened 2017-10-26
copy python scripts to docker image

omerlh:
All the python scripts are not copied to the bare image, this PR fixed it.
I'm closing this PR - there is no Python installed, and it is intentionally...

### DIFF
--- a/build/docker/Dockerfile-bare
+++ b/build/docker/Dockerfile-bare
@@ -31,4 +31,6 @@ ENV ZAP_PATH /zap/zap.sh
 ENV HOME /home/zap/
 ENV ZAP_PORT 8080
 
+COPY zap* /zap/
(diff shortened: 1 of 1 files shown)


thc202:
Right, was going to ask that :)

thc202:
(The scripts could still be useful if someone is creating an image (with python) based on this one, but they could use the stable in that case?)

omerlh:
Or getting it `wget`, which is how I found out python is missing...
But I think that we need either:
 - Change the script so it will connect to remote Zap host and then run it on a different machine.
 - Use [Glue](https://github.com/OWASP/glue) to run and process the results. Glue has some advantages, like filtering and reporting - so I think this the preferred option. 
What do you think?

PR #3990, opened 2017-10-27
rename function removeRequestDefaultViewSelectorFactoryAndDefaultView…

juhakivekas:
The removeRequestDefaultViewSelectorFactoryAndDefaultViewSelectorsAdded does not do what it says, but does what removeRequestDefaultViewSelectors would be expected to do.

This deprecates the old function but does not fix it to do what it should, because that would require the function arguments to include the name of the DefaultViewSelectorFactory to be deleted.

### DIFF
--- a/src/org/zaproxy/zap/view/HttpPanelManager.java
+++ b/src/org/zaproxy/zap/view/HttpPanelManager.java
@@ -112,7 +112,15 @@ public class HttpPanelManager {
 	public void removeRequestDefaultViewSelectorFactory(String componentName, String defaultViewSelectorFactoryName) {
 		requestPanels.removeDefaultViewSelectorFactory(componentName, defaultViewSelectorFactoryName);
 	}
-	
+
+	public void removeRequestDefaultViewSelectors(String componentName, String defaultViewSelectorName, Object options) {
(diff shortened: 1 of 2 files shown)


kingthorin:
Build is failing.

src/org/zaproxy/zap/view/HttpPanelManager.java
```diff
@@ -112,7 +112,12 @@ public void removeRequestViews(String componentName, String viewName, Object opt
 	public void removeRequestDefaultViewSelectorFactory(String componentName, String defaultViewSelectorFactoryName) {
 		requestPanels.removeDefaultViewSelectorFactory(componentName, defaultViewSelectorFactoryName);
 	}
-	
+
+	public void removeRequestDefaultViewSelectors(String componentName, String defaultViewSelectorName, Object options) {
+		requestPanels.removeDefaultViewSelectors(componentName, defaultViewSelectorName, options);
+	}
+
+	@Deprecated
```
thc202: Please also add a `@deprecated` JavaDoc tag pointing to new method (also add `(TODO add version)` at the beginning to later indicate the version in which was deprecated).

juhakivekas:
fixed build by renaming usages too, added propet javadoc annotaions

thc202:
Thanks!

PR #3992, opened 2017-10-30
Allow to enable code folding in body text views

thc202:
Change request/response body (syntax highlight) text views to allow to
enable code folding, makes it easier to see big HTML/XML/JSON bodies.


thc202:
(From talks in IRC.)

PR #3993, opened 2017-10-30
Added checkbox secure to callback option dialog. 

denniskniep:
This allows the usage of https when the callbackadress is requested with the method getCallbackAddress(). 

Only https in Ui and for creating the callback adress. Need additionally PR #3975 to work as expected.

### DIFF
--- a/src/org/zaproxy/zap/extension/callback/ExtensionCallback.java
+++ b/src/org/zaproxy/zap/extension/callback/ExtensionCallback.java
@@ -98,12 +98,13 @@ public class ExtensionCallback extends ExtensionAdaptor implements
 
     public String getCallbackAddress() {
         String addr = this.getCallbackParam().getRemoteAddress();
-        if (addr.contains(":")) {
-            // Looks like its IPv6
-            return "http://[" + addr + "]:" + actualPort + "/";
-        }
-        // Looks like IPv4
-        return "http://" + addr + ":" + actualPort + "/";
+        boolean ipv6 = addr.contains(":");
+        String hostname = ipv6 ? "[" + addr + "]" : addr;
+
+        boolean isSecure = this.getCallbackParam().isSecure();
+        String scheme = isSecure ? "https" : "http";
+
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/callback/ExtensionCallback.java
```diff
@@ -98,12 +98,14 @@ private void restartServer(int port) {
 
     public String getCallbackAddress() {
         String addr = this.getCallbackParam().getRemoteAddress();
+        boolean isSecure = this.getCallbackParam().isSecure();
+        String schema = isSecure ? "https" : "http";
```
thc202: `scheme`.

src/org/zaproxy/zap/extension/callback/ExtensionCallback.java
```diff
@@ -98,12 +98,14 @@ private void restartServer(int port) {
...
         if (addr.contains(":")) {
             // Looks like its IPv6
-            return "http://[" + addr + "]:" + actualPort + "/";
+            return schema + "://[" + addr + "]:" + actualPort + "/";
```
thc202: Worth remove the code duplication: `addr = "[" + addr + "]";` (and drop the IPv4 comment).

denniskniep: You mean this way?

```     
boolean ipv6 = addr.contains(":");
String ipAddr = ipv6 ? "[" + addr + "]" : addr;   
return schema + "://" + ipAddr + ":" + actualPort + "/";
```

thc202: Was suggesting:
```Java
if (addr.contains(":")) {
    // Looks like its IPv6
    addr = "[" + addr + "]";
}
return scheme + "://" + addr + ":" + actualPort + "/";
```
but that's fine as well (although the variable `ipAddr` should have other name, `addr` might be a hostname).

thc202:
Thanks!

PR #4005, opened 2017-11-03
Use %USERPROFILE% in zap.bat

### LINKED ISSUE #4004: zap.bat should use %USERPROFILE% instead of %HOMEDRIVE%%HOMEPATH%
ZAP Version: 2017-10-30 (weekly release)

## Suggenstion and Reasons

**zap.bat** should use `%USERPROFILE%` instead of `%HOMEDRIVE%%HOMEPATH%` to locate the file ".ZAP_JVM.properties".

There are two reasons:
1. [The user's default ZAP directory](https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsJvm) containing ".ZAP_JVM.properties" is [determined by "user.home" Java system property](https://github.com/zaproxy/zaproxy/blob/w2017-10-30/src/org/parosproxy/paros/Constant.java#L326) which usually equals to `%USERPROFILE%`, not `%HOMEDRIVE%%HOMEPATH%`.
2. `%HOMEDRIVE%%HOMEPATH%` can be different to `%USERPROFILE%`. (Actually I experienced it on a Windows machine managed by Active Directory. Of course, it caused an error same as #3211) See [about these variables](https://stackoverflow.com/questions/606483/what-is-the-meaning-of-these-windows-environment-variables-homedrive-homepath#8797855)

### Detail about reason 1

How "user.home" property is set is rather complicated, please see [stackoverflow answer](https://stackoverflow.com/questions/16889940/on-windows-7-how-does-java-jvm-set-user-home-system-property#16889969).

Here is summary:
* In Java 8: It equals to `%USERPROFILE%`
* In Java 7:  It seems to equal to `%USERPROFILE%` by default.

## Impact by adopting suggestion
* Users who have had `%HOMEDRIVE%%HOMEPATH%` different to `%USERPROFILE%` and put ".ZAP_JVM.properties" in only `%HOMEDRIVE%%HOMEPATH%` will get an error same as #3211

thc202:
Change zap.bat to use %USERPROFILE% to check for the ZAP_JVM.properties,
that's the (default) variable used to set the property "user.home" (used
for the path of (default) ZAP home and the properties file).

Fix #4004 - zap.bat should use %USERPROFILE% instead of
%HOMEDRIVE%%HOMEPATH%

### DIFF
--- a/src/zap.bat
+++ b/src/zap.bat
@@ -1,5 +1,5 @@
-if exist "%HOMEDRIVE%%HOMEPATH%\OWASP ZAP\.ZAP_JVM.properties" (
-	set /p jvmopts=< "%HOMEDRIVE%%HOMEPATH%\OWASP ZAP\.ZAP_JVM.properties"
+if exist "%USERPROFILE%\OWASP ZAP\.ZAP_JVM.properties" (
+	set /p jvmopts=< "%USERPROFILE%\OWASP ZAP\.ZAP_JVM.properties"
 ) else (
 	set jvmopts=-Xmx512m
 )


psiinon:
I'm good with this, but shouldnt we also check for the file in the old location, and if it exists copy t to the new one? Otherwise previous config settings will be lost.

thc202:
@kingthorin you mean the issue label?

@psiinon I didn't see the need to do further changes because in the two Windows versions (XP and 7) I tested the old location was the same as the new one, granted that's not a big sample. If you prefer I can try add the check/copy.

(In any case I'd like to mention this change in the release notes.)

kingthorin:
Yes, sorry for being vague I was in a rush.

thc202:
No problem. Even if minor I'd prefer that to show up in the release notes (just in case) but I'm OK with changing that to minor too (specially if the release notes mention the change in more detail).

kingthorin:
Ok, leave the labels as they are then, that makes sense.

psiinon:
Checked on Windows 10 and they are also the same, so I'm good with that :)

PR #4007, opened 2017-11-05
docker bare - copy policies folder 

omerlh:
the policies folder was missing, now I added it.

### DIFF
--- a/build/docker/Dockerfile-bare
+++ b/build/docker/Dockerfile-bare
@@ -26,6 +26,8 @@ RUN echo "http://dl-3.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
 #Change to the zap user so things get done as the right person (apart from copy)
 USER zap
 
+COPY policies /home/zap/.ZAP/policies/
+
 ENV PATH $JAVA_HOME/bin:/zap/:$PATH
 ENV ZAP_PATH /zap/zap.sh
 ENV HOME /home/zap/


build/docker/Dockerfile-bare
```diff
@@ -26,6 +26,8 @@ RUN echo "http://dl-3.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories
 #Change to the zap user so things get done as the right person (apart from copy)
 USER zap
 
+COPY policies /home/zap/.ZAP/policies/
```
thc202: There's an issue here (caught while testing for the push of the image), ZAP fails to start (with zap user) because of write permissions in the `.ZAP` dir, will open a PR shortly to address it (apply same fix as in other images).

omerlh: My mistake, tested only that it builds - not running. Maybe worth adding some CI for that...

thc202: Right, I also just tested that before merging. PR to address that opened, #4009.

Yes, not sure if with Travis CI.

omerlh: We could add some basic sanity test - `docker run` and then call the `JSON/core/view/version/?zapapiformat=JSON&formMethod=GET` to see Zap's actually working. I can try working on a PR if that make sense.

thc202: That sounds good to me, we could change Travis CI to do that (but that should run conditionally, building/running the images on every change will take a lot more time/resources than needed).

thc202:
Latest image includes the policy files.

PR #4011, opened 2017-11-06
Differentiate GET/POST array parameters

### LINKED ISSUE #2496: ZAP only report the first alert in array
Hello zap team,

Thank you very much for making this useful tool.

I identified another bug while I was verifying the fix of bug #2153.

Steps to reproduce:

1) Put this php script to http://localhost/test.php

[test.php.txt](https://github.com/zaproxy/zaproxy/files/273209/test.php.txt)

2) Submit this form through ZAP.

[posttest.html.txt](https://github.com/zaproxy/zaproxy/files/273211/posttest.html.txt)

3) Active Scan to history previous 2) request.(to get simple example, Active Scan Policy:Cross Site Scripting(Reflected)=Default, other=OFF)

I confirmed that the bug #2153 is fixed in zaproxy-w2016-05-16. (Thank you!!!)

But, I've found other bug.

In 'Alerts' tab, a XSS found in result. But only this case is detected.

[case1.txt](https://github.com/zaproxy/zaproxy/files/273224/case1.txt)

Actually, there is other XSS in scan history.These case is not in detect result.

[case2-5.txt](https://github.com/zaproxy/zaproxy/files/273226/case2-5.txt)

Is these results shuld be in 'Alerts' tab and 'HTML Report' like case1?

(Because an array value from HTML form has different mean by order in array.
ex. param[0]: name , param[1]: address, param[2]: phone number...
So, only param[0] is in scan result, causes miss the XSS in param[1],param[2]...)

Regards.

thc202:
Change VariantAbstractQuery to use indexed names for array parameters,
for example, if param[]=x&param[]=y use as names param[0] and param[1],
respectively.
Add tests to assert the expected behaviour of VariantAbstractQuery and
normalise some of its behaviour (that is, validate that valid parameters
are set, prevent modifications to internal parameter list, and do not
accumulate parameters on successive calls).

Fix #2496 - ZAP only report the first alert in array

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/VariantAbstractQuery.java
+++ b/src/org/parosproxy/paros/core/scanner/VariantAbstractQuery.java
@@ -30,24 +30,32 @@
 // ZAP: 2016/05/04 Changes to address issues related to ParameterParser
 // ZAP: 2016/05/26 Use non-null String for names and values of parameters, scanners might not handle null names/values well
 // ZAP: 2016/09/13 Issue 2863: Attack query string even if not originally specified
+// ZAP: 2017/11/06 Use indexed names for array parameters (Issue 2496).
 
 package org.parosproxy.paros.core.scanner;
 
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


thc202:
Will update the help pages to mention this change.

PR #4023, opened 2017-11-10
Prevent Commas in Port Number Strings

kingthorin:
Just tested the new multi proxy functionality out quickly, noticed that when you try to create a
duplicate entry the error message port number had commas.


thc202:
Thanks!

psiinon:
Good spot!

Issue #4022, opened 2017-11-10
Zap scripting question


ChrisFernandez:
https://security.stackexchange.com/questions/173162/zap-proxy-script-to-modify-headers-automatically
Basically I am looking for a zap script to modify response codes in the headers before it gets to my browser, so I do not have to edit manually after each try.
Thanks

kingthorin:
Well you could use the replaced extension (check the market place) or you could use an HttpSender script.

https://github.com/zaproxy/community-scripts/tree/master/httpsender

For future reference the ZAP User Group https://groups.google.com/group/zaproxy-users and Developer Group https://groups.google.com/group/zaproxy-develop are much better places for questions.

ChrisFernandez:
Thanks <3 I will dig into that example and join the groups.

thc202:
For help about the add-on:
https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsReplacerReplacer

Although not as flexible in this case it's easier/faster to use a Replacer rule than a script.

kingthorin:
Sorry that earlier response should have said "Replacer extension"

https://github.com/zaproxy/zap-extensions/wiki/HelpAddonsReplacerReplacer

ChrisFernandez:
Thank you! I'll try it tonight and let you guys know and close the ticket.

kingthorin:
If you're set on using a script you could build an HttpSender script that does something like:
```javascript
function sendingRequest(msg, initiator, helper) {
	// Nothing to do here
}

function responseReceived(msg, initiator, helper) {
     msg.setResponseHeader(msg.getResponseHeader().toString().replace('404','200'))
}
```

thc202:
I'd suggest using a Proxy script (or check the `initiator` to just apply to the proxy?), otherwise it might affect the results of the active scanners (or other tools).

PR #4026, opened 2017-11-10
Allow to set the status code and reason

thc202:
Change HttpResponseHeader to allow to set the status code and reason
phrase.

Makes use cases like #4022 more straightforward.


thc202:
Forgot to update an assert value in one of the tests.

thc202:
Test passing.

PR #4028, opened 2017-11-10
Warn when check for updates fails

thc202:
Change ExtensionAutoUpdate to notify the callback even if failed to
obtain the latest information (e.g. because of I/O errors or other non
insecure related errors), to have the manage add-ons dialogue show a
warning dialogue saying that the check for updates failed. Also, change
to log the (last) exception as warning (instead of debug) to show/log
the cause of the issue by default.
Add JavaDoc to CheckForUpdateCallback.


thc202:
(The warning was being shown in older versions than 2.4.0.)

PR #4040, opened 2017-11-14
Add Scroll Pane for Multiproxy

kingthorin:
* Add a scroll pane for the proxy options dialog.
* Trimmed html content from `options.proxy.local.label.browser` resource value, to facilitate linewrap.
* OptionsCallbackPanel removed unused code (and reference to the above resource key) [As a Clean Coders contribution ;) ]


src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
```diff
@@ -38,6 +39,8 @@
 public class OptionsProxiesPanel extends AbstractParamPanel {
 
     private static final long serialVersionUID = 1L;
+    
+    private static JScrollPane proxiesScrollPane;
```
thc202: The static can be removed.

src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
```diff
@@ -57,10 +60,19 @@ public OptionsProxiesPanel(ExtensionProxies extension) {
...
         this.add(getMainProxyPanel(), LayoutHelper.getGBC(0, 0, 1, 0.0));
         this.add(getSecurityProtocolsPanel(), LayoutHelper.getGBC(0, 1, 1, 0.0));
-        this.add(getProxiesOptionsPanel(), LayoutHelper.getGBC(0, 2, 1, 1.0, 1.0));
+        this.add(getScrollPane(), LayoutHelper.getGBC(0, 2, 1, 1.0, 1.0));
+    }
+    
+    private JScrollPane getScrollPane() {
+        if (proxiesScrollPane == null) {
+            proxiesScrollPane = new JScrollPane();
+            proxiesScrollPane.setViewportView(getProxiesOptionsPanel());
```
thc202: Although it requires some more changes I'd suggest adding to the whole panel instead of just the table (as it would be always scrollable no matter the size of the dialogue).

src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
```diff
@@ -57,10 +60,19 @@ public OptionsProxiesPanel(ExtensionProxies extension) {
 
     private void initialize() {
         this.setName(Constant.messages.getString("proxies.options.title"));
+        this.add(getScrollPane());
```
thc202: This can be removed? (Being added later.)

src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
kingthorin: Whoops yup, that was from an earlier iteration. Sorry

kingthorin:
Latest push is broken, causing an NPE on startup. I pushed a broken version as I hope to work on it at lunch.

kingthorin:
Should be good now. Thanks @thc202 for your assistance earlier.

src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
```diff
@@ -58,9 +63,26 @@ public OptionsProxiesPanel(ExtensionProxies extension) {
...
         this.setLayout(new GridBagLayout());
-        this.add(getMainProxyPanel(), LayoutHelper.getGBC(0, 0, 1, 0.0));
-        this.add(getSecurityProtocolsPanel(), LayoutHelper.getGBC(0, 1, 1, 0.0));
-        this.add(getProxiesOptionsPanel(), LayoutHelper.getGBC(0, 2, 1, 1.0, 1.0));
+        this.add(getScrollPane(), LayoutHelper.getGBC(0, 0, 1, 1.0, 1.0));
+    }
+    
+    private JScrollPane getScrollPane() {
+        if (proxiesScrollPane == null) {
+            proxiesScrollPane = new JScrollPane();
```
thc202: This could use an empty border, `proxiesScrollPane.setBorder(BorderFactory.createEmptyBorder());`.

src/org/zaproxy/zap/extension/proxies/OptionsProxiesPanel.java
```diff
@@ -58,9 +63,26 @@ public OptionsProxiesPanel(ExtensionProxies extension) {
...
+            proxiesScrollPane.setViewportView(getScrollPanel());
+        }
+        return proxiesScrollPane;
+    }
+    
+    private JPanel getScrollPanel() {
+        if (scrollPanel == null) {
+            scrollPanel = new JPanel();
+            scrollPanel.setLayout(new VerticalLayout());
```
thc202: The layout could be passed in the constructor.

thc202:
Thank you!

PR #4047, opened 2017-11-16
Correct Sites scope filter rendering in macOS

### LINKED ISSUE #3917: Show only URLS in scope - Not working for Sites in macOS
It appears that this issue was brought up before - #3675, and has resurfaced. 

I'm running Zap 2.6.0  primarily on Windows 10 but I the issue also happens with Kali Linux.

zap_showurlsscope

thc202:
Change SiteMapPanel to force the query of the row height from the cell
renderer when using macOS L&F, to ensure the filtered nodes are not
shown (with empty space).

Fix #3917 - Show only URLS in scope - Not working for Sites in macOS

Thanks to @psiinon :)


thc202:
@psiinon or @ScatteredThoughts, it would be great if you could test in macOS (the L&F check seems correct but it's not tested).

ScatteredThoughts:
@thc202 Tested on my work Mac and it looks good!

thc202:
@ScatteredThoughts thank you!

PR #4048, opened 2017-11-16
Log correct address/port when stopping alt proxy

### LINKED ISSUE #4039: MultiProxy Logging Discrepancy
Just noticed a minor logging issue with the new multiproxy functionality:
```
7730 [ZAP-BootstrapGUI] INFO org.zaproxy.zap.extension.proxies.ExtensionProxies  - Starting alt proxy server: localhost:8081
.....
23174 [Thread-9] INFO org.zaproxy.zap.extension.proxies.ExtensionProxies  - Stopping alt proxy server: localhost:8080
```

First there's no logging of the main proxy starting. Then at shutdown there main proxy is referred to as an alt and there's no logging for the actual alt.

multiproxy

thc202:
Change ExtensionProxies to log the key of the proxy being shutdown,
instead of using the ParamProxy which does not contain the correct data.

Fix #4039 - MultiProxy Logging Discrepancy

### DIFF
--- a/src/org/zaproxy/zap/extension/proxies/ExtensionProxies.java
+++ b/src/org/zaproxy/zap/extension/proxies/ExtensionProxies.java
@@ -88,7 +88,7 @@ public class ExtensionProxies extends ExtensionAdaptor implements OptionsChanged
         super.stop();
         // Stop all of the running servers
         for (Entry<String, ProxyServer> entry : proxyServers.entrySet()) {
-            stopProxyServer(entry.getValue());
+            stopProxyServer(entry.getKey(), entry.getValue());
         }
         proxyServers.clear();
     }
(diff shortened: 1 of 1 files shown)


thc202:
They were already correct (or it's missing something?).

thc202:
OK, I see, not sure, do we need to? (We didn't before, only when daemon.)

kingthorin:
I didn’t see them when I encountered and reported the issue, will dbl check.

thc202:
As you were using the GUI it's the expected behaviour, we just log the start of the main proxy when in daemon and we don't log the stop in any of the modes (GUI, daemon). I can change that if you think it's better to do it.

kingthorin:
Nope that’s makes sense

PR #4050, opened 2017-11-17
Use main connection settings in additional proxies

thc202:
Change ExtensionProxies to set the connection settings (for example,
connection timeout, outgoing proxy) into the additional proxies created,
otherwise it would use default values.


thc202:
(I guess in time we could allow to tweak that per proxy.)

PR #4055, opened 2017-11-20
New splash screen

psiinon:
The progress bar should be blue, but that looks non trivial :/
The background image also doesnt scale as well as it should, eg if you
set the font size to 20 or more. NHot sure why this is - any
suggestions?


thc202:
Setting a `BasicProgressBarUI` to the progress bar allows to override the colour (but there's no colour glow/gradient).

src/org/zaproxy/zap/view/BackgroundImagePanel.java
```diff
@@ -75,7 +87,9 @@ public final void setBackgroundImage(URL imageUrl) {
     protected void paintComponent(Graphics g) {
         if (img != null) {
             setOpaque(false);
-            g.drawImage(img, 0, 0, DisplayUtils.getScaledSize(img.getWidth()), DisplayUtils.getScaledSize(img.getHeight()), null);
+            g.drawImage(img, 0, 0,
```
thc202: Setting rendering hints seems to help with the scaling artifacts:
```Java
Map<RenderingHints.Key, Object> hints = new HashMap<RenderingHints.Key, Object>();
hints.put(RenderingHints.KEY_INTERPOLATION, RenderingHints.VALUE_INTERPOLATION_BICUBIC);
hints.put(RenderingHints.KEY_RENDERING, RenderingHints.VALUE_RENDER_QUALITY);
hints.put(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
((Graphics2D) g).addRenderingHints(hints);
```
(From `org.fife.ui.rtextarea.BufferedImageBackgroundPainterStrategy`.)

psiinon:
Thanks @thc202 - updated to fix the scaling and set the color. It now looks like this: https://imgur.com/a/sTV51

psiinon:
Added the OTF license - we should also update the credits

PR #4056, opened 2017-11-20
Use realm as domain for NTLM authentication

thc202:
Change HttpAuthenticationMethodType to use the realm as domain for NTLM
authentication, instead of using the hostname of the server. If an
invalid domain is provided it might lead to an authentication failure
(which would most likely happen since the hostname is always set).


thc202:
This fixes the issue reported in the mailing list (where a VM was kindly provided):
https://groups.google.com/d/msg/zaproxy-users/sOjj98kHbpQ/QgwtmA9bAgAJ

PR #4061, opened 2017-11-21
Update NTLM engine implementation

thc202:
Update to latest version from upstream project HttpComponents Client
(branch 4.6.x, commit f8a26dffded21be28e1c1ae73f12bb0f42ad29f2), to
improve interoperability with other servers.


thc202:
It's probably better to diff against the original (I can post a diff if it helps).

That's most likely caused by the client (browser) trying to authenticate (which then confuses ZAP). If you send the requests directly from ZAP does that still happen?

kingthorin:
When I ran the spider it worked as expected without those sequence messages. 

Don’t worry about the diff, I’m happy.

PR #4062, opened 2017-11-21
Honour proxy exclusions in additional proxies

thc202:
Change ExtensionProxies to have the additional proxies honour proxy
exclusions (session and global).
Change Messages.properties to update label and API descriptions to refer
that the exclusions applies to the local proxies (not just one/main
local proxy).


thc202:
Yeah, that makes sense. Switched.

PR #4063, opened 2017-11-21
Validate that -cmd and -daemon are not both set

thc202:
Change CommandLine to validate that -cmd and -daemon are not set at the
same time, the arguments are mutually exclusive (the former stops ZAP
after executing the command line arguments and the latter keeps ZAP
running indefinitely).
Add test to assert the new behaviour.


thc202:
(Per comments in the users mailing list.)

PR #4066, opened 2017-11-22
Improve error message in script API

thc202:
Change an error message in ScriptAPI to mention the actual error/issue
when the script to be executed does not have the correct type.

### DIFF
--- a/src/org/zaproxy/zap/extension/script/ScriptAPI.java
+++ b/src/org/zaproxy/zap/extension/script/ScriptAPI.java
@@ -195,7 +195,10 @@ public class ScriptAPI extends ApiImplementor {
 				throw new ApiException(ApiException.Type.DOES_NOT_EXIST, ACTION_PARAM_SCRIPT_NAME);
 			}
 			if (!script.getType().getName().equals(ExtensionScript.TYPE_STANDALONE)) {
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/script/ScriptAPI.java
```diff
@@ -195,7 +195,10 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 				throw new ApiException(ApiException.Type.DOES_NOT_EXIST, ACTION_PARAM_SCRIPT_NAME);
 			}
 			if (!script.getType().getName().equals(ExtensionScript.TYPE_STANDALONE)) {
-				throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, ACTION_PARAM_SCRIPT_NAME);
+				throw new ApiException(
+						ApiException.Type.ILLEGAL_PARAMETER,
+						"Parameter " + ACTION_PARAM_SCRIPT_NAME + " does not select a " + ExtensionScript.TYPE_STANDALONE
```
kingthorin: Would “is not a” or “does not match a” make more sense.

thc202: Yeah, changed.

PR #4075, opened 2017-11-23
Updated mac icons

psiinon:


psiinon:
No, it only seems to be viewable on the mac.
But I have just updated it to copy the icns file to the build dir

PR #4077, opened 2017-11-23
Allow explicitly add OverrideMessageProxyListener

thc202:
Change ExtensionHook to allow to add the OverrideMessageProxyListener,
without getting the list of the listeners (which should not have been
public).


thc202:
(Per comment in the dev mailing list.)

PR #4083, opened 2017-11-27
Replaced the 'TODO add version's

psiinon:


psiinon:
Note that this is in the new release-2.7.0 branch, which was a copy of develop but with the versions.txt file updated

thc202:
There are some entries that were not replaced (using different case, `Add version`).

psiinon:
Good spot - think I've got those now.

PR #4113, opened 2017-12-04
Validate risk ID in core API views

thc202:
Change CoreAPI to reject invalid risk IDs when handling the views
"alerts" and "numberOfAlerts".


thc202:
(Per comments in IRC channel.)

PR #4141, opened 2017-12-11
Allow to search for a String in AbstractParamContainerPanels

### LINKED ISSUE #3358: Enhancement: Options Panel Filter
This idea came from an email thread:

Quote @kingthorin:
>I've also been thinking in both the context configuration tree and more specifically the options panel it'd be awesome if we had a filter. For example in the options panel type local or proxy and the tree is filtered to only show "Local Proxy". Initially this only needs to be based on tree entries, down the road I can see users wanting to be able to filter/search specific options themselves.

Quote @psiinon:
>The options filter/search could be tricky to implement, but would be really useful to users.
If we have to make a set of core changes to enable it, then maybe we should create a (set of) class(es) which represents an option, in a similar way to how we represent API end points.
This might allow us make config screens much easier to produce - most of them could be generated on the fly automatically which a small number of custom ones for more complex interfaces.
It would also allow us to expose the config values via the UI, ie the ones that can be specified via the '-config' cmdline option.

denniskniep:
…(OptionsDialog, PolicyDialog, SessionDialog) and also highlights the findings.

Fix #3358

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/AbstractParamContainerPanel.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/AbstractParamContainerPanel.java
@@ -25,6 +25,7 @@
 // ZAP: 2016/10/27 Explicitly show other panel when the selected panel is removed.
 // ZAP: 2017/06/23 Ensure panel with validation errors is visible.
 // ZAP: 2017/09/03 Cope with Java 9 change to TreeNode.children().
+// ZAP: 2017/12/09 Issue 3358: Added SearchPanel.
 // ZAP: 2018/01/08 Allow to expand the node of a param panel.
 // ZAP: 2018/03/26 Ensure node of selected panel is visible.
 // ZAP: 2018/04/12 Allow to check if a param panel is selected.
@@ -41,8 +42,11 @@ import java.awt.Insets;
 import java.awt.Rectangle;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.awt.event.KeyAdapter;
+import java.awt.event.KeyEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.Hashtable;
@@ -51,6 +55,7 @@ import javax.swing.JButton;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JSplitPane;
+import javax.swing.JToolBar;
 import javax.swing.JTree;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeCellRenderer;
@@ -63,6 +68,12 @@ import org.zaproxy.zap.extension.help.ExtensionHelp;
 import org.zaproxy.zap.utils.DisplayUtils;
 import org.zaproxy.zap.utils.FontUtils;
 import org.zaproxy.zap.utils.ZapTextField;
+import org.zaproxy.zap.view.panelsearch.ComponentHighlighterProvider;
+import org.zaproxy.zap.view.panelsearch.ComponentSearchProvider;
+import org.zaproxy.zap.view.panelsearch.SearchAndHighlight;
+import org.zaproxy.zap.view.panelsearch.items.AbstractComponentSearch;
+import org.zaproxy.zap.view.panelsearch.items.TreeNodeElement;
+import org.zaproxy.zap.view.panelsearch.items.TreeUtils;
 
 public class AbstractParamContainerPanel extends JSplitPane {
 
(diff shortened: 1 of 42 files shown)


denniskniep:
Here is a first impression how the changes would improve the Options-Dialog:
final_searchoptions

denniskniep:
Furthermore the search field is available at the Scan Policy Manager and the Session Properties:
final_activescanpolicydialog

denniskniep:
During development I encountered some problems:

1. The yellow border which indicates that there is a finding inside of a table or combobox can sometimes not reverted. I cache the original border, set the border to yellow and try to restore it to the original border. But that does not work sometimes. Then the yellow border is still there. 
I implemented a workaround which sets the border to "empty". I think that is no problem until we insert a table with a custom border.
see [here](https://github.com/zaproxy/zaproxy/pull/4141/files#diff-9f3bb93445f6f78011275ff9b9b4b667R98)

2. I have a similar problem with `TreeCellRenderer`, `TableCellRenderer` and `ListCellRenderer`. I can not determine the original value for each item. So I can not use the original value as default, if its not highlighted. But same here. Currently it works, if I set the Background to null. There is currently no control with a CustomRenderer that breaks this workaround.
see 
[TreeCellRenderer](https://github.com/zaproxy/zaproxy/pull/4141/files#diff-ed190859fa868dc6cfe7e809ebcf8a22R92)
[TableCellRenderer](https://github.com/zaproxy/zaproxy/pull/4141/files#diff-34d2bb3d7c12ceb3d2eb2feb82db17f3R123)
[TreeCellRenderer](https://github.com/zaproxy/zaproxy/pull/4141/files#diff-cdc2a3dff638f878bdedec186f5fee50R111)

Has anyone an Idea how to handle this? Or can we live with this two workarounds?

kingthorin:
@denniskniep are you working on the build failure issues?

denniskniep:
@kingthorin yeah, now passing.

src/org/zaproxy/zap/view/panelsearch/FoundComponent.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public void addParent(Object parent){
+        parents.add(parent);
+    }
+
+    public List<Object> getParents() {
+        return new ArrayList<>(parents);
+    }
+
+    public <T> T getParentAtCasted(int index) {
```
kingthorin: Should this be `asCasted`?

src/org/zaproxy/zap/view/panelsearch/HighlighterUtils.java
```diff
@@ -0,0 +1,102 @@
...
+import java.awt.Color;
+
+public final class HighlighterUtils {
+
+    private static final String OPAQUE = "Opaque";
+    private static final String BACKGROUND = "Background";
+    private static final String BORDER = "Border";
+    private static final String TITLE = "title";
+
+    public static final Color DefaultHighlightColor = new Color(255, 204, 0);
```
kingthorin: Constants should be full caps.

denniskniep:
@kingthorin Thanks for your feedback. Fixed it now.

psiinon:
@denniskniep sorry, it has conflicts again, but this time we should aim to get it merged asap.
@thc202 any other feedback before this can be merged?

thc202:
This should be extensible, add-ons should be able to override/extend the search functionality (e.g. not traverse the component tree each time, the panels are pretty much static; add support for other UI components).

There's a component leak, which prevents the add-ons from being unloaded when there's a highlighted search. There are other implementation details that need polishing IMO (e-.g. `AbstractParamContainerPanel` exposing its internal state, it should be able to search itself).

denniskniep:
> This should be extensible, add-ons should be able to override/extend the search functionality 

Now an implementation of the `AbstractParamPanel` can implement `ComponentSearchProvider` and/or `ComponentHighlighterProvider` and inject his own Components into the Search and/or Highlight process. It could add generic support for an other UI component or add a component that searches itself.

> There's a component leak, which prevents the add-ons from being unloaded when there's a highlighted search.

Thank you, good catch. Everytime an `AbstractParamPanel` is removed (that's happening when the Extension is unloaded) the highlights are completely removed.

> AbstractParamContainerPanel exposing its internal state

I solved this with an inner class. It is no longer exposing the internal state.


@thc202 Are the fixes okay for you and what you expected? 

Finally I can squash the changes.

psiinon:
This pull request **introduces 1 alert** when merging 528c537e0a1f82a566638bf36b65f2526a86cd63 into 4c2e67c5be4c41210a4976e898fb54f01e660e4c - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-c3ca939f5cbad5731fca55e12ed5dbf72dfcb862)

**new alerts:**

* 1 for Dereferenced variable may be null

---

*Comment posted by [LGTM.com](https://lgtm.com)*

thc202:

thc202:
Fixed the conflicts.

thc202:
@denniskniep thanks!

thc202:
For the record, some of the classes/state (e.g. helper/utility classes) while public/mutable should not be considered part of the public API just internal implementation details.

Issue #3879, opened 2017-09-11
Auto Set Active Imported Certificate


davewichers:
I had previously requested (in issue #3163):

In ZAP Options, when you import a new PKCS#12 cert (and probably works the same for the other types as well), after successful import, it takes you back to the KeyStore tab (which is good). But then to make that the Active cert the user has to:

a) Click on the cert in the left column,
b) Click on the username in the right column,
c) Click on the Set Active button

Steps a) and b) are now automatic, but the user still has to do step c).

Is there any reason why we don't automatically make the imported certificate active? If not, can we make step c) above happen automatically too?

kingthorin:
To be clear I'm not against this idea, I've only worked with client certs a little bit and even then usually single user single cert.

If you had multiple certs I guess there's a scenario in which you'd import one (because you received it or whatever), but not immediately want to make it "active" for an existing user user.... I suppose it's fair to assume that if you're importing it you'd want to use it. But, it is an assumption. I can also see that a user might import and cert and then realize they had the wrong user selected and have to redo both.... I dunno... food for thought.

davewichers:
ZAP doesn't remember certs between uses of ZAP, so you have to import the
cert EVERY time you use it (if you need one). At a minimum, we should
automatically set active the cert if its the 1st one imported. But I really
think, by default, we should set the latest one imported as active.  Either
of these would be an improvement over what happens now and I think would
cover 99+% of the usage of certs. (The 1st (and usually only) cert
imported).

kingthorin:
I have no ability to test with pkcs#11 to make similar mods, however I was able to test with pkcs#12. So I propose we do something like the following. Basically, when a cert is added if it's the only cert then it is also activated. 

I was torn as to activating the latest cert, but it's definitely reasonable to activate the "only" cert. I can easily see instances where a user might import more than one cert at a time or import secondary/tertiary certs while still using the original one (with plans to use the others in the future). However, I wasn't able to convince myself of making the latest one the active one.... either way it's better than it was and saves a click or two in getting going which is nice.

```diff
diff --git a/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java b/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
index 9834c37..38c7108 100644
--- a/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
[...]

PR #4146, opened 2017-12-12
Issue 3879 - Auto Set Active Imported Certificate

kingthorin:
Automatically activate the first client cert imported.

Fixes zaproxy/zaproxy#3879

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
@@ -31,6 +31,7 @@
 // ZAP: 2017/08/16 Tidy up usage of CertificateView.
 // ZAP: 2017/08/16 Show error message if failed to activate the certificate.
 // ZAP: 2017/08/17 Reduce code duplication when showing cert/keystore errors
+// ZAP: 2017/12/12 Use first alias by default (Issue 3879).
 
 package org.parosproxy.paros.extension.option;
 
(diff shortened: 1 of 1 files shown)


kingthorin:
Wow you were fast on that. I switch tabs and came back and it was labeled 👍

thc202:
Should I push here if I get the PKCS#11 part working (or add diff to issue)?

kingthorin:
I'm fine with you updating this PR. Whichever your prefer. If the PKCS#11 part is burdensome we can always tackle it separate/later.

thc202:
As you mentioned in the issue, it was the same code changes. I moved the code to existing method (which was already being called by both implementations). I also changed the logic to check if there's just one cert instead of just one keystore, as for the latter it would automatically active a cert even if there was more than one (from my understanding of the comments in the issue that was not intended?).

Feel free to fixup, drop, or change the commit as you see fit.

kingthorin:
Looks good to me. Thanks! I'll test it a bit more tomorrow to see the multiple key handling. Maybe I didn't complete adding the second key, or canceled out of the dialog or something...

kingthorin:
Tested this a bit this morning, the PKCS#12 stuff seems to be working as expected. Good with me squashing it, and I guess I'll ask psiinon to review it?

thc202:
Sounds good.

kingthorin:
@psiinon since both @thc202 and I worked on this could you review and merge?

psiinon:
of course :)

Issue #4171, opened 2017-12-20
Spider/Active Scan Hangs scanning https://observatory.mozilla.org redirect to non-existent port


claudijd:
I was originally starting with running zap-full-scan.py scans against a number of hosts, one of which was https://observatory.mozilla.org and the active scan was seeming running indefinitely.  I moved to the UI to try and get more information on this and I worked with @thc202 a fair bit in #websectools trying to debug this.  We ultimately didn't find a resolution, but putting all my debug information/intel here in the hopes it eventually finds a root cause and we can work past this.

Reproduction Steps

1. Open up the ZAP UI (I was running 2.7.0 release on Mac OS Sierra)
2. URL to Attack: https://observatory.mozilla.org
3. You will then witness it move through and complete the Spider quite quickly and move on to the Active Scan
4. In the active scan it will start with the Path Traversal plugin, make up to about ~50 requests and then the request counter will stop increasing.

Troubleshooting/Debug Context

1. We originally thought this was some IPS/WAF solution, but I have verified with the service owner this is not the case and I verified with other MiTM proxies and this behavior was not observed there.  We believe we ruled this out.

2. We ran ZAP and got stack traces from all the threads.  Based on this @thc202 concluded that the scanner was waiting, not timing out

https://irccloud.mozilla.com/pastebin/0MYRL5jb/

3.  @thc202 also just realized that http://observatory.mozilla.org/images/favicons which redirects to http://observatory.mozilla.org:58443/images/favicons/ and believes this is the source of the issue.

claudijd:
Ok, after some looking around it appears the issue is the redirect to a non-active port.

Things I think we could do to make this work better...

1.  Restrict the scope to the service being scanned (ie. if we ask it to scan :443, it will not observe redirects and add them to the scan scope automatigally, maybe this is a prompt/alert the user somehow?)

2.) Go fix this redirect thing in Observatory, which I have raised with @april (it seems this is the internal listening port)

april:
At the very least, ZAP should definitely not hang attempting to get the resource any longer than however long it or the webserver takes to timeout.

thc202:
Closing, this should be addressed with #4173.

PR #4193, opened 2017-12-29
Refactor redirection validation in active scans

thc202:
Change AbtractPlugin and Analyser to use HostProcess to validate the
redirections, to reduce code duplication.
Change HostProcess to provide a HttpRedirectionValidator and a
HttpRequestConfig for validation of redirects.
Change HttpRedirectionValidator to have notifyMessageReceived as a
default method, the method is provided for convenience and not all
implementation require it, allowing the interface to be functional.
Change DefaultHttpRedirectionValidator to not implement the now default
method.


thc202:
Yes, related to #4171 (although it does not move it further).

kingthorin:
Ok, thanks :)

PR #4199, opened 2018-01-01
Address exception in Scan Progress dialogue

thc202:
Fix MissingResourceException(s) in Scan Progress dialogue after
uninstalling an add-on with scanners used in a previous active scan.

Change PluginStats to cache the name of the scanner.
Change ScanProgressItem to use the name from PluginStats (instead of
Plugin, the latter might no longer be available). Also, change to use
PluginStats were possible instead of going through the HostProcess.
Change HostProcess to initialise the the PluginStats with the name of
the Plugin.


psiinon:

thc202:
Conflict addressed.

PR #4201, opened 2018-01-02
Alert Tab UI Tweaks

kingthorin:
AlertViewPanel.java > Disable horizontal scrolling, enable line wrap for text areas.

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
+++ b/src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
@@ -42,6 +42,7 @@ import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JSpinner;
 import javax.swing.JTextField;
+import javax.swing.ScrollPaneConstants;
 import javax.swing.border.TitledBorder;
 
 import org.apache.log4j.Logger;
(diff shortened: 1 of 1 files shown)


kingthorin:
@psiinon look good?

psiinon:
yep :)

src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -155,6 +156,7 @@ private void initialize() {
 	private JScrollPane getAlertPane() {
 		if (alertPane == null) {
 			alertPane = new JScrollPane();
+			alertPane.setHorizontalScrollBarPolicy(ScrollPaneConstants.HORIZONTAL_SCROLLBAR_NEVER);
```
thc202: This is an improvement but there's still an issue, the text areas are allowed to expand but they don't shrink, so if you resize the panel "too much" you might no longer see some of the data after restoring it to the previous size. I guess each text area needs its own scroll pane(?).

src/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -274,6 +276,7 @@ public void actionPerformed(ActionEvent e) {
 			alertUrl.setLineWrap(true);
 			
 			alertDescription = createZapTextArea();
+			alertDescription.setLineWrap(true);
```
thc202: It seems this is already being done in `createZapTextArea`.

kingthorin:
Will look into further changes

kingthorin:
This was discussed a bit today via IRC. For the record: https://mozilla.logbot.info/websectools/20180222

PR #4208, opened 2018-01-03
Add NameValuePair Types

kingthorin:
Extend the list to include 4 new types to cover newer parsing of multipart/form-data, builds on: #3814

I started the numbering at 33 to provide slack space to the bitshifting in ScannerParam.java (which currently maxes at 8, but could easily have two additional entries).

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/NameValuePair.java
+++ b/src/org/parosproxy/paros/core/scanner/NameValuePair.java
@@ -22,6 +22,7 @@
 // ZAP: 2014/01/06 Issue 965: Support 'single page' apps and 'non standard' parameter separators
 // ZAP: 2014/02/08 Used the same constants used in ScanParam Target settings
 // ZAP: 2016/02/22 Add hashCode, equals and toString methods. Remove redundant instance variable initialisations.
+// ZAP: 2018/01/03 Added type constants for revised multipart/form-data handling
 package org.parosproxy.paros.core.scanner;
 
 public class NameValuePair {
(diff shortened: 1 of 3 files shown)


src/org/parosproxy/paros/core/scanner/NameValuePair.java
```diff
@@ -32,6 +32,10 @@
     public static final int TYPE_COOKIE = ScannerParam.TARGET_COOKIE;
     public static final int TYPE_HEADER = ScannerParam.TARGET_HTTPHEADERS;
     public static final int TYPE_POST_DATA = ScannerParam.TARGET_POSTDATA;
+    public static final int TYPE_MULTIPART_DATA_PARAM = 33;
```
thc202: Worth adding a since JavaDoc tag, also, mention to which part of the content is the type referring to? (Same comment for the other types.)

kingthorin: Will do

kingthorin:
Done & yes (done)

kingthorin:
I just noticed that there is a [NameValuePairUnitTest](https://github.com/zaproxy/zaproxy/blob/f2fb322ad21f0a5072accb0663426df18c2106c3/test/org/parosproxy/paros/core/scanner/NameValuePairUnitTest.java), but there doesn't seem to be anything that specifically tests the type assignment. Glad to hear thoughts/ideas on what I could/should add.

thc202:
I don't think it needs additional tests, the type is being tested/verified when the `NameValuePair` is created (it can't be changed after that).

src/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -91,7 +92,11 @@ public void setMessage(HttpMessage msg) {
 				String value = part.replaceAll(boundary + partHeaderLine + HttpHeader.CRLF + HttpHeader.CRLF, "");
 				value = value.replaceAll(HttpHeader.CRLF + "(" + Pattern.quote(getBoundary(contentType)) + "--"
 						+ HttpHeader.CRLF + ")?$", ""); // Strip final boundary
-				extractedParameters.add(new NameValuePair(NameValuePair.TYPE_POST_DATA, name, value, position));
+				if (isFileParam) {
+					extractedParameters.add(new NameValuePair(NameValuePair.TYPE_MULTIPART_DATA_FILE_PARAM, name, value, position));
+				} else {
+					extractedParameters.add(new NameValuePair(NameValuePair.TYPE_POST_DATA, name, value, position));
```
thc202: Should be `TYPE_MULTIPART_DATA_PARAM`? We should assert the types in `VariantMultipartFormParametersUnitTest.shouldExtractParametersFromAllParts`.
In `shouldNotAllowToModifyReturnedParametersList` it could be changed to use one of the multipart constants (just for consistency).

src/org/parosproxy/paros/core/scanner/NameValuePair.java
```diff
@@ -32,6 +33,30 @@
     public static final int TYPE_COOKIE = ScannerParam.TARGET_COOKIE;
     public static final int TYPE_HEADER = ScannerParam.TARGET_HTTPHEADERS;
     public static final int TYPE_POST_DATA = ScannerParam.TARGET_POSTDATA;
+    /*
```
thc202: It's missing a `*` for JavaDoc (`/**`).

thc202:
Thanks!

PR #4212, opened 2018-01-03
Do not create filter directory

thc202:
Change Constant to no longer create the filter directory, not needed,
the filter functionality is no longer available. Also, deprecate the
FOLDER_FILTER constant.

Related to #4191 - Remove filter extension from core functionality


thc202:
Not break binary compatibility. Yes, some of the core classes are still using it (add-ons could be using it too, although unlikely). I'd prefer leave it for now and remove it along with the filter extension in a following version.

kingthorin:
Fair enough.

PR #4213, opened 2018-01-03
Exit immediately if not able to create home dir

thc202:
Change Constant to exit immediately if not able to create the home dir,
instead of continuing and attempting to copy home files, which would
fail and "hide" (with an exception stack trace) the actual error. Also,
add more checks (exists/canWrite) to provide a more appropriate error
message.


psiinon:

thc202:
Done.

psiinon:
Ta :)

PR #4218, opened 2018-01-04
Do not notify of session change on error

thc202:
Change Control to not notify the extensions if an error occurred while
changing (saving/opening) the session, it would only lead to more errors
as the extensions try to access the session.

### DIFF
--- a/src/org/parosproxy/paros/control/Control.java
+++ b/src/org/parosproxy/paros/control/Control.java
@@ -67,6 +67,7 @@
 // ZAP: 2017/03/16 Allow to initialise Control without starting the Local Proxy.
 // ZAP: 2017/06/07 Allow to persist the session properties (e.g. name, description).
 // ZAP: 2017/08/31 Use helper method I18N.getString(String, Object...).
+// ZAP: 2018/01/04 Do not notify extensions if failed to change the session.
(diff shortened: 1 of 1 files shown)


src/org/parosproxy/paros/control/Control.java
```diff
@@ -504,19 +505,31 @@ public void createAndOpenUntitledDb() throws ClassNotFoundException, Exception {
...
+
 		if (lastCallback != null) {
 			lastCallback.sessionOpened(file, e);
 			lastCallback = null;
 		}
 		
 	}
 
+	/**
+	 * Notifies the extensions that the session changed, iff the given exception is {@code null}.
```
kingthorin: Little typo on `iff`

thc202: That was on purpose for "if and only if" but not necessary?

kingthorin: Huh, I've never seen that before. I'm sure it's "technically" correct. I think it's probably clear enough as "if" but if not then it should probably actually say "if and only if".

thc202: OK, changed to "if".

Issue #1429, opened 2015-06-04
Option to add custom ZAP header


zapbot:
```
Option to add a custom (ie user defined) header to flag ZAP requests.
This will allow the user to identify which requests came from ZAP, and do things like
configure WAF rules to permit them.
```

Original issue reported on code.google.com by `psiinon` on 2014-12-10 13:10:26

kingthorin:
There are various options to accomplish this. The Plugin_ID header, scripts, etc.

Can we close this one?

thc202:
How about adding a (template) HTTP Sender script that does that (and close the issue with it)?

thc202:
(We could also update the help pages to mention it.)

PR #4239, opened 2018-01-09
Issue 1429 - Add script template for ZAP Header

kingthorin:
AddZapHeader.js > New template which adds a ZAP header to all requests. eg: `X-ZAP-Initiator: 3`

Fixes zaproxy/zaproxy#1429


thc202:
Thank you!

PR #4252, opened 2018-01-11
Dont restrict callback urls to the specified addresses

psiinon:
Callback urls by their very nature are on the target domain and should be accessible from it.
Luckily all of the current add-ons have changed to use the ExtensionCallback.
Have also removed the debuging of the full API response - this is very noisy and is probably only needed when developing API calls, at which point it can be added back.


psiinon:
Updated to fix the failing tests.

PR #4315, opened 2018-01-27
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


src/lang/vulnerabilities_fil_PH.xml
```diff
@@ -158,7 +158,7 @@ Many file inclusion problems occur because the programmer assumed that certain i
 	<reference>http://cwe.mitre.org/data/definitions/98.html</reference>
 </vuln_item_wasc_5>
 
-<vuln_items>wasc_6</vuln_items>
+<vuln_items>ang wasc_6</vuln_items>
```
kingthorin: Won't this break things?

thc202: Nice catch! Fixed in Crowdin.

Issue #2528, opened 2016-06-02
Feature request: Passive scanning during fuzzing


kurobeats:
Just a couple: 
1. Option to apply passive scanner rules to fuzzing activities.
2. Saving all session files into a zip archive (or any other format like 7z) to keep files consolidated and decrease storage size.

kingthorin:
Note: #2 is a duplicate of zaproxy/zaproxy#373

kingthorin:
Changing title, removing second part since it's a duplicate.

PR #4318, opened 2018-01-28
Enhancement - Allow addons to "opt-in" for passive scanning

kingthorin:
Changes to facilitate addons "opting-in" their HistoryTypes for passive scanning. 
Leverages that functionality to fix zaproxy/zaproxy#2528

Also see `HistoryReference`, eg: `public static final int TYPE_FUZZER =
8;`


kingthorin:
Is it worth including `addApplicableHistoryTypes(Set<Integer>)` and a plural remove method as well? (Leveraging addAll/removeAll)

thc202:
Could you describe how the add-ons are expected to use this functionality?

kingthorin:
Sure I pictured the fuzz extension having an options panel with a "Passive Scan Requests/Responses" checkbox. When checked it'd `PluginPassiveScanner.addApplicableHistoryType(HistoryReference.TYPE_FUZZER);` when uncheck it'd `PluginPassiveScanner.removeApplicableHistoryType(HistoryReference.TYPE_FUZZER);`.

Or something along those lines.

Would it be better if the actions were in the Extension class? Or have I over simplified things too much, will it not work as I'm envisioning?

thc202:
OK, I thought this was going to work differently (based on the current implementation), some of the methods need a static modifier then.

I think it's OK this way, but ideally #2528 would be enabled per scan instead of applying to all of them (although that would require other changes).

I could also see something like this as a core option, and allow to set it per scanner (similar to Initiators tab of Replacer add-on).

kingthorin:
Ok, so it seems like there are three ideas there. Shall we keep this or abandon it for one of the later two? 

As-is I've only got 15 or 20mins in it so it's not a big deal to drop it.

kingthorin:
@psiinon, thoughts/feedback?

psiinon:
As per our irc chat, I see advantages in both, in which case this global options is a good start.

thc202:
I'd prefer see this as a core option than having the add-ons add/set the IDs, it would be just one place to tweak that (instead of having an option in each add-on?).

kingthorin:
DO NOT MERGE.

1) The options panel still isn't right. (GroupLayout is confusing, I'm fairly certain I'm overlooking something trivial though).
2) If I access something like example.org/jane I get a passive alert for missing CSP. If I then fuzz jane (even making it null or blank so that a 200 happens). The logging shows that the fuzzer messages are scanned but no alert pops.

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -141,7 +147,14 @@ public void run() {
...
 								}
-								if (scanner.isEnabled() && scanner.appliesToHistoryType(href.getHistoryType())) {
+								int hrefHistoryType = href.getHistoryType();
+								if (pscanOptions.isScanFuzzerMessages()) {
+									addApplicableHistoryType(HistoryReference.TYPE_FUZZER);
+									addApplicableHistoryType(HistoryReference.TYPE_FUZZER_TEMPORARY);
+									logger.info("Here " + href.getURI().toString());
+								}
+								if (scanner.isEnabled() && scanner.appliesToHistoryType(hrefHistoryType)
+										&& getApplicableHistoryTypes().contains(hrefHistoryType)) {
```
thc202: This one should be an or.

kingthorin:
All fixed up and ready for review.

kingthorin:
Do I need to control the enable state of the button based on whether or not the fuzzer is installed or is that overkill?

thc202:
As mentioned in IRC I don't think that's necessary (e.g. replacer does not do that).

kingthorin:
Thanks thc202! Sorry if I forgot or misunderstood something we talked about.

thc202:
No, just answering here for the record.

kingthorin:
@psiinon following conversations on IRC I've made further changes since your last review. Could you take a look again?

src/lang/Messages.properties
```diff
@@ -1849,7 +1849,8 @@ pscan.options.dialog.scanner.field.label.enabled                 = Enabled:
 pscan.options.dialog.scanner.field.label.name                    = Name:
 pscan.options.dialog.scanner.field.label.type                    = Type:
 pscan.options.main.name = Passive Scanner
-pscan.options.main.label.scanOnlyInScope = Only scan messages in scope:
+pscan.options.main.label.scanOnlyInScope = Only scan messages in scope
+pscan.options.main.label.scanFuzzerMessages = Include traffic from the Fuzzer when passive scanning
```
thc202: The space at the end could be removed (although it does not make much difference).

src/org/zaproxy/zap/extension/pscan/PassiveScanParam.java
```diff
@@ -61,6 +62,14 @@
      */
     private boolean scanOnlyInScope;
     
+    /**
+     * Flag that indicates whether or not the passive scan should be performed on traffic generated by the fuzzer.
+     * <p>
+     * Default is {@code false}, fuzzer traffic is not scanned.
+     * @since TODO Add version
```
thc202: This "since" tag could be removed (only really needed for public/exposed API).

src/org/zaproxy/zap/extension/pscan/PassiveScanParam.java
```diff
@@ -155,4 +165,30 @@ public void setScanOnlyInScope(boolean scanOnlyInScope) {
...
+     *
+     * @param scanFuzzerMessages
+     *            {@code true} if the scan should be performed on traffic
+     *            generated by the fuzzer, {@code false} otherwise.
+     * @since TODO Add version
+     * @see #isScanFuzzerMessages()
+     */
+    public void setScanFuzzerMessages(boolean scanFuzzerMessages) {
+        this.scanFuzzerMessages = scanFuzzerMessages;
+        getConfig().setProperty(SCAN_FUZZER_MESSAGES_KEY, Boolean.valueOf(scanFuzzerMessages));
```
thc202: The (explicit) boxing should be removed (per new guidelines).

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -141,7 +146,13 @@ public void run() {
 								if (shutDown) {
 									return;
 								}
-								if (scanner.isEnabled() && scanner.appliesToHistoryType(href.getHistoryType())) {
+								int hrefHistoryType = href.getHistoryType();
+								if (pscanOptions.isScanFuzzerMessages()) {
```
thc202: It should also remove them (when disabled), but instead of doing that here it can be called by `PassiveScanParam` when changed.

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -141,7 +146,13 @@ public void run() {
...
+									addApplicableHistoryType(HistoryReference.TYPE_FUZZER);
+									addApplicableHistoryType(HistoryReference.TYPE_FUZZER_TEMPORARY);
+								}
+								if (scanner.isEnabled() && scanner.appliesToHistoryType(hrefHistoryType)
```
thc202: It needs parentheses for the types (otherwise the enabled could be bypassed, or was that intended?).

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -141,7 +146,13 @@ public void run() {
...
+										|| getApplicableHistoryTypes().contains(hrefHistoryType)) {
```
thc202: This should use `optedInHistoryTypes` (to just check additionally the opted-in not all, e.g. the scanner might not want to apply to all of defaults in the first place).

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -278,4 +289,58 @@ public void sessionAboutToChange(Session session) {
...
+	/**
+	 * Returns the set of History Types which have "opted-in" to be applicable
+	 * for passive scanning.
+	 * 
+	 * @return a set of {@code Integer} representing all of the History Types
+	 *         which have "opted-in" for passive scanning.
+	 * @since TODO add version
+	 */
+	public static Set<Integer> getOptedInHistoryTypes() {
+		return optedInHistoryTypes;
```
thc202: `Collections.unmodifiableSet(optedInHistoryTypes)`?

src/org/zaproxy/zap/extension/pscan/PassiveScannerOptionsPanel.java
```diff
@@ -42,27 +44,22 @@
...
-
-        layout.setAutoCreateGaps(true);
-        layout.setAutoCreateContainerGaps(true);
-
-        scanOnlyInScopeCheckBox = new JCheckBox();
-        JLabel scanOnlyInScopeLabel = new JLabel(messages.getString("pscan.options.main.label.scanOnlyInScope"));
-        scanOnlyInScopeLabel.setLabelFor(scanOnlyInScopeCheckBox);
-
-        layout.setHorizontalGroup(
-                layout.createSequentialGroup().addComponent(scanOnlyInScopeLabel).addComponent(scanOnlyInScopeCheckBox));
```
thc202: Replace with a `createParallelGroup` (and add `scanOnlyInScopeCheckBox` and `scanFuzzerMessagesCheckBox`).
Same for the vertical group but with a `createSequentialGroup`.

src/org/zaproxy/zap/extension/pscan/PassiveScannerOptionsPanel.java
```diff
@@ -42,27 +44,22 @@
...
-
-        layout.setVerticalGroup(
-                layout.createParallelGroup(GroupLayout.Alignment.BASELINE)
-                        .addComponent(scanOnlyInScopeLabel)
-                        .addComponent(scanOnlyInScopeCheckBox));
+        setLayout(new GridBagLayout());
+
+        scanOnlyInScopeCheckBox = new JCheckBox(messages.getString("pscan.options.main.label.scanOnlyInScope"));
```
thc202: The spaces at the end should be removed.

src/org/zaproxy/zap/extension/pscan/PassiveScannerOptionsPanel.java
```diff
@@ -42,27 +44,22 @@
...
+
+        scanOnlyInScopeCheckBox = new JCheckBox(messages.getString("pscan.options.main.label.scanOnlyInScope"));        
+        scanFuzzerMessagesCheckBox = new JCheckBox(messages.getString("pscan.options.main.label.scanFuzzerMessages"));
+        
+        int row = 0;
+        add(scanOnlyInScopeCheckBox, LayoutHelper.getGBC(0, row, 1, 0.0D, 0));
+        row++;
+        add(scanFuzzerMessagesCheckBox, LayoutHelper.getGBC(0, row, 1, 0.0D, 0));
+        
+        add(new JLabel(), LayoutHelper.getGBC(0, 10, 1, 0.5D, 1.0D)); // Spacer
```
thc202: This could be removed if still using the `GroupLayout`.

src/org/zaproxy/zap/extension/pscan/PassiveScanParam.java
kingthorin: :+1:

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -141,7 +147,9 @@ public void run() {
 								if (shutDown) {
 									return;
 								}
-								if (scanner.isEnabled() && scanner.appliesToHistoryType(href.getHistoryType())) {
+								int hrefHistoryType = href.getHistoryType();
+								if (scanner.isEnabled() && (scanner.appliesToHistoryType(hrefHistoryType)
+										|| getOptedInHistoryTypes().contains(hrefHistoryType))) {
```
thc202: This could use `optedInHistoryTypes` directly, to avoid the `Collections` wrapping.

src/org/zaproxy/zap/extension/pscan/PassiveScannerOptionsPanel.java
```diff
@@ -52,17 +52,17 @@ public PassiveScannerOptionsPanel(I18N messages) {
...
-                layout.createSequentialGroup().addComponent(scanOnlyInScopeLabel).addComponent(scanOnlyInScopeCheckBox));
-
-        layout.setVerticalGroup(
-                layout.createParallelGroup(GroupLayout.Alignment.BASELINE)
-                        .addComponent(scanOnlyInScopeLabel)
-                        .addComponent(scanOnlyInScopeCheckBox));
+                layout.createParallelGroup()
+                    .addComponent(scanOnlyInScopeCheckBox)
+                    .addComponent(scanFuzzerMessagesCheckBox));
+                layout.setVerticalGroup(
```
thc202: This (and following statements) should be moved to the left two levels?

src/org/zaproxy/zap/extension/pscan/PassiveScanParam.java
```diff
@@ -155,4 +165,37 @@ public void setScanOnlyInScope(boolean scanOnlyInScope) {
...
+     * @param scanFuzzerMessages
+     *            {@code true} if the scan should be performed on traffic
+     *            generated by the fuzzer, {@code false} otherwise.
+     * @since TODO Add version
+     * @see #isScanFuzzerMessages()
+     */
+    public void setScanFuzzerMessages(boolean scanFuzzerMessages) {
+        this.scanFuzzerMessages = scanFuzzerMessages;
+        getConfig().setProperty(SCAN_FUZZER_MESSAGES_KEY, scanFuzzerMessages);
+        if (scanFuzzerMessages) {
```
thc202: Just noticed that with this change we now need to execute this logic after reading from the file. (Extract a method and call from both places?)

kingthorin: Sure, is it he file read it n this class?

kingthorin: What file?

thc202: Sorry, I was referring to the `parse` method.

kingthorin: Thanks for clarifying, will look into it.

thc202:
Thank you!

PR #4322, opened 2018-01-29
Fix API issues with pconn connections

psiinon:


src/org/parosproxy/paros/core/proxy/ProxyThread.java
```diff
@@ -451,13 +452,19 @@ protected void processHttp(HttpRequestHeader requestHeader, boolean isSecure) th
...
 			    }
 			}
-
-			if (parentServer.isEnableApi() &&
-					API.getInstance().handleApiRequest(requestHeader, httpIn, httpOut, isRecursive(requestHeader))) {
-				// It was an API request
-				return;
+			if (parentServer.isEnableApi()) {
+				msg = API.getInstance().handleApiRequest(requestHeader, httpIn, httpOut, isRecursive(requestHeader));
+				if (msg != null) {
```
thc202: Should have another if (to just return) for when the `msg` is empty? We shouldn't notify if the API request was not allowed or is already handled.

src/org/parosproxy/paros/core/proxy/ProxyThread.java
```diff
@@ -451,13 +452,19 @@ protected void processHttp(HttpRequestHeader requestHeader, boolean isSecure) th
...
+					ZapGetMethod method = (ZapGetMethod) msg.getUserObject();
```
thc202: Does it worth get/check the method? (It should never be present, the API does not set it?)

psiinon:
Hopefully fixed those comments

thc202:
Thanks!

Issue #1161, opened 2015-06-04
[Enable Session Tracking (Cookie)] and [Use current tracking session] toggle button


zapbot:
```
What steps will reproduce the problem?
1. Check [Enable Session Tracking (Cookie)] in [Edit] menu.
2. Click  [Manual Request Editor...] in [Tools] menu.
    At this point, [Use current tracking session] toggle button is still off.
3. Close Manual Request Editor.
4. Open Manual Request Editor.
    At this point, [Use current tracking session] toggle button is on.
5. Close Manual Request Editor and uncheck [Enable Session Tracking (Cookie)] in [Edit]
menu.
6. Open Manual Request Editor.
    At this point, [Use current tracking session] toggle button is still on.
7. Reopen Manual Request Editor.
    At this point, [Use current tracking session] toggle button is still on. 

What is the expected output? What do you see instead?
- At the point of step 2, [Use current tracking session] toggle button should be on.
- At the point of step 6 and 7, [Use current tracking session] toggle button should
be off.

What version of the product are you using? On what operating system?

ZAP2.3.0.1
Mac OS X 10.9.2 

Please provide any additional information below.

- In the branches/2.3/src/org/parosproxy/paros/extension/state/ExtensionState.java#105,
  it might be good to call "isSelected" method instead of "isEnabled".
- In the branches/2.3/src/org/parosproxy/paros/extension/manualrequest/ManualRequestEditorDialog.java#99,
  "getMessageSender().cleanup()" is called in "windowClosing" method, but it might
be good to call "getMessageSender().cleanup()" at the point of window opening to reload
the current toggle's status before doing Manual Request task.

```

Original issue reported on code.google.com by `yukisov` on 2014-04-23 14:09:11

kingthorin:
I don't seem to see this option in the Edit menu any more. Was it removed at some point? Or is some particular config required to see/enable it?

I do not it is still listed in the Help content.

kingthorin:
I guess it's over here -> `Model.getSingleton().getOptionsParam().getConnectionParam().isHttpStateEnabled()`

thc202:
The option was moved to other place #3394 (see also, #4280).

kingthorin:
Since it's a global option now shall we just remove the button from the resend dialog?

psiinon:
Not if it still works (and if it doesnt then we should fix it) - being able to override a global option from a specific dialog is really useful

kingthorin:
Ok fair enough.

[...]

PR #4340, opened 2018-02-01
Fix Issue 1161 - Manual Request SessionTracking Button Sync

kingthorin:
Add overridden windowActivated method when initializing. Calls `getMessageSender().cleanup()` so that the state of the Session Tracking button is properly set to be the same as the current global config option setting.

Fixes zaproxy/zaproxy#1161

### DIFF
--- a/src/org/parosproxy/paros/extension/manualrequest/ExtensionManualRequestEditor.java
+++ b/src/org/parosproxy/paros/extension/manualrequest/ExtensionManualRequestEditor.java
@@ -41,6 +41,7 @@
 // ZAP: 2016/06/20 Removed unnecessary/unused constructor
 // ZAP: 2017/04/07 Added getUIName()
 // ZAP: 2017/06/06 Clear dialogues in EDT.
+// ZAP: 2018/02/23 Issue 1161: Fix Session Tracking button sync
 
 package org.parosproxy.paros.extension.manualrequest;
 
@@ -65,6 +66,7 @@ import org.zaproxy.zap.extension.httppanel.Message;
 public class ExtensionManualRequestEditor extends ExtensionAdaptor implements SessionChangedListener {
 	
 	private Map<Class<? extends Message>, ManualRequestEditorDialog> dialogues = new HashMap<>();
+	private ManualHttpRequestEditorDialog httpSendEditorDialog;
 	
 	/**
 	 * Name of this extension.
@@ -87,7 +89,7 @@ public class ExtensionManualRequestEditor extends ExtensionAdaptor implements Se
         super.initView(view);
 
         // add default manual request editor
-        ManualRequestEditorDialog httpSendEditorDialog = new ManualHttpRequestEditorDialog(true, "manual", "ui.dialogs.manreq");
+        httpSendEditorDialog = new ManualHttpRequestEditorDialog(true, "manual", "ui.dialogs.manreq");
         httpSendEditorDialog.setTitle(Constant.messages.getString("manReq.dialog.title"));
         
         addManualSendEditor(httpSendEditorDialog);
@@ -139,6 +141,7 @@ public class ExtensionManualRequestEditor extends ExtensionAdaptor implements Se
 			}
 			
 			extensionHook.addSessionListener(this);
+			extensionHook.addOptionsChangedListener(httpSendEditorDialog);
 		}
 	}
 
--- a/src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
+++ b/src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
@@ -18,6 +18,7 @@
 package org.parosproxy.paros.extension.manualrequest.http.impl;
 
 import java.awt.EventQueue;
+import java.awt.event.ItemEvent;
 import java.io.IOException;
 import java.net.Socket;
 import java.net.UnknownHostException;
@@ -202,10 +203,6 @@ public class HttpPanelSender implements MessageSender {
             delegate.shutdown();
             delegate = null;
         }
-
-        final boolean isSessionTrackingEnabled = Model.getSingleton()
-                .getOptionsParam().getConnectionParam().isHttpStateEnabled();
-        getButtonUseTrackingSessionState().setEnabled(isSessionTrackingEnabled);
     }
 
     private HttpSender getDelegate() {
@@ -239,6 +236,7 @@ public class HttpPanelSender implements MessageSender {
                     .getResource("/resource/icon/fugue/cookie.png"))); // Cookie
             useTrackingSessionState.setToolTipText(Constant.messages
                     .getString("manReq.checkBox.useSession"));
+            useTrackingSessionState.addItemListener(e -> setUseTrackingSessionState(e.getStateChange() == ItemEvent.SELECTED));
         }
         return useTrackingSessionState;
     }
@@ -310,5 +308,16 @@ public class HttpPanelSender implements MessageSender {
         public URI getInvalidRedirection() {
             return invalidRedirection;
         }
+    }
+    
+    private void setUseTrackingSessionState(boolean shouldUseTrackingSessionState) {
+        if (delegate != null) {
+            delegate.setUseGlobalState(shouldUseTrackingSessionState);
+        }
+    }
+    
+    public void setButtonTrackingSessionStateEnabled(boolean enabled) {
+        getButtonUseTrackingSessionState().setEnabled(enabled);
+        getButtonUseTrackingSessionState().setSelected(enabled);
     }
 }
--- a/src/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
+++ b/src/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
@@ -52,9 +52,10 @@ import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.extension.OptionsChangedListener;
 import org.parosproxy.paros.extension.manualrequest.ManualRequestEditorDialog;
-import org.parosproxy.paros.extension.manualrequest.MessageSender;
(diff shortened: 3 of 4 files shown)


src/org/parosproxy/paros/extension/manualrequest/ManualRequestEditorDialog.java
```diff
@@ -100,6 +101,10 @@ public ManualRequestEditorDialog(boolean isSendEnabled, String configurationKey)
 
 	protected void initialize() {
 		addWindowListener(new WindowAdapter() {
+			@Override
+			public void windowActivated(WindowEvent e) {
+				getMessageSender().cleanup();
```
thc202: It would be better if the state of the button was tied to the state of the option (and this could be handled as a HTTP implementation detail). If you want to keep this I'd suggest adding a new (default) method to the interface with a name more accordingly to the behaviour provided (does not seem correct to "cleanup" each time the dialogue is focused, even though the WebSocket implementation is not doing anything).

kingthorin: So you mean check the global option here and set the ‘selected’ value of the button (instead of calling cleanup)?

thc202: It would not be here, in `HttpPanelSender`(or ManualHttpRequestEditorDialog?) by having it listen to the option change (`OptionsChangedListener`?) and update the button when that happens.

src/org/parosproxy/paros/extension/manualrequest/ExtensionManualRequestEditor.java
```diff
@@ -65,6 +66,7 @@
 public class ExtensionManualRequestEditor extends ExtensionAdaptor implements SessionChangedListener {
 	
 	private Map<Class<? extends Message>, ManualRequestEditorDialog> dialogues = new HashMap<>();
+	ManualHttpRequestEditorDialog httpSendEditorDialog = new ManualHttpRequestEditorDialog(true, "manual", "ui.dialogs.manreq");
```
thc202: - `private`;
 - Declare here, but instantiate in the `initView` as before (to still not create it if there's no view).

src/org/parosproxy/paros/extension/manualrequest/ExtensionManualRequestEditor.java
```diff
@@ -136,6 +137,7 @@ public void hook(ExtensionHook extensionHook) {
 		if (getView() != null) {
 			for (Entry<Class<? extends Message>, ManualRequestEditorDialog> dialogue : dialogues.entrySet()) {
 				extensionHook.getHookMenu().addToolsMenuItem(dialogue.getValue().getMenuItem());
+				extensionHook.addOptionsChangedListener(httpSendEditorDialog);
```
thc202: This should be done outside the loop, after `addSessionListener`?

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -311,4 +308,12 @@ public URI getInvalidRedirection() {
             return invalidRedirection;
         }
     }
+    
+    public void setUseTrackingSessionState (boolean shouldUseTrackingSessionState) {
+        getButtonUseTrackingSessionState().setEnabled(shouldUseTrackingSessionState);
+        delegate.setUseGlobalState(shouldUseTrackingSessionState);
```
thc202: - Needs a null check for `delegate`.
 - This should be done inside the following if statement (otherwise it could be set without the button being selected).
 - Per the talk in IRC, yeah, we also need to add an item listener to the button to enable/disable the global state in the sender (previously you had to close/open the dialogue to active it...).

src/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
```diff
@@ -566,10 +567,15 @@ private JSplitPane createSplitPane(int orientation) {
 	}
 	
 	public void addPersistentConnectionListener(PersistentConnectionListener listener) {
-		((HttpPanelSender) getMessageSender()).addPersistentConnectionListener(listener);
+		(getMessageSender()).addPersistentConnectionListener(listener);
```
thc202: Could be removed the surrounding parentheses too. (Same comment for the following change.)

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
kingthorin: Should I just remove the condition and simplify to:
```Java
public void setUseTrackingSessionState (boolean shouldUseTrackingSessionState) {
    getButtonUseTrackingSessionState().setSelected(shouldUseTrackingSessionState);
    getDelegate().setUseGlobalState(shouldUseTrackingSessionState);
}
```
Whether it's false or not set it, and use getDeligate so that we don't have to worry about null handling.

kingthorin: GAH, I think this should be using setSelected not setEnabled.

I can use getDelegate() to address the null handling. I think there's other stuff missing too...I'll have to tackle further changes next week I think.

thc202: Fine with the use of `getDelegate` (though with null check it would avoid creating it). The current usage of setEnabled/setSelected is correct.

kingthorin: Ok I figured if we were checking we’d want it to exist.

kingthorin: Also I don’t think the setEnabled is right because if it starts off it can never be toggled....

thc202: It can if the option is enabled (otherwise the user is not supposed to interact with the button).

kingthorin: Oh I figured they should be able to toggle on or off at will totally separate from the global option, but to be honest it was a pretty shallow thought ;)

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -311,4 +322,17 @@ public URI getInvalidRedirection() {
             return invalidRedirection;
         }
     }
+    
+    private void setUseTrackingSessionState(boolean shouldUseTrackingSessionState) {
+        if (!shouldUseTrackingSessionState) {
```
thc202: This if can be removed from this method (should now be done in `setButtonTrackingSessionStateEnabled`).

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -311,4 +322,17 @@ public URI getInvalidRedirection() {
...
         }
     }
+    
+    private void setUseTrackingSessionState(boolean shouldUseTrackingSessionState) {
+        if (!shouldUseTrackingSessionState) {
+            if (delegate != null) {
+                delegate.setUseGlobalState(shouldUseTrackingSessionState);
+            }
+        }
+        getButtonUseTrackingSessionState().setSelected(shouldUseTrackingSessionState);
```
thc202: This should be moved to `setButtonTrackingSessionStateEnabled` (done when `enabled` is `false`).

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -311,4 +322,17 @@ public URI getInvalidRedirection() {
...
+    }
+    
+    public void setButtonTrackingSessionStateEnabled(boolean enabled) {
+        getButtonUseTrackingSessionState().setEnabled(enabled);
```
thc202: If `enabled` is `false` we should call `setUseTrackingSessionState` (with `false`).

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -239,6 +238,18 @@ private JToggleButton getButtonUseTrackingSessionState() {
...
+                  public void itemStateChanged(ItemEvent itemEvent) {
+                    int state = itemEvent.getStateChange();
+	                    if (state == ItemEvent.SELECTED) {
+	                      setUseTrackingSessionState(true);
+	                    } else {
+	                      setUseTrackingSessionState(false);
+	                    }
+                    }
+                };
+              useTrackingSessionState.addItemListener(btnItemnListener);
```
thc202: Could use a lambda:
```Java
useTrackingSessionState.addItemListener(e -> setUseTrackingSessionState(e.getStateChange() == ItemEvent.SELECTED));
```

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -15,9 +15,12 @@
  * See the License for the specific language governing permissions and 
  * limitations under the License. 
  */
+// ZAP: 2018/02/23 Issue 1161: Fix Session Tracking button sync
```
thc202: Can be removed, ZAP class :)

src/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
kingthorin: In `setButtonTrackingSessionStateEnabled` I want to set it no matter what...

I wanted to separate the enable and selected setting as the button needs to be set disabled when initially displayed, but not when the user presses the button.

thc202: I'm not following, the button needs to be deselected when the button is disabled (to effectively disable the option).

thc202:
Thank you!

PR #4342, opened 2018-02-02
Clean up / Fix JavaDoc errors

sigee:


src/org/zaproxy/zap/extension/spider/SpiderPanel.java
```diff
@@ -142,7 +142,7 @@
 	 * <p>
 	 * Lazily initialised.
 	 * 
-	 * @see #getAddedNodesTableScrollPane()
+	 * @see #getAddedUrlsTableScrollPane()
```
thc202: Better to rename the method.

src/org/zaproxy/zap/spider/filters/ParseFilter.java
```diff
@@ -62,7 +62,7 @@ public FilterResult filtered(HttpMessage responseMessage) {
 	 *
 	 * @since 2.7.0
 	 * @see #NOT_FILTERED
-	 * @see #ParseFilter(String)
+	 * @see #filtered(HttpMessage)
```
thc202: It should be `#FilterResult(String)` instead (though Java 8 does not process it correctly).

thc202:
Guess the ZAP comments can be skipped, @kingthorin, @psiinon wdyt?

kingthorin:
I think so, it's a non-functional change.

kingthorin:
Please do an interactive rebase and fix-up or squash into a single "atomic" commit: https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#what-we-zap-team-expect-from-you .

- https://www.instructables.com/id/How-to-Rebase-Interactive-With-Eclipse-eGit/

thc202:
Thanks!

thc202:
@sigee how would you like to be credited?

Issue #4277, opened 2018-01-17
Warn if a # is specified in a break point URL


psiinon:
If someone includes a URL fragment in a contains match then its probably not going to work as the fragment is not sent by the browser.
https://security.stackexchange.com/questions/177816/how-to-intercept-xmlhttprequest-with-owaps-zap/177825#177825

varshav21:
I am a beginner and I would like to take on the issue.Can anybody give me suggestions on how to approach this issue?

psiinon:
@varshav21 

1. Find the dialog in the UI which allows you to add break points
1. Search for some of the text in the dialog in the src folder
1. Make a note of the i18n key
1. Search for the i18n key in the src folder

You'll hopefully now have found the relevant code - have you got that far yet?

varshav21:
Thank you so much for your response, I believe I have found the relevant code.

psiinon:
Where have you found the text in the dialog? Can you post a link to it here?

kingthorin:
> I am stuck on finding i18n key, can you please explain how do I do that?

All strings in the ZAP UI are internationalized (translated) into various other languages. These appear in the various messages.properties files. 

So if you take a few unique word from the UI (such as those displayed in the add breakpoint dialog or panel), search the code base for it, you'll find a messages.properties entry like: `core.view.add.breakpoint.label=Add Breakpoint` (or whatever....thats a fictitious example). You can then take the key `core.view.add.breakpoint.label` and search the code base for its use and find the java source code files it's used in.

varshav21:
I think this is the i18n key
break.api.action.addHttpBreakpoint=Adds a custom HTTP breakpont. The string is the string to match. Location may be one of\: url, request_header, request_body, response_header or response_body. Match may be\: contains or regex. Inverse (match) may be true or false. Lastly, ignorecase (when matching the string) may be true or false.
I've found the key in  zaproxy/src/org/zaproxy/zap/extension/brk/BreakAPI.java

psiinon:
Thats for the API, which isnt what you want in this case.
You need to be very specific when searching for strings - I cant see "Adds a custom HTTP breakpont" in the desktop UI anywhere, its only in the API UI.
A better string to look for is the title of the popup: "Add Break Point"

varshav21:
[...]

PR #4349, opened 2018-02-05
Warn if a # is specified in a break point URL(Issue #4277)

varshav21:


src/org/zaproxy/zap/extension/brk/impl/http/BreakAddEditDialog.java
```diff
@@ -146,6 +146,9 @@ public String validateFields() {
 			}
 		}
 		return null;
+		if(FIELD_STRING.contains("#")==true){
```
thc202: - It should do the contains on the value of the field (e.g. `getStringValue(FIELD_STRING)`).
 - The comparison to `true` could be removed.
 - Note that this check needs to be done before the above return statement.

src/org/zaproxy/zap/extension/brk/impl/http/BreakAddEditDialog.java
```diff
@@ -146,6 +146,9 @@ public String validateFields() {
...
+			return Constant.messages.getString("brk.brkpoint.edit.title");
```
thc202: It needs to be a new/custom message that indicates the actual problem.

kingthorin:
@varshav21 are you able to address the feedback?

varshav21:
Thank you so much, yes I am working on it.Do I add a new custom key in messages.properties file for the return message?

kingthorin:
Yes. Something like `brk.brkpoint.warn.urlfragment=Pattern shouldn't include URL fragment (#)` or something along those lines.

varshav21:
Why did the Travis CI build fail when I pushed commits to this branch but was successful when I pushed the same commits to the master branch and started a separate pull request?

thc202:
This one has different file changes (note that the class has leading plus chars in some lines), removing them should pass the build.

thc202:
BTW, the new validation should also check if the `FIELD_LOCATION` is `url`(?) and `Match` is `contains`?

varshav21:
Should I take the key for url and contains from the messages.properties file and check if they are equal to strings FIELD_LOCATION and FIELD_MATCH respectively in the if condition?

thc202:
For that we can use the methods `strToLoc` and `strToMatch`, then compare against the `Location.url` and `Match.contains`.

src/org/zaproxy/zap/extension/brk/impl/http/BreakAddEditDialog.java
```diff
@@ -145,7 +145,10 @@ public String validateFields() {
 				return Constant.messages.getString("brk.brkpoint.error.regex"); 
 			}
 		}
-		return null;
+		if( (this.getStringValue(FIELD_STRING)).contains("#") && (Location.url.equals(this.strToLoc(this.getStringValue(FIELD_LOCATION)))) && (Match.contains.equals(this.strToMatch(this.getStringValue(FIELD_MATCH)))) ){
```
thc202: Some of the parentheses could be removed (e.g. the ones surrounding `this.getStringValue(FIELD_STRING)` and following operands).

src/org/zaproxy/zap/extension/brk/impl/http/BreakAddEditDialog.java
```diff
@@ -145,7 +145,10 @@ public String validateFields() {
...
+			return Constant.messages.getString("brk.brkpoint.warn.urlfragment");
+		}
+		return null;
```
thc202: The tabs added at the end of the line should be removed.

thc202:
Thank you!

thc202:
@varshav21 how would you like to be credited?

varshav21:
Thank you very much for your guidance. Varsha Venkatesh is fine.

PR #4380, opened 2018-02-12
Add API view to get context's URLs

### LINKED ISSUE #4364: Allow to obtain context's URLs through the API
From OWASP ZAP Developer Group: https://groups.google.com/d/topic/zaproxy-develop/HaQLAcSOZhI/discussion

It should be possible to obtain the URLs of a context through the ZAP API (similar to UI #3253).

thc202:
Add a view to ContextAPI that allows to obtain all URLs that belong to
a given context.
Add description to the view in Messages.properties file.

Fix #4364 - Allow to obtain context's URLs through the API

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ContextAPI.java
+++ b/src/org/zaproxy/zap/extension/api/ContextAPI.java
@@ -20,15 +20,19 @@ package org.zaproxy.zap.extension.api;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import net.sf.json.JSONObject;
 
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.model.HistoryReference;
 import org.parosproxy.paros.model.Model;
+import org.parosproxy.paros.model.SiteNode;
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/extension/api/ContextAPI.java
```diff
@@ -296,6 +302,23 @@ public ApiResponse handleApiView(String name, JSONObject params)
 			}
 			result = resultList;
 			break;
+		case VIEW_URLS:
+			resultList = new ApiResponseList(name);
+			Set<String> addedUrls = new HashSet<>();
+			for (SiteNode node : getContext(params).getNodesInContextFromSiteTree()) {
+				HistoryReference href = node.getHistoryReference();
+				if (HistoryReference.getTemporaryTypes().contains(href.getHistoryType())) {
```
kingthorin: When we were talking the other day about passive scanning temp fuzzer messages you mentioned that alerts on them  weren't an issue. If a URL is added or represented only by an alert on a temp message would it be missed here?

thc202: Yes. I added that check to keep the same behaviour as the GUI functionality, I'm fine changing both places to ignore the type? (Maybe it's better, we already do that in other endpoints, e.g. #4176).

kingthorin: I think that’s probably best (change both).

thc202: OK, will do.

PR #4384, opened 2018-02-13
Remove unnecessary boxing / unboxing (Java 5 type of warnings)

sigee:


sigee:
@psiinon,
I have added the comments to almost every class under org.parosproxy.paros packages package except:
- org.parosproxy.paros.extension.encoder.Base64
- org.parosproxy.paros.extension.history.HistoryFilter
- org.parosproxy.paros.extension.option.DatabaseParam
- org.parosproxy.paros.extension.option.ProxyExcludedDomainsTableModel

because these are ZAP and its related class files or "This code is from Public Domain" and any ot them had ZAP: YYY/MM/DD type of comments however they have been changed.
Should I add the comments to these files also or it is not needed?

thc202:
It's not necessary, only the Paros classes need the ZAP comment.

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -410,7 +410,7 @@ private void applyConnectionParams(final HttpMethod method) throws IOException {
             param = this.conn.getParams().getParameter(HttpConnectionParams.SO_TIMEOUT);
         }
         if (param != null) {
-            timeout = ((Integer)param).intValue();
+            timeout = (Integer) param;
```
thc202: Would you mind reverting this change? (Be good to keep the changes to these forked classes to a minimum necessary for our use cases.)

sigee: @thc202,
You are right. I'll revert this one.

sigee:
@thc202,
I've added "If it is possible avoid introduce the un/boxings in new code." to the Dev guidelines wiki page.

src/org/apache/commons/httpclient/HttpMethodDirector.java
```diff
@@ -410,7 +410,7 @@ private void applyConnectionParams(final HttpMethod method) throws IOException {
             param = this.conn.getParams().getParameter(HttpConnectionParams.SO_TIMEOUT);
         }
         if (param != null) {
-            timeout = ((Integer)param).intValue();
+            timeout = ((Integer) param).intValue();
```
thc202: Could this change be reverted as well?

thc202:
Thanks for adding the guideline entry.

thc202:
Thanks!

PR #4395, opened 2018-02-14
Improve baseline error messages

psiinon:


psiinon:
Updated to include the other scripts. I know there duplicated code, but I'm not sure its worth pulling htat out just because they have the same options.
But I did add the missing 'decode' in the other 2 scripts...

thc202:
Travis CI didn't run??

Issue #4391, opened 2018-02-14
Enhancement: Params Tab - Multipart Form Params


kingthorin:
I see that this came up in IRC the other day: https://mozilla.logbot.info/websectools/20180212

It's great that one of our users scripted a way around the current limitation, however, it may as well actually be part of the tool.

Now that the multipart form params handling has been revised (various recent PRs [1][2]). It seems pretty easy to add those params. 

My question is how would we want the file part of multipart form posts handled? I'm thinking by filename. Having the file content in the params window doesn't seem practical....

I'm thinking something like this around L309 in ExtensionParams.java
```
VariantMultipartFormParameters params2 = new VariantMultipartFormParameters();
params2.setMessage(msg);
for (NameValuePair nvp : params2.getParamList()) {
	if (Arrays.asList(NameValuePair.TYPE_MULTIPART_DATA_PARAM, NameValuePair.TYPE_MULTIPART_DATA_FILE_NAME)
			.contains(nvp.getType())) {
		persist(sps.addParam(site, new HtmlParameter(HtmlParameter.Type.form, nvp.getName(), nvp.getValue()),
				msg));
	}
}
```

Seem reasonable?

[1] https://github.com/zaproxy/zaproxy/pull/3814
[2] https://github.com/zaproxy/zaproxy/pull/4208

psiinon:
yeah, that sounds good.

thc202:
How about using other type than `form` (to differentiate them)?

kingthorin:
Sure, I can add another type as well.

PR #4397, opened 2018-02-14
Fix Issue 4391 - Include MultipartFormPost Params on Params Tab

kingthorin:
Update the Params Extension to include multipart/form-data parameters. The "value" for file parameters is the filename. Leverage `Pattern.quote()` for existing search functionality. Remove inline block comments.

Fixes zaproxy/zaproxy#4391

### DIFF
--- a/src/org/zaproxy/zap/extension/params/ExtensionParams.java
+++ b/src/org/zaproxy/zap/extension/params/ExtensionParams.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Pattern;
 
 import javax.swing.tree.TreeNode;
 
@@ -38,6 +39,8 @@ import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.control.Control.Mode;
+import org.parosproxy.paros.core.scanner.NameValuePair;
+import org.parosproxy.paros.core.scanner.VariantMultipartFormParameters;
 import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.db.RecordParam;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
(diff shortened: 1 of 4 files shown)


kingthorin:
Update of `/zap-core-help/src/help/zaphelp/contents/ui/tabs/params.html` to follow after this is merged.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -305,6 +307,16 @@ public boolean onHttpRequestSend(HttpMessage msg) {
 			persist(sps.addParam(site, param, msg));
 		}
 		
+		VariantMultipartFormParameters params2 = new VariantMultipartFormParameters();
+		params2.setMessage(msg);
+		for (NameValuePair nvp : params2.getParamList()) {
+			if (Arrays.asList(NameValuePair.TYPE_MULTIPART_DATA_PARAM, NameValuePair.TYPE_MULTIPART_DATA_FILE_NAME)
```
thc202: This could be created once before the for loop (or moved to a constant, or just compare the types).

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -411,6 +423,9 @@ protected void searchForSelectedParam() {
 						extSearch.search(/*".*" + */item.getName() + "=.*", ExtensionSearch.Type.Header, true, false);
 				} else if (HtmlParameter.Type.header.equals(item.getType())) {
 					extSearch.search(item.getName() + ":.*", ExtensionSearch.Type.Header, true, false);
+				} else if (HtmlParameter.Type.multipart.equals(item.getType())) {
+					extSearch.search("(?i)\\s*content-disposition\\s*:.*\\s+name\\s*\\=?\\s*\\\"?" + item.getName(),
```
thc202: We should quote the name, it might affect the regex (same applies to other types though).

kingthorin: Do you mean `+ "\""+ item.getName() + "\""`?

src/org/zaproxy/zap/extension/params/ExtensionParams.java
kingthorin: Changed it to compare `nvp.getType()` vs the two types (ORed).

src/org/zaproxy/zap/extension/params/ExtensionParams.java
thc202: With `Pattern.quote`.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
thc202: Looks good.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
kingthorin: Ok so:
`extSearch.search("(?i)\\s*content-disposition\\s*:.*\\s+name\\s*\\=?\\s*\\\"?" + Pattern.quote(item.getName()), ExtensionSearch.Type.Request, true, false);` ?

Shall I go ahead and change the others while I'm at it?

thc202: Yeah. I think so.

kingthorin: Thanks @thc202. I’m tied up in a presentation all afternoon, so I’ll tackle that this evening.

thc202: OK. Regarding the inline comments, +1 :)

thc202:
Thanks!

PR #4398, opened 2018-02-14
Allow to configure the irrelevant parameters for the Spider

### LINKED ISSUE #4388: Allow to configure the irrelevant parameters for the Spider
From IRC channel: https://mozilla.logbot.info/websectools/20180207

It should be possible to configure the irrelevant parameters (e.g. session tokens)  when canonicalising the URLs found during the spider scan, currently they are hardcoded:
https://github.com/zaproxy/zaproxy/blob/fc0cbcbad2ad982de35060260e2c2ac6685f9ee3/src/org/zaproxy/zap/spider/URLCanonicalizer.java#L60-L64
and
https://github.com/zaproxy/zaproxy/blob/fc0cbcbad2ad982de35060260e2c2ac6685f9ee3/src/org/zaproxy/zap/spider/URLCanonicalizer.java#L305

(Ideally the session tokens should be retrieved from the HTTP Sessions options.)

AleksandrPea:
Fix #4388.

* Added irrelevantParameters property to `SpiderParam`.
* Tweaked URLCanonicalizer to take into account irrelevant parameters. 
* Made changes on UI side to allow custom configuration for irrelevant parameters.

Spider options screenshot:
selection_001

Spider dialog screenshot:
selection_002

Please, review:
@thc202

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
@@ -34,6 +34,7 @@ import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 import javax.swing.JScrollPane;
 import javax.swing.JSlider;
+import javax.swing.ScrollPaneConstants;
 import javax.swing.SortOrder;
 import javax.swing.border.Border;
 import javax.swing.border.EmptyBorder;
@@ -46,6 +47,7 @@ import org.parosproxy.paros.model.OptionsParam;
 import org.parosproxy.paros.view.AbstractParamPanel;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.utils.ZapNumberSpinner;
+import org.zaproxy.zap.utils.ZapTextArea;
 import org.zaproxy.zap.view.AbstractMultipleOptionsTablePanel;
 import org.zaproxy.zap.view.LayoutHelper;
 import org.zaproxy.zap.view.PositiveValuesSlider;
@@ -83,6 +85,7 @@ public class OptionsSpiderPanel extends AbstractParamPanel {
     private JCheckBox chkAcceptCookies;
     private DomainsAlwaysInScopeMultipleOptionsPanel domainsAlwaysInScopePanel;
     private DomainsAlwaysInScopeTableModel domainsAlwaysInScopeTableModel;
+    private ZapTextArea irrelevantUrlParameters;
 
     private JComboBox<org.zaproxy.zap.spider.SpiderParam.HandleParametersOption> handleParameters =
             null;
@@ -189,6 +192,20 @@ public class OptionsSpiderPanel extends AbstractParamPanel {
             innerPanel.add(getChkParseGit(), gbc);
             innerPanel.add(getHandleODataSpecificParameters(), gbc);
 
+            ZapTextArea irrelevantUrlParameters = getIrrelevantUrlParameters();
+            JLabel label =
+                    new JLabel(
+                            Constant.messages.getString(
+                                    "spider.options.label.irrelevantUrlParameters"));
+            label.setLabelFor(irrelevantUrlParameters);
+            JScrollPane irrelevantUrlParametersScrollPane = new JScrollPane();
+            irrelevantUrlParametersScrollPane.setVerticalScrollBarPolicy(
+                    ScrollPaneConstants.VERTICAL_SCROLLBAR_AS_NEEDED);
+            irrelevantUrlParametersScrollPane.setViewportView(irrelevantUrlParameters);
+
+            innerPanel.add(label, gbc);
+            innerPanel.add(irrelevantUrlParametersScrollPane, gbc);
+
             JScrollPane scrollPane = new JScrollPane(innerPanel);
             scrollPane.setBorder(BorderFactory.createEmptyBorder());
 
@@ -223,6 +240,8 @@ public class OptionsSpiderPanel extends AbstractParamPanel {
         getChkParseGit().setSelected(param.isParseGit());
         getComboHandleParameters().setSelectedItem(param.getHandleParameters());
         getHandleODataSpecificParameters().setSelected(param.isHandleODataParametersVisited());
+        getIrrelevantUrlParameters().setText(param.getIrrelevantUrlParametersAsString());
+        getIrrelevantUrlParameters().discardAllEdits();
     }
 
     @Override
@@ -252,6 +271,7 @@ public class OptionsSpiderPanel extends AbstractParamPanel {
                 (org.zaproxy.zap.spider.SpiderParam.HandleParametersOption)
                         getComboHandleParameters().getSelectedItem());
         param.setHandleODataParametersVisited(getHandleODataSpecificParameters().isSelected());
+        param.setIrrelevantUrlParameters(getIrrelevantUrlParameters().getText());
     }
 
     /**
@@ -486,6 +506,14 @@ public class OptionsSpiderPanel extends AbstractParamPanel {
             domainsAlwaysInScopeTableModel = new DomainsAlwaysInScopeTableModel();
         }
         return domainsAlwaysInScopeTableModel;
+    }
+
+    private ZapTextArea getIrrelevantUrlParameters() {
+        if (irrelevantUrlParameters == null) {
+            irrelevantUrlParameters = new ZapTextArea();
+            irrelevantUrlParameters.setLineWrap(true);
+        }
+        return irrelevantUrlParameters;
     }
 
     /** A renderer for properly displaying the name of the HandleParametersOptions in a ComboBox. */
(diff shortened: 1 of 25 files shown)


src/org/zaproxy/zap/spider/URLCanonicalizer.java
```diff
@@ -96,17 +91,18 @@ private URLCanonicalizer(){}
...
 	/**
 	 * Gets the canonical url, starting from a relative or absolute url found in a given context (baseURL).
 	 * 
 	 * @param url the url string defining the reference
 	 * @param baseURL the context in which this url was found
+	 * @param irrelevantParameters parameters that are skipped when canonicalizing
 	 * @return the canonical url
 	 */
-	public static String getCanonicalURL(String url, String baseURL) {
+	public static String getCanonicalURL(String url, String baseURL, Set<String> irrelevantParameters) {
```
thc202: - It's probably better to use a functional interface (`Predicate<String>`?) instead of a passing a `Set`, it provides more flexibility.
- We should aim at keep binary compatibility (same for other public methods).

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
```diff
@@ -89,6 +90,7 @@
 	private JCheckBox chkAcceptCookies;
 	private DomainsAlwaysInScopeMultipleOptionsPanel domainsAlwaysInScopePanel;
 	private DomainsAlwaysInScopeTableModel domainsAlwaysInScopeTableModel;
+	private ZapTextArea irrelevantUrlParameters;
```
thc202: We have been moving away from free text options, as they are not as flexible as they should. For example, in this case I was expecting that the user could allow to specify a regular expression not just an exact string (it would also allow to remove the hardcoded "utm_" check).

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
```diff
@@ -224,7 +239,7 @@ public void initParam(Object obj) {
 		getChkParseGit().setSelected(param.isParseGit());
 		getComboHandleParameters().setSelectedItem(param.getHandleParameters());
 		getHandleODataSpecificParameters().setSelected(param.isHandleODataParametersVisited());
-
+		getIrrelevantUrlParameters().setText(param.getIrrelevantUrlParametersAsString());
```
thc202: The undoable edits should be discarded (also after saving).

src/org/zaproxy/zap/spider/URLCanonicalizer.java
```diff
@@ -469,6 +468,11 @@ private static String normalizePath(final String path) {
 		return path.replace("%7E", "~").replace(" ", "%20");
 	}
 
+	private static boolean isSessionToken(String paramName) {
+		return Control.getSingleton().getExtensionLoader().getExtension(ExtensionHttpSessions.class)
```
thc202: - As mentioned in the issue the extension is null if disabled.
 - This class should not depend on the extension, it should be up to the callers to include this additional check :)

src/org/zaproxy/zap/spider/parser/SpiderParser.java
```diff
@@ -93,10 +95,12 @@ protected void notifyListenersPostResourceFound(HttpMessage message, int depth,
 	 * @param depth the depth
 	 * @param localURL the local url
 	 * @param baseURL the base url
+	 * @param irrelevantUrlParameters url parameters that are skipped when canonicalizing
 	 */
-	protected void processURL(HttpMessage message, int depth, String localURL, String baseURL) {
+	protected void processURL(HttpMessage message, int depth, String localURL, String baseURL,
+							  Set<String> irrelevantUrlParameters) {
```
thc202: Since all parsers end up passing the same irrelevant parameters I'd suggest adding a new constructor `SpiderParser(SpiderParam)` (add deprecate the default) instead of adding this.

src/org/zaproxy/zap/spider/URLCanonicalizer.java
AleksandrPea: Please clarify your point about passing predicate instead of a passing a Set, because I can't get how we can pass collection using functional interface that returns boolean.

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
AleksandrPea: What do you mean under undoable edits? When user type something but doesn't save it?

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
AleksandrPea: So, if I understand correctly there is no problem in using `ZapTextArea`, but I need to include interpreting its text as regexp, don't I?

src/org/zaproxy/zap/spider/URLCanonicalizer.java
thc202: You pass the predicate, which makes use of the `Set`, e.g. `param -> mySet.contains(param)` (I'd expect the predicate to also make use of the `ExtensionHttpSessions` though).

thc202: (Also, it can be shared by all `SpiderParser`s.)

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
thc202: Yeah, `ZapTextArea` keeps track of edits that can be undone, those should be discarded when the dialogue is closed and when initialising.

src/org/zaproxy/zap/spider/URLCanonicalizer.java
AleksandrPea: So, do I need remove checking for sessionToken totally including hardcoded values which I've already deleted?

thc202: Yeah, "older" methods should use the old values (it can be used a default predicate with those though).

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
thc202: Technically there's no problem but it would be better (usability) if we didn't. @psiinon @kingthorin wdyt?

We can also handle the UI changes.

psiinon: Agree that we should really be using the newer control which allows the user to add, change and delete items. But if that does look too hard then as @thc202 said we can make that change.

AleksandrPea: Ok, I've understood. Please, let me do it myself)

src/org/zaproxy/zap/spider/URLCanonicalizer.java
AleksandrPea: Yeah, great idea - I'll make changes. Thanks!

kingthorin:
@AleksandrPea are you able to address @thc202's feedback and fix the conflict?

AleksandrPea:
@kingthorin , probably yes. However, I got stuck doing UI part: you have already created the same component which I need for this feature, but it was constructed in such way that doesn't allow to be easy extended(from my point of view). I tried to generalize it but it requires many code changes which incredibly reduces readability and increases complexity.
@thc202 @psiinon, I apologize for my overconfidence. Could you help me with UI?

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
kingthorin: @AleksandrPea further changes coming?

kingthorin:
Anyone got advice for @AleksandrPea ?

thc202:
IMO it's better to use what's currently in place as it allows to implement this feature already. It might not be optimal and lead to some code duplication but refactoring that is not in the scope of this PR. Anyway, could you provide more details about the extensibility issues?

kingthorin:
@AleksandrPea any further input/changes coming?

AleksandrPea:
@kingthorin sorry, I'm not able to end up with this.

thc202:
No problem, we'll tackle the remaining changes. Thanks for the contribution! How would you like to be credited?

AleksandrPea:
@thc202 , my full name is Oleksandr Horokh. Thank you!

thc202:
The credits are now updated, thank you! :)

thc202:

thc202:

thc202:
Ready to review, all comments addressed except the one mentioned above.

src/org/zaproxy/zap/extension/spider/OptionsSpiderPanel.java
thc202: Being done in zaproxy/zap-extensions#3991.

psiinon (approved):

PR #4402, opened 2018-02-15
Add option to show request/response in same tab

### LINKED ISSUE #4331: Allow to show Request and Response in the same tab
Hi, 

I think would be great if requester and request tab could share the same response tab. For laptops with a small screen, I have to change from normal layout to full layout when I'm working with the requester.

Best regards,

RLobo

thc202:
Add an option to toolbar (and entry in Options > Display), to show the
Request and Response panels in the same tab.

Fix #4331 - Allow to show Request and Response in the same tab

Tasks:
 - [x] Enable in Full Layout.
 - [x] Move icon position, to be consistent with other tabs.


thc202:
Still WIP (e.g. shortcut to select or focus the request tab are not working properly) but good enough for feedback about the L&F.

juhakivekas:
This looks great. I like that doubleclicking to fullscreen works even if full layout isn't supported. Tiny detail: all other tabs (except request) have the icon on the left side of the tab name, looks little inconsistent to my eye.

thc202:
Right, we could enable this (along with "Tabs Side by Side") for full layout, the current ones were disable because they didn't make sense in full layout. Agreed, the icon in that position does not look good.

thc202:
And thanks for feedback :)

thc202:
Conflict resolved. It's now ready for review.

juhakivekas:
Sweet :candy:

sbzo:
Thanks, guys!!!! I will wait for the weekly release.

PR #4409, opened 2018-02-17
FIX - Issue #4277 :Warn if a # is specified in a break point URL

harjas27:


kingthorin:
We now have two PRs for the same change:
https://github.com/zaproxy/zaproxy/pull/4349

thc202:
@harjas27 thanks for the contribution, but as mentioned in the previous comment the issue is already being addressed in the other pull request. Let us know if there's other issue you would like to work on.

PR #4415, opened 2018-02-19
Added the ID for the NoSQL add-on and a new constructor for the VariantAbstractRPCQuery class

LuigiCasciaro:


src/org/parosproxy/paros/core/scanner/NameValuePair.java
```diff
@@ -57,6 +58,13 @@
      * @since TODO add version
      */
     public static final int TYPE_MULTIPART_DATA_FILE_CONTENTTYPE = 36;
+    /**
+     * The content-type portion of a application/json file parameter
```
thc202: Might need a tweak, something like `An application/json parameter.`?

thc202:
Thanks!

kingthorin:
Sorry, I should have squashed when merging. I noticed the second commit too late :(

thc202:
No worries :)

PR #4430, opened 2018-02-23
Allow to add AbstractPanel tabs with custom index

thc202:
Change TabbedPanel2 to allow to add an AbstractPanel tab with a custom
index, to add/remove the tab and keep it always in the same position.


kingthorin:
Oh no, my script submission made you do changes....

PR #4432, opened 2018-02-23
Allow to get the current parameter in IV scripts

thc202:
Change VariantCustom to allow Input Vector scripts to get the current
parameter being tested so that they can reliably know which parameter is
being tested (instead using custom names or the name being injected,
which might not be the same as the current parameter).
Change Input Vector default template.js to make use of the new method.


thc202:
(Related to zaproxy/community-scripts/pull/65.)

PR #4440, opened 2018-02-26
Added method to unregister event consumer from all publishers

psiinon:
Also added some basic unit tests.

### DIFF
--- a/src/org/zaproxy/zap/eventBus/EventBus.java
+++ b/src/org/zaproxy/zap/eventBus/EventBus.java
@@ -56,7 +56,14 @@ public interface EventBus {
 	void registerConsumer (EventConsumer consumer, String publisherName, String[] eventTypes);
 	
 	/**
-	 * Unregister the consumer
+	 * Unregister the consumer from all publishers
+	 * @param consumer the consumer of events
+	 * @since TODO add version
+	 */
+	void unregisterConsumer(EventConsumer consumer);
+	
+	/**
+	 * Unregister the consumer from the specified publisher
 	 * @param consumer the consumer of events
 	 * @param publisherName the name of the publisher
 	 */
--- a/src/org/zaproxy/zap/eventBus/SimpleEventBus.java
+++ b/src/org/zaproxy/zap/eventBus/SimpleEventBus.java
@@ -3,8 +3,10 @@ package org.zaproxy.zap.eventBus;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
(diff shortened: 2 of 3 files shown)


src/org/zaproxy/zap/eventBus/EventBus.java
```diff
@@ -56,12 +56,18 @@
 	void registerConsumer (EventConsumer consumer, String publisherName, String[] eventTypes);
 	
 	/**
-	 * Unregister the consumer
+	 * Unregister the consumer from all publishers
 	 * @param consumer the consumer of events
-	 * @param publisherName the name of the publisher
 	 */
-	void unregisterConsumer(EventConsumer consumer, String publisherName);
+	void unregisterConsumer(EventConsumer consumer);
```
thc202: Could have a since tag.

src/org/zaproxy/zap/eventBus/EventBus.java
```diff
@@ -56,12 +56,18 @@
...
 	
+    /**
+     * Unregister the consumer from the specified publisher
+     * @param consumer the consumer of events
+     * @param publisherName the name of the publisher
+     */
+    void unregisterConsumer(EventConsumer consumer, String publisherName);
```
thc202: Needs indentation tweaks.

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -127,31 +128,54 @@ public void registerConsumer(EventConsumer consumer, String publisherName,
...
-			if (publisher != null) {
-				publisher.removeComsumer(consumer);
-			} else {
-				// Check to see if its cached waiting for the publisher
-				for (RegisteredConsumer regCon : this.danglingConsumers) {
-					if (regCon.getConsumer().equals(consumer)) {
-						this.danglingConsumers.remove(regCon);
-						break;
-					}
+			log.debug("unregisterConsumer " + consumer.getClass().getCanonicalName());
```
kingthorin: Check `isDebugEnabled()`?

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -127,31 +128,54 @@ public void registerConsumer(EventConsumer consumer, String publisherName,
...
+			for (Entry<String, RegisteredPublisher> entry : nameToPublisher.entrySet()) {
+			    entry.getValue().removeComsumer(consumer);
+			}
+			// Check to see if its cached waiting for a publisher
+			for (RegisteredConsumer regCon : this.danglingConsumers) {
+				if (regCon.getConsumer().equals(consumer)) {
+					this.danglingConsumers.remove(regCon);
+					break;
```
kingthorin: There can only be one? (below too)

psiinon: That was copied from the existing code :) I guess in theory a consumer could get added more than once, but it def wouldnt be good practice as they'd then get multiple events. @thc202 views?

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
thc202: It does avoid the creation of the string (less garbage mainly), here does not make much difference since this will not be not called that often I guess.

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
thc202: Yeah, that sounds like a bug, we could throw when the same consumer is registered more than once for the same publisher?

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -127,31 +128,54 @@ public void registerConsumer(EventConsumer consumer, String publisherName,
...
-					if (regCon.getConsumer().equals(consumer)) {
-						this.danglingConsumers.remove(regCon);
-						break;
-					}
+			log.debug("unregisterConsumer " + consumer.getClass().getCanonicalName());
+			for (Entry<String, RegisteredPublisher> entry : nameToPublisher.entrySet()) {
+			    entry.getValue().removeComsumer(consumer);
+			}
+			// Check to see if its cached waiting for a publisher
+			for (RegisteredConsumer regCon : this.danglingConsumers) {
```
thc202: We should extract a method for this.

src/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -127,31 +128,54 @@ public void registerConsumer(EventConsumer consumer, String publisherName,
...
+					break;
 				}
 			}
 		} finally {
 			regMgmtLock.unlock();
 		}
 	}
 
+    @Override
+    public void unregisterConsumer(EventConsumer consumer, String publisherName) {
```
thc202: This (the whole method) could keep the same indentation char.

psiinon:
Must admit I'd forgotten about this :/
Now updated.

PR #4441, opened 2018-02-26
Validate i18n time stamp patterns

thc202:
Add test to check that the internationalised time stamp patterns are
valid.

---
The build will fail, some of the patterns are not correct and others have the char `'` incorrectly escaped. Plan to fix that in Crowdin.

### DIFF
--- a/test/org/zaproxy/zap/utils/TimeStampLocaleUnitTest.java
+++ b/test/org/zaproxy/zap/utils/TimeStampLocaleUnitTest.java
@@ -1,0 +1,119 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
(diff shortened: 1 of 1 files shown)


thc202:
It seems the right place for that, not all patterns will be the same for all locales (e.g. `DD-MM` vs `MM-DD`). I think we could hardcode the ISO 8601 pattern as that's the same for all of them?

kingthorin:
Ok I'll put that on my todo list :zap:

thc202:
OK, thanks.

thc202:
The incorrect patterns are now fixed in Crowdin.

psiinon:
Unit tests still failing.

thc202:
Now passing (needed a rebase to pick latest Crowdin changes).

PR #4449, opened 2018-02-28
Reload scripts if they have changed on disk

### LINKED ISSUE #2950: Script change monitor
When a script is modified outside ZAP, it does not recognize the the update.

psiinon:
Fixes #2950
This will not play well with any scripts being displayed in the script console, so that will need to be changed as well. I'll raise another bug for that.

### DIFF
--- a/src/org/zaproxy/zap/extension/script/ScriptWrapper.java
+++ b/src/org/zaproxy/zap/extension/script/ScriptWrapper.java
@@ -20,9 +20,13 @@
 
 package org.zaproxy.zap.extension.script;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.io.Writer;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
 
 import javax.script.ScriptException;
 
@@ -52,9 +56,11 @@ public class ScriptWrapper {
 	private boolean error = false;
 	private boolean loadOnStart = false;
 	private File file;
+	private long lastModified;
 	private String lastErrorDetails = "";
 	private Exception lastException = null;
 	private Writer writer = null;
+	private Charset charset = ExtensionScript.DEFAULT_CHARSET;
 	
 	public ScriptWrapper() {
 	}
@@ -77,7 +83,7 @@ public class ScriptWrapper {
 		this.engineName = engineName;
 		this.type = type;
 		this.enabled = enabled;
-		this.file = file;
+		this.setFile(file);
 	}
 	
 	public ScriptWrapper(String name, String description,
@@ -91,7 +97,7 @@ public class ScriptWrapper {
 		}
 		this.type = type;
 		this.enabled = enabled;
-		this.file = file;
+		this.setFile(file);
 	}
 
 	protected ScriptWrapper(String name, String description, String engineName, String typeName, boolean enabled, File file) {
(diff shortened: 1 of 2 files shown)


juhakivekas:
As long as I'm not actively editing the same script in vim and the console this works just as I'd expect :+1:

kingthorin:
Ok, how would you expect that situation to pan out, because it'll definitely come up.

juhakivekas:
I'd expect the most recently edited version to always be used, regardless of what editor the newest version was made with. If I haven't edited the console contents, then I expect the version from disk to be ran. Currently this is not always the case:

- run a standalone script `print('foo')` and save it
- run the script with the UI, `foo` will be printed
- edit and save in vim `print('bar')`
- run the script with the UI, `bar` will be printed
- run the script again, `foo` will be printed. The console version will be ran even if the script hasn't been edited in the console

This is not the behaviour in enabled/disabled type scripts, so I assume Run forces a read of the editor contents.

I'd also expect the script UI to either update editor contents when a new version has been loaded or to show an indication that it is not the version currently in use by eg. removing colours or fading the panel.

As psiinon mentioned in #4450 prompting users in conflicting cases could be done. If the console and disk version have been updated since last run, then ask what to use and keep. At best this will inform users to stick to one editor. For enable/disable type scripts this might be a bit of an annoyance as the prompts will come without interaction. Again, one of the best outcomes of this is to make users stick to one-script-one-editor even quicker.

src/org/zaproxy/zap/extension/script/ScriptWrapper.java
```diff
@@ -102,6 +108,7 @@ protected ScriptWrapper(String name, String description, String engineName, Stri
 		this.typeName = typeName;
 		this.enabled = enabled;
 		this.file = file;
```
thc202: Remove?

src/org/zaproxy/zap/extension/script/ScriptWrapper.java
```diff
@@ -292,4 +302,55 @@ void setPreviouslyEnabled(boolean enabled) {
...
+	/**
+	 * Saves the latest contents into the configured file
+	 * @throws IOException
+	 * @since TODO add version
+	 */
+	void saveScript() throws IOException {
+		try ( BufferedWriter bw = Files.newBufferedWriter(file.toPath(), ExtensionScript.DEFAULT_CHARSET)) {
+			bw.append(getContents());
+		}
+		this.lastModified = file.lastModified();
```
thc202: Should update the `charset` too (to the default).

psiinon:
OK, have tweaked this based on the related implementation in the script console: https://github.com/zaproxy/zap-extensions/pull/1441
Hopefully it all not works as expected?

src/org/zaproxy/zap/extension/script/ScriptWrapper.java
```diff
@@ -292,4 +301,55 @@ void setPreviouslyEnabled(boolean enabled) {
...
+	}
+	
+	/**
+	 * Saves the latest contents into the configured file
+	 * @throws IOException
+	 * @since TODO add version
+	 */
+	void saveScript() throws IOException {
+		try ( BufferedWriter bw = Files.newBufferedWriter(file.toPath(), charset)) {
```
thc202: I liked the previous behaviour more :) We should save with the default always because we will try to read with that in the next ZAP run. We just need to update it for following (re)loads.

psiinon: Updated

thc202:
Thanks!

thc202:
Thanks!

PR #4438, opened 2018-02-25
Allow to select the look and feel

nhegde610:
#2964 
These are the changes I have done. I would like to get some feedback on it. Should I add support for more look and feel ?

PR #4461, opened 2018-03-01
Remove unnecessary resource message

thc202:
Remove a resource message that was not being shown in the UI (it was
only used internally in a UI panel). Also, tweak the panels that were
using it to not require a name.


thc202:
(Can wait for #4438, which might lead to conflicts.)

PR #4471, opened 2018-03-01
Update Webswing to 2.5 in Docker images

### LINKED ISSUE #3553: Update Webswing in Docker images
Update Webswing to version 2.5 to address the issues:
 - https://bitbucket.org/meszarv/webswing/issues/97 (fixed in 2.4, but following is a blocker)
 - https://bitbucket.org/meszarv/webswing/issues/214

that prevent some of the dialogues from being used (e.g. in Options > API, in Session Properties...).

(Pending release of version 2.5.)

thc202:
Update Docker build files to download Webswing 2.5, remove its demos,
and use a directory without version for Webswing (less changes required
when updating).
Update Webswing configuration to 2.5, ZAP will now be started/available,
for example, under:
  `http://localhost:8080/zap/`
instead of:
  `http://localhost:8080/?anonym=true&app=ZAP`

ZAP will still be accessed with an anonymous account, no other account
will be available.
Update Webswing start up script, zap-webswing.sh, to use the new  dir
and to not pick the user.properties file, there's none to.

Fix #3553 - Update Webswing in Docker images

### DIFF
--- a/build/docker/Dockerfile-live
+++ b/build/docker/Dockerfile-live
@@ -64,9 +64,14 @@ RUN git clone --depth 1 https://github.com/zaproxy/zaproxy.git && \
 	cp -R /zap-src/zaproxy/build/zap/* /zap/ && \
 	rm -rf /zap-src/* && \
 	cd /zap/ && \
-	curl -s -L https://bitbucket.org/meszarv/webswing/downloads/webswing-2.3-distribution.zip > webswing.zip && \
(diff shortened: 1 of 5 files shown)


thc202:
Tested stable and weekly images and seem to be working as expected, still pending live image (takes a lot of time to build).

thc202:
For the record, live image is also working fine.

PR #4477, opened 2018-03-02
Enable help in Add/Edit Break Point dialogue

thc202:
Change BreakAddEditDialog to declare the help index, to have the
corresponding help page accessible directly from the dialogue.


thc202:
Will also open a PR to update the help of the dialogue.

PR #4489, opened 2018-03-07
Move docker dir to a top level dir

thc202:
Move the docker dir from build dir to a top level directory, to leave
the build dir just for build (transient) files/data, while the build dir
is not removed with the current build tool it will with the new one.

Needs update:
 - [x] `hub.docker.com/r/owasp/zap2docker-live/`;
 - [ ] `zap-mgmt-scripts` (zapbot/zap-mgmt-scripts#14);
 - [x] `zaproxy` (wiki);
 - [ ] `zaproxy.blogspot.com`;
 - [ ] ?

Let me know if it's missing anything, plan to update them when merged (except for the first which should be done just before merging, ideally).


kingthorin:
Is this safe to merge or does it need to wait for the checklist items?

thc202:
Yeah, let me prepare the changes in the other places first.

thc202:
It can be merged now, main updates done.

thc202:
All tasks done. Let me know if you spot anything  that still needs to be updated.

Issue #4450, opened 2018-02-28
Script Console should handle scripts changed on disk


psiinon:
Related to https://github.com/zaproxy/zaproxy/issues/2950
If https://github.com/zaproxy/zaproxy/pull/4449 is merged then the Script Console will need to change to handle scripts that have changed on disk.
The code also needs to handle scripts that have been changed but not saved - these might not be the currently displayed script.
The script currently being displayed should also be handled if its changed on disk even if its not been changed via the console,
In all cases the user should be prompted to ask if they want to change to see the new contents, and informed if any changes they made via the console will be lost.

thc202:
Being handled in zaproxy/zap-extensions#1441.

thc202:
Fixed in zaproxy/zap-extensions#1441.

PR #4493, opened 2018-03-08
Reload script of changed on disk in getScript()

psiinon:


denniskniep:
It would be helpful to know why you changed this :) . Was there a concrete problem?
Is it correct that the method `List<ScriptWrapper> getScripts(ScriptType type)` does not need reload?

thc202:
> Should we raise an issue or is this expected/known?

I'm not sure how you disabled the file watcher(?) so depending on what changes you did that might be expected/known. If the user does not decide which changes it wants ZAP defaults to what's in the Script Console tab (instead of the file), this is part of #4449 and #4450. If you don't change the script in ZAP then ZAP will keep using the latest changes from the file.

> Was there a concrete problem?

This is to ensure that the script contains the latest changes from the file when obtained from the extension, but maybe it should be up to the caller to refresh it before use(? maybe a method in `ScriptWrapper`).

thc202:
Yeah, that's also known (and yes, it's "odd" to execute once with the new script), we did not see the need to change that though, as no data is lost (file or Script Console) and that the user would use ZAP or the editor, not both (so this would not happen that often but still catch mistakes when editing in both places).

PR #4502, opened 2018-03-13
Add more events and data

psiinon:


src/org/parosproxy/paros/extension/history/ProxyListenerLogEventPublisher.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2017 The ZAP Development Team
```
kingthorin: 2018

src/org/parosproxy/paros/extension/history/ProxyListenerLog.java
```diff
@@ -144,6 +144,8 @@ private void createAndAddMessage(HttpMessage msg, int type) {
         extension.addHistory(historyRef);
 
         addToSiteMap(historyRef, msg);
+        
+        ProxyListenerLogEventPublisher.getPublisher().publishHrefAddedEvent(historyRef);
```
thc202: This change requires a ZAP comment. I was wondering if we could use `HistoryReferenceEventPublisher` instead? Too generic maybe?

src/org/parosproxy/paros/extension/history/ProxyListenerLogEventPublisher.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.parosproxy.paros.model.HistoryReference;
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.eventBus.Event;
+import org.zaproxy.zap.eventBus.EventPublisher;
+
+/**
+ * A {@link EventPublisher} of {@link ProxyListenerLog} events.
+ * <p>
+ * Notifies of changes in the {@code HistoryReference}'s state/data.
```
thc202: This needs update.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -284,6 +284,14 @@ private void publishAlertEvent(Alert alert, String event) {
...
         map.put(AlertEventPublisher.ALERT_ID, Integer.toString(alert.getAlertId()));
         map.put(AlertEventPublisher.HISTORY_REFERENCE_ID, Integer.toString(alert.getSourceHistoryId()));
+        map.put(AlertEventPublisher.NAME, alert.getName());
+        map.put(AlertEventPublisher.URI, alert.getUri().toString());
+        map.put(AlertEventPublisher.PARAM, alert.getParam());
+        map.put(AlertEventPublisher.RISK, Integer.toString(alert.getRisk()));
+        map.put(AlertEventPublisher.RISK_STRING, Alert.MSG_RISK[alert.getRisk()]);
+        map.put(AlertEventPublisher.CONFIDENCE, Integer.toString(alert.getConfidence()));
+        map.put(AlertEventPublisher.CONFIDENCE_STRING, Alert.MSG_CONFIDENCE[alert.getConfidence()]);
+        map.put(AlertEventPublisher.SOURCE, alert.getSource().getI18nKey());
```
thc202: I'd suggest returning the ID instead, the key is more likely to change than the ID.

src/org/zaproxy/zap/extension/brk/BreakEventPublisher.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public void publishInactiveEvent(Message msg) {
+        this.publishEvent(BREAK_POINT_INACTIVE, msg, Collections.emptyMap());
+    }
+
+    private void publishEvent(String event, Message msg, Map<String, String> parameters) {
+        Map<String, String> map = new HashMap<String, String>();
+        map.putAll(parameters); // Could be an empty map
+        map.put(MESSAGE_CLASS, msg.getClass().getCanonicalName());
```
thc202: Maybe we should use a specific ID/token (e.g. "HTTP", "WebSocket") instead of the class name?

src/org/parosproxy/paros/extension/history/ProxyListenerLog.java
psiinon: I tried that, and it generated way too many events. This is more for someone trying to reproduce the History list, which is why I went for another publisher. But happy to reconsider...

src/org/zaproxy/zap/extension/brk/BreakEventPublisher.java
psiinon: I went for the class as until we can update the websocket add-on with post 2.7.0 changes it wont easily be able to supply this. However I could add a new Message method and default that to the class name? I prefer the idea of a more easily readable token..

thc202: Adding a new method to `Message` sounds good to me.

src/org/parosproxy/paros/extension/history/ProxyListenerLog.java
thc202: Right, fine as is :)

psiinon:
Updated based on the comments.

thc202:
Thanks!

PR #4510, opened 2018-03-15
Update ID 40033 for MongoDB (NoSQL Injection)

LuigiCasciaro:


src/org/zaproxy/zap/model/Tech.java
```diff
@@ -160,4 +160,4 @@ public int compareTo(Tech o) {
         
         return this.toString().compareTo(o.toString());
     }
-}
+}
```
kingthorin: Could you revert this or exclude this file?

thc202:
@LuigiCasciaro note that the commits are using different accounts (or not tied to your own).

src/org/zaproxy/zap/model/Tech.java
LuigiCasciaro: Of course, but it's strange because it does not change anything between the red and green lines.

thc202: It removed a newline.

thc202:
Thanks!

PR #4511, opened 2018-03-15
Remove AbstractPanel from internal list, always

thc202:
Change `TabbedPanel2` to override `removeTabAt(int)` and remove the
`AbstractPanel` from the internal list in this method (instead of in
`removeTab(AbstractPanel)`) to ensure the panels are removed always,
thus avoid leaking them. The method `removeTabAt(int)` ends up being
called by `removeTab(AbstractPanel))`, `remove(Component)`, and
`remove(int)`.


thc202:
(Per comments in zaproxy/zap-extensions#1486.)

PR #4516, opened 2018-03-17
Fixes #4515; AuthenticationScript (Python, Ruby, Groovy) does not work

### LINKED ISSUE #4515: AuthenticationScript (Python, Ruby, Groovy) does not work
Creating a new AuthenticationScript (Python, Ruby or Groovy) that misses the Methods `getLoggedInIndicator` and `getLoggedOutIndicator`. Fails when it gets loaded. 
Because it tries to first load `AuthenticationScriptV2`. Some ScriptEngines do not verify if all Interface Methods are contained in the script. Some ScriptEngines (Python, Ruby, Groovy) loads successfully AuthenticationScriptV2. But without the methods `getLoggedInIndicator()` `getLoggedOutIndicator()`.
Than an exception occures when the methods are invoked. Exception depends on the ScriptEngine but something like `MethodNotFoundException`

The ScriptEngine JavaScript has no problems, because it does some verification in the `getInterface()` Method. That kind of verification are not implemented by the other ScriptEngines.

denniskniep:


src/org/zaproxy/zap/authentication/ScriptBasedAuthenticationMethodType.java
```diff
@@ -452,7 +453,8 @@ private void loadScript(ScriptWrapper scriptW, boolean adaptOldValues) {
 				getScriptsExtension().handleScriptException(scriptW, e);
 				log.error("Error while calling authentication script", e);
 				warnAndResetPanel(Constant.messages.getString(
-						"authentication.method.script.dialog.error.text.loading", e.getMessage()));
+						"authentication.method.script.dialog.error.text.loading",
+						ThrowableUtils.getFirstNotNullMessageThroughRootCauses(e)));
```
thc202: Use `ExceptionUtils.getRootCauseMessage` instead.

denniskniep: Thanks :)

thc202:
Thanks!

Issue #2033, opened 2015-11-06
Generate swagger / openapis spec for ZAP API


psiinon:
see http://swagger.io/ and https://openapis.org/

SkypLabs:
Hi,

I was thinking about writing an OpenAPI definition from scratch for ZAP's own web API. After some research on a precedent initiative, I stubbled upon this GH issue but I am not sure to fully understand it. Am I wrong to say that the efforts tracked by this issue aim to generate an OpenAPI definition directly from the implementation?

Many thanks.

thc202:
That's the intention, and it's already being done to some extent: https://github.com/zaproxy/zap-api-docs/blob/master/openapi.yaml

PR #4520, opened 2018-03-19
Added swagger file with some operations

omerlh:
I've created a basic swagger file with 3 operations I needed for Glue. 
What left?
- [ ] Added swagger linter on CI (For example, [this](https://www.npmjs.com/package/swagger-cli) is a good option
- [ ] Publish the swagger somewhere
- [ ] Acceptance tests for the API (maybe using [this](https://apigee.com/about/blog/api-technology/swagger-test-templates-test-your-apis))
- [ ] Generate ruby client

#2033

### DIFF
--- a/swagger/swagger.yaml
+++ b/swagger/swagger.yaml
@@ -1,0 +1,155 @@
+swagger: "2.0"
+info:
+  description: "REST API for Zaproxy. The API let you control all the functionality of Zaproxy, and ease the process of integrating Zaproxy into CI."
+  version: "1.0.0"
+  title: "OWASP Zaproxy REST API"
+  termsOfService: "http://swagger.io/terms/"
+  contact:
+    email: "apiteam@swagger.io"
+  license:
+    name: "Apache 2.0"
+    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
+basePath: "/JSON"
+tags:
+- name: "pscan"
+  description: "Passive Scan related functionality"
+  externalDocs:
+    description: "Find out more on the wiki"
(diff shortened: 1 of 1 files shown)


thc202:
What's the main purpose? Generate clients? Document the API?

omerlh:
My motivation is a Ruby client, the code on Glue use Curl which kinda works as I find while writing tests for it. It will also helps with documentation, of course.

psiinon:
All of the other API clients use code generation.
I'd recommend either generating the Ruby client in the same way or generating the swagger file.
The current generators are all in https://github.com/zaproxy/zaproxy/tree/develop/src/org/zaproxy/zap/extension/api
Please dont do it by hand :)

omerlh:
I am familiar with both options. Generating the swagger automatically is not a simple task, same as creating ruby client generator. I was thinking to create it manually as a step forward automatic generation of the swagger.

psiinon:
How about discussing this on the dev list first?
I'm not sure I like the idea of PRs just raised for the purposes of discussion. I'm concerned that they will make it harder for us to review PRs.
I also wouldnt want us to merge a PR that could require us to manually maintain a swagger spec going forwards.
Other thoughts appreciated...

kingthorin:
I agree.
https://groups.google.com/forum/#!forum/zaproxy-develop

psiinon:
Have chatted about this with @thc202 as well, and it looks like we're all in agreement.
PRs should be, in the view of their authors, ready to merge.
Happy for you to create a tracker issue and/or develop this in your own fork, but I dont think we should have incomplete tracker PRs.

PR #4552, opened 2018-03-28
Log target in Docker API scans

thc202:
Change zap-api-scan.py to log the target when obtained from ZAP (after
importing from file) and when normalising it (strip the path component),
to help know what target is the script really going to use.


thc202:
This is related to a thread in OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/nsjH7hEThh4/discussion

PR #4554, opened 2018-03-28
Do not try to scan without URLs in Docker API scan

thc202:
Change zap-api-scan.py to terminate sooner if no URLs were successfully
imported into ZAP, attempting to continue and scan would lead to more
"cryptic" errors in ZAP (as no valid URL would be used) and in the
script itself (when attempting to use the invalid scan ID).

---
Related to thread in OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/nsjH7hEThh4/discussion

### DIFF
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -61,6 +61,10 @@ from zapv2 import ZAPv2
 from zap_common import *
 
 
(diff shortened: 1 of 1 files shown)


psiinon:
Looks good but now conflicting.

PR #4562, opened 2018-04-02
Persist copy of proxied message

thc202:
Change ProxyListenerLog to persist a copy of the message forward to the
client to avoid following listeners to change it, following listeners
are not expected to change the message to be persisted, only the one
being forwarded.


thc202:
Tweaked title to reflect better the problem.

PR #4565, opened 2018-04-03
Fix main layouts

thc202:
Change WorkbenchPanel to hide the panels instead of relying on the
behaviour of the UI components (which removes the panel from the current
parent before adding it to itself), as that would end up removing them
from the internal list of panels of the TabbedPanel2 (per changes done
in #4511).
Change TabbedPanel2 to allow to hide all the tabs and add a panel hidden
(adds to the internal list without displaying the tab), to support the
above case.

### DIFF
--- a/src/org/zaproxy/zap/view/TabbedPanel2.java
+++ b/src/org/zaproxy/zap/view/TabbedPanel2.java
@@ -244,14 +244,33 @@ public class TabbedPanel2 extends TabbedPanel {
 					this.addTab(c.getName(), null, c, true, true, index);
 				}
 				this.hiddenTabs.remove(c);
+				handleHiddenTabListTab();
 			}
-			
-		} else {
-			if (! this.hiddenTabs.contains(c)) {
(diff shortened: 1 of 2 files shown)


src/org/zaproxy/zap/view/TabbedPanel2.java
```diff
@@ -277,6 +296,21 @@ public void addTab(AbstractPanel panel) {
 	    addTab(panel, panel.getTabIndex());
 	}
 
+	/**
+	 * Adds the given panel, whose tab will be hidden.
+	 * <p>
+	 * This method effectively overrides the the {@link AbstractPanel#isHideable() hideable} and {@link AbstractPanel#isPinned()
+	 * pinned} state of the panel.
```
kingthorin: There's a double "the" before the link.

thc202: Fixed, thank you.

PR #4569, opened 2018-04-04
Move ISO8601 Timestamp format out of properties

kingthorin:
In follow-up to: https://github.com/zaproxy/zaproxy/pull/4441

Also simplify things by removing `timestamp.format.default` and simply using `timestamp.format.datetime` in those instances.

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -2310,9 +2310,7 @@ technologies.os = OS
 technologies.scm = SCM
 technologies.ws = WS
 
-timestamp.format.default = yyyy-MM-dd HH:mm:ss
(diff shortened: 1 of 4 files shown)


thc202:
LGTM, but was wondering if we could also merge "default" and "datetime", do you recall why they are separate? I don't... :/

thc202:
I think we are already handling the malformed issue with `SAFE_TIME_STAMP_FORMAT`. So, we could remove `timestamp.format.default` and use `timestamp.format.datetime` instead?

kingthorin:
Sure, change inbound.

thc202:
Thanks!

Issue #3485, opened 2017-05-04
Minor Enhancement: Indicate amount of Java memory actually used in JVM Options Panel


davewichers:
I've run out of Java Heap space when running ZAP for multiple days on a large site. So, I need to increase the amount of memory I provide to ZAP when I run it. Problem is that I don't know how much it's using currently.

It would be helpful if the Options-->JVM Panel displayed all of the current Java memory options to help guide the user as to what to increase them to. And to display them in the way the user would need to set them in Java. For example:

Current Java Max Heap Size Set to:    -Xmx1500m
Current Java Initial Heap Size Set to:  -Xms256m (or "default for Java") if not set.

Displaying this information this way is helpful in 2 ways:
1) They know what its currently set to.
2) They know the syntax for changing it if they want to.

I see that it does display any new settings that are set, it just doesn't display the default values.

Not required, but if you want to really be clever you could remember that it was changed, but ZAP hasn't been restarted yet, and indicate something like: Pending Change: list whatever it is (ZAP must be restarted for this to take affect)

If its easy, it might also be helpful to display the total amount of memory available on the machine on the same panel.

PR #4572, opened 2018-04-05
Display Memory Info on JVM Options Panel

kingthorin:
OptionsJvmPanel > Updated with necessary ZapLabels and getters to display the headings and human readable (eg: MiB/GiB) values. Values (labels) are updated by calling update methods in the overridden initParam function.
SizeBytesStringValue > Added supplemental constructor that accepts a boolean for the useJustBytesUnit flag. (Thus providing a means of creating and simultaneously setting the flag, vs. using 2 steps in
calling code.)
Messages.properties > Updated with the necessary supporting keys and
messages.

Fixes zaproxy/zaproxy#3485

### DIFF
--- a/src/org/parosproxy/paros/extension/option/OptionsJvmPanel.java
+++ b/src/org/parosproxy/paros/extension/option/OptionsJvmPanel.java
@@ -29,12 +29,16 @@ import java.util.List;
 
 import javax.swing.JLabel;
 import javax.swing.JPanel;
+import javax.swing.JSeparator;
+import javax.swing.SwingConstants;
 
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.view.AbstractParamPanel;
 import org.parosproxy.paros.view.View;
+import org.zaproxy.zap.utils.ZapLabel;
 import org.zaproxy.zap.utils.ZapTextField;
 import org.zaproxy.zap.view.LayoutHelper;
+import org.zaproxy.zap.view.renderer.SizeBytesStringValue;
 
 /**
  * The JVM options panel.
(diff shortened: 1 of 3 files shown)


kingthorin:
Related to: https://github.com/zaproxy/zaproxy/issues/3485

I can add a fix tag after someone has had a peak, I wasn't fully comfortable with the implementation.

kingthorin:
Ok this one is a bit better, I generalized the update function. I still don't like that I have to pass the Constant strings twice, but maybe I'm over thinking it....

src/org/parosproxy/paros/extension/option/OptionsJvmPanel.java
```diff
@@ -93,8 +109,45 @@ private ZapTextField getJvmOptionsField() {
...
+	private JLabel getMaxMemoryLabel() {
+		if (maxMemoryLabel == null) {
+			maxMemoryLabel = new JLabel();
+			updateMemoryLabel(maxMemoryLabel, Constant.messages.getString("jvm.options.memory.max",
+					sbsv.getString(Runtime.getRuntime().maxMemory())));
+		}
+		return maxMemoryLabel;
+	}
+	
+	private void updateMemoryLabel(JLabel labelToUpdate, String text) {
```
thc202: Instead of text it could require the key and the value, so in this method it would do:
`labelToUpdate.setText(Constant.messages.getString(key, sbsv.getString(value)))`

kingthorin:
Updated.
image

src/org/parosproxy/paros/extension/option/OptionsJvmPanel.java
```diff
@@ -93,8 +109,41 @@ private ZapTextField getJvmOptionsField() {
 		return jvmOptionsField;
 	}
 	
+	private JLabel getSizeMemoryLabel() {
+		if (sizeMemoryLabel == null) {
+			sizeMemoryLabel = new JLabel();
+			updateMemoryLabel(sizeMemoryLabel, "jvm.options.memory.size", Runtime.getRuntime().totalMemory());
```
thc202: I guess we could remove these, `initParam` is always called before showing the panel, wdyt?

kingthorin: Sure. I’ll remove the updates from the getters and put a comment leading the updates in init so that it’s obvious for the next guy 🙂

thc202: btw, does it worth using `ZapLabel`s instead? (to allow select and copy)

kingthorin: Sure, will tweak that too.

thc202:
Thank you!

PR #4574, opened 2018-04-06
Latest files from Crowdin

zapbot:
Autogenerated PR from zapbot


src/lang/Messages_bs_BA.properties
```diff
@@ -872,7 +872,7 @@ context.list.table.name=Kontekst
 context.list.table.inscope=U Opsegu
 context.name.prefix.label=Kontekst\:
 context.new.title=Novi kontekst
-context.prefixName=Context\: {0}
+context.prefixName=Komentar\: {0}
```
thc202: Typo? Based on the above it should be `Kontekst`.

kingthorin: Agreed

thc202: Added a comment in Crowdin.

PR #4581, opened 2018-04-09
Skip Access Control responses for HTTP Sessions

thc202:
Change class ExtensionHttpSessions to skip/ignore the responses from
Access Control, as with other automated responses they would end up
flood the HTTP Sessions tab with a lot of unnecessary sessions (e.g.
when testing with anonymous).


psiinon:
looks good but now conflicting

thc202:
Addressed.

Issue #4123, opened 2017-12-06
Passive Scan Tags interpolation


PenelopeFudd:
I've created a Passive Scan Tag that matches a request header, and I'd like to use part of the value as a tag.

The Request Header Regex is: "X-Amz-Target: com.amazon.(.*)"
The Configuration is currently: "AmazonAPI"
I'd like it to be: "AmazonAPI $1" or "AmazonAPI \1"

Actually, it'd be awesome if I could recreate AWS command-lines from interpreting the headers.  That's starting to sound like a script, akin to the curl_command_generator.js script!

kingthorin:
@penelopeFudd do you think it would be enough to provide for use of just the first capturing group?

For example:
Assume the following tag regex `(\d{3})` for the following Tag rule:
![image](https://user-images.githubusercontent.com/7570458/38101382-292e0e5c-334e-11e8-9f94-01c9242ab51e.png)

Given 3 inputs:
```
<html>
abc1234
<!-- bleh -->
</html>
```

```
<html>
333
</html>
```

```
<html>
fred
</html>
```

Would produce tags like:
![image](https://user-images.githubusercontent.com/7570458/38101522-8198e1d4-334e-11e8-86a4-a30032658d7a.png)

That might not be as specific as you were thinking when you raise this issue but hopefully it's a good enough illustration.

Capturing group 0 is the whole match, group 1 is the first group in the regex. 
Here's an online example:
http://www.regexplanet.com/share/index.html?share=yyyyyky9par
<sub>Note: You might have to select java then hit test when you visit the link.</sub>
<sub>Note 2: Notice the behavior when you change the pattern to `abc(\d{3})` and hit test.

kingthorin:
@psiinon @thc202 what do you think about implementing just based on group 1 to provide some flexibility?

thc202:
How would that work? If there's a group it's automatically appended to the tag?
(Note that it can be used `Matcher.appendReplacement` to automatically replace the group references in the tag.)

thc202:
Sounds good.

kingthorin:
Here's my initial code: https://github.com/zaproxy/zaproxy/compare/develop...kingthorin:tag-interpolation
It currently only applies to response bodies (but making it applicable to requests and headers isn't a big deal), it could maybe also use some clean up of the var/const names. 

[...]

PR #4583, opened 2018-04-11
Passive Scan Tag Interpolation

kingthorin:
RegexAutoTagScanner > Changed such that capturing group 1 is inserted to the tag (configuration) replacing the group reference `$1`.

Fixes zaproxy/zaproxy#4123

### DIFF
--- a/src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
+++ b/src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
@@ -187,7 +187,7 @@ public class RegexAutoTagScanner extends PluginPassiveScanner {
 					msg.getRequestHeader().toString());
 			if (m.find()) {
 				// Scanner matches, so do what it wants...
-				matched(msg, id);
+				matched(m, msg, id);
 				return;
 			}
 		}
@@ -196,7 +196,7 @@ public class RegexAutoTagScanner extends PluginPassiveScanner {
 					msg.getRequestHeader().getURI().toString());
 			if (m.find()) {
 				// Scanner matches, so do what it wants...
-				matched(msg, id);
+				matched(m, msg, id);
 				return;
 			}
 		}
@@ -216,7 +216,7 @@ public class RegexAutoTagScanner extends PluginPassiveScanner {
 					msg.getResponseHeader().toString());
 			if (m.find()) {
 				// Scanner matches, so do what it wants...
-				matched(msg, id);
+				matched(m, msg, id);
 				return;
 			}
 		}
@@ -225,7 +225,7 @@ public class RegexAutoTagScanner extends PluginPassiveScanner {
 					msg.getResponseBody().toString());
 			if (m.find()) {
 				// Scanner matches, so do what it wants...
-				matched(msg, id);
+				matched(m, msg, id);
 				return;
 			}
 		}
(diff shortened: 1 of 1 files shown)


kingthorin:
Example:
</img>

</img>

src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
```diff
@@ -310,9 +310,17 @@ public boolean equals(Object object) {
...
-	private void matched(HttpMessage msg, int id) {
+	private void matched(Matcher matcher, HttpMessage msg, int id) {
+		String tag = getConf();
 		if (tagHistoryType(msg.getHistoryRef().getHistoryType())) {
-			parent.addTag(id, this.getConf());
+			if (matcher.groupCount() > 0) {
+				StringBuffer result = new StringBuffer();
+				matcher.appendReplacement(result, tag);
+				tag = result.substring(matcher.start());
+				parent.addTag(id, tag);
```
thc202: This could be done outside/after the if/then branches.

Should the replaced tag also be used in the stats?

kingthorin: Good question. I would think based on the tag firing not necessarily based on each variation if the tag. But that's just a quick gutt reaction.

thc202:
Thanks!

src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
thc202: Yeah, makes sense.

src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
```diff
@@ -310,9 +310,15 @@ public boolean equals(Object object) {
...
     
-	private void matched(HttpMessage msg, int id) {
+	private void matched(Matcher matcher, HttpMessage msg, int id) {
+		String tag = getConf();
 		if (tagHistoryType(msg.getHistoryRef().getHistoryType())) {
-			parent.addTag(id, this.getConf());
+			if (matcher.groupCount() > 0) {
+				StringBuffer result = new StringBuffer();
+				matcher.appendReplacement(result, tag);
+				tag = result.substring(matcher.start());
```
denniskniep: Alternative could be
`tag = matcher.group().replaceAll(matcher.pattern().pattern(), tag);`

kingthorin:
Thanks!

src/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScanner.java
thc202: Yeah, that's probably better, it avoids allocating a (potential) bigger string. Though it could avoid compiling the pattern again and replace first (there's just one to replace):
`matcher.pattern().matcher(matcher.group()).replaceFirst(tag);`

kingthorin: K I’ll tweak that

kingthorin:
Tweaked as discussed.

denniskniep:
Thanks

Issue #3447, opened 2017-04-28
Memory errors when running active scan


RomHartmann:
Seems to be partially identical to https://github.com/zaproxy/zaproxy/issues/1283

Steps to create error:

1)  Use docker image [`owasp/zap2docker-stable`](https://github.com/zaproxy/zaproxy/blob/develop/build/docker/Dockerfile-stable)
2)  `zap.sh -daemon -port 8090 -host 127.0.0.1 -config api.disablekey=true`
3)  `zap-cli -p 8090 open-url http://www.plumblink.co.za/`
4)  `zap-cli -p 8090 spider http://www.plumblink.co.za/`
5)  `zap-cli -p 8090 active-scan -r http://www.plumblink.co.za/`

And then after scanning for some time I get the below trace.

Can I either increase the size of the DB or plug in my own?

Just a note, the scan succeeds on other urls that I have tested

```
2017-04-28T07:57:12.748768013Z INFO:__main__:starting daemon...
2017-04-28T07:57:12.795957006Z INFO:__main__:daemon not running yet, waiting...
2017-04-28T07:57:14.211171930Z Found Java version 1.8.0_121
2017-04-28T07:57:14.322029997Z Available memory: 3699 MB
2017-04-28T07:57:14.322464530Z Setting jvm heap size: -Xmx924m
2017-04-28T07:57:17.431836231Z 1005 [main] INFO org.zaproxy.zap.DaemonBootstrap  - OWASP ZAP 2.6.0 started 28/04/17 07:57:17

...
2017-04-28T09:51:08.057026890Z 6831630 [ZAP-PassiveScanner] ERROR org.zaproxy.zap.extension.params.ExtensionParams  - java.sql.SQLDataException: data exception: string data, right truncation
2017-04-28T09:51:08.057049509Z org.parosproxy.paros.db.DatabaseException: java.sql.SQLDataException: data exception: string data, right truncation
2017-04-28T09:51:08.057053796Z 	at org.parosproxy.paros.db.paros.ParosTableParam.update(Unknown Source)
2017-04-28T09:51:08.057057429Z 	at org.zaproxy.zap.extension.params.ExtensionParams.persist(Unknown Source)
2017-04-28T09:51:08.057069917Z 	at org.zaproxy.zap.extension.params.ExtensionParams.onHttpRequestSend(Unknown Source)
[...]

PR #4601, opened 2018-04-17
Fix Issue 3447 - Params Table Truncation

kingthorin:
When DatabaseException is caught check if the cause is SQLDataException. If so log an info message (if debug is enabled log the exception message and stack trace). If the cause is not SQLDataException log an error as per the existing behavior.

Since the values field (column) is ~8MB it should be rare that this occurs. (In fact to test/recreate I had to temporarily edit the DB .script file to make the values column ~80KB and reload a page with a ~10KB header value a bunch of times.) While the user may want to be aware of what's happening it is unlikely to impact their use of ZAP for "normal" use cases.

Fixes zaproxy/zaproxy#3447


src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -350,7 +351,15 @@ private void persist(HtmlParameterStats param) {
 						param.getId(), param.getTimesUsed(), setToString(param.getFlags()), setToString(param.getValues()));
 			}
 		} catch (DatabaseException e) {
-			logger.error(e.getMessage(), e);
+			if (e.getCause() instanceof SQLDataException) {
```
kingthorin: Should I also (or only) check that `e.getCause().getMessage().contains("truncation");`?

kingthorin:
DOH, sorry I missed the label, I'm horrible with that :cry:

src/org/zaproxy/zap/extension/params/ExtensionParams.java
thc202: Probably better.

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -350,7 +351,15 @@ private void persist(HtmlParameterStats param) {
 						param.getId(), param.getTimesUsed(), setToString(param.getFlags()), setToString(param.getValues()));
 			}
 		} catch (DatabaseException e) {
-			logger.error(e.getMessage(), e);
+			if (e.getCause() instanceof SQLDataException) {
+				logger.info("Could not add or update param: " + param.getName());
```
thc202: `warn`?

src/org/zaproxy/zap/extension/params/ExtensionParams.java
```diff
@@ -350,7 +351,15 @@ private void persist(HtmlParameterStats param) {
 						param.getId(), param.getTimesUsed(), setToString(param.getFlags()), setToString(param.getValues()));
 			}
 		} catch (DatabaseException e) {
-			logger.error(e.getMessage(), e);
+			if (e.getCause() instanceof SQLDataException) {
+				logger.info("Could not add or update param: " + param.getName());
+				logger.info("It is likely that the length of one of the data elements exceeded the column size.");
```
thc202: Worth appending the exception's message.

thc202:
Thanks!

PR #4648, opened 2018-05-07
Active Scan & Spider Title Caps Tweaks - Issue 2000

kingthorin:
Per on-going zaproxy/zaproxy#2000


thc202:
Spider (normal and AJAX) dialogue also needs tweaks (e.g. `Show Advanced Options` / `Starting Point`), you going to change that as well?

kingthorin:
Hadn't looked, but ya I can tweak that and include it here too. Update inbound.

thc202:
Thanks!

PR #4655, opened 2018-05-08
Track the directories with scripts

thc202:
Change ExtensionScript to add the scripts of the directories also when a
script engine or type is added (they were already being removed). This
ensures that the scripts of the directories are added when an add-on
with script engines or types is installed, updated, or finished
loading (the case of Python engine added asynchronously).


thc202:
This affects the Python scripts (e.g. HUNT) bundled in the Community Scripts add-on, they are not added. :/

PR #4668, opened 2018-05-14
Make Passive Scan Tags management cleaner

juhakivekas:
Delete the "Type" field from table and modify/add dialogs because it's always TAG. The contents of the UI element weren't even used before, so functionally nothing changed. Rename "Configuration" field to "Tag" to be less general and more clear.

### DIFF
--- a/src/org/zaproxy/zap/extension/pscan/DialogModifyAutoTagScanner.java
+++ b/src/org/zaproxy/zap/extension/pscan/DialogModifyAutoTagScanner.java
@@ -56,10 +56,7 @@ class DialogModifyAutoTagScanner extends DialogAddAutoTagScanner {
     @Override
     protected void init() {
         getNameTextField().setText(scanner.getName());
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/pscan/DialogModifyAutoTagScanner.java
```diff
@@ -56,10 +56,6 @@ protected boolean validateName(String name) {
     @Override
     protected void init() {
         getNameTextField().setText(scanner.getName());
-        getRequestUrlRegexTextField().discardAllEdits();
```
thc202: It should call `getNameTextField().discardAllEdits();` instead.

juhakivekas: fixd

thc202:
Thanks!

Issue #3563, opened 2017-05-22
Enhancement: Add New Abstract Plugin Type


kingthorin:
I would be handy to have an active scan plugin type that runs once per directory (or once per node that could have children).

thc202:
Thanks for raising the issue :)

PR #4672, opened 2018-05-15
Move (non-dist) doc files to top level directory

thc202:
Move non-distribution doc files to top level directory, those are not
bundled with ZAP.
Update links to one of the files in scripts bundled in Docker images.


thc202:
That's a good question, I think the former, the SQLi plugin still set the properties but that's it, `Kb` is not much used. Maybe something to deprecate and remove?

kingthorin:
I don't think the whole thing should be removed. I was just wondering about that doc piece. I see the value in having it, and ZAP could definitely leverage it more.

image

thc202:
OK. I was not counting that one, `AbstractAppFilePlugin` will be removed once #3563 lands.

kingthorin:
Oh ok, I wasn't assuming any specific timing for that one so figured it may as well be doc'd.

Plus I had idea of tech detection or tracking (kinda like the SQLi one) in the back of my head.

PR #4677, opened 2018-05-17
Ensure strings that are JSON are not auto converted

### LINKED ISSUE #4580: Problem with scanner when post parameters are json
Hi

Some time ago, i asked how could i trigger scans with active scanner through ZAP API for a POST request. The answer was to use a command like this

ascan.scan(t.get("url"), "", "", "", t.get("type"), t.get("params"))

Where params is something like this:

name=test&id=1

However, im working with an application that sends POST parameters as a json. 

{'name': 'test', 'id':1}

When running that command, it fails when triggering the scanner, as it says that this url doesnt exist

Is there any way to run a scanner on a POST request with this format? If not, should this be an enhancement?

Thanks

### LINKED ISSUE #4590: ZAP API converts data to wrong format
Hi

I did open a bug the other day about json compatibility in post requests #4580 

I have seen another bug related to this. When the ZAP API (in python) returns that data, the JSON is formatted as a dictionary instead of having a JSON format

So, if the post body was {"name": "test", "id":"1"}

The API returns this data as {'name': 'test', 'id':'1'}

And therefore, it is not recognised as a JSON anymore

psiinon:
Fixes #4590

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ApiResponseElement.java
+++ b/src/org/zaproxy/zap/extension/api/ApiResponseElement.java
@@ -18,8 +18,10 @@
 package org.zaproxy.zap.extension.api;
 
 import net.sf.json.JSON;
+import net.sf.json.JSONException;
 import net.sf.json.JSONNull;
 import net.sf.json.JSONObject;
+import net.sf.json.JSONSerializer;
 
 import org.apache.commons.lang.StringEscapeUtils;
 import org.w3c.dom.Document;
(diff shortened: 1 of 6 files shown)


psiinon:
Ugh, sure that test was working before I checked it in :/
Should know better than to issue a PR last thing ;)
Will look at it asap...

psiinon:
Fixed

psiinon:
Removed #4580 from the comment as this PR wont fix that in all cases

Issue #2182, opened 2016-01-18
Add support for dynamic fields in Form-based authentication method


tony1016:
According to [this discussion](https://github.com/zaproxy/zaproxy/issues/2093#issuecomment-172026232),zaproxy can not handle user_token properly in DVWA,please fix it.

thc202:
Fixed in #4682.

PR #4682, opened 2018-05-21
support reauthentication in the presence of an anti-CSRF token for post based authentication method type.

KajanM:
see #2182

### DIFF
--- a/src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
+++ b/src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
@@ -66,6 +66,8 @@ import org.parosproxy.paros.network.HttpSender;
 import org.parosproxy.paros.view.SessionDialog;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.authentication.UsernamePasswordAuthenticationCredentials.UsernamePasswordAuthenticationCredentialsOptionsPanel;
+import org.zaproxy.zap.extension.anticsrf.AntiCsrfToken;
+import org.zaproxy.zap.extension.anticsrf.ExtensionAntiCSRF;
 import org.zaproxy.zap.extension.api.ApiDynamicActionImplementor;
 import org.zaproxy.zap.extension.api.ApiException;
 import org.zaproxy.zap.extension.api.ApiResponse;
@@ -159,6 +161,8 @@ public abstract class PostBasedAuthenticationMethodType extends AuthenticationMe
 		private SiteNode loginSiteNode = null;
 		private String loginRequestURL;
 		private String loginRequestBody;
+		
+		private ExtensionAntiCSRF extAntiCsrf;
 
 		/**
 		 * Constructs a {@code PostBasedAuthenticationMethod} with the given data.
@@ -275,16 +279,16 @@ public abstract class PostBasedAuthenticationMethodType extends AuthenticationMe
 			HttpMessage msg;
 			try {
 				msg = prepareRequestMessage(cred);
+				// Make sure the message will be sent with a good WebSession that can record the changes
+				if (user.getAuthenticatedSession() == null)
+					user.setAuthenticatedSession(sessionManagementMethod.createEmptyWebSession());
+				msg.setRequestingUser(user);
+				
+				replaceAntiCsrfTokenValueIfRequired(msg, user);
 			} catch (Exception e) {
 				LOGGER.error("Unable to prepare authentication message: " + e.getMessage(), e);
 				return null;
 			}
-
-			// Make sure the message will be sent with a good WebSession that can record the changes
-			if (user.getAuthenticatedSession() == null)
-				user.setAuthenticatedSession(sessionManagementMethod.createEmptyWebSession());
-			msg.setRequestingUser(user);
-
 			// Clear any session identifiers
 			msg.getRequestHeader().setHeader(HttpRequestHeader.COOKIE, null);
 
@@ -314,6 +318,51 @@ public abstract class PostBasedAuthenticationMethodType extends AuthenticationMe
 
 			// Return the web session as extracted by the session management method
 			return sessionManagementMethod.extractWebSession(msg);
+		}
+		
+		/**
+		 * Modifies the input {@code requestMessage} by replacing fresh anti-CSRF token value in the request body.  
+		 * It first checks if the input {@code requestMessage} has an anti-CSRF token.
+		 * If yes, then it makes a {@code GET} request to the login page in order to renew the anti-CSRF token.
+		 * Then it modifies the input {@code requestMessage} with the regenerated anti-CSRF token value.
+		 * If the first {@code POST} login request does not have anti-CSRF token
+		 * then the input {@code requestMessage} is left as it is.
+		 * 
+		 * @param requestMessage the login request message with correct credential
+		 * @param user the user of the context
+		 * @throws IOException if problem occurs when sending login request to get login page.
+		 */
+		private void replaceAntiCsrfTokenValueIfRequired(HttpMessage requestMessage, User user) throws IOException {
+			if(extAntiCsrf == null) {
+				extAntiCsrf = Control.getSingleton().getExtensionLoader().getExtension(ExtensionAntiCSRF.class);
+			}
+			List<AntiCsrfToken> antiCsrfTokensOfFirstLoginMsg = null;
+  			if(extAntiCsrf != null) {
+				antiCsrfTokensOfFirstLoginMsg = extAntiCsrf.getTokens(requestMessage);
+			}
+			AntiCsrfToken antiCsrfTokenOfFirstLoginMsg = null;
+			if(antiCsrfTokensOfFirstLoginMsg != null && antiCsrfTokensOfFirstLoginMsg.size() > 0) {
+				antiCsrfTokenOfFirstLoginMsg = antiCsrfTokensOfFirstLoginMsg.get(0);
+			}
+			
+			if (antiCsrfTokenOfFirstLoginMsg != null) {
+				HttpMessage loginMsgToRegenerateAntiCsrfToken = antiCsrfTokenOfFirstLoginMsg.getMsg().cloneAll();
+				loginMsgToRegenerateAntiCsrfToken.setRequestingUser(user);
+				
+				// Clear any session identifiers
+				loginMsgToRegenerateAntiCsrfToken.getRequestHeader().setHeader(HttpRequestHeader.COOKIE, null);
+				
+				getHttpSender().sendAndReceive(loginMsgToRegenerateAntiCsrfToken);
+				AuthenticationHelper.addAuthMessageToHistory(loginMsgToRegenerateAntiCsrfToken);
+
+				String regeneratedAntiCsrfTokenValue = extAntiCsrf.getTokenValue(loginMsgToRegenerateAntiCsrfToken, antiCsrfTokenOfFirstLoginMsg.getName());
+				String requestBody = requestMessage.getRequestBody().toString();
+				if (regeneratedAntiCsrfTokenValue != null) {
+					requestBody = requestBody.replace(
+							antiCsrfTokenOfFirstLoginMsg.getValue(), paramEncoder.apply(regeneratedAntiCsrfTokenValue));
+					requestMessage.getRequestBody().setBody(requestBody);
+				}
+			}
 		}
 
 		/**


src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -159,6 +163,9 @@ protected PostBasedAuthenticationMethodType(
 		private SiteNode loginSiteNode = null;
 		private String loginRequestURL;
 		private String loginRequestBody;
+		
+		private ExtensionAntiCSRF extAntiCsrf;
+		private boolean requireAnitCsrfTokenWhenLogin;
```
thc202: There's a typo in anti.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -236,6 +243,7 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
 				if (requestBody != null) {
 					requestMessage.getRequestBody().setBody(requestBody);
 					requestMessage.getRequestHeader().setHeader(HttpHeader.CONTENT_LENGTH, null);
+					replaceAntiCsrfTokenAndCookieValueIfRequired(requestMessage);
```
thc202: This should be done also when not using a `SiteNode`. Any reason not to do this in `authenticate` instead?

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +260,43 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+			AntiCsrfToken antiCsrfTokenOfFirstLoginMsg = null;
+			if(antiCsrfTokensOfFirstLoginMsg != null && antiCsrfTokensOfFirstLoginMsg.size() > 0) {
+				antiCsrfTokenOfFirstLoginMsg = antiCsrfTokensOfFirstLoginMsg.get(0);
+			}
+			
+			if (antiCsrfTokenOfFirstLoginMsg != null) {
+				this.requireAnitCsrfTokenWhenLogin = true;
+				URI loginURI = new URI(loginRequestURL, true, "UTF-8");
+				HttpMessage loginMsgToRegenerateAntiCsrfToken = new HttpMessage(loginURI);
+				loginMsgToRegenerateAntiCsrfToken.setRequestingUser(null);
```
thc202: Why not the user? We do want to use the same session.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +260,43 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+				antiCsrfTokensOfFirstLoginMsg = extAntiCsrf.getTokens(requestMessage);
+			}
+			AntiCsrfToken antiCsrfTokenOfFirstLoginMsg = null;
+			if(antiCsrfTokensOfFirstLoginMsg != null && antiCsrfTokensOfFirstLoginMsg.size() > 0) {
+				antiCsrfTokenOfFirstLoginMsg = antiCsrfTokensOfFirstLoginMsg.get(0);
+			}
+			
+			if (antiCsrfTokenOfFirstLoginMsg != null) {
+				this.requireAnitCsrfTokenWhenLogin = true;
+				URI loginURI = new URI(loginRequestURL, true, "UTF-8");
```
thc202: Shouldn't this use the URL from `antiCsrfTokenOfFirstLoginMsg`? (It might not be the same.)

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +260,43 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+				HttpMessage loginMsgToRegenerateAntiCsrfToken = new HttpMessage(loginURI);
+				loginMsgToRegenerateAntiCsrfToken.setRequestingUser(null);
+				try {
+					getHttpSender().sendAndReceive(loginMsgToRegenerateAntiCsrfToken);
+				} catch (IOException e) {
```
thc202: Should not catch it? If the request requires a fresh anti-csrf token there's no point trying to authenticate.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +260,43 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+					e.printStackTrace();
```
thc202: We should log instead (print only when in cmd line mode). (Note that it should also return.)

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +260,43 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+				String regeneratedAntiCsrfTokenValue = extAntiCsrf.getTokenValue(loginMsgToRegenerateAntiCsrfToken, antiCsrfTokenOfFirstLoginMsg.getName());
+
+				String requestBody = requestMessage.getRequestBody().toString();
+				if (regeneratedAntiCsrfTokenValue != null) {
+					requestBody = requestBody.replace(
+							antiCsrfTokenOfFirstLoginMsg.getValue(), paramEncoder.apply(regeneratedAntiCsrfTokenValue));
+					requestMessage.getRequestBody().setBody(requestBody);
+				}
+				TreeSet<HtmlParameter> cookieParams = loginMsgToRegenerateAntiCsrfToken.getResponseHeader().getCookieParams();
+				requestMessage.setCookieParams(cookieParams);
```
thc202: This should not be required if using the users' session, the cookies are set into the request when the message is sent.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -286,7 +331,9 @@ public WebSession authenticate(SessionManagementMethod sessionManagementMethod,
 			msg.setRequestingUser(user);
 
 			// Clear any session identifiers
-			msg.getRequestHeader().setHeader(HttpRequestHeader.COOKIE, null);
+			if(!this.requireAnitCsrfTokenWhenLogin) {
```
thc202: This should not be needed if using the users' session.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -296,7 +343,11 @@ public WebSession authenticate(SessionManagementMethod sessionManagementMethod,
 
 			// Send the authentication message
 			try {
-				getHttpSender().sendAndReceive(msg);
+				if(this.requireAnitCsrfTokenWhenLogin) {
```
thc202: Why?

thc202:
Which applications where used to test this?

KajanM:
This change allows ZAP to reauthenticate successfully in the presence of an anti-CSRF token in a login message. This is only tested against latest DVWA[1] available at GitHub. However, there is a problem in session state management. 

[1] https://github.com/ethicalhack3r/DVWA

After setting up necessary form-based authentication configuration, I logged out and made a request to localhost/dvwa/index.php.
## headers of the reqeusted message to index.php after signing out.
### Request Header
image

### Response Header
image

## make a request to login page to regenerate anti-CSRF token
### Request Header
image

###Response Header
image

###Anti-CSRF token value in response body
image

## authenticate with credentials and token value
### Request Header
image

### Response Header
image

## Redirect to index.php with regenerated session id
### Request Header
image

### Response Header
image

### Response body showing we are logged in
image

After this step cookies from this message is not correctly transferred to the session.
image

image

KajanM:
Where do I have to make change to set cookie values to the session?

thc202:
You don't directly, they are set by sending the message with a user.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: I am replacing the old token value with the new one. To get the old one I need the history reference to the very first post request to login site.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: Actually, I logged out to perform testing. When logging out the application redirects to the login page with a new token.
With the `loginSiteNode` I could not get the new token. It always gives the very first token sent with the post request to login site(when authenticating using browser).
That's why I make a new request and get the token from that.
BTW, I do not understand why there's no point in trying to authenticate.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
thc202: Whether the authentication was set up with a `SiteNode` or manually does not matter for the token generation (`AntiCsrfToken` already has the message to regenerate the token).

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: `antiCsrfTokenOfFirstLoginMsg` is got from passing the `HttpMessage` of `loginSiteNode` to the `extAntiCsrf.getToken(HttpMessage)`.
as URL of `antiCsrfTokenOfFirstLoginMsg` is referring to the `POST` method and I want a fresh 'GET`, I opted to this. But as this is just URI, I can change this. Did not get the point where URL created using `loginRequestURL` and URL from `antiCsrfTokenOfFirstLoginMsg` may not be the same. Any example?

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
thc202: Yes, the `loginSiteNode` does not need to be used for the regeneration of the token, we should use the message from the `AntiCsrfToken` (which has the correct URL).

Because the authentication will fail (why try it anyway?).

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: Well, DVWA redirects to index page when successfully authenticated. I assumed this is the general case. With the redirect response header I do not get anything useful to mark as a logged-in/logged-out indicator. 
image

I marked "You have logged in as 'admin'" in index page as logged-in indicator. Without redirection as the logged in indicator is not present in the response, the authentication is marked as failed.

thc202: That should also be part of the logged in indicator (e.g. `Location: index.php`), if not other parts of ZAP will fail to know that the message is authenticated (not everything follows redirections automatically).

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
thc202: The `antiCsrfTokenOfFirstLoginMsg` should point to a GET not POST, could you confirm that?

That happens when the login form is not submitted to the same page.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: Initially, I went with `Location: index.php`. But then I found the logged in/ logged out indicator must be something that is common among all pages. Correct me if I am wrong. As each response goes through `isAuthenticated` process(?) where the response header and body are checked for any match of the `loggedInIndicator` or `loggedOutIndicator` value, this value must be something that presents in all pages.
Though I marked in index.php the logout link and logged in message appear on all pages. That's why I moved from `Location: index.php`. Am I missing anything?

thc202: The logged in/out indicators can be composed to include both cases, e.g.:
```
(Location: index.php)|(You have logged in as)
```

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: sorry for the misunderstanding, as `loginSiteNode` is referring to `POST` and I got `antiCsrfTokenOfFirstLoginMsg` from the history reference of `loginSiteNode` I thought the `antiCsrfTokenOfFirstLoginMsg` also pointing to `POST`.

`antiCsrfTokenOfFirstLoginMsg` is pointing to `GET`.
image

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: then, can I rely on `loginRequestBody` and `loginRequestURL` in `PostBasedAuthenticationMethod` to get the `AntiCsrfToken` using `extAntiCsrf.getTokens(reqBody, reqUri)`?

thc202: Yes, if `loginRequestBody` and `loginRequestURL` have the expected contents it should work.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
KajanM: Initially, I put this to prevent cookies of `POST` request to `login.php`getting replaced by old values instead of the new values from the response of `GET` request to `login.php`.  When refactoring I messed up.
Is there any way to get the user from this scope?
As the first `GET`, `POST` to `login.php` is made from the browser when `forcedUser` mode is disabled, they do not have the `user` info set.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
thc202: The problem is that we don't have any guarantees that's still fresh/valid (specially if using another session), so the authentication would fail and we would have to retry it after getting a valid one.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
thc202: We need to pass the user from `authenticate` to `prepareRequestMessage` and then to this method (we could also call `replaceAntiCsrfTokenAndCookieValueIfRequired` from `authenticate` after preparing the message).

Yeah, that's not a problem, we just need that when re-authenticating.

KajanM:
suggestions are addressed and working fine with DVWA. 

Still, I could not get rid of setting cookies manually. as the response of `loginMsgToRegenerateAntiCsrfToken` is the login page that is not authenticated, cookie values of that message are not transferred to the session. So setting the user did not solve the issue.  Is there any other way to overcome this?

thc202:
We should use the user as that's how it's expected for the session/cookies to be passed around. There are a couple of changes needed to make that work though:
 1. We need to create the user's session before refreshing the token, otherwise it might not have any session. This: https://github.com/zaproxy/zaproxy/blob/5134e9bec57f6d6fbb750614612221973aa288bd/src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java#L284-L285 should be moved to before the try-catch.
 2. The cookies of `loginMsgToRegenerateAntiCsrfToken` should be reset before the message is sent, to ensure it uses the cookies from the session (and the session registers the ones set). Ideally this should be done by the session management method but for now it's fine to do that manually, like it was already being done for the other message: https://github.com/zaproxy/zaproxy/blob/5134e9bec57f6d6fbb750614612221973aa288bd/src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java#L288-L289

This also means that `requireAntiCsrfTokenWhenLogin` is no longer needed. Let me know how it goes.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -252,6 +259,39 @@ protected HttpMessage prepareRequestMessage(UsernamePasswordAuthenticationCreden
...
+				antiCsrfTokenOfFirstLoginMsg = antiCsrfTokensOfFirstLoginMsg.get(0);
+			}
+			
+			if (antiCsrfTokenOfFirstLoginMsg != null) {
+				this.requireAntiCsrfTokenWhenLogin = true;
+				HttpMessage loginMsgToRegenerateAntiCsrfToken = antiCsrfTokenOfFirstLoginMsg.getMsg().cloneAll();
+//				loginMsgToRegenerateAntiCsrfToken.setRequestingUser(user);
+				
+				getHttpSender().sendAndReceive(loginMsgToRegenerateAntiCsrfToken);
+
```
thc202: Worth calling here `AuthenticationHelper.addAuthMessageToHistory(loginMsgToRegenerateAntiCsrfToken);` to show the message to the user.

src/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRF.java
```diff
@@ -222,7 +222,7 @@ public boolean requestHasToken(String reqBody) {
 		return this.getTokens(msg.getRequestBody().toString(), msg.getRequestHeader().getURI().toString());
 	}
 	
-	private List<AntiCsrfToken> getTokens(String reqBody, String targetUrl) {
+	public List<AntiCsrfToken> getTokens(String reqBody, String targetUrl) {
```
thc202: I was thinking more about this, it might be better to just use `getTokens(HttpMessage)` (with `requestMessage`) instead of making public. This allow us to use the same method for JSON messages (and others, which we can tell by its `Content-Type`) once the support is added.

KajanM:
many thanks, suggestions are addressed.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -29,6 +29,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeSet;
```
thc202: This import can be removed no longer used (same for `HtmlParameter`).

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -315,6 +321,51 @@ public WebSession authenticate(SessionManagementMethod sessionManagementMethod,
...
+		
+		/**
+		 * Modifies the input {@code requestMessage} by replacing fresh anti-CSRF token value in the request body.  
+		 * It first checks if the input {@code requestMessage} has an anti-CSRF token.
+		 * If yes, then it makes a {@code GET} request to the login page in order to renew the anti-CSRF token.
+		 * Then it modifies the input {@code requestMessage} with the regenerated anti-CSRF token value.
+		 * If the first {@code POST} login request does not have anti-CSRF token
+		 * then the input {@code requestMessage} is left as it is.
+		 * 
+		 * @param	requestMessage	the login request message with correct credential
```
thc202: It does not worth align the parameters (and their descriptions), that adds unnecessary work (to keep them aligned when they change, also it does not make any difference for rendered JavaDoc).

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -315,6 +321,51 @@ public WebSession authenticate(SessionManagementMethod sessionManagementMethod,
...
+				antiCsrfTokensOfFirstLoginMsg = extAntiCsrf.getTokens(requestMessage);
+			}
+			AntiCsrfToken antiCsrfTokenOfFirstLoginMsg = null;
+			if(antiCsrfTokensOfFirstLoginMsg != null && antiCsrfTokensOfFirstLoginMsg.size() > 0) {
+				antiCsrfTokenOfFirstLoginMsg = antiCsrfTokensOfFirstLoginMsg.get(0);
+			}
+			
+			if (antiCsrfTokenOfFirstLoginMsg != null) {
+				HttpMessage loginMsgToRegenerateAntiCsrfToken = antiCsrfTokenOfFirstLoginMsg.getMsg().cloneAll();
+				//to make sure cookies from the loginMsgToRegenerateAntiCsrfToken response
```
thc202: I think this comment could be removed the docs of `setRequestingUser` already imply that.

src/org/zaproxy/zap/authentication/PostBasedAuthenticationMethodType.java
```diff
@@ -315,6 +321,51 @@ public WebSession authenticate(SessionManagementMethod sessionManagementMethod,
...
+				//is transferred to the user's web session
+				//so that subsequent requests will have correct cookie values set
+				loginMsgToRegenerateAntiCsrfToken.setRequestingUser(user);
+
```
thc202: We should remove the cookies before sending the `loginMsgToRegenerateAntiCsrfToken`, otherwise the session might not be correctly updated.

thc202: This was not applied, note that I was referring to `loginMsgToRegenerateAntiCsrfToken` (for `msg` it can be done where it was).

KajanM:
fixed

thc202:
Thanks!

PR #4694, opened 2018-05-25
Provide details of why the add-on is not valid

thc202:
Change ExtensionAutoUpdate to provide details of why an add-on is not
valid when installing a local add-on, before it would just say that the
add-on is not valid, now it will show the actual reason why it is not
valid.
Change AddOn to provide the result of the validation to allow callers to
show a more appropriate message, also, avoid reading the manifest twice
by using the manifest already read when validating.

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -803,7 +803,14 @@ cfu.warn.addOnSameVersion = The same version of the add-on is already installed:
 cfu.warn.addOnNotRunnable.message = The add-on will not run until the following requirements are fulfilled:
 cfu.warn.addOnNotRunnable.question = Continue with the installation?
 cfu.warn.cantload      = Can not load the specified add-on\:\nNot before \= {0}\nNot from \= {1}
-cfu.warn.invalidAddOn = The selected file is not a valid ZAP add-on.
+cfu.warn.invalidAddOn = The selected file is not a valid ZAP add-on{0}
+cfu.warn.invalidAddOn.invalidPath = :\nThe path is not valid.
+cfu.warn.invalidAddOn.noZapExtension = :\nThe file does not have a "zap" extension.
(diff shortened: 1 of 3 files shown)


src/lang/Messages.properties
```diff
@@ -803,7 +803,14 @@ cfu.warn.addOnSameVersion = The same version of the add-on is already installed:
 cfu.warn.addOnNotRunnable.message = The add-on will not run until the following requirements are fulfilled:
 cfu.warn.addOnNotRunnable.question = Continue with the installation?
 cfu.warn.cantload      = Can not load the specified add-on\:\nNot before \= {0}\nNot from \= {1}
-cfu.warn.invalidAddOn = The selected file is not a valid ZAP add-on.
+cfu.warn.invalidAddOn = The selected file is not a valid ZAP add-on{0}
+cfu.warn.invalidAddOn.noName = :\nThe file has no name.
```
kingthorin: What's this mean? How can you select a file with no name?

kingthorin:
In ExtensionAutoUpdate should a message be logged as well as displaying the dialog? Should view or viewdelegate be checked first (how are command line or API installations handled)?

src/lang/Messages.properties
thc202: That happens when you try to validate/create an add-on with a null or root path (though not through the UI which just allows to select files). I guess the message could be changed to the path is not valid instead.

thc202:
Not much would be logged, all the information is already provided by the message. This (installation of local add-on) is currently being done only through the UI.

src/lang/Messages.properties
kingthorin: That seems more meaningful

kingthorin:
Ok

src/lang/Messages.properties
thc202: Changed.

PR #4712, opened 2018-05-31
Add Passive Scan Footer Icon/Status

kingthorin:
ExtensionPassiveScan > Add ScanStatus object and appropriate initialization and utility methods.
Messages.properties > Add key/value for tooltip.
PassiveScanThread > Add call to update ScanStatus display in run().
pscan.png > Icon for use in the footer status area.

Animated GIF example:
pscan_cap


src/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
```diff
@@ -61,6 +62,7 @@
     private PassiveScanThread pst = null;
     private boolean passiveScanEnabled;
     private PassiveScanParam passiveScanParam;
+    private static ScanStatus scanStatus = null;
```
thc202: The `static` modifier can be removed.

src/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
```diff
@@ -80,6 +82,9 @@ public ExtensionPassiveScan() {
     private void initialize() {
         this.setOrder(26);
         this.setName(NAME);
+        if (View.isInitialised()) {
```
thc202: Is this really needed here? (It can be initialised later in `hook`.)

src/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
```diff
@@ -548,6 +554,16 @@ public void loadFrom(Configuration conf) {
         }
 	}
 
+    public ScanStatus getScanStatus() {
```
thc202: - `protected`.
 - The counter should be reset when the session changes (otherwise it might show an outdated number, from previous session).

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -103,6 +105,7 @@ public void run() {
 		// Get the last id - in case we've just opened an existing session
 		currentId = this.getLastHistoryId();
 		lastId = currentId;
+
```
thc202: Revert?

src/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -143,6 +146,10 @@ public void run() {
 						
 						for (PassiveScanner scanner : scannerList.list()) {
 							try {
+								if (View.isInitialised()) {
```
thc202: This should be done outside the loop (the message is still being scanned here thus the count will not change), maybe after line 190?

kingthorin:
Review addressed, thanks!

src/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
```diff
@@ -472,6 +475,7 @@ private void stopPassiveScanThread() {
     @Override
     public void sessionChanged(Session session) {
         startPassiveScanThread();
+        getScanStatus().setScanCount(0);
```
thc202: Should check that the view is initialised.

kingthorin:
That one's done too :zap:

thc202:
Thank you!

Issue #4729, opened 2018-06-06
Param tab unable to "Flag as Anti-CSRF token"


kingthorin:
**Describe the bug**
From the params tab via the context menu I'm unable to "Flag as Anti-CSRF token" on a cookie value....

**To Reproduce**
1. Launch ZAP
2. Browse OWASP Pixi a bit
3. Review the Params tab a bit
4. Note there are two CSRF token cookies.
5. Right click on them and note that "Flag as Anti-CSRF token" is not enabled for use.

**Expected behavior**
The menu item should be enabled for use.

**Screenshots**
![image](https://user-images.githubusercontent.com/7570458/41009061-b7d044e8-68fc-11e8-87b3-9958383aea53.png)

**Software versions**
 - ZAP: Dev Build
 - OS: Win 7
 - Java:
java version "1.8.0_171"
Java(TM) SE Runtime Environment (build 1.8.0_171-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.171-b11, mixed mode)
 - Browser: Firefox 60.0.1

**Additional context**
Also note the menu item is subject to an Issue 2000 title caps discrepancy as well.

thc202:
Cookies are not supported yet.

kingthorin:
Ok, I hadn't taken the time to look at the code yet.

thc202:
How/where are they being set?

kingthorin:
I had started with a quick scan. It looks like they were set in response to the first request by the spider for robots.txt.
```
HTTP/1.1 404 Not Found
X-Powered-By: Express
Set-Cookie: _csrf=jrW3ZM6dC9v9wQGBdXdrlgDw; Path=/
Set-Cookie: CSRF-TOKEN=4tisFkvH-1-mapgoz88EoaKRyFYWGOc2DUuo; Path=/
Content-Security-Policy: default-src 'self'
X-Content-Type-Options: nosniff
Content-Type: text/html; charset=utf-8
Content-Length: 149
Date: Tue, 05 Jun 2018 23:56:44 GMT
Connection: keep-alive
```

Issue #1133, opened 2015-06-04
Anti-CSRF Token check in URL


zapbot:
```
hey all,

i'm currently facing a problem with tokens which are generated with every request from
the server and ZAProxy seems not to handle that properly, to make valid requests, e.g.
for fuzzing a special parameter.

The client request looks like this:

POST http://webapp.de/ptlweb/WebPortal?timeout=4931&applId=etretr&frame=content&wp_token_etretr=1783017392
HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:24.0) Gecko/20100101 Firefox/24.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: de-de,de;q=0.8,en-us;q=0.5,en;q=0.3
Referer: http://webapp.de/ptlweb/WebPortal?timeout=4931&applId=etretr&frame=content&wp_token_etretr=1075322898
Cookie: JSESSIONID=60CC585862E3194B3A3F78C736645B40.glebtcb21201_instanz_165; opttan_data=width_opttan_m%3D186
Connection: keep-alive
Content-Type: application/x-www-form-urlencoded
Content-Length: 72
Host: webapp.de

bezVerzeichnis=testdirectory&event___erzeugenVerzeichnis=Verzeichnis+erstellen


the value 1783017392 in wp_token_etretr changes with every request, so when i try to
fuzz e.g. the parameter
bezVerzeichnis with the same value in wp_token_etretr again nothing happens.


In the response body i find the post request like this:

<div class="profiler-item"><div class="profiler-line clearfix"><span class="profiler-name">POST
http://webapp.de/etrweb/eTresor[requestParams]{bezVerzeichnis=testdirectory, event___erzeugenVerzeichnis=Verzeichnis
erstellen, pid=45CE45259A457A289BBCFA3BFE09F020.glebtcb21201_165, wpsid=957D8B2EA2F3603839C91AA8CDAF4401.glebtcb21201_165,
sprache=de, wpuri=/ptlweb/WebPortal, frame=content, kinNr=4931, applId=etretr, wp_token_etretr=1783017392,
timeout=4931}</span><span class="profiler-offset">6 ms</span><span class="profiler-duration">14
ms</span></div></div>



this part of the application uses the wp_token_etretr token another part uses another
token e.g. requestId like here:


[...]

PR #4730, opened 2018-06-06
Params Tab - Add/Remove Param Name as Anti-CSRF Token for Any Param

kingthorin:
Messages.properties > Address zaproxy/zaproxy#2000 title caps issues and clarify that the menu item simply adds the 'param' name as a known Anti-CSRF token.
PopupMenuAddAntiCSRF/PopupMenuAddAntiCSRF > Change conditional in isEnableForComponent to accommodate all types of params.

The java files contain mixed indentation, I tried to keep it consistent.

Fixes zaproxy/zaproxy#4729

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -1823,8 +1823,8 @@ options.script.dialog.dirs.remove.title                = Remove Script Directori
 output.panel.mnemonic		= o
 output.panel.title 			= Output
 
-params.anticrsf.add.popup	= Flag as Anti-CSRF token
-params.anticrsf.remove.popup	= Unflag as Anti-CSRF token
+params.anticrsf.add.popup	= Add Name as Anti-CSRF Token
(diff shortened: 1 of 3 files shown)


thc202:
Cookies are still not supported, although they should be already sent if a session is enabled. Is this just to add the parameter name as an anti-csrf token?

kingthorin:
Ya that's all I was after.

thc202:
Sounds good to me, I'd suggest changing the name of the menu to e.g. "Add parameter name as Anti-CSRF Token", the current name implies more than just the name being added (which is the token being properly handled, which only the form is), also, might worth just remove the restriction of what parameter types can be added (e.g. #1133)?

kingthorin:
Will do.

kingthorin:
Tweaked.

thc202:
Thanks!

@psiinon PTAL at the new changes.

PR #4754, opened 2018-06-12
Remove HttpUserAgent class

thc202:
Remove the HttpUserAgent class and its files, the class and files are no
longer in use (since 2012?), they were used briefly in some passive
scanners to raise alerts on just some User-Agents.


thc202:
It might happen that 3rd party add-ons are using that class, but I didn't find one while searching in GitHub.

Issue #4771, opened 2018-06-18
Bundle required files in the zap.jar


thc202:
As discussed in IRC, bundle required files (e.g. Messages.properties, vulnerabilities.xml, config.xml...) in zap.jar to make it easier to use in functional tests (e.g. add-ons, #4755) and run with just zap.jar+libraries (e.g. from zap-extensions project along with the add-ons being developed).

ZAP should try first read the required files from the file system (as is now) only if not found it reads the ones bundled in the JAR.

PR #4773, opened 2018-06-18
Bundle Messages.properties and vulnerabilities.xml

thc202:
Change build.xml to include the files Messages.properties and
vulnerabilities.xml in the zap.jar.
Change I18N and Vulnerabilities to fallback to the bundled files if the
corresponding files are not in the file system.
Tweak LocaleUtils to not attempt to read the lang directory (for
Messages.properties) as it might no longer exist and if it exists log a
warn (instead of error) if no files are found (it would use the bundled
file).
Change VulnerabilitiesLoader to allow to load a vulnerabilities.xml file
using an InputStream (e.g. the bundled one) and return an empty list
of vulnerabilities if the lang directory is not found (to fallback to
bundled file).
Change ZapXmlConfiguration to allow to read a configuration from an
InputStream.

Part of #4771 - Bundle required files in the zap.jar

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -153,6 +153,10 @@
 			<fileset dir="${src}/lang" excludes="**/.svn/** **/_svn/**" />
 		</copy>
 
+		<copy todir="${build}/org/zaproxy/zap/resources/">
+			<fileset dir="${src}/lang" includes="Messages.properties,vulnerabilities.xml" />
+		</copy>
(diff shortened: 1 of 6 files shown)


thc202:
These are the most important for zap-extension's tests (which are failing in Java 9+ also because of the reliance on `ClassLoaderUtil`).

PR #4791, opened 2018-06-25
Properly re-parent panels in TabbedPanel2

### LINKED ISSUE #4789: Break tab is not automatically shown in full layout when breakpoint is hit
**Describe the bug**

When the "full layout" is active, the Break tab is not yet added and global break is enabled, the Break tab is not shown when a breakpoint is hit. This means that users are not aware that there's an HTTP request still waiting for interaction, it's just hanging there.

Manually adding the Break tab then shows the hanging request so users can decide what to do with it.

**To Reproduce**
1. Start ZAP
1. Activate "full layout" by clicking on the icon in the icon list
1. Break tab should not be added already
1. Activate global break
1. Make an HTTP request
1. Nothing happens, Break tab does not appear
1. Add the Break tab manually via the + sign in the tab bar
1. Intercepted HTTP request is shown

**Expected behavior**

Break tab should appear automatically in "full layout" mode.

**Software versions**
 - ZAP: 2.7.0
 - OS: Debian stable 9.4 ("stretch")
 - Java: openjdk 1.9, Debian version number: `9~b181-4~bpo9+1`, package `openjdk-9-jre`

thc202:
Change TabbedPanel2 to set itself as parent also to non-visible panels
when the visible panels are set to ensure those panels have the correct
parent (not just the visible ones). Also, don't re-parent if the panel
being added is hidden, as it means it's being shown (thus child) of
other TabbedPanel2.

Fix #4789 - Break tab is not automatically shown in full layout when
breakpoint is hit

### DIFF
--- a/src/org/zaproxy/zap/view/TabbedPanel2.java
+++ b/src/org/zaproxy/zap/view/TabbedPanel2.java
@@ -308,7 +308,7 @@ public class TabbedPanel2 extends TabbedPanel {
 	 * @see #setVisible(Component, boolean)
 	 */
 	public void addTabHidden(AbstractPanel panel) {
-		this.addTab(panel.getName(), panel.getIcon(), panel, panel.isHideable(), false, panel.getTabIndex());
+		this.addTab(panel.getName(), panel.getIcon(), panel, panel.isHideable(), false, panel.getTabIndex(), false);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


thc202:
This also fixes the issue in dev/weekly.

thc202:
Updated to fix another case, when ZAP was started in full layout it would have the same problem (i.e. not show the break panel when it should).

PR #4807, opened 2018-07-02
Add all dangling event consumers

thc202:
Change SimpleEventBus to add all dangling consumers, not just the first
found, also, delay the (actual) registration of the publisher to finish
adding all dangling consumers.
Add test to assert the expected behaviour.


psiinon:
Now conflicting

thc202:
Resolved.

PR #4821, opened 2018-07-06
Bundle files for report generation

thc202:
Change CoreAPI, ExtensionCompare, and ReportLastScan to fallback to the
bundled XSL files if not found in the file system.
Change ReportGenerator to allow to use a custom StreamSource (for the
XSL data).
Change build.xml to include the XSL files used for report generation.

Related to #4771 - Bundle required files in the zap.jar


thc202:
Conflict resolved.

Issue #4753, opened 2018-06-12
Exporting response bodies of many messages at the same time


blurpshell:
Hi, the body of the response of a site I'm testing is purely JSON. Is it possible to export the response body of many select responses at a time. I see you can export the raw data of one response, but is there a way to select many and export? Thanks.

kingthorin:
You could use something like this:
https://github.com/zaproxy/community-scripts/blob/fd9328d55ddba835eb2af8b727fdab0abc7de866/extender/HTTP%20Message%20Logger.js

To log only the responses.

If it's already in history/sites tree you can also traverse those and save the responses, check here for inspiration: https://github.com/zaproxy/community-scripts/tree/master/standalone

thc202:
Ref: https://groups.google.com/d/topic/zaproxy-users/4UgVqIkYHso/discussion :)

kingthorin:
Dev refs: 
* https://github.com/zaproxy/zaproxy/blob/develop/src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java (Needs changes)
* https://github.com/zaproxy/zaproxy/blob/develop/src/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java (Seems to handles multi-select)

PR #4828, opened 2018-07-10
Fix Issue 4753 - Allow exporting multiple responses

kingthorin:
PopupMenuExportRsponse > Remove check for single selection. Add loop for selected history references, add boundary similar to the Message Export functionality.
Messages.properties > Remove no longer needed multi-select warning, and tweak menu entry to imply singular or plural.

Fixes zaproxy/zaproxy#4753


src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
```diff
@@ -66,11 +67,6 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
                     return;
         	    }
 
-                if (hrefs.size() > 1) {
-                    extension.getView().showWarningDialog(Constant.messages.getString("history.export.response.single.warning"));	// ZAP: i18n
-                    return;
-                }
-
                 HttpMessage msg = null;
```
thc202: The following message validations should be removed as well? (We should validate all or none?)

src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
```diff
@@ -130,6 +128,8 @@ private void exportHistory(HttpMessage msg, BufferedOutputStream bos) {
 
         try {
             if (!msg.getResponseHeader().isEmpty()) {
+                String boundry = CRLF + "==== " + msg.getHistoryRef().getHistoryId() + " ==========" + CRLF;
```
thc202: There's a typo.

src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
```diff
@@ -46,6 +46,7 @@
 public class PopupMenuExportResponse extends JMenuItem {
 
 	private static final long serialVersionUID = 1L;
+	private static final String CRLF = System.lineSeparator();
```
thc202: Worth renaming it (to newline or something like that), it might not be `\r\n`.

src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
kingthorin: Best to remove or migrate inside loop?

thc202: I think it's fine to remove them, but we could read the message inside a try/catch block to log warning and continue though(?)

kingthorin:
Fixed

src/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
```diff
@@ -102,7 +85,16 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
         	    BufferedOutputStream bos = null;
                 try {
                     bos = new BufferedOutputStream(new FileOutputStream(file, isAppend));
-           	        exportHistory(msg, bos);
+                    for (HistoryReference href : hrefs) {
+                        HttpMessage msg = null;
+                        try {
+                            msg = href.getHttpMessage();
+                        } catch (Exception e1) {
```
thc202: Sorry, it's probably better to not catch the exception, that way the user is informed (UI) that an error occurred.

kingthorin: So totally remove the catch or put the dialog back in?

thc202: Remove the catch (the exception is being caught in a following catch block).

kingthorin:
Done

PR #4831, opened 2018-07-11
Don't require an options button in scan panels

thc202:
Change ScanPanel2 to allow to not add the options button, as not all
scanners might need one.


src/org/zaproxy/zap/view/ScanPanel2.java
```diff
@@ -250,7 +265,10 @@ protected GridBagConstraints getGBC(int gridx, int gridy, double weightx, Insets
 	 * <p>
 	 * The method must return the new coordinates of the current cell, after the elements have been
 	 * added.
-	 * </p>
+	 * <p>
+	 * <strong>Note:</strong> It is added a button at the end of the tool bar to show the Options dialogue, unless
```
kingthorin: Maybe `A button is added at the end....`

thc202:
Change done, thank you!

PR #4833, opened 2018-07-12
Added alert api endpoints

psiinon:
The 'alerts' view is a copy of the core one - plan is to deprecate the core ones somehow.
I think there will be a load more alert API endpoints that will be useful, such as ones for creating and changing alerts..

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/AlertAPI.java
+++ b/src/org/zaproxy/zap/extension/alert/AlertAPI.java
@@ -1,0 +1,499 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ * 
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ * 
+ * Copyright 2018 The ZAP Development team
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License. 
+ * You may obtain a copy of the License at 
+ * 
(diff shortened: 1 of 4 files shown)


src/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -0,0 +1,204 @@
...
+            ApiResponseList resultList = new ApiResponseList(name);
+            result = resultList;
+
+            // 0 (RISK_INFO) -> 3 (RISK_HIGH)
+            ApiResponseList[] list = new ApiResponseList[4];
+            for (int i = 0; i < list.length; i++) {
+                list[i] = new ApiResponseList(Alert.MSG_RISK[i]);
+            }
+
+            AlertTreeModel model = extension.getTreeModel();
```
thc202: It could reuse the same logic even more (like in core done with the `Processor`).

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -515,7 +515,7 @@ private void refreshAlert(Session session) throws DatabaseException {
                 }
             }
         }
-        siteTree.nodeStructureChanged((SiteNode) siteTree.getRoot());
+        siteTree.nodeStructureChanged(siteTree.getRoot());
```
thc202: The branch could be rebased to include these changes already done.

thc202:
Thanks!

PR #4836, opened 2018-07-13
Add option to prevent windows from grabbing focus from other apps

psiinon:


src/org/parosproxy/paros/view/View.java
```diff
@@ -218,6 +218,11 @@
      * Lazily initialised in {@link #getDefaultDeleteKeyStroke()}.
      */
     private KeyStroke defaultDeleteKeyStroke;
+    
+    /**
+     * If true the the UI can request focus from other applications, if false then it should not
```
thc202: Typo in then.

src/org/parosproxy/paros/view/View.java
```diff
@@ -1042,4 +1047,13 @@ public void displayMessage(Message message) {
             getResponsePanel().setMessage(httpMessage, true);
         }
     }
+
+
+    public boolean isCanGetFocus() {
```
thc202: It could have a since JavaDoc tag (and a little description, like the var has), same for following method.

psiinon:
Updated based on the comments received.

thc202:
Thank you!

PR #4868, opened 2018-07-23
Reduce code duplication in ExtensionKeyboard

thc202:
Extract a method to set the configured accelerator to a menu item.
Deprecate ZapMenuItem method with a typo (used in the refactored code).


kingthorin:
Use of the deprecated method `getIdenfifier` needs to be replaced to address build failures.

thc202:
Of course, done. Thank you!

PR #4887, opened 2018-07-30
Remove unused build targets and bundled file

thc202:
Remove targets dist-with-source and copy-source-to-build, they are not
used moreover the ZAP sources (and JavaDoc) are available from Maven
Central.
Remove unused bundled file xml/log4j.properties.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -131,8 +131,6 @@
 		<mkdir dir="${dist}" />
 		<!--copy file="lib.jar" tofile="${dist}/lib.jar"/-->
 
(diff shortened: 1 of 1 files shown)


kingthorin:
The xml reference was removed but not the file?

thc202:
Right, the file is still used in other location.

thc202:
It's also used in that one but not bundled (i.e. it's read from the file system).

kingthorin:
Ok, I took the commit comment "_Remove unused bundled file xml/log4j.properties._" to mean the file itself should be gone too....

Issue #2023, opened 2015-11-03
Allow to set client certificate (PKCS#12) through the ZAP API


hxd86:
Although some methods in javadoc such as setClientCertLocation(String clientCertLocation) and isUseClientCert() indicate ZAP is able to add/select client certificates with API, it is not possible to do it. 

This is related to discussion at https://groups.google.com/d/msg/zaproxy-users/7bh-zVzzGHw/FTcBUWksDgAJ

MJUi:
The command line doesn't allow specifying a client certificate either.

Related to this question : https://groups.google.com/forum/#!topic/zaproxy-users/WyBc0vbIjPc

adrianbarton:
Additionally, ZAP is not able to pick up/load certificates passed in at runtime e.g. using -Djavax.net.ssl.keystore.

victorbordo:
@thc202 Has anyone expressed interest in adding this functionality to the API recently?

sbrawner:
ME!

kingthorin:
@sbrawner Great we'll watch for your PR

sbrawner:
wait I meant I'd be interested in _having_ this functionality

kingthorin:
Awwwwww. Ok. Thanks for clarifying.

Though we are happy to accept contributions 👍 😁

Calbrilliant:
@kingthorin I've been looking into this and it seems like this functionality was purposefully disabled. Any idea why?

thc202:
@victorbordo not that I'm aware of.

@Calbrilliant which functionality specifically?

Calbrilliant:
The use of client certs. at least according my understanding of the following snippet in zaproxy/src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```
        // always turn off client cert
        setUseClientCert(false);
        setClientCertLocation("");
```

thc202:
That's because the cert options are not persisted (e.g. pin/pass, active cert...), it does not mean the cert functionality is disabled.

thc202:
@Calbrilliant if you are interested in working on this issue we can help/guide with what/where needs to be one.

Calbrilliant:
From what i can see that snippet prevents it from parsing any of the commandline options I am parsing. I would like to work on this issue, so your guidance would definitely be appreciated.

thc202:
Which command line options? If you want to add support through the command line (`-config` arg?), yes, that needs to be adjusted.

[...]

PR #4892, opened 2018-08-01
Allow to set client certificate from CLI

spoki0:
Added support for setting the client certificate from CLI, as well as persisting it.
Related to #2023, but that was mainly regarding the API interface.

I'm not too worried about using String for password, as that is being logged from the -config param anyway.


psiinon:
Thanks for this PR :)
Can you add some details of how this should be used?
Its using the -config cmdline option which means its essentially undocumented and therefore people wont know it there or how to use it. We can mitigate that with a FAQ or updating the help, so having something to cut-n-paste into one of those would be very useful. It will also help us when testing your PR ;)

spoki0:
It requires all 3 of the following parameters to be set, and should then enable the chosen certificate.
-config certificate.use=1
-config certificate.clientCertLocation=/path/to/certificate.p12 
-config certificate.clientCertPassword=certificate-password

The following flag will persist the settings such that next startup automatically load the settings.
-config certificate.persist=1

The latter one is mentioned in logs when client certificate is enabled by CLI, and will give a warning that the certificate password is being stored in configuration. If the persist flag is not set, the values will be cleared, and this is the default option.

I'm afraid I haven't done extensive testing on error handling, and mainly focused on a .p12 file, as that is my use-case in our CI/CD pipelines.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +70,80 @@ public OptionsParamCertificate() {
...
+
+                SSLContextManager contextManager = getSSLContextManager();
+                int ksIndex = contextManager.loadPKCS12Certificate(certPath, certPass);
+                contextManager.unlockKey(ksIndex, ksIndex, certPass);
+                contextManager.setDefaultKey(ksIndex, ksIndex);
+
+                setActiveCertificate();
+                setEnableCertificate(true);
+
+                logger.info("Client Certificate enabled from CLI");
```
kingthorin: Does the user always need to see this? Or would debug be more appropriate?

spoki0: It was meant to be informational on the same level as the other log outputs I see.

kingthorin: Fair enough, if anyone feels it's too "chatty" or whatever in the future it can always be changed.

spoki0:
Sadly, I couldn't really figure out how to change the GUI. That was my initial approach, to 'trigger' the GUI and set the client certificate that way. 

Any pointers on where and how I'd go about to check the boxes in the GUI?

kingthorin:
@spoki0 thanks for following up, sorry for the slow reply. I'll have to dig a bit, let me see what I can sort out.

thc202:
Yes, basically it just needs to do that (in `OptionsCertificatePanel.initParam`).

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -46,12 +52,15 @@
     private static final String CERTIFICATE_BASE_KEY = "certificate";
 
     private static final String USE_CLIENT_CERT = CERTIFICATE_BASE_KEY + ".use";
+    private static final String PERSIST_CLIENT_CERT = CERTIFICATE_BASE_KEY + ".persist";
     private static final String CLIENT_CERT_LOCATION = CERTIFICATE_BASE_KEY + ".clientCertLocation";
+    private static final String CLIENT_CERT_PASSWORD = CERTIFICATE_BASE_KEY + ".clientCertPassword";
```
thc202: It might be better to be more explicit and indicate which cert is being configured (`clientPkcs12CertLocation`, `clientPkcs12CertPassword`?), for if/when we happen to allow to specify PKCS#11 cert as well.

psiinon:
@spoki0 are you going to be able to apply the changes suggested?

spoki0:
@psiinon Yes, sorry for the delay.
The certificate should now show up in the GUI if set from CLI. It's a bit crude logic wise. Essentially check if the keystore contain a value on initialize, which it only should if set from CLI, and force update the UI when opening the panel.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
spoki0: From my understanding, PKCS#11 is an interface rather than a certificate+key combination. The GUI also request a driver and pin, rather than the certificates directly.

spoki0:
@kingthorin @psiinon @thc202 
Am I by chance missing a step to re-submit the pull-request for review?

kingthorin:
Nope I’d simply missed this somehow. I’ll try to test it this week.

kingthorin:
@spoki0 thanks for the new commit!

spoki0:
@kingthorin You're welcome.
I was sure I had fixed it previously, so I'm a bit annoyed at myself by now.

kingthorin:
Been there, no worries.

kingthorin:
Tested this again today, seems to be working as expected.

spoki0:
I assume @thc202 needs to accept the review as well? (or disagree with my comment)
Also, how do I go about updating the documentation with the following configuration properties?

> It requires all of the following parameters to be set, and should then enable the chosen certificate.
> -config certificate.use=1
> -config certificate.clientCertLocation=/path/to/certificate.p12
> -config certificate.clientCertPassword=certificate-password
> 
> The following flag will persist the settings such that next startup automatically load the settings.
> -config certificate.persist=1

kingthorin:
Two people from the core team to to review/approve it for merge.

This is the core help content that should be updated (separate PR) with those details:
- https://github.com/zaproxy/zap-core-help/blob/master/src/help/zaphelp/contents/ui/dialogs/options/certificate.html

(Which becomes - https://github.com/zaproxy/zap-core-help/wiki/HelpUiDialogsOptionsCertificate at release.)

kingthorin:
@psiinon @thc202 if this is good to go I can tackle the core help PR to help streamline getting this finished...

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +70,80 @@ public OptionsParamCertificate() {
...
         setAllowUnsafeSslRenegotiationSystemProperty(allowUnsafeSslRenegotiation);
     }
 
+    /**
+     *  Saves the client cert settings if the flag is set explicitly.
+     *  Only works for the CLI currently.
+     */
+    private void saveClientCertSettings(){
+
+        if (getInteger(PERSIST_CLIENT_CERT, 0) > 0){
```
thc202: This should be a boolean.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +70,80 @@ public OptionsParamCertificate() {
...
+        int enableClientCert = getInteger(USE_CLIENT_CERT, 0);
+        String certPath = getString(CLIENT_CERT_LOCATION, "");
+        String certPass = getString(CLIENT_CERT_PASSWORD, "");
+
+        if(enableClientCert!=0 && !certPath.isEmpty() && !certPass.isEmpty()) {
+            try {
+
+                SSLContextManager contextManager = getSSLContextManager();
+                int ksIndex = contextManager.loadPKCS12Certificate(certPath, certPass);
+                contextManager.unlockKey(ksIndex, ksIndex, certPass);
```
thc202: The actual cert used should be configurable as well.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +70,80 @@ public OptionsParamCertificate() {
...
+                contextManager.setDefaultKey(ksIndex, ksIndex);
+
+                setActiveCertificate();
+                setEnableCertificate(true);
+
+                logger.info("Client Certificate enabled from CLI");
+                logger.info("Use -config certificate.persist=1 to save settings");
+
+            } catch (IOException ex) {
```
thc202: These could be merged and use a more informative message (e.g. "The client certificate was not enabled because of an error."), the type is already logged.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +70,80 @@ public OptionsParamCertificate() {
...
+        }
+    }
+
+    /**
+     * Enables ClientCertificate from -config CLI parameters
+     * Requires location, password and a flag to use client certificate.
+     */
+    private void clientCertCheck(){
+
+        int enableClientCert = getInteger(USE_CLIENT_CERT, 0);
```
thc202: This can also be converted to a boolean, now that it will be in effective use.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
thc202: Yes, they are initialised differently but both provide certificates, IMO it's better if the keys are explicit (e.g. `certificate.pkcs12.path`, `certificate.pkcs12.password`, `certificate.pkcs12.index`, `certificate.pkcs11.driver`, `certificate.pkcs11.pin`, `certificate.pkcs11.index`...) than not. Also, it's easier for maintenance if they are under different elements.

kingthorin:
Rebased current and addressed feedback.

Arguments are now like:
```
-config certificate.use=true
-config certificate.pkcs12.path=/path/to/file.p12
-config certificate.pkcs12.password=WhateverThePasswordIs
-config certificate.persist=1
```

src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
```diff
@@ -991,10 +999,21 @@ private OptionsCertificatePanel getContentPane() {
 	public void initParam(Object obj) {
 		OptionsParam options = (OptionsParam) obj;
 		OptionsParamCertificate certParam = options.getCertificateParam();
-		useClientCertificateCheckBox.setSelected(certParam.isUseClientCert());
```
thc202: This can be kept (and remove in the if/else), it's being done always.

src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
```diff
@@ -991,10 +999,21 @@ private OptionsCertificatePanel getContentPane() {
 	public void initParam(Object obj) {
 		OptionsParam options = (OptionsParam) obj;
 		OptionsParamCertificate certParam = options.getCertificateParam();
-		useClientCertificateCheckBox.setSelected(certParam.isUseClientCert());
+		
+		// Should only run once after startup if client certificate is set from commandline
+		if(overrideEnableClientCertificate) {
+			certParam.setEnableCertificate(true);
```
thc202: We should not need to call this, the `OptionsParamCertificate` already enables the certificate.

src/org/parosproxy/paros/extension/option/OptionsParamCertificate.java
```diff
@@ -61,14 +72,75 @@ public OptionsParamCertificate() {
...
+                SSLContextManager contextManager = getSSLContextManager();
+                int ksIndex = contextManager.loadPKCS12Certificate(certPath, certPass);
+                contextManager.unlockKey(ksIndex, certIndex, certPass);
+                contextManager.setDefaultKey(ksIndex, certIndex);
+
+                setActiveCertificate();
+                setEnableCertificate(true);
+
+                logger.info("Client Certificate enabled from CLI");
+                logger.info("Use -config certificate.persist=1 to save settings");
```
thc202: This needs update, `true` instead of `1`. We could also omit the log if already persisting?

src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
thc202: Sorry for overlooking this (and following comment) in the previous review.

thc202:
Commits can be squashed.

src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
kingthorin: Not a problem at all

src/org/parosproxy/paros/extension/option/OptionsCertificatePanel.java
thc202: From talks in IRC, this does need to be called (for when persist is false).

psiinon:
Yeah, many thanks @spoki0 and @kingthorin !

thc202:
@spoki0 how would you like to be credited?

spoki0:
Nice, though I didn't have time to fix the code as according to the review feedback these past two days, I hope my contribution was still of some use. Also, thanks for the help and pointers in adding this feature.

@thc202 
Jonathan Ness
Or would I need to do a pull request and update it that way?

kingthorin:
We'll take care of it, and don't worry we know how life gets busy. Thanks for the contribution!

PR #4898, opened 2018-08-02
core changes required for authentication helper add on v1.1.0

KajanM:
- [x] required changes for guided configuration
- [x] required changes for authentication status scanning
- [x] required changes for automated configuration

Above changes are parallelly applied and manually tested in authenticationhelper add-on v2.
You can check them out at
- https://github.com/KajanM/zap-extensions/tree/authenticationhelper-addon-1.1.0
- https://github.com/KajanM/zap-extensions/tree/authenticationhelper-v2-with-updated-zap

Some changes are not used by the add-on, however, they are added assuming they may help other developers.

PR #4900, opened 2018-08-03
Validate index/state when setting custom tab panel

thc202:
Change StandardFieldsDialog to validate the tab index and dialogue state
when setting a custom tab panel.

---
Spotted while checking #4898.


psiinon:
Not merging yet pending LGTM checks...

PR #4922, opened 2018-08-14
Provide ZAP as a snap

### LINKED ISSUE #4603: Provide ZAP as snap
It would be nice if we could provide ZAP via snap package manager (https://snapcraft.io)

One benefit is, that we could provide different channels:
* weekly
* stable
....may be more

psiinon:
Fixes #4603 

Note that this doesnt specify any channels, just defines the 2.7.0 snap.
Its currently defined as grade: devel : https://docs.snapcraft.io/build-snaps/upload


psiinon:
Just remembered I meant to add a ZAP icon, will update this PR..

psiinon:
Just added the icon and changed it so that it applies to all architectures.

OK to publish the snap on https://snapcraft.io/store?

psiinon:
The snap was rejected with one error when I uploaded it :(
I've asked a question on the forum: https://forum.snapcraft.io/t/how-can-you-include-java-in-a-snap/6877

psiinon:
Changed this to not specify the architecture, as that was failing as we're including the JDK.
The snap I've generated using this has passed review and has been published: https://snapcraft.io/zaproxy
I also changed the directory to just 'snap' in case we want to use Lauchpad to create the snaps: "Launchpad can build snap packages using snapcraft, from any Git or Bazaar branch on Launchpad that has a snap/snapcraft.yaml, snapcraft.yaml, or .snapcraft.yaml file at its top level. "

snap/README.md
```diff
@@ -0,0 +1,39 @@
+This directory contains all of the files required to build a ZAP 2.7.0 https://snapcraft.io/ snap.
+
+Docker is used to build the snap as snapcraft is only supported on a limited number of linux distros.
+
+In order to build the snap run:
+
+`./build-snap.sh`
```
thc202: We could use `docker run -it -v "$PWD:$PWD" -w "$PWD" snapcore/snapcraft snapcraft` directly, which builds the snap in the local file system (in this case we should add the snap dirs and the snap file to `.gitignore`).

snap/snapcraft.yaml
```diff
@@ -0,0 +1,32 @@
...
+apps:
+  zaproxy:
+    command: desktop-launch $SNAP/wrapper
+
+parts:
+  zaproxy:
+    after: [desktop-glib-only]
+    source: https://github.com/zaproxy/zaproxy/releases/download/2.7.0/ZAP_2.7.0_Linux.tar.gz
+    plugin: dump
+    build: |
```
thc202: The `parts` could be simplified if we use the [jdk plugin](https://github.com/snapcore/snapcraft/blob/master/snapcraft/plugins/jdk.py), e.g.:
```yaml
parts:
  java:
    plugin: jdk
    source: .
    build-packages:
      - ca-certificates-java
  zaproxy:
    after: [desktop-glib-only]
    plugin: dump
    source: https://github.com/zaproxy/zaproxy/releases/download/2.7.0/ZAP_2.7.0_Linux.tar.gz
```
we wouldn't need to specify additional packages and the `wrapper` would no longer be needed (the env variables are already set).

psiinon:
Updated based on feedback.
I was hoping that build-snap.sh could call the docker image with build-snap-in-docker.sh but I couldnt get that to work :/ Suggestions for fixing that appreciated :)

thc202:
Thank you!

Issue #4933, opened 2018-08-16
New/Updated Global Excludes


kingthorin:
I accidentally left Firefox running overnight, with ZAP set as proxy, it was left with a single tab open and benign content:
![image](https://user-images.githubusercontent.com/7570458/44214936-3cced680-a13f-11e8-98ec-918d27aae009.png)

Earlier yesterday I'd added two new Global Excludes:
* `^https?://.*\.cdn\.mozilla\.(org|net|com)/.*$` - `Site - Mozilla CDN (requests such as getpocket)`
* `^https?://.*\.telemetry\.mozilla\.(org|net|com)/.*$` - `Site - Mozilla Firefox browser telemetry`

Could use updating:
* `^https?://(safebrowsing-cache|sb-ssl|sb|safebrowsing\.clients)\.google\.com/.*$` - `Site - Google malware detector updates`
  * I'd suggest `^https?://(?:safebrowsing-cache|sb-ssl|sb|safebrowsing).*\.(?:google|googleapis)\.com/.*$`

From the screenshot above, it seems that the following could be added:
* *.services.mozilla.com
* *.adblockplus.org
* *.gvt1.com (this seems to be Google related)

Should the Regex for global excludes be using non-capturing groups? [`(?:)`] (Is there efficiency to be gained there?) 
Also many of the patterns use `|` so perhaps re2j should be leveraged here too?

thc202:
Worth doing some tests, but in this case it's not a big problem I guess (compared to responses, the URLs are not that big).

kingthorin:
Good point. Shall we go ahead and close this one then?

On a separate but related note are we ready to stop calling this functionality Beta?

thc202:
I'm good with that.

Yes, it's about time! :)

PR #4936, opened 2018-08-16
Global Excludes Additions and Updates

kingthorin:
Per zaproxy/zaproxy#4933

- Added a few new entries.
- Updated the google safebrowsing entry.
- Made groups non-capturing.
- Changed the ordering in the tld groups for some Mozilla patterns (making the first entry `com` as it seems to be more relevant based on my recent Firefox experiences)


thc202:
Thanks!

Issue #3522, opened 2017-05-11
Enhancement: Allow Search to be Case Sensitive


davewichers:
I'd like all the search options to support case sensitive search. Insensitive by default is fine. This includes the search tab at the bottom, and search within a specific request/response or anywhere else search is supported.

On a related note, do we support regex based search? If not, that would be good too, but harder and probably deserves a separate ticket if we decide its worth it.

thc202:
Search tab supports regular expressions, the "Find" dialogues usually don't.

thc202:
(And some of the "Find" dialogues allow to specify if the match is case sensitive or not.)

davewichers:
OK. Sounds like we are really close. Can we just then make sure that ALL the search functions allow the user to specify Case Sensitive search? Ideally, they'd all just use the same dialog where possible so they'd be the same by definition.

thc202:
Sure. Right, we should normalise that.

PR #4938, opened 2018-08-17
Show message in Find dialogue if string not found

### LINKED ISSUE #4935: Minor 'find' enhancement - when searching in request/response
When find doesn't find anything in a request or response, it doesn't say anything. The only thing I notice it does is scroll back to the top of the request or response (whatever you are searching in). And sometimes that's not obvious.

Can you enhance the behavior of find within text fields like this so that if it doesn't find anything, it explicitly indicates that.

thc202:
Change FindDialog to show a message (in the dialogue) if the string was
not found, also, remove hardcoded sizes to allow to resize the dialogue
at will and log the exception (instead of printing).

Fix #4935.


davewichers:
Excellent! And thanks for being so fast! I Previously asked for case sensitive search (#3522). Any chance you could knock that out too? At least for the Find dialog, since you were just in that code?

PR #4943, opened 2018-08-21
Global Excludes No Longer Beta

kingthorin:
Per zaproxy/zaproxy#4933


thc202:
Some help pages need to be updated.

kingthorin:
Yup core help update coming too.

thc202:
Thanks!

PR #4962, opened 2018-08-30
Persist History tab filter

### LINKED ISSUE #6141: Search Filter configuration is not loaded from persisted session if ZAP was exited.
**Describe the bug**
If user configured Search tab Filter and the persisted the session, the Filter configuration can be reloaded correctly only if ZAP wasn't closed. If user closes and restarts the ZAP, then opens the persisted session, the Filter configuration will be empty. As loading configuration works while not exiting, this is not a feature request, this is bug.

**To Reproduce**
1. Start ZAP
2. Navigate any site with .css .js resources.
3. Use the following regexes in the exclusion filter of the History tab.
.*\.js
.*\.js\?.*
.*\.mp3
.*\.css.*
.*\.woff.*
.*\.ttf.*
.*\.eot.*
.*\.otf.*
.*\.svg.*
.*\.map.*
.*\.mp4.*
4. Persist the session
5. Exit the ZAP
6. Restart the ZAP
7. Open the persisted session
8. Filter will be OFF.

**Expected behavior**
As filter can be saved to persisted session file and loaded, it should also load after restarting the ZAP too.

**Screenshots**
%pn-2020-08-25-13-11-34

**Software versions**
- ZAP 2.9.0 w2020-08-17
- OS: Windows 10 x64 2004
- Java: Oracle JRE 1.8.0_251 both x32 x64

I'm not sure if this error is connected to filter loading.
[log.txt](https://github.com/zaproxy/zaproxy/files/5123062/log.txt)

**Additional context**
Reproduces since long ago.
Doesn't occur if ZAP wasn't quit and restarted.

Yes

denniskniep:
This PR adds the possibility to store and load the Filter in the HistoryTab. 

There is a special Filter "_Set_Date_To_Now", which adds the current time to the current filter. This is good for temp. excluding all existing requests in the HistoryTab.

image

filterload

Fix #6141.


kingthorin:
I got to test this today, overall it seems good to me. Two quick bits of feedback:
1. It seems like the filter dialog could use a call to `pack` (ex: when you click the funnel the first time, the two "Alert" fields come up with scroll bars).
1. If a Filter is "saved" (named) when the dialog is closed ("Apply" clicked) should the dropdown in the history tab be set to that new filter name? [Seems logical to me, I'm not sure how much of a pain in the rear it might be to do so programmatically.]

kingthorin:
@denniskniep any thoughts on those two items?

denniskniep:
> 2. If a Filter is "saved" (named) when the dialog is closed ("Apply" clicked) should the dropdown in the history tab be set to that new filter name? [Seems logical to me, I'm not sure how much of a pain in the rear it might be to do so programmatically.]

Problem with that is that the Filter "_Set_Date_To_Now" adds only the current date to the filter and leave the other fields as they are. 
Furthermore you could modify any filter in the dialog which you applied. Then it is no more euqal to the named filter from the combobox. 

Sorry for the late reply

denniskniep:
@thc202 @kingthorin @psiinon 
Would this PR be merged if it is changed to the correct structure (Is it worth to migrate)?

psiinon:
I'm happy to see it merged :)

thc202:
Changed to target `main`.

PR #4963, opened 2018-08-30
Allow to specify the URI of the login page

KajanM:
addressing https://github.com/zaproxy/zaproxy/pull/4696#discussion_r210227296

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/SessionDialog.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/SessionDialog.java
@@ -71,7 +71,7 @@ public class SessionDialog extends AbstractParamDialog {
     /** This method initializes this dialog. */
     private void initialize() {
         if (Model.getSingleton().getOptionsParam().getViewParam().getWmUiHandlingOption() == 0) {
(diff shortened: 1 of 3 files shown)


src/lang/Messages.properties
```diff
@@ -482,6 +482,7 @@ authentication.method.jb.name = JSON-based Authentication
 authentication.method.jb.popup.login.request = {0} : JSON-based Auth Login Request
 authentication.method.pb.field.label.postData				= Login Request POST Data (if any):
 authentication.method.pb.field.label.postDataRequired = Login Request POST Data *:
+authentication.method.pb.field.lable.loginFormUrl			= URL to GET login page(form) *:
```
thc202: The field is not mandatory, it's possible to leave it empty which leads to issues later when creating the message. Maybe it shouldn't be? If not needed the user can clear it?

thc202:

thc202:
Addressed the issue, it will now fallback to the login URL if the page URL is not provided. (Renamed some variables/methods for consistency, use "login page url" everywhere.)

thc202:
Thanks!

PR #4971, opened 2018-09-05
Made active scan progress consistent with spider, show progress of 100 if scan is stopped

rezen:
Currently, if you stop an active scan, the API will report the progress as the percent it was stopped at. This behavior is not consistent with the spider, which will report the progress status of 100 in the API if it was stopped. This PR addresses that inconsistency.


kingthorin:
Relevant spider snippet:
https://github.com/zaproxy/zaproxy/blob/13772bb0221ed711201689bccd3bd0d03b6d1860/src/org/zaproxy/zap/extension/spider/SpiderAPI.java#L554-L561

thc202:
Also, as pointed out earlier by @rezen, this prevents infinite loops in the API consumers (e.g. `zap-full-scan.py`) when the active scan times out (e.g. max duration) as they base the logic on the scan progress (related spider change #2469).

thc202:
Thanks!

PR #4996, opened 2018-09-21
Prevent duplicated API elements

thc202:
Change ApiImplementor to validate that the ApiElement (view, action, or
other) being added does not use a duplicated name, it would not work
properly.

### DIFF
--- a/src/org/zaproxy/zap/extension/api/ApiImplementor.java
+++ b/src/org/zaproxy/zap/extension/api/ApiImplementor.java
@@ -83,14 +83,33 @@ public abstract class ApiImplementor {
 	}
 
 	public void addApiView (ApiView view) {
+		validateApiElement(apiViews, view);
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/api/ApiImplementor.java
```diff
@@ -83,14 +83,33 @@ public int compare(Method method, Method otherMethod) {
 	}
 
 	public void addApiView (ApiView view) {
+		validateApiElement(apiViews, view);
 		this.apiViews.add(view);
 	}
+
+	/**
+	 * Validates that the name of the given {@code ApiElement} does not already exists in the given list of {@code ApiElement}.
```
kingthorin: The `s` can be dropped from the end of `exist`.

thc202:
Thanks, change done.

Issue #4781, opened 2018-06-21
Tasks for 2.8.0 release


thc202:
Tracker issue for tasks to be done immediately before/after the 2.8.0 release.

  - [x] Update [WebSwing](https://github.com/zaproxy/zaproxy/wiki/WebSwing) wiki page, to have both images use the same link (that is, the one from the weekly) to access ZAP.
  - [x] Tweak https://github.com/zaproxy/zaproxy/wiki/FAQAddonVersions, per https://github.com/zaproxy/zaproxy/pull/5010#issuecomment-425134612.
  - Done in:
    - https://github.com/zaproxy/zaproxy/wiki/FAQAddonVersions/_compare/2184733%5E...2184733
    - https://github.com/zaproxy/zaproxy/wiki/FAQAddonVersions/_compare/4bd81e7%5E...4bd81e7

PR #5010, opened 2018-09-27
Show Version in Marketplace Installed Addons Tab

kingthorin:
I noticed this morning that the hover text shows the details of the update if one is available. I believe this change makes things clearer in the dialog and will also simplify: 
https://github.com/zaproxy/zaproxy/wiki/FAQAddonVersions

Before:
image

After:
image


kingthorin:
Note: I'll update the FAQ (incl. screenshots) after/if this is merged.

thc202:
Looks good. Would you mind adding a task in #4781?

PR #5027, opened 2018-10-03
Restore info message in Input Vectors tab

thc202:
Restore message lost in previous changes, change CustomScanDialog to set
the reason why the Input Vectors are disabled to OptionsVariantPanel.
Change OptionsVariantPanel to show the reason when the variants are
disabled.


thc202:
(Lost in a refactoring 27f60de5c26a4ced647b1ee2bc0e65292a2ec2d4.)

PR #5034, opened 2018-10-05
Add button to install all add-ons from marketplace

thc202:
Add a button, shown in dev mode only, to install all add-ons available
in the marketplace, to not require selecting them one by one.


psiinon:
Out of interest whats the use case for this?
I know we have a command line option for this, but that was intended more for testing than anything else.
Has anyone asked for this?

thc202:
Personal use mainly, I usually want to install all of them when trying changes that might affect the add-ons (and this changes makes it a lot faster).

psiinon:
Maybe only enable this in dev mode then?
I cant see it being very useful for most ZAP users - the UI is complicated enough without them having absolutely everything installed :/

thc202:
Yeah, change done.

PR #5082, opened 2018-10-26
Fixes 5065; Listening for changes of the message in the Request/Response panels

### LINKED ISSUE #5065: Listening for changes of the message in the Request/Response panels
I want to get a notification if the HttpMessage in the HttpPanel is changed. For example, the user selects a new message in the history tab.

Maybe we could add a subscribe mechanism in the `View` class. And every time the method `displayMessage` is called all listeners will be notified.

Other suggestions are welcome

denniskniep:
Fixes #5065.

### DIFF
--- a/src/org/parosproxy/paros/extension/ExtensionHookView.java
+++ b/src/org/parosproxy/paros/extension/ExtensionHookView.java
@@ -31,6 +31,7 @@ import java.util.Collections;
 import java.util.List;
 
 import org.parosproxy.paros.view.AbstractParamPanel;
+import org.zaproxy.zap.extension.httppanel.DisplayedMessageChangedListener;
 import org.zaproxy.zap.view.ContextPanelFactory;
 
 /**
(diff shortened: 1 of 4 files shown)


src/org/parosproxy/paros/view/View.java
```diff
@@ -1048,6 +1049,22 @@ public void displayMessage(Message message) {
         }
     }
 
+    public void addRequestPanelMessageChangedListener(MessageChangedListener messageChangedListener){
```
thc202: Lets expose this through the `ExtensionHookView` instead, less changes for the add-ons that use it.

src/org/zaproxy/zap/extension/httppanel/MessageChangedListener.java
```diff
@@ -0,0 +1,27 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.zaproxy.zap.extension.httppanel;
+
+public interface MessageChangedListener {
```
thc202: - I'd suggest renaming it to `MessageDisplayedListener`(?), to avoid any misunderstandings with the editable panels (where the messages themselves can be changed).
 - Worth adding a short description of the listener (and `@since` tag), it should also link to `HttpPanel`'s method that allows to add the listener and mention the `ExtensionHookView`.

src/org/zaproxy/zap/extension/httppanel/MessageChangedListener.java
```diff
@@ -0,0 +1,27 @@
...
+
+    void messageChanged(Message oldMessage, Message newMessage);
```
thc202: The method should be documented, specially when/that the messages are null.

thc202: (This would also have to be renamed, `messageDisplayed`?)

src/org/zaproxy/zap/extension/httppanel/MessageChangedListener.java
denniskniep: Is
```suggestion
public interface DisplayedMessageChangedListener {
```
okay for you?

thc202: Yeah, that's clear too.

denniskniep:
@thc202 resolved your review comments

thc202:
Review comments addressed.

thc202:
Thanks!

PR #5085, opened 2018-10-28
Persist column configuration in History tab

### LINKED ISSUE #4788: Persist column configuration in History tab
First: Thank you very much for your work on ZAP!

**Describe the bug**

The HTTP request history tab has several columns. Although by default most columns are shown, users can configure which columns are displayed and in which order. This  information is lost when ZAP is quit, so users have to do this configuration every time.

**To Reproduce**
1. Start ZAP
1. Click on the History tab
1. Configure the columns by clicking in the header row on the rightmost icon
1. Disable the "Id" column, it isn't shown any more
1. Move the Request column to the front
1. Quit ZAP
1. Start ZAP again
1. "Id" column is shown again, Method column is third column, all configuration for the columns was reset

**Expected behavior**

Column configuration for History tab should be the same after restart.

**Software versions**
 - ZAP: 2.7.0
 - OS: Debian stable 9.4 ("stretch")
 - Java: openjdk 1.9, Debian version number: `9~b181-4~bpo9+1`, package `openjdk-9-jre`

dafenix:
Fix #4788.

ToDo
- [x]  store column position in zap configuration file
- [x]  remember hidden columns after restart
- [x]  restore column position in table
- [x]  change persistance to default behaviour to avoid build issues with jaxb  
- [x]  implement "reset to default" for table columns in column configuration popup
- [x] restore column configuration even if current language changes

image

image


dafenix:
@kingthorin Ok i take care of it. Can you please assign this issue to me?

kingthorin:
This is a PR you opened not an issue 😀

kingthorin:
Disregard, PRs are assignable too 😀 just had to switch from the mobile site to the desktop site so I could see the functionality.

kingthorin:
@dafenix there's still a bunch of style things like ifs without curly bracers, annotations on same lines instead of line before, etc. Would you rather I pulled the code and tweaked it (then you'd have to fetch from origin and build on top if there's other stuff), or would you rather I just review it and you make the changes?

dafenix:
@kingthorin sorry for that i will fix the remaining styling issues.

dafenix:
I have cleaned what i found. Please let me know if i missed anything.

kingthorin:
I've updated this PR to account for the new repo format. The change seems to work as expected. The only thing I couldn't sort out is there the reset functionality comes into play.

@dafenix (or anyone else) if you're going to make further changes please ensure you fetch/reset from origin to get the updated branch before making further modifications (don't want to clobber it with your old local branch).

thc202:
The following change was lost during the update:
[develop...dafenix:b5d5f3fad212e60d6a81ddf62505029ee13676ba#diff-0e82862e82f895ae1ce2db9b2312fc61R131](https://github.com/zaproxy/zaproxy/compare/develop...dafenix:b5d5f3fad212e60d6a81ddf62505029ee13676ba#diff-0e82862e82f895ae1ce2db9b2312fc61R131)

kingthorin:
Thanks! That did it.

thc202:
This needs to documentation/JavaDoc (it's easy to get the order of statements wrong or forget to persist the config), ideally we should change the API to be less error prone.

zap/src/main/java/org/parosproxy/paros/extension/history/HistoryTable.java
```diff
@@ -34,7 +34,14 @@ public HistoryTable() {
         setAutoCreateColumnsFromModel(false);
 
         setName("History Table");
+        super.loadColumnConfiguration("HistoryTable");
+        if (this.config == null) {
```
thc202: This check could be done in `loadColumnConfiguration`, less checks needed by extending classes and we could also remove the instance variable `config` (which seems to be used only to signal that the persistence was previously done).

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -64,6 +70,13 @@
     private boolean autoScroll;
     private AutoScrollAction autoScrollAction;
     private StickyScrollbarAdjustmentListener autoScrollScrollbarAdjustmentListener;
+    private String persistanceIdentifier;
+    protected TableColumnConfiguration config = null;
+    private static final String COLUMN_NAME = "ColumnName";
```
thc202: These constants could be moved to after `serialVersionUID`.

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -76,6 +89,36 @@ public ZapTable(TableModel dataModel) {
...
+        TableColumnConfiguration config =
+                TableColumnConfiguration.loadFromConfig(persistanceIdentifier);
+        this.persistanceIdentifier = persistanceIdentifier;
+        this.config = config;
+        if (config != null) {
+            config.apply(this);
+        }
+    }
+
+    protected void applyAdditionalColumnConfigurations() {}
```
thc202: `applyDefaultColumnConfigurations`

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -334,4 +378,224 @@ public void actionPerformed(ActionEvent e) {
...
+            if (tableColumnConfig.getColumnConfigs().size() == 0) {
+                return null;
+            }
+            return tableColumnConfig;
+        }
+
+        private static String getElementBaseKey(String persistanceIdentifier, int index) {
+            return COLUMN_CONFIGURATION
+                    + persistanceIdentifier
+                    + "(0)."
```
thc202: This index could be removed (there's only one).

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -334,4 +378,224 @@ public void actionPerformed(ActionEvent e) {
...
+            TableColumnConfiguration tableColumnConfig = loadFromConfig(persistanceIdentifier);
+            if (tableColumnConfig == null) {
+                return;
+            }
+            HierarchicalConfiguration config = (HierarchicalConfiguration) getConfig();
+            config.clearTree(COLUMN_CONFIGURATION + persistanceIdentifier);
+        }
+    }
+
+    static class TableColumnConfig {
```
thc202: `private`?

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -76,6 +89,36 @@ public ZapTable(TableModel dataModel) {
         init();
     }
 
+    public void persistColumnConfiguration() {
+        if (StringUtils.isEmpty(this.persistanceIdentifier)) {
+            return;
+        }
+        config = new TableColumnConfiguration();
+        List<javax.swing.table.TableColumn> columns =
```
thc202: The package could be removed, already being imported (same in other places).

zap/src/main/java/org/zaproxy/zap/view/ZapTable.java
```diff
@@ -334,4 +378,224 @@ public void actionPerformed(ActionEvent e) {
...
+        }
+
+        @Override
+        public void actionPerformed(ActionEvent e) {
+            table.createDefaultColumnsFromModel();
+            table.applyAdditionalColumnConfigurations();
+        }
+    }
+
+    public static class TableColumnConfiguration {
```
thc202: `private`?

sbzo:
Thanks @dafenix and @thc202 for the changes. Anyone knows when this changes will be pushed to ZAP_D for testing?. This changes are very helpful for many users. 

Best regards,

RLobo.

thc202:

kingthorin:
Rebased current and addressed a bunch of the previous review comments.

psiinon:
Whats left to do on this PR?

thc202:
For a start this https://github.com/zaproxy/zaproxy/pull/5085#issuecomment-554361072

thc202:
We talked about this in IRC and the plan is to merge as is and follow up with other PRs as needed.

psiinon (approved):
Works for me!

thc202:
Thank you @dafenix!

Issue #5044, opened 2018-10-10
NodeJS API Upgrade


binarymist:
Working on this now:
https://github.com/zaproxy/zaproxy/blob/develop/src/org/zaproxy/zap/extension/api/NodeJSAPIGenerator.java
https://www.npmjs.com/package/zaproxy

psiinon:
Great :)
FYI we'd like to move the NodeJS API to https://github.com/zaproxy/zap-api-nodejs :)
The NodeJSAPIGenerator can stay where it is though.

binarymist:
Yip, changes (first round) have already been made in source. Just need to run the generator.

binarymist:
Any idea why there is no `issues` tab on [zap-api-nodejs](https://github.com/zaproxy/zap-api-nodejs/issues)?

kingthorin:
https://github.com/zaproxy/zap-api-nodejs/blob/master/LICENSE
https://github.com/zaproxy/zaproxy/blob/develop/LICENSE

Both Apache License 2.0???

binarymist:
Yes, those two are @kingthorin , but the License in the link I provided above is MIT

thc202:
Issues are not enabled since we manage them here, in the main repo. That's the same for all API clients.

thc202:
I don't know why MIT was chosen.

kingthorin:
> Yes, those two are @kingthorin , but the License in the link I provided above is MIT

Weird, maybe if you dig through the history/blame there's something about it...

binarymist:
Todo future PR: @binarymist:
  - [ ] Remove zaproxy/nodejs files
  - [ ] Review of all generated files

thc202:
The repo has been added to snyk.

binarymist:
Can you please add the snyk badge to the README @thc202 ?

thc202:
Sure.

binarymist:
Then I guess:  
1. `npm version 1.0.0-rc.1`
2. `npm publish`
3. Update `CHANGELOG.md`

psiinon:
0. Add to https://lgtm.com/ (done)
4. Publicise via ZAP groups, twitter etc :)

psiinon:
lgtm has analysed the code now. Looks good until you show the files it excludes: https://lgtm.com/projects/g/zaproxy/zap-api-nodejs/latest/files/?sort=name&dir=ASC&mode=heatmap&showExcluded=true
[...]

PR #5088, opened 2018-10-29
Node api generation

binarymist:
- [x] Provided backwards compatibility for those wishing to continue using callbacks, while those wishing to take advantage of promises can now do so
  - [x] The apiKey is now only required on the initial ZapClient instantiation

Addresses #5044

### DIFF
--- a/src/org/zaproxy/zap/extension/api/NodeJSAPIGenerator.java
+++ b/src/org/zaproxy/zap/extension/api/NodeJSAPIGenerator.java
@@ -31,12 +31,17 @@ import java.util.ResourceBundle;
 
 public class NodeJSAPIGenerator extends AbstractAPIGenerator {
     
+    /**
+	 * Default output directory in zap-api-nodejs project.
+	 */
+    private static final String DEFAULT_OUTPUT_DIR = "../zap-api-nodejs/src/zapv2/";
+
     private final String HEADER = 
             "/* Zed Attack Proxy (ZAP) and its related class files.\n" +
             " *\n" +
             " * ZAP is an HTTP/HTTPS proxy for assessing web application security.\n" +
             " *\n" +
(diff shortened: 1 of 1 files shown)


binarymist:
Switching to tag 2.7.0 as suggested in https://github.com/zaproxy/zap-api-nodejs/pull/1#issuecomment-433961996 yields:

thc202:
Are all the libraries under the lib directory in the module's dependencies?

binarymist:
I haven't removed any

thc202:
Right, but you might need to add.

src/org/zaproxy/zap/extension/api/NodeJSAPIGenerator.java
```diff
@@ -31,12 +32,17 @@
 
 public class NodeJSAPIGenerator extends AbstractAPIGenerator {
     
+    /**
+	 * Default output directory in zap-api-nodejs project.
+	 */
+    private static final String DEFAULT_OUTPUT_DIR = "../zap-api-nodejs/src/zapv2/";
```
thc202: Is the path correct? In zaproxy/zap-api-nodejs#1 the files are being added directly to `src` dir (i.e. not to `src/zapv2`).

binarymist: Yes, I didn't see any reason for a zapv2 directory, that's what versioning is for, unless there is something I've missed?

thc202: I'm good with that, we should drop the `zapv2` from this path then.

binarymist: Just realised I missed this, thought I'd got it, but obviously hadn't
fe8074c

thc202:
Thanks!

PR #5089, opened 2018-10-29
Include Alert API when generating API clients

thc202:
Change ApiGeneratorUtils to include the Alert API, introduced in #4833.


psiinon:
@kingthorin this look ok?

PR #5091, opened 2018-10-30
Update AntiCsrfParam.java

kurobeats:
Added a common anti-csrf token


kingthorin:
Thanks

thc202:
Existing configurations will not benefit from this change unless the options are reset, does it worth adding the token when updating the config?

Issue #5060, opened 2018-10-16
Custom Vectors don't work properly if plugin ID is injected in the message


thc202:
From OWASP ZAP User Group: https://groups.google.com/d/topic/zaproxy-users/oHT-BMHwyM0/discussion

Steps to reproduce the issue:
1. Start ZAP and proxy a POST request;
2. Enable the option "Active Scan" > "Inject plugin ID in header for all active scan requests";
3. Attack the POST request with a Custom Vector in the request body;
4. Note that the payload is not injected in the selected position.

Also, depending on the position selected it might lead to exceptions, e.g.:
```
[ZAP-ActiveScanner-0] ERROR org.parosproxy.paros.core.scanner.VariantUserDefined  - 
org.parosproxy.paros.network.HttpMalformedHeaderException
	at org.parosproxy.paros.network.HttpHeader.setMessage(HttpHeader.java:183)
	at org.parosproxy.paros.network.HttpRequestHeader.setMessage(HttpRequestHeader.java:256)
	at org.parosproxy.paros.network.HttpRequestHeader.setMessage(HttpRequestHeader.java:282)
	at org.parosproxy.paros.core.scanner.VariantUserDefined.setParameter(VariantUserDefined.java:90)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.setParameter(AbstractAppParamPlugin.java:286)
	at org.zaproxy.zap.extension.ascanrules.TestCrossSiteScriptV2.scan(TestCrossSiteScriptV2.java:168)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scan(AbstractAppParamPlugin.java:272)
	at org.zaproxy.zap.extension.ascanrules.TestCrossSiteScriptV2.scan(TestCrossSiteScriptV2.java:113)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scanVariant(AbstractAppParamPlugin.java:200)
	at org.parosproxy.paros.core.scanner.AbstractAppParamPlugin.scan(AbstractAppParamPlugin.java:171)
	at org.parosproxy.paros.core.scanner.AbstractPlugin.run(AbstractPlugin.java:341)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

ZAP
Version 2.7.0 (and older).

thc202:
Fixed in #5092.

kingthorin:
[...]

PR #5092, opened 2018-10-30
Fix misaligned custom vectors indices

setzamora:
This is related to #5060 

The injection of the X-ZAP-Scan-ID header happens at the `init` phase of `AbstractPlugin` altering the payload before setting the parameters. 

I moved it to the `sendAndReceive` where it is called after the `setParameter`. As a result, the parameters are now interpolated correctly.

Cheers

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -148,9 +148,6 @@ public abstract class AbstractPlugin implements Plugin, Comparable<Object> {
     public void init(HttpMessage msg, HostProcess parent) {
         this.msg = msg.cloneAll();
         this.parent = parent;
-        if (this.parent.getScannerParam().isInjectPluginIdInHeader()) {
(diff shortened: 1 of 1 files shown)


src/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -272,6 +269,11 @@ protected void sendAndReceive(HttpMessage message, boolean isFollowRedirect, boo
                 }
             }
         }
+        
+        //ZAP: Set the ZAP scan id header here instead at the init phase
```
thc202: We should not add ZAP comments in the code, they should be after the license header (in this case seems unnecessary the existing one already covers it).

setzamora: @thc202 

This is noted. I just updated the pull request.

thc202:
Thanks!

thc202:

setzamora:
> How would you like to be credited?
> https://github.com/zaproxy/zap-core-help/wiki/HelpCredits

Joset Zamora (@setzamora)

Thanks @thc202

PR #5096, opened 2018-10-31
Removed superfluous old node api files

binarymist:
Removed api files as discussed in IRC with @thc202, and checklist item on #5044


thc202:
Thanks!

PR #5114, opened 2018-11-12
Correct -m flag description in full scan script

rezen:
Currently the default spider duration is unbound and it should be 1 minute (per the usage flag).

### DIFF
--- a/docker/zap-full-scan.py
+++ b/docker/zap-full-scan.py
@@ -80,7 +80,7 @@ def usage():
     print('    -c config_file    config file to use to INFO, IGNORE or FAIL warnings')
     print('    -u config_url     URL of config file to use to INFO, IGNORE or FAIL warnings')
     print('    -g gen_file       generate default config file(all rules set to WARN)')
-    print('    -m mins           the number of minutes to spider for (default 1)')
+    print('    -m mins           the number of minutes to spider for (defaults to no limit)')
     print('    -r report_html    file to write the full ZAP HTML report')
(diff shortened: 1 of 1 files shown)


thc202:
It should be updated the help instead, in the full scan script the default is to leave the spider unconstrained.

rezen:
Ah, okay, I wasn't sure if the flag or the default was wrong. I wonder if having a reasonable constraint might be good from a new user experience perspective. If a new user runs a scan without a configuring context structure for example, the spider could go infinitely. On the other hand, new users probably aren't running the full-scan?

rezen:
I was testing out the full scan (on Juice Shop) and kept running into `java.lang.OutOfMemoryError: Java heap space` and then realized ah-ha! the spider wasn't stopping which led me to check out the script.

kingthorin:
@resend are you going to modify as per @thc202's suggestion?

thc202:
Changed to correct the description instead of changing the default value. @rezen please raise an issue if you still think that the default needs to be changed.

kingthorin:
Thanks @thc202 and @rezen

cudandres:
Playing catchup! I think from the average user perspective, having time bounds on the spider is friendlier for usability?

rezen:
Playing catchup! I think from the average user perspective, having time bounds on the spider is friendlier for usability?

psiinon:
I'm not so sure. "Full scan" to me implies no time limit, so having to change from the to make it a really full scan sounds counter intuitive.

PR #5119, opened 2018-11-14
Fix and tweaks to global excludes

### LINKED ISSUE #5118: Global Excludes Modify Description Doesn't Work
**Describe the bug**
If you modify an existing Global Exclude to update the description the changes are not accepted. There's no error but they just don't stick.

**To Reproduce**
1. Launch ZAP. (2.7.0 or latest Dev build)
2. Go into Options and "Global Exclude URL" [With or without [Beta] tag depending on build]
3. Select an existing entry, click the Modify... button.
4. Add some text to the "Description" and click the Modify button.
5. Back in the Options panel note that no update occurred. (Repeat step 3 with the same entry, note the description wasn't change... so not just a table update issue, the change was seemingly discarded.)

Note: Modifying the regex _does_ seem to behave as expected.
Note2: When you Add... a new entry you can create it with no description, however when modifying the Modify button is disabled when the description field is cleared.

**Expected behavior**
Modifying a Global Exclude to update the description should work :+1: 

**Software versions**
 - ZAP: 2.7.0 & Dev
 - OS: Kali/Win7

thc202:
Change GlobalExcludeURLParamToken to:
 - Properly implement equals method to ensure the changes done to the
 description are applied. Update hashCode accordingly.
 - Use the enabled parameter passed in the constructor.

Change DialogAddToken/DialogModifyToken to discard the edits of the
description after being set, also, for DialogAddToken do not ignore the
case when comparing the regular expression, the case matters.

Fix #5118 - Global Excludes Modify Description Doesn't Work

### DIFF
--- a/src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
+++ b/src/org/zaproxy/zap/extension/globalexcludeurl/DialogAddToken.java
@@ -123,7 +123,7 @@ class DialogAddToken extends AbstractFormDialog {
     protected boolean validateFields() {
         String tokenName = getRegexTextField().getText();
         for (GlobalExcludeURLParamToken t : tokens) {
-            if (tokenName.equalsIgnoreCase(t.getRegex())) {
+            if (tokenName.equals(t.getRegex())) {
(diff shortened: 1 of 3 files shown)


kingthorin:
Thanks!

PR #5120, opened 2018-11-15
Add Accept header

### LINKED ISSUE #4961: Include proper Accept header when requesting OpenAPI definitions
From https://mozilla.logbot.info/websectools/20180816
(Raising issue so that this is not forgotten, in case the pull request is not open in the end.)

The OpenAPI add-on does not add the Accept header which might lead to the target to serve other content than the API definition.

Add-on:
OpenAPI Support

tushar-97:
Fix #4961

### DIFF
--- a/src/org/parosproxy/paros/network/HttpHeader.java
+++ b/src/org/parosproxy/paros/network/HttpHeader.java
@@ -38,6 +38,7 @@
 // ZAP: 2017/03/02 Issue 3226: Added API Key and Nonce headers
 // ZAP: 2018/02/06 Make the lower/upper case changes locale independent (Issue 4327).
 // ZAP: 2018/04/24 Add JSON Content-Type.
+// ZAP: 2018/11/15 Add Accept header for HTTP requests.
 
 package org.parosproxy.paros.network;
 
@@ -69,6 +70,7 @@ public abstract class HttpHeader implements java.io.Serializable {
     public static final String IF_NONE_MATCH = "If-None-Match";
     public static final String USER_AGENT = "User-Agent";
     public static final String ACCEPT_ENCODING = "Accept-Encoding";
+    public static final String ACCEPT = "Accept";
     // ZAP:  the correct case is "Cache-Control", not "Cache-control"
     public static final String CACHE_CONTROL = "Cache-Control";
     public static final String PRAGMA = "Pragma";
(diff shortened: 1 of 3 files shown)


thc202:
Probably better to add it to `HttpRequestHeader` instead, more appropriate place.

The commit message should be amended, this change is not required for the issue (just good to have).

tushar-97:
Sure. Will change the grammar for the commit message. Its misleading.

src/org/parosproxy/paros/network/HttpHeader.java
```diff
@@ -69,6 +70,7 @@
     public static final String IF_NONE_MATCH = "If-None-Match";
     public static final String USER_AGENT = "User-Agent";
     public static final String ACCEPT_ENCODING = "Accept-Encoding";
+    public static final String ACCEPT = "Accept";
```
thc202: We should also add a since JavaDoc tag, make it easier to know when it was introduced (similar to the ones in `HttpResponseHeader`, e.g. `CSP`).

kingthorin:
Seems fine to me.

kingthorin:
Yes that diff seems ok.

src/org/parosproxy/paros/network/HttpRequestHeader.java
```diff
@@ -70,6 +71,13 @@
 
 public class HttpRequestHeader extends HttpHeader {
 
+    /**
+     * The {@code Accept} response header.
```
thc202: This needs a tweak, request instead.

tushar-97: I will make it a point to not push changes before I go to sleep. Will update

kingthorin: Don't worry @tushar-97 we've all been there..

tushar-97:
@kingthorin @thc202 made requested changes. This has become very messy but all looks good now.

kingthorin:
LGTM. Don’t worry about the number of commits it can be squashed at merge time.

Though sorting out appending commits might be a good move for the future.

In Eclipse/eGit it’s simple you just hit the little A button in the commit message panel. At the flu I’m pretty sure it’s just “commit -a”.

PR #5132, opened 2018-12-04
Connection Options Title Caps Tweaks

kingthorin:
* Consistent use of brackets for units (General Config - Timeout & DNS - TTL Successful Queries)
* Title caps on "Use Proxy Chain" and "Proxy Authentication"
* Title cap on "User Name" [which makes it the same as Context > User >Add]

Part of zaproxy/zaproxy#2000


thc202:
The corresponding help page needs to be updated.

PR #5142, opened 2018-12-10
Use Java 9+ APIs for macOS GUI changes

thc202:
Change OsXGui to use the newer APIs when running with Java 9+.
Remove handler implementations and related API library, now using
reflection for both APIs.

Part of #2602 - Java 9
Reported in #4669 - Unable to launch ZAP from Mac OS High Sierra


thc202:
It would be great if someone with a macOS could verify that this works fine with Java 8 and 9+.

psiinon:
Merging for the weekly as I dont htink anyone else in the core team has a Mac to test with ;)

thc202:
Thanks!

PR #5148, opened 2018-12-12
Show add-on details in a panel instead of tool tip

thc202:
Change Manage Add-ons dialogue to show the add-on details in a bottom
panel in the tabs Installed and Marketplace, to give more space and
properly render the changes of the add-ons (e.g. word wrap, links).
Change ZapLabel to disable caret updates instead of resetting the
position after setting the text to avoid scroll jumps.
Change AddOn to read the info URL also when creating the add-on from a
file.

### DIFF
--- a/src/lang/Messages.properties
+++ b/src/lang/Messages.properties
@@ -728,6 +728,7 @@ cfu.label.oldzap = This version of ZAP was created over a year ago!\nZAP is upda
 running with an out of date release.\n\nCheck for a new version now?\n\n
 cfu.label.outofdateaddons = Add-ons out of date?
 cfu.label.outofdatezap = ZAP out of date!
+cfu.label.selectAddOnForDetails = Select an add-on above to see more details.
 cfu.label.zap.border   = ZAP Core
 cfu.manage.title    		   = Manage Add-ons
 cfu.options.dialog.dirs.remove.title	= Remove Add-on Directories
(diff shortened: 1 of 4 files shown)


kingthorin:
I like it and it seems to work as intended. However, perhaps the initial sizing could be tweaked?

image

thc202:
You mean the whole dialogue or just the bottom panel?

kingthorin:
Well mainly the bottom panel,  but once you get into that it might mean both since the top could become too small?

thc202:
Removed the hardcoded dialogue size (and use preferred sizes) and gave more space to the bottom panel, let me know.

kingthorin:
I think it looks much better!
With something selected:
image
Initial open:
image

I don't want to be overly nit-picky so feel free to ignore this suggestion. But I kind of wonder if the fields should be slightly re-ordered to keep the more user important info up top. I know that means that the changes field might bump things around depending on what the user has selected but that shouldn't be a huge issue.
(Ex: Drop Id/Author to the bottom? Not Before/Not From at the bottom? [Is that checked when things are displayed? I mean do you see items displayed that can't be installed/used with the current version?])

psiinon:
Agree that this looks much better, but that the more relevant info should be higher.
Maybe if nothing is selected display a message like "Select an add-on above to see more details" rather than blank fields?

thc202:
Thanks for the feedback/suggestions, very welcome!

The fields were reordered to show more relevant ones first (let me know if you prefer other order) and it's now showing a message if no add-on is selected.

kingthorin:
Seems great to me!

thc202:
Done more suggestions, hide empty fields and show the info URL.

psiinon:
Big improvement to what we had before - many thanks :)

PR #5157, opened 2018-12-20
Allow to install a local add-on through the API

thc202:
Add an API endpoint to install a local (file system) add-on when in dev
mode, allowing to set up build tasks that (re)install the add-on being
developed.


psiinon:
Like this - can see it being very useful for us :)

PR #5164, opened 2018-12-27
Allow shortcut and callback implementors to override headers

psiinon:
Around line 535 addCustomHeaders is only called if 'impl' is not null, meaning that it wont be called for shortcut or callback implementors.
This will allow them to override any headers they need to.


psiinon:
Required for https://github.com/zaproxy/zap-hud/pull/315

PR #5178, opened 2019-01-09
Hook Help for Callbacks Panel

kingthorin:
When all is said and done
https://github.com/zaproxy/zap-core-help/pull/219 will be presented when
pressing F1 and the Callbacks tab has focus


src/org/zaproxy/zap/extension/callback/ExtensionCallback.java
```diff
@@ -94,6 +95,7 @@ public void hook(ExtensionHook extensionHook) {
                     getCallbackPanel());
             extensionHook.getHookView().addOptionPanel(
                     getOptionsCallbackPanel());
+            ExtensionHelp.enableHelpKey(getOptionsCallbackPanel(), "ui.tabs.callbacks");
```
thc202: Actually should be `getCallbackPanel()` instead?

kingthorin: That’s what I get for rushing, tweak coming

PR #5181, opened 2019-01-10
Add max alerts per rule option to the API

thc202:
Add pscan endpoints to get/set the value of max alerts per rule.


src/lang/Messages.properties
```diff
@@ -1906,10 +1906,12 @@ pscan.api.action.disableAllScanners = Disables all passive scanners
 pscan.api.action.enableScanners = Enables all passive scanners with the given IDs (comma separated list of IDs)
 pscan.api.action.disableScanners = Disables all passive scanners with the given IDs (comma separated list of IDs)
 pscan.api.action.setScannerAlertThreshold = Sets the alert threshold of the passive scanner with the given ID, accepted values for alert threshold: OFF, DEFAULT, LOW, MEDIUM and HIGH
+pscan.api.action.setMaxAlertsPerRule = Sets the maximum number of alerts a passive scan rule should raise.
```
kingthorin: Is this really a "should" situation? (Should seems uncertain, are there circumstances under which it wouldn't apply?) Or could this say "...can raise."? (Same for the View value below).

thc202: I was using the description of the option (from the help page zaproxy/zap-core-help#210), which refers that it might raise more than the max because of threading. Happy to change to can here.

kingthorin: Nope if there's a way it could be more than max then leaving as "should" is fine. Tks.

PR #5185, opened 2019-01-11
Remove data of option that no longer exists

thc202:
Remove label and update logic of an older option that no longer exists.


psiinon:

thc202:
Addressed.

PR #5189, opened 2019-01-14
Fix and tweaks to DynamicFieldsPanel

thc202:
Change default constructor to call an existing one to ensure that it's
used the correct layout manager, that is, GridBagLayout instead of
default FlowLayout.
Set the labels to the fields and extract a method that adds a field
(same logic for required and optional fields).
Add a vertical glue to the panel to ensure the fields are shown at the
top (instead of centred).


thc202:
Pushed again to trigger it.

psiinon:
That seemed to do it :) Just waiting for LGTM to finish...

kingthorin:
Thanks

PR #5201, opened 2019-01-25
Updated zap-api-scan

Crazyabout:
added the -S option in the try block

### DIFF
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -152,7 +152,7 @@ def main(argv):
     fail_inprog_count = 0
 
     try:
-        opts, args = getopt.getopt(argv, "t:f:c:u:g:m:n:r:J:w:x:l:hdaijp:sz:P:D:S:T:O:", ["hook="])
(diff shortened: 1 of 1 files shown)


docker/zap-api-scan.py
```diff
@@ -152,7 +152,7 @@ def main(argv):
     fail_inprog_count = 0
 
     try:
-        opts, args = getopt.getopt(argv, "t:f:c:u:g:m:n:r:J:w:x:l:hdaijp:sz:P:D:T:O:", ["hook="])
+        opts, args = getopt.getopt(argv, "t:f:c:u:g:m:n:r:J:w:x:l:hdaijp:sz:P:D:S:T:O:", ["hook="])
```
psiinon: Have you actually tried this out?
The additional colon will mean that it will require another parameter, so I think this will fail.

Crazyabout: My bad , was to eager to fix this ASAP.

thc202:
Thanks!

PR #5206, opened 2019-01-30
Validate Alert data

thc202:
Change ExtensionAlert to validate that the alert being raised has a
URI and HTTP message, otherwise it would fail or not work as intended
later when those fields are needed.

### DIFF
--- a/src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
+++ b/src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
@@ -39,6 +39,7 @@ import java.util.Vector;
 import javax.swing.JTree;
 import javax.swing.tree.TreePath;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import org.parosproxy.paros.Constant;
 import org.parosproxy.paros.control.Control;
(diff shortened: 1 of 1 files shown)


src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -74,6 +75,7 @@
 		SessionChangedListener, XmlReporterExtension, OptionsChangedListener {
 
     public static final String NAME = "ExtensionAlert";
+    private static final Logger logger = Logger.getLogger(ExtensionAlert.class);
```
kingthorin: Const? (Full caps)

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -224,6 +229,17 @@ public void run() {
         }
     }
 
+    private static boolean isInvalid(Alert alert) {
```
kingthorin: Negative logic on purpose? Does a `isValid` already exist?

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: Didn't change the case as that would require update a lot of places.

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
thc202: Yes, to read better in the above if statement. Can change to `isValid` if preferred.

thc202:
No, it was with front end scanners (fix in zaproxy/zap-extensions#1928).

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: Ok

src/org/zaproxy/zap/extension/alert/ExtensionAlert.java
kingthorin: I guess it isn't a huge issue, just surprised me.

PR #5207, opened 2019-01-30
Hook OverrideMessageProxyListener also on install

thc202:
Change ExtensionLoader to hook the OverrideMessageProxyListener when the
add-on is installed.


thc202:
This can be reproduced when using zaproxy/zap-hud#310 (the HTTP sites would no longer be upgraded to HTTPS).

PR #5232, opened 2019-02-25
Reset bug counter on session changes

thc202:
Change ExtensionLog4j to reset the bug counter when the session changes,
to make it easier to spot any new issues.


psiinon:
Looks good but holding off approving it until the Docker weekly image has built to keep it in step with the version uploaded ;)

PR #5236, opened 2019-03-01
Expose the local add-ons through the ZAP API

thc202:
Allow to obtain all the local add-ons, installed or not. Also, provide
the installation status.
Extract a method to reduce code duplication.


kingthorin:
What’s a local addon that isn’t installed?

thc202:
It's an add-on that it's available to ZAP locally, that was not yet installed (e.g. not all dependencies are fulfilled).

kingthorin:
Tks

PR #5249, opened 2019-03-11
Do not check for updates on start up in dev mode

thc202:
Unnecessary, in dev mode it is assumed that ZAP and/or the add-ons are
being developed and should be the latest.

### DIFF
--- a/src/org/zaproxy/zap/extension/autoupdate/ExtensionAutoUpdate.java
+++ b/src/org/zaproxy/zap/extension/autoupdate/ExtensionAutoUpdate.java
@@ -723,6 +723,10 @@ public class ExtensionAutoUpdate extends ExtensionAdaptor implements CheckForUpd
     }
     
     public void alertIfNewVersions() {
+        if (Constant.isDevMode()) {
+            return;
+        }
+
     	// Kicks off a thread and pops up a window if there are new versions.
     	// Depending on the options the user has chosen.
     	// Only expect this to be called on startup and in desktop mode


psiinon:
Not quite so sure of this.
I usually just deploy the add-on(s) I'm working on rather than all of them.
It will also mean we dont get stats on people running ZAP in dev mode (with cfu turned on)

phoenix24:
> Unnecessary, in dev mode it is assumed that ZAP and/or the add-ons are
> being developed and should be the latest.

flags like these should be driven by configuration, and not hardcoded as a function of the environment. 👎

thc202:
> > Unnecessary, in dev mode it is assumed that ZAP and/or the add-ons are
> > being developed and should be the latest.
> 
> flags like these should be driven by configuration, and not hardcoded as a function of the environment. -1

I'm happy to split the dev mode from dev build, since unlike the others this change means loss of functionality.

> I usually just deploy the add-on(s) I'm working on rather than all of them.

Yeah, I might be stretching it, but if you need a new version you can still check for updates anytime.

> It will also mean we dont get stats on people running ZAP in dev mode (with cfu turned on)

Agreed, although I'm not sure how useful it is to know that I start ZAP multiple times a day ;) (I use a clean home many times).

Would splitting the dev mode from dev build be a good compromise? That would not be much different than setting the config to disable CFU, just making it more convenient.

thc202:
Closing, superseded by #5354.

PR #5300, opened 2019-04-02
Ignore "javax.faces.ViewState" parameter in active scans

JavanXD:
Field "javax.faces.ViewState" is used by JavaServer Faces (JSF)

### DIFF
--- a/src/org/parosproxy/paros/core/scanner/ScannerParam.java
+++ b/src/org/parosproxy/paros/core/scanner/ScannerParam.java
@@ -275,6 +275,7 @@ public class ScannerParam extends AbstractParam {
             addScannerParamFilter("__EVENTVALIDATION", NameValuePair.TYPE_POST_DATA, "*");
             addScannerParamFilter("__EVENTTARGET", NameValuePair.TYPE_POST_DATA, "*");
             addScannerParamFilter("__EVENTARGUMENT", NameValuePair.TYPE_POST_DATA, "*");
+            addScannerParamFilter("javax.faces.ViewState", NameValuePair.TYPE_POST_DATA, "*");
             addScannerParamFilter("(?i)jsessionid", NameValuePair.TYPE_UNDEFINED, "*");
             addScannerParamFilter("cfid", NameValuePair.TYPE_COOKIE, "*");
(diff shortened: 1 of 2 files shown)


kingthorin:
ViewState isn’t anti-CSRF

kingthorin:
Basically for both .Net ViewState and JSF ViewState an argument can be made that it provides implicit CSRF protection, however, that's not what it is meant for and also there are explicit controls aimed at CSRF prevention.

Some reference material:

- https://www.beyondjava.net/jsf-viewstate-and-csrf-hacker-attacks
- https://www.oracle.com/webfolder/technetwork/tutorials/obe/java/JSF-CSRF-Demo/JSF2.2CsrfDemo.html
- https://security.stackexchange.com/questions/8744/does-asp-net-viewstate-implicitly-prevent-csrf-attacks-what-does-this-mean-for
- https://www.we45.com/blog/3-csrf-remedies-that-are-not-as-effective-as-you-thought-they-were

You might want to add it here though:
image

JavanXD:
Thanks for making this clear. In my case, it was mistakenly but consistently used as a CSRF token.

kingthorin:
Note, there's nothing that keeps you (or any user) from manually adding it as a CSRF token and having ZAP treat it as such. I just don't think it should be added to the default list (for the reasons outlined above). However, if you want to add it to the ignored parameter list that probably makes sense (since testing/fuzzing it likely just breaks things).

kingthorin:
Let us know if you'll be updating this PR or if it should just be closed (I hope you'll update it :+1: ).

:zap:

JavanXD:
I updated.

thc202:
@JavanXD thanks! How would you like to be credited?

JavanXD:
@thc202 As 'Javan Rasokat' would be just fine.

In the future, especially for the HUD, hopefully more contributions will follow from me. Btw I ❤️  the ZEST scripts.

psiinon:
@JavanXD we have weekly HUD conf calls - would you like to be invited to them? For more details see https://groups.google.com/d/msg/zaproxy-hud/jHVyiKE9xSI/j_vgbimIAwAJ although the times up in the air again

Issue #5304, opened 2019-04-04
Improve error message when failed to connect to outgoing proxy


osandadeshan:
**Describe the bug**
When trying do do a quick scan, "Failed to attack the URL: host not found" error will be visible

**To Reproduce**
1. Open Zap 2.7.0
2. Input http://newtours.demoaut.com/ as the URL to attack
3. Click on Attack button
4. See error "Failed to attack the URL: host not found"

**Expected behavior**
Should be able to do a quick atack

**Screenshots**
![image](https://user-images.githubusercontent.com/9147189/55539861-4feae280-56df-11e9-99e0-d5b6a9451939.png)

**Software versions**
 - ZAP: 2.7.0
 - OS: Windows 10 64bit
 - Java: openjdk version "1.8.0_171"
 - Browser: Chrome 73

kingthorin:
Can you access the target in a browser?

osandadeshan:
@kingthorin 
You mean Launch browser button on the below. Then it will open a chrome instance and "Explore your application with ZAP" will be visible on the html body

kingthorin:
I mean when you encounter that error are you able to access the target at all? The message seems to indicate a DNS issue.

osandadeshan:
No I have able to access the web page http://newtours.demoaut.com/
Also I used another web sites as well. But not working

kingthorin:
Ok proxy your browser through ZAP and let us know if that works.

osandadeshan:
@kingthorin 
I have added manual proxy to firefox and when I go to the URL this shows in the web page body
`ZAP Error [java.net.UnknownHostException]: Default

Stack Trace:
java.net.UnknownHostException: Default
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184)
	at java.net.PlainSocketImpl.connect(PlainSocketImpl.java:172)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at java.net.Socket.<init>(Socket.java:434)
	at java.net.Socket.<init>(Socket.java:286)
	at org.apache.commons.httpclient.protocol.DefaultProtocolSocketFactory.createSocket(DefaultProtocolSocketFactory.java:80)
[...]

PR #5313, opened 2019-04-09
Improve error message when failed to connect to outgoing proxy

kingthorin:
Update `ProxyThread.setErrorResponse(HttpMessage msg, String responseStatus, Exception cause, String errorType)` to check the type of exception, if proxy chain is enabled, and if the error has the same hostname as the outbound proxy.

Fixes zaproxy/zaproxy#5304


thc202:
Thanks!

PR #5321, opened 2019-04-16
Updated an alert reference link

tjtanjin:
The old reference link for the alert "Web Browser XSS Protection Not Enabled" no longer works. Updated with a new link.

### DIFF
--- a/examples/ZAP_2.4.3_report-merged.html
+++ b/examples/ZAP_2.4.3_report-merged.html
@@ -2855,7 +2855,7 @@ Host: localhost:8080
 <tr bgcolor="#e8e8e8" valign="top">
 <td width="20%"><font size="2" face="Arial, Helvetica, sans-serif">
 <p>Reference</p>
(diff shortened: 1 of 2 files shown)


kingthorin:
It would make much more sense to change it at the source.

https://github.com/zaproxy/zap-extensions/blob/2fbc18b4c86b508f8dca4dcc83eb68e3a62addfe/addOns/pscanrules/src/main/resources/org/zaproxy/zap/extension/pscanrules/resources/Messages.properties#L7

thc202:
@tjtanjin would you mind updating the above file instead? I'm not sure we need to update these example reports, iirc, they were added to just show the structure (merged vs unmerged).

tjtanjin:
@kingthorin Thank you so much for your help with the link. I have made a pull request for it here: https://github.com/zaproxy/zap-extensions/pull/2000

kingthorin:
Thanks for contributing!

I guess this PR can be closed now?

PR #5324, opened 2019-04-22
Enhancement: Spider handle Hyperlink Auditing Functionality

kingthorin:
After having come across:
* https://www.bleepingcomputer.com/news/software/major-browsers-to-prevent-disabling-of-click-tracking-privacy-risk/
* https://html.spec.whatwg.org/multipage/links.html#hyperlink-auditing

I thought I would add this.

### DIFF
--- a/src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
+++ b/src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
@@ -122,12 +122,14 @@ public class SpiderHtmlParser extends SpiderParser {
 		List<Element> elements = source.getAllElements(HTMLElementName.A);
 		for (Element el : elements) {
 			resourcesfound |= processAttributeElement(message, depth, baseURL, el, "href");
+			resourcesfound |= processAttributeElement(message, depth, baseURL, el, "ping");
 		}
 
 		// Process AREA elements
 		elements = source.getAllElements(HTMLElementName.AREA);
 		for (Element el : elements) {
 			resourcesfound |= processAttributeElement(message, depth, baseURL, el, "href");
+			resourcesfound |= processAttributeElement(message, depth, baseURL, el, "ping");
 		}
 
 		// Process Frame Elements
@@ -204,7 +206,15 @@ public class SpiderHtmlParser extends SpiderParser {
 			return false;
 		}
 
-		processURL(message, depth, localURL, baseURL);
+		if (!attributeName.equalsIgnoreCase("ping")) {
+			processURL(message, depth, localURL, baseURL);
+		} else {
+			for (String pingURL : localURL.split("\\s")) {
+				if (!pingURL.isEmpty()) {
+					processURL(message, depth, pingURL, baseURL);
+				}
+			}
+		}
 		return true;
 	}
 
--- a/test/org/zaproxy/zap/spider/parser/SpiderHtmlParserUnitTest.java
+++ b/test/org/zaproxy/zap/spider/parser/SpiderHtmlParserUnitTest.java
@@ -166,6 +166,99 @@ public class SpiderHtmlParserUnitTest extends SpiderParserTestUtils {
     }
 
     @Test
+    public void shouldFindUrlsInAnchorPingElements() {
+        // Given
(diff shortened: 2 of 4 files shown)


src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -204,7 +206,15 @@ private boolean processAttributeElement(HttpMessage message, int depth, String b
 			return false;
 		}
 
-		processURL(message, depth, localURL, baseURL);
+		if (!attributeName.equalsIgnoreCase("ping")) {
+			processURL(message, depth, localURL, baseURL);
+		} else {
+			for (String pingURL : localURL.split(" ")) {
+				if (!pingURL.equals(" ")) {
```
kingthorin: Since the `ping` attribute allows multiple URLs separated by space I need an elegant way to handle the `(same page)` scenario while also handling the fact that the space delimited list might have multiple spaces which are irrelevant. Any ideas?

Also need to ensure that the ping URLs start with http(s).

Note, I haven't yet added the UnitTests for area handling until I get this part sorted.

kingthorin: I've been thinking about the `(same page)` scenario(s) vs. handling multiple spaces in the list. I think the best that can be done is:
* Check the length of the array resulting from split. If it's a single item then process as normal (`(same page)` scenario will work). If it's multiple then only process if not space (as above) and a `(same page)` scenario might be missed.
  * It seems unlikely that many people/frameworks would choose to process the pings (by snooping headers/POST content etc at the same endpoint as is hosting the content), and do so by choosing to not specify the destination at all (ex: `ping=""`). [Not impossible but perhaps improbable?]

If we just document this behavior, is that sufficient? Or implement it and see what kind of feedback we get?

thc202: What do you mean by `(same page)`?

kingthorin: If you look at the unittests, there's a `(same page)` variant that's essentially an empty string `""`

thc202: Sorry, missed that. In that case the list of strings is empty (if I'm reading https://infra.spec.whatwg.org/#split-on-ascii-whitespace correctly), so no pings to handle.

thc202: It's not necessary to ignore non-HTTP(S) URLs, the spider already does that.

src/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -204,7 +206,15 @@ private boolean processAttributeElement(HttpMessage message, int depth, String b
 			return false;
 		}
 
-		processURL(message, depth, localURL, baseURL);
+		if (!attributeName.equalsIgnoreCase("ping")) {
+			processURL(message, depth, localURL, baseURL);
+		} else {
+			for (String pingURL : localURL.split("\\s")) {
+				if (!pingURL.equals("")) {
```
thc202: Could use `isEmpty()` instead.

test/org/zaproxy/zap/spider/parser/SpiderHtmlParserUnitTest.java
```diff
@@ -165,6 +165,99 @@ public void shouldFindUrlsInAElements() {
...
+                        "http://pong.example.com/"));//Trailing slash is added on host
+    }
+
+    @Test
+    public void shouldFindUrlsInAreaPingElements() {
+        // Given
+        SpiderHtmlParser htmlParser = new SpiderHtmlParser(new SpiderParam());
+        TestSpiderParserListener listener = createTestSpiderParserListener();
+        htmlParser.addSpiderParserListener(listener);
+        HttpMessage messageHtmlResponse = createMessageWith("AElementsWithPingSpiderHtmlParser.html");
```
thc202: File name was not updated.

kingthorin:
Fixed

thc202:
Thanks!

Issue #5335, opened 2019-05-02
Add feedback on search progress


zestysoft:
**Is your feature request related to a problem? Please describe.**
When doing a search, there is currently no indication if the search is active or has completed.

**Describe the solution you'd like**
Some type of feedback on search status, especially if no results are found.

**Describe alternatives you've considered**
Wait for an undetermined amount of time for results that may never appear (because there are no matches).

psiinon:
Agreed, I've fallen foul of this more than a few times ;)

PR #5344, opened 2019-05-07
Add Search Feedback Progress Indicator

kingthorin:
- Search panel now has a status label (`Searching`/`Complete`).
- Search button is now disabled when searching (to prevent accidental re-starts)
- The cursor now shows busy (wait) when searching and on (over) the panel.

Fixes zaproxy/zaproxy#5335


src/org/zaproxy/zap/extension/search/ExtensionSearch.java
```diff
@@ -298,6 +298,7 @@ public void sessionScopeChanged(Session session) {
 		public void sessionChanged(final Session session) {
 			if (EventQueue.isDispatchThread()) {
 				getSearchPanel().resetSearchResults();
+				getSearchPanel().getSearchStatusLabel().setText("");
```
kingthorin: Food for thought:

- Should the search "term" also be reset between loads? (I can see it being handy both ways)
- Do you think there's any point ensuring the search button is in enabled state? (I couldn't get a long enough running search to determine if loading a new session while searching might result in it being disabled.)

kingthorin:
Animated gif (search is done twice in this video so that different things can be pointed to and the cursor can be moved around):

srch_cap

src/org/zaproxy/zap/extension/search/ExtensionSearch.java
kingthorin: I'll add the fix keyword after answer/tweaks :zap:

PR #5350, opened 2019-05-10
Enable custom (script) input vectors by default.

psiinon:
Script input vectors can be enabled and disabled individually.
Having all of them turned off by default just gives the user one more thing they have to remember to do and doesnt really add any benefit.


psiinon:
@kingthorin I'm not sure I follow you :/ In the code they are called custom input vectors, but in practice this is the setting used to globally enable/disable active scan scripts.

psiinon:
Nm, @thc202 has explained :)

psiinon:
Have updated the commit message

PR #5356, opened 2019-05-14
Add -silent command line option

### LINKED ISSUE #5354: New silent option to disable all calls home
ZAP doesnt make many 'calls home', just the check-for-updates and loading the browser launch page.
However we know some people _really_ dont like them, and so we should add a new 'silent' option that disables all of these checks, for now and in the future.
This should be documented in a FAQ along with details of the calls home we actually make.
I think this can just be a config file option, not UI needed (anyone feel free to disagree).

psiinon:
Fixes #5354 

Will need to update specific add-ons and the helpas well.

### DIFF
--- a/src/org/parosproxy/paros/CommandLine.java
+++ b/src/org/parosproxy/paros/CommandLine.java
@@ -80,6 +80,7 @@ public class CommandLine {
     public static final String LOWMEM = "-lowmem";
     public static final String EXPERIMENTALDB = "-experimentaldb";
     public static final String SUPPORT_INFO = "-suppinfo";
+    public static final String SILENT = "-silent";
 
     /**
      * Command line option to disable the default logging through standard output.
@@ -112,6 +113,7 @@ public class CommandLine {
     private boolean displaySupportInfo = false;
     private boolean lowMem = false;
     private boolean experimentalDb = false;
+    private boolean silent = false;
     private int port = -1;
     private String host = null;
     private String[] args;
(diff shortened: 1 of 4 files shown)


src/org/parosproxy/paros/CommandLine.java
```diff
@@ -502,6 +507,13 @@ public String getHelp() {
     public boolean isNoStdOutLog() {
         return noStdOutLog;
     }
+    
+    /*
+     * Returns true if ZAP should not make any unsolicited requests, eg check-for-updates etc
+     */
+    public boolean isSilent() {
```
thc202: Could have the since tag.

src/org/parosproxy/paros/Constant.java
```diff
@@ -1264,6 +1266,17 @@ public static void setDevMode(boolean devMode) {
         Constant.devMode = devMode;
     }
     
+    /**
+     * Sets whether or not ZAP should be 'silent' ie not make any unsolicited requests.
+     * <p>
+     * <strong>Note:</strong> This method should be called only by bootstrap classes.
+     * 
+     * @param devMode {@code true} if ZAP should be silent, {@code false} otherwise.
```
thc202: - Parameter name needs to be updated.
 - Could have since tags (same for getter).
 - Needs ZAP comment.

psiinon:
Hopefully fixed those.

Issue #4780, opened 2018-06-21
Tidy up dependencies


thc202:
Tracker issue to tidy up the dependencies, used by the `zaproxy` project, for 2.8.0 release.

From OWASP ZAP Developer Group: https://groups.google.com/d/topic/zaproxy-develop/wN6HtAKOFfw/discussion

Remove dependencies not used by core nor add-ons (in `zap-extensions`):
 - [x] Apache HttpComponents HttpCore/HttpClient. Done in #4794.

Remove dependencies in use but not really needed:
 - [x] BrowserLauncher, it's used in core as fallback and in one add-on (not currently available in the marketplace). This is effectively replaced with a JRE class (`java.awt.Desktop`). Done in #4782, zaproxy/zap-extensions#1683, and zaproxy/zap-extensions#1685.
 - [x] JavaFX runtime, only required by an add-on. It would rely on the JRE (e.g.
OracleJDK, OpenJDK+OpenJFX). Done in #4785.

kingthorin:
Since this is 3/3 can it be closed?

thc202:
Still plan to check other dependencies (e.g. diffutils) and update others.

kingthorin:
Ping :grin:

thc202:
This will be done after the migration to Gradle.

thc202:
Closing, all changes done.

PR #5374, opened 2019-05-24
Remove usage of JDOM library

thc202:
Use Commons Configuration to read/write the XML files instead of JDOM,
the former is used in more places.
The library will still be made available to add-ons, until they are
changed to bundled it or use other library.
Add tests to assert that the corresponding XML files are correctly
handled.
Correct typos in XML elements of the drivers (no impact on functionality
the value of the element had the default value).

Part of #4780 - Tidy up dependencies


thc202:
Changed the scope of the dependency to be available only at runtime (avoids being reintroduced in ZAP accidentally and forces add-ons targeting the newer ZAP version to either depend on it or change to other library).

PR #5392, opened 2019-06-03
Update LEGALNOTICE

thc202:
Update with latest libraries/versions (#4780).


thc202:
I'm fine either way (IANAL though), if we don't want/need that info harcoded we can use a Gradle plugin to generate a report (e.g. https://github.com/jk1/Gradle-License-Report ).

PR #5396, opened 2019-06-05
Remove unnecessary class

thc202:
Use already bundled library to obtain the hex MD5 hash instead.


psiinon:

thc202:
Conflicts addressed.

PR #5408, opened 2019-06-10
Updated snapcraft file for 2.8.0

psiinon:
The snap build is actually failing for me right now, so still looking into that.
But these changes will be needed in any case.


psiinon:
Ah, looks like we have to call `apt-get update` before calling 'snapcraft` in the docker image :/
Will require some tweaking to script, but this PR seems to be ok.

PR #5466, opened 2019-07-10
Add utility methods to Alert class

kingthorin:
* Added isValidRisk and isValidConfidence convenience methods.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
@@ -54,6 +54,7 @@
 // ZAP: 2017/09/15 Initialise the source from the RecordAlert always.
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
+// ZAP: 2019/07/10 Add utility methods isValidRisk(int) and isValidConfidence(int)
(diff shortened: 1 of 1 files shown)


kingthorin:
I considered implementing this by creating a collection (list or set) for each element (risk/confidence) and then using a `.contains` check but I wasn't sure if there was a benefit to that vs this implementation.

kingthorin:
Example use case, situations such as:
image

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -892,4 +892,37 @@ public void setSource(Source source) {
...
+     * Checks if a value {@code int} is between {@value #RISK_INFO} (RISK_INFO) and {@value
+     * #RISK_HIGH} (RISK_HIGH)
+     *
+     * @return true if the checked risk ({@code int}) is in the range, false otherwise
+     * @since TODO add version
+     * @see #RISK_INFO
+     * @see #RISK_HIGH
+     */
+    public static boolean isValidRisk(int risk) {
+        if (risk >= RISK_INFO && risk <= RISK_HIGH) {
```
thc202: - Could return the result of the expression instead of using an if.
 - Needs ZAP comment.

kingthorin:
Tweaked.

thc202:
Thanks!

PR #5470, opened 2019-07-11
Coverity Fixes

kingthorin:
* SpiderHtmlFormParser - CID 157242 Unused variable. Don't assign defaultValue in getDefaultTestValue` if it already has a value.
* SpiderHtmlFormParserUnitTest - Updated to account for new changes.
* FormNoDefaultValues.html - Removed radio buttons as one is always selected (representing a default value).

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
@@ -333,7 +333,7 @@ public class SpiderHtmlFormParser extends SpiderParser {
 
         // Get its value(s)
         List<String> values = field.getValues();
-        String defaultValue;
+        String defaultValue = null;
 
         // If the field has a value attribute present(Predefined value)
         // Should store the value being submitted to be passed to the ValueGenerator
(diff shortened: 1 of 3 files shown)


kingthorin:
It looks like the SpiderHtmlFormParser modification is causing a unit test failure. Looking at the output "unknown" vs. an actual value it seems the new behavior is superior and the unit test should be updated to align. Can someone confirm or correct me if I'm wrong?

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -347,26 +347,27 @@ private static void addAll(SortedSet<FormControl> formControls, List<Element> el
...
+                    Iterator<String> iterator = predefValues.iterator();
                     defaultValue = iterator.next();
+
+                    // If there are more values, don't use the first, as it usually is a "No select"
+                    // item
+                    if (iterator.hasNext()) {
+                        defaultValue = iterator.next();
+                    }
                 }
             }
```
thc202: After this add the `defaultValue = DEFAULT_EMPTY_VALUE` if still null.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlFormParser.java
```diff
@@ -347,26 +347,27 @@ private static void addAll(SortedSet<FormControl> formControls, List<Element> el
...
 
-        } else {
-            defaultValue = values.get(0);
+        } else if (defaultValue == null) {
+            defaultValue = values.get(0) == null ? DEFAULT_EMPTY_VALUE : values.get(0);
```
thc202: This could be as before (`values.get(0)` is non-null).

kingthorin:
It seems like the latest round of modifications help further:
* Expected defaultValue f though fieldAttributes are id male and value m?
```
    Expected: is <uri=http://example.com/, targetUri=http://example.org/post, fieldName=gender, defaultValue=f, values=[m, f], fieldAttributes={name=gender, id=male, type=radio, Control Type=RADIO, value=m}, formAttributes={action=http://example.org/post, method=POST, atta=valueA}>
```
* Now actually aligns defaultValue m, with fieldAttributes id male and value m. 
```
         but: was <uri=http://example.com/, targetUri=http://example.org/post, fieldName=gender, defaultValue=m, values=[m, f], fieldAttributes={name=gender, id=male, type=radio, value=m, Control Type=RADIO}, formAttributes={action=http://example.org/post, method=POST, atta=valueA}>
```
Unless I'm missing something that makes more sense.

The other two UnitTest failures are also now defined instead of unknown (as previously observed) [searching the Travis log for "unknown" and highlighting occurrences makes it stand out nicely.]

zap/src/test/java/org/zaproxy/zap/spider/parser/SpiderHtmlFormParserUnitTest.java
```diff
@@ -1125,7 +1124,8 @@ public void shouldSetValuesToFieldsWithNoValueWhenParsingPostForm() {
         boolean completelyParsed = htmlParser.parseResource(msg, source, BASE_DEPTH);
         // Then
         assertThat(completelyParsed, is(equalTo(false)));
-        assertThat(listener.getNumberOfResourcesFound(), is(equalTo(9)));
+        assertThat(listener.getNumberOfResourcesFound(), is(equalTo(8)));
+        System.out.println(listener.getResourcesFound().get(5));
```
thc202: Debug?

kingthorin: Whoops, yup, will fix that tomorrow.

kingthorin:
Leftover debug removed.

thc202:
Thank you!

PR #5501, opened 2019-07-26
More LGTM Fixes

kingthorin:
ExtensionLoader - Relocate hook null check earlier in hookScannerHook(Scanner). LGTM Rule: Dereferenced variable may be null. Specific finding: https://lgtm.com/projects/g/zaproxy/zaproxy/snapshot/6a2d4338c766f041ed0773d398eb609bc0319312/files/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java?sort=name&dir=ASC&mode=heatmap#x89e8c7fad7ce7e45:1

Analyser - Remove parent null check in sendAndReceive(HttpMessage). LGTM Rule: Dereferenced variable may be null. Specific finding: https://lgtm.com/projects/g/zaproxy/zaproxy/snapshot/6a2d4338c766f041ed0773d398eb609bc0319312/files/zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java?sort=name&dir=ASC&mode=heatmap#x4e1d1877edf3b8eb:1

HttpBreakpointMessage - Modify match(Message, boolean, boolean) remove null assignment on src and add default case which falls through to case url. LGTM Rule: Dereferenced variable may be null. Specific findings: https://lgtm.com/projects/g/zaproxy/zaproxy/snapshot/6a2d4338c766f041ed0773d398eb609bc0319312/files/zap/src/main/java/org/zaproxy/zap/extension/brk/impl/http/HttpBreakpointMessage.java?sort=name&dir=ASC&mode=heatmap#xf6c79c952cfcbf71:1 and https://lgtm.com/projects/g/zaproxy/zaproxy/snapshot/6a2d4338c766f041ed0773d398eb609bc0319312/files/zap/src/main/java/org/zaproxy/zap/extension/brk/impl/http/HttpBreakpointMessage.java?sort=name&dir=ASC&mode=heatmap#x31c7ce8d1a9199c6:1

StatusUI - Modify StatusUI(AddOn.Status, String) to use this.status which won't be null instead of the inbound status which might be null. LGTM Rule: Dereferenced variable may be null. Specific findings: https://lgtm.com/projects/g/zaproxy/zaproxy/snapshot/6a2d4338c766f041ed0773d398eb609bc0319312/files/zap/src/main/java/org/zaproxy/zap/view/StatusUI.java#xc78703fd791c978a:1

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java
@@ -41,6 +41,7 @@
 // ZAP: 2017/12/29 Rely on HostProcess to validate the redirections.
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
+// ZAP: 2019/07/26 Remove null check in sendAndReceive(HttpMessage). (LGTM Issue)
 package org.parosproxy.paros.core.scanner;
 
 import java.io.IOException;
@@ -516,11 +517,7 @@ public class Analyser {
 
         httpSender.sendAndReceive(msg, parent.getRedirectRequestConfig());
         requestCount++;
-
-        // ZAP: Notify parent
-        if (parent != null) {
-            parent.notifyNewMessage(msg);
-        }
+        parent.notifyNewMessage(msg);
     }
 
     public int getDelayInMs() {
--- a/zap/src/main/java/org/zaproxy/zap/view/StatusUI.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/StatusUI.java
@@ -45,7 +45,7 @@ public class StatusUI implements Comparable<StatusUI> {
         }
 
         if (stringRepresentation == null || stringRepresentation.isEmpty()) {
-            this.stringRepresentation = status.toString();
+            this.stringRepresentation = this.status.toString();
         } else {
             this.stringRepresentation = stringRepresentation;
         }
--- a/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java
@@ -83,6 +83,7 @@
 // ZAP: 2019/03/15 Issue 3578: Added Helper options for Import menu
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
+// ZAP: 019/07/25: Relocate null check to be earlier in hookScannerHook(scan) [LGTM issue].
 package org.parosproxy.paros.extension;
 
 import java.awt.Component;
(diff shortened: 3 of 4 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java
```diff
@@ -514,11 +515,11 @@ private void sendAndReceive(HttpMessage msg) throws HttpException, IOException {
...
 
-        httpSender.sendAndReceive(msg, parent.getRedirectRequestConfig());
-        requestCount++;
-
-        // ZAP: Notify parent
         if (parent != null) {
+            httpSender.sendAndReceive(msg, parent.getRedirectRequestConfig());
+            requestCount++;
+
+            // ZAP: Notify parent
```
kingthorin: Shall I remove the blank and comment? (The comment seems unnecessary)

zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java
```diff
@@ -514,11 +515,11 @@ private void sendAndReceive(HttpMessage msg) throws HttpException, IOException {
...
             parent.notifyNewMessage(msg);
```
kingthorin: Shall I remove the blank and comment? (The comment seems unnecessary)

thc202: Yes, we can also remove the parent null check, it will never happen here (it would break earlier).

zap/src/main/java/org/zaproxy/zap/view/StatusUI.java
```diff
@@ -39,7 +39,7 @@
      */
     public StatusUI(AddOn.Status status, String stringRepresentation) {
         if (status == null) {
-            this.status = AddOn.Status.unknown;
+            this.status = status = AddOn.Status.unknown;
```
thc202: It could be changed the later one to `this.status` instead.

kingthorin: You mean just change the order? `status = this.status = AddOn.Status.unknown`

zap/src/main/java/org/parosproxy/paros/core/scanner/Analyser.java
kingthorin: parent is instantiated null at L102.

thc202: Yes, but it would fail in `analyse` not here (only the constructor with parent is being used, so the parent is never null anyway).

zap/src/main/java/org/zaproxy/zap/view/StatusUI.java
thc202: The one in `this.stringRepresentation = status.toString();`.

kingthorin:
Yes, I broke the PR and commit message. I'll fix that once LGTM is done this run.

thc202:
Thanks!

PR #5522, opened 2019-08-14
Removed unused Model params

### LINKED ISSUE #5297: Unused parameter : "model" in ReportLastScan methods
most methods of **ReportLastScan.java** has a parameter:
`org.parosproxy.paros.model.ModelModel model` , but it is actually unused. those methods just pass this parameter to: `public void generate(StringBuilder report, Model model) throws Exception` where it is not used at all.

Flo4l:
Removed unused Model params in `ReportLastScan.java` methods.

Closes #5297

Signed-off-by: Florian Albers <Florian.Albers@adesso.de>

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/report/ReportLastScan.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/report/ReportLastScan.java
@@ -35,6 +35,7 @@
 // ZAP: 2018/07/04 Fallback to bundled XSL files.
 // ZAP: 2018/07/09 No longer need cast on SiteMap.getRoot
 // ZAP: 2019/06/05 Normalise format/style.
+// ZAP: 2019/08/15 Issue 5297: Removed unused model params.
 package org.parosproxy.paros.extension.report;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


zap/src/main/java/org/parosproxy/paros/extension/report/ReportLastScan.java
```diff
@@ -84,18 +85,18 @@ public ReportLastScan() {}
...
     @Deprecated
-    public File generate(String fileName, Model model, String xslFile) throws Exception {
+    public File generate(String fileName, String xslFile) throws Exception {
         StringBuilder sb = new StringBuilder(500);
-        this.generate(sb, model);
+        this.generate(sb);
         return ReportGenerator.stringToHtml(sb.toString(), xslFile, fileName);
     }
 
-    public File generate(String fileName, Model model, ReportType reportType) throws Exception {
```
thc202: We need to keep binary compatibility, for the add-ons.

Flo4l: Force pushed required changes.

thc202: Sorry I was not clear, the comment applies to the other public methods as well. We should deprecate the public methods with unused parameter and add new ones without the parameter (the initial commit was just missing the old/deprecated methods).

Flo4l: My bad. Should be fine now.

zap/src/main/java/org/parosproxy/paros/extension/report/ReportLastScan.java
```diff
@@ -87,15 +88,24 @@ public ReportLastScan() {}
      *     Model model, ReportType reportType)}
      */
     @Deprecated
-    public File generate(String fileName, Model model, String xslFile) throws Exception {
```
thc202: This change should be reverted (the linked method in the JavaDoc could point to the new method as before).

zap/src/main/java/org/parosproxy/paros/extension/report/ReportLastScan.java
```diff
@@ -125,7 +135,7 @@ private static File generateReportWithXsl(String report, String reportFile, Stri
         }
     }
 
-    public void generate(StringBuilder report, Model model) throws Exception {
```
thc202: This needs to be kept/deprecated.

thc202:
Thanks!

PR #5524, opened 2019-08-14
Check binary compatibility of zap.jar

thc202:
Add JApicmp Gradle Plugin to check the binary compatibility of zap.jar,
to ensure it's not accidentally broken.


kingthorin:
What's the plan for when/if things are purposefully broken? ex: Something that's been deprecated for ages is finally removed.

psiinon:
Ditto ;)

thc202:
Incompatibilities can be ignored.

PR #5525, opened 2019-08-15
Check that core APIs are added to generator

thc202:
Add test that checks that all core APIs are present in the generator.


thc202:
Build now passing in Travis CI (reworked the test to load the extensions "manually").

PR #5546, opened 2019-08-27
PassiveScanner Default Methods

kingthorin:
Create empty default methods for `scanHttpRequestSend(HttpMessage, int)` and  `scanHttpResponseReceive(HttpMessage, int)` so that scanners which often only scan request OR response, don't require an empty overridden method.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanner.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanner.java
@@ -25,9 +25,9 @@ import org.parosproxy.paros.network.HttpMessage;
 
 public interface PassiveScanner {
 
-    void scanHttpRequestSend(HttpMessage msg, int id);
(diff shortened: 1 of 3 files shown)


thc202:
It needs to be ignored, looking at that...

kingthorin:
Possible relevant ref: https://github.com/siom79/japicmp/issues/201

thc202:
It is.

kingthorin:
I tried briefly (~10min) to sort this out this morning but I think maybe the `overrideCompatibilityChangeParameters` functionality isn’t exposed in the gradle plugin yet or I just couldn’t find the right syntax/combo. I did find some maven examples but they all seem to be based on an XML config file.

thc202:
fyi, I already have the changes necessary to accept these changes. With the Gradle plugin you need to add custom rule.

kingthorin:
Great, okay I can rebase once that PR is done and merged. Thanks @thc202!

kingthorin:
Rebased.

thc202:
Thank you!

PR #5553, opened 2019-08-31
Cleanups

rhowe:
A bunch of somewhat scattered cleanups.

I think these are all still Java8 compatible (at least that's what IntelliJ thinks).

Clear out unused code in `ZapNumberSpinner` and do some misc. tidyups there.
Remove the unused Undo Manager stuff from text components
Replace some `for` loops with `foreach` loops or Streams
Remove the long-unused `HtmlContextAnalyser` class


thc202:
The code removed is actually in use (or might be) by add-ons.

rhowe:
> The code removed is actually in use (or might be) by add-ons.

Ah, I did wonder how exposed it would be. Does this extend to the Undo Manager stuff as well?

rhowe:
Closing in favour of https://github.com/zaproxy/zaproxy/pull/5554

thc202:
> Does this extend to the Undo Manager stuff as well?

Yes, it applies to those changes too (in any case it would have to be deprecated first).

PR #5554, opened 2019-08-31
Replace for loops with foreach and Streams

rhowe:
We can replace some for loops with more modern mechanisms

Also it looks like another couple of small linter fixes sneaked in, but they look pretty innocuous and probably not worth a separate PR

Signed-off-by: Russell Howe <rhowe@siksai.co.uk>

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/httppanel/HttpPanel.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/httppanel/HttpPanel.java
@@ -33,7 +33,6 @@ import java.util.Comparator;
 import java.util.Hashtable;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map.Entry;
 import java.util.regex.Pattern;
 import javax.swing.AbstractAction;
 import javax.swing.Action;
@@ -238,11 +237,7 @@ public abstract class HttpPanel extends AbstractPanel implements Tab {
         Message oldMessage = this.message;
         this.message = msg;
 
-        for (Iterator<Entry<String, HttpPanelComponentInterface>> it =
-                        components.entrySet().iterator();
-                it.hasNext(); ) {
-            HttpPanelComponentInterface component = it.next().getValue();
-
+        for (HttpPanelComponentInterface component : components.values()) {
             if (!component.isEnabled(message)) {
                 if (enabledComponents.contains(component)) {
                     disableComponent(component);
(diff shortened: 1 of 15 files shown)


zap/src/main/java/org/zaproxy/zap/extension/httppanel/HttpPanel.java
```diff
@@ -238,10 +238,10 @@ public void setMessage(Message msg) {
         Message oldMessage = this.message;
         this.message = msg;
 
-        for (Iterator<Entry<String, HttpPanelComponentInterface>> it =
-                        components.entrySet().iterator();
-                it.hasNext(); ) {
-            HttpPanelComponentInterface component = it.next().getValue();
+        for (Entry<String, HttpPanelComponentInterface> stringHttpPanelComponentInterfaceEntry :
```
thc202: Any reason why the name is so verbose? Could be just `entry`, it could also be changed to just iterate the values.

zap/src/main/java/org/zaproxy/zap/utils/ClassLoaderUtil.java
```diff
@@ -76,8 +76,8 @@ public static void addURL(URL u) throws IOException {
 
         URLClassLoader sysLoader = (URLClassLoader) ClassLoader.getSystemClassLoader();
         URL[] urls = sysLoader.getURLs();
-        for (int i = 0; i < urls.length; i++) {
-            if (StringUtils.equalsIgnoreCase(urls[i].toString(), u.toString())) {
+        for (URL url : urls) {
```
thc202: Not sure worth changing a deprecated class.

zap/src/main/java/org/zaproxy/zap/utils/ZapTextComponentUndoManager.java
```diff
@@ -49,7 +49,7 @@
  *   <li>{@code UndoManagerPolicy.ALWAYS_DISABLED}
  * </ul>
  *
- * The policy can be changed with the method {@code setUndoManagerPolicy}.
+ * <p>The policy can be changed with the method {@code setUndoManagerPolicy}.
```
thc202: This could be reverted.

rhowe: Ah yes, that was from the other, bigger PR. IntelliJ wanted a `<p>` but it's pretty random to have it in this changeset

zap/src/main/java/org/zaproxy/zap/extension/httppanel/HttpPanel.java
rhowe: Must be how the IDE autogenerated it and yes, I agree on the values thing.. amendment coming right up

kingthorin:
Needs a spotlessApply

thc202:
Thank you!

PR #5557, opened 2019-09-03
Request Plugin Id for SRI missing integrity attribute Scanner

cotonne:
Requesting Plugin Id for a new scanner in zap-extension.
Proposition after some talks on the owasp slack/project-zap channel.

### DIFF
--- a/docs/scanners.md
+++ b/docs/scanners.md
@@ -177,6 +177,7 @@ Scanners:
 
 90001   Insecure JSF ViewState
 90002   Java Serialized Object
+90003   SRI integrity missing attribute
 90011   Charset Mismatch
 90017	XSLT Injection
 90018   SQL Injection SQLMap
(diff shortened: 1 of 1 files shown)


docs/scanners.md
```diff
@@ -177,6 +177,7 @@ Scanners:
 
 90001   Insecure JSF ViewState
 90002   Java Serialized Object
+90003   SRI integrity missing attribute
```
thc202: The `missing` should be at the end?

kingthorin: Good catch. Also the `I` in the accronym is technically `Integrity` which is redundant to include twice so how about either:
`SRI Attribute Missing`
or
`Subresource Integrity Attribute Missing`

thc202: Last one looks good to me.

thc202:
Thanks!

PR #5558, opened 2019-09-03
Fix Title Caps for Active Scan Tools Options

JFM9:
Fix Title Caps for Active Scan options found under Tools -> Options -> Active Scan.

Related to [Issue #2000](https://github.com/zaproxy/zaproxy/issues/2000).

Signed-off-by: JFM9 <JFM9@users.noreply.github.com>


kingthorin:
Thanks for this, I'll go through them tonight. 

With just a quick look it seems like the guidance in: https://github.com/zaproxy/zaproxy/wiki/DevGuidelines#style-guidelines isn't quite specific enough. Some of these are probably better handled by just putting a period on the end.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -409,12 +409,12 @@ ascan.custom.warn.disabled	= Input vectors disabled via the Custom Vectors tab
 
 ascan.desc                     = Active scanner, heavily based on the original Paros active scanner, but with additional tests added
 
-ascan.options.anticsrf.label   = Handle anti-CSRF tokens (experimental functionality)
+ascan.options.anticsrf.label   = Handle Anti-CSRF Tokens (Experimental Functionality)
```
kingthorin: I think leaving this one as it was and just adding a period probably makes more sense.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -409,12 +409,12 @@ ascan.custom.warn.disabled	= Input vectors disabled via the Custom Vectors tab
...
 ascan.options.apply.label		= Apply
-ascan.options.attackOnStart.label = Allow Attack mode to be enabled on startup
-ascan.options.attackPrompt.label = In Attack mode prompt to rescan nodes when scope changed
-ascan.options.attackRescan.label = In Attack mode always rescan nodes when scope changed
-ascan.options.delayInMs.label  = Delay when scanning in milliseconds:
+ascan.options.attackOnStart.label = Allow Attack Mode to Be Enabled on Startup
+ascan.options.attackPrompt.label = In Attack Mode Prompt to Rescan Nodes When Scope Changed
+ascan.options.attackRescan.label = In Attack Mode Always Rescan Nodes When Scope Changed
+ascan.options.delayInMs.label  = Delay When Scanning in Milliseconds:
```
kingthorin: I think a period works for these four too. You could leave the caps on "Attack Mode" since that's a name.

kingthorin: For the delay one maybe: `Delay When Scanning (In Milliseconds):` because that seems to fit better with the style of the other two max settings?

JFM9: These changes will be in the next commit

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
JFM9: Personally I see this one as title more than a sentence. I can change it if you think that would be preferable.

kingthorin: Hmmm I'm torn (and it seems I missed the "Inject" one). 

I'm thinking for consistency it makes sense for the green parts to be sentences and the blue parts to be titles. I'm open to discussion though:

image

thc202: Agreed.

JFM9: Yup. I agree that the "Inject" one should also be a sentence.

The the sake of fluidity, I would also agree that "Handle anti..." should also be a sentence.

thc202:
Thank you!

kingthorin:
@jpm9 let us know how you want to appear in the credits: https://github.com/zaproxy/zap-core-help/wiki/HelpCredits#zap-extended-team

Name/handle/link/no-link ..... have a look at what others have done and let us know.

JFM9:
@kingthorin, "Justin Matte (JFM9)" is fine.

Thanks!

PR #5566, opened 2019-09-07
Refactor SimpleEventBus

rhowe:
Bring this class up to date with Java 8, making better use of the Collections
and Streams APIs and reducing the use of varargs a little.

Signed-off-by: Russell Howe <rhowe@siksai.co.uk>

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
+++ b/zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
@@ -22,10 +22,8 @@ package org.zaproxy.zap.eventBus;
 import java.security.InvalidParameterException;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -39,9 +37,8 @@ import org.apache.logging.log4j.Logger;
  */
 public class SimpleEventBus implements EventBus {
 
-    private Map<String, RegisteredPublisher> nameToPublisher =
-            new HashMap<String, RegisteredPublisher>();
-    private List<RegisteredConsumer> danglingConsumers = new ArrayList<RegisteredConsumer>();
+    private Map<String, RegisteredPublisher> nameToPublisher = new HashMap<>();
+    private List<RegisteredConsumer> danglingConsumers = new ArrayList<>();
 
     /**
      * The {@code Lock} for registration management (register and unregister) of publishers and
@@ -61,18 +58,19 @@ public class SimpleEventBus implements EventBus {
         }
         regMgmtLock.lock();
         try {
-            if (nameToPublisher.get(publisher.getPublisherName()) != null) {
+            String publisherName = publisher.getPublisherName();
+            if (nameToPublisher.get(publisherName) != null) {
                 throw new InvalidParameterException(
                         "Publisher with name "
-                                + publisher.getPublisherName()
+                                + publisherName
                                 + " already registered by "
                                 + nameToPublisher
-                                        .get(publisher.getPublisherName())
+                                        .get(publisherName)
                                         .getPublisher()
                                         .getClass()
                                         .getCanonicalName());
             }
-            log.debug("registerPublisher " + publisher.getPublisherName());
+            log.debug("registerPublisher " + publisherName);
 
             RegisteredPublisher regProd = new RegisteredPublisher(publisher, eventTypes);
 
@@ -80,7 +78,7 @@ public class SimpleEventBus implements EventBus {
             // Check to see if there are any cached consumers
             danglingConsumers.removeIf(
                     regCon -> {
-                        if (regCon.getPublisherName().equals(publisher.getPublisherName())) {
+                        if (regCon.getPublisherName().equals(publisherName)) {
                             consumers.add(regCon);
                             return true;
                         }
@@ -88,7 +86,7 @@ public class SimpleEventBus implements EventBus {
                     });
 
             regProd.addConsumers(consumers);
-            nameToPublisher.put(publisher.getPublisherName(), regProd);
+            nameToPublisher.put(publisherName, regProd);
         } finally {
             regMgmtLock.unlock();
         }
@@ -102,11 +100,11 @@ public class SimpleEventBus implements EventBus {
 
         regMgmtLock.lock();
         try {
-            log.debug("unregisterPublisher " + publisher.getPublisherName());
-            RegisteredPublisher regPub = nameToPublisher.remove(publisher.getPublisherName());
-            if (regPub == null) {
+            String publisherName = publisher.getPublisherName();
+            log.debug("unregisterPublisher " + publisherName);
+            if (nameToPublisher.remove(publisherName) == null) {
                 throw new InvalidParameterException(
-                        "Publisher with name " + publisher.getPublisherName() + " not registered");
+                        "Publisher with name " + publisherName + " not registered");
             }
         } finally {
             regMgmtLock.unlock();
@@ -154,9 +152,7 @@ public class SimpleEventBus implements EventBus {
         regMgmtLock.lock();
         try {
             log.debug("unregisterConsumer " + consumer.getClass().getCanonicalName());
-            for (Entry<String, RegisteredPublisher> entry : nameToPublisher.entrySet()) {
-                entry.getValue().removeConsumer(consumer);
-            }
+            nameToPublisher.values().forEach(publisher -> publisher.removeConsumer(consumer));
             // Check to see if its cached waiting for a publisher
             removeDanglingConsumer(consumer);
         } finally {
@@ -165,12 +161,8 @@ public class SimpleEventBus implements EventBus {
     }
 
     private void removeDanglingConsumer(EventConsumer consumer) {
-        Iterator<RegisteredConsumer> iter = danglingConsumers.iterator();
-        while (iter.hasNext()) {
-            if (iter.next().getConsumer().equals(consumer)) {
-                iter.remove();
-            }
-        }
+        danglingConsumers.removeIf(
+                registeredConsumer -> registeredConsumer.getConsumer().equals(consumer));
     }
 
     @Override
@@ -187,11 +179,11 @@ public class SimpleEventBus implements EventBus {
                             + " for "
                             + publisherName);
             RegisteredPublisher publisher = nameToPublisher.get(publisherName);
-            if (publisher != null) {
-                publisher.removeConsumer(consumer);
-            } else {
+            if (publisher == null) {
                 // Check to see if its cached waiting for the publisher
                 removeDanglingConsumer(consumer);
+            } else {
+                publisher.removeConsumer(consumer);
             }
         } finally {
             regMgmtLock.unlock();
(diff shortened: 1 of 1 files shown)


psiinon:
@rhowe thanks for submitting this :) Did you see the LGTM alert above?

rhowe:
> @rhowe thanks for submitting this :) Did you see the LGTM alert above?

Yeah, it's a bit spurious I think - the collection contents aren't accessed directly but `registerPublisher` calls `removeIf` on it so the contents are used as a marker to populate another collection - LGTM and the IntelliJ IDEA linter both got a bit confused by this.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -284,11 +255,10 @@ public String getPublisherName() {
 
     private static class RegisteredPublisher {
         private EventPublisher publisher;
-        private String[] eventTypes;
+        private List<String> eventTypes;
```
preetkaran20: is it possible to make it as Set/LinkedHashSet so that contains complexity can be reduced.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -284,11 +255,10 @@ public String getPublisherName() {
...
         private List<RegisteredConsumer> consumers = new CopyOnWriteArrayList<>();
```
preetkaran20: one doubt i have, do we need CopyOnWriteArrayList here ?

thc202: Yes, #4827.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
thc202: We don't need order here so `HashSet` would be enough (though I doubt that this a performance issue, there are not that many events).

preetkaran20: I thought there are more event types.

thc202: Well it depends on the publisher some have more than others, but each one does not have many (less than 10).

thc202: Maybe that's already enough to see a difference, will check...

preetkaran20: oh.. ok. then makes sense.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
kingthorin: Marking resolved, evidence for necessity provided.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
rhowe: It comes in as an array in `registerConsumer` anyway and we never modify it. In fact, maybe it should just be made immutable.

rhowe: I just pushed an update which makes these `eventType` lists immutable

preetkaran20:
> This pull request **introduces 1 alert** when merging [a3e79d6](https://github.com/zaproxy/zaproxy/commit/a3e79d64de65b716f132c01c44b7c38cf9a55b5f) into [4b47f14](https://github.com/zaproxy/zaproxy/commit/4b47f14fe9b308fff683d73694552743366e55a3) - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-bcf21a9c018382878fbeb2edc8b6bf513062ab2e)
> 
> **new alerts:**
> 
> * 1 for Container contents are never accessed

I think this is a false positive as you are modifying container too.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
preetkaran20: IMHO, I think in registerConsumer and in registerPublisher we are running contains search and incase there are 10 consumers each with 10 event types then in worst case it will be 100 searches in ``` getConsumer ``` call. As we are already converting array to List we can instead convert to Set if we are not accessing using index.

preetkaran20:
Except Performance related discussion. LGTM.

thanks,
Karan

rhowe:
> > This pull request **introduces 1 alert** when merging [a3e79d6](https://github.com/zaproxy/zaproxy/commit/a3e79d64de65b716f132c01c44b7c38cf9a55b5f) into [4b47f14](https://github.com/zaproxy/zaproxy/commit/4b47f14fe9b308fff683d73694552743366e55a3) - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-bcf21a9c018382878fbeb2edc8b6bf513062ab2e)
> > **new alerts:**
> > 
> > * 1 for Container contents are never accessed
> 
> I think this is a false positive as you are modifying container too using removeIf.

Yep, think we can call this one a bug in the linter - shows up in IntelliJ and LGTM

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
rhowe: Sure, sounds good to me

rhowe:
> Except Performance related discussion. LGTM.
> 
> thanks,
> Karan

We're now using a `HashSet`

preetkaran20:
> > Except Performance related discussion. LGTM.
> > thanks,
> > Karan
> 
> We're now using a `HashSet`

Yes, I have seen the changes. LGTM. Thanks

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -19,16 +19,14 @@
...
+import static java.util.Arrays.asList;
+
 import java.security.InvalidParameterException;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
+import java.util.*;
```
kingthorin: Please use specific imports not wildcard.

rhowe: Argh, I blame IDEA

kingthorin: No worries. Intelli-J tends to do that on default settings as well.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
thc202: For the record, with 100 searches it's still a rounding error, you would need way more to see the difference (though a Set is clearly the best choice, good performance in all cases).

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -60,34 +63,33 @@ public void registerPublisher(EventPublisher publisher, String... eventTypes) {
         }
         regMgmtLock.lock();
         try {
-            if (this.nameToPublisher.get(publisher.getPublisherName()) != null) {
+            final String publisherName = publisher.getPublisherName();
```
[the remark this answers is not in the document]
rhowe: I added `final` as it's referenced by the lambdas. Whilst it's not strictly necessary, it seemed like a helpful hint. The compiler doesn't seem to really care though.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -60,34 +63,33 @@ public void registerPublisher(EventPublisher publisher, String... eventTypes) {
...
                                         .getCanonicalName());
             }
-            log.debug("registerPublisher " + publisher.getPublisherName());
+            log.debug("registerPublisher " + publisherName);
 
-            RegisteredPublisher regProd = new RegisteredPublisher(publisher, eventTypes);
+            RegisteredPublisher regProd =
+                    new RegisteredPublisher(publisher, new HashSet<>(asList(eventTypes)));
 
-            List<RegisteredConsumer> consumers = new ArrayList<>();
```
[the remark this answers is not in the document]
rhowe: This list only exists so that we can add a bunch of consumers in one go, but we don't need this batch functionality and if we add them one by one we can remove `RegisteredPublisher.addConsumers` and simplify `RegisteredPublisher` to only have one way to add consumers.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -203,78 +204,54 @@ public void publishSyncEvent(EventPublisher publisher, Event event) {
...
         private EventConsumer consumer;
-        private String[] eventTypes;
+        private Set<String> eventTypes;
         private String publisherName;
 
-        public RegisteredConsumer(EventConsumer consumer, String[] eventTypes) {
+        RegisteredConsumer(EventConsumer consumer, Set<String> eventTypes) {
             this.consumer = consumer;
-            this.eventTypes = eventTypes;
+            this.eventTypes = unmodifiableSet(eventTypes);
```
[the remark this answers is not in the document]
rhowe: We don't need to, no, but I thought it would be nice to make our intentions clear. It still irks me that the immutability ot collections isn't part of the type system :(

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
thc202: > but we don't need this batch functionality

It avoids changing `consumers` multiple times.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
thc202: I think it's implied/clear that's not intended to be modified (why would we change the event types that the consumer wants?).

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
rhowe: Ah, I see - since it's a `CopyOnWriteArrayList` that's more expensive

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
rhowe: Actually there are no methods on `RegisteredConsumer` that let you mutate it anyway so yeah, it can go

kingthorin:
Seems to need a spotlessApply

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
```diff
@@ -284,36 +261,33 @@ public String getPublisherName() {
...
-        public List<RegisteredConsumer> getConsumers() {
-            return consumers;
-        }
-
-        public void addConsumers(List<RegisteredConsumer> consumers) {
-            this.consumers.addAll(consumers);
+        List<RegisteredConsumer> getConsumers(String eventType) {
+            return consumers.stream()
+                    .filter(consumer -> consumer.wantsEvent(eventType))
+                    .collect(Collectors.toList());
```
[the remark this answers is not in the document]
thc202: This is still creating the collection.

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
thc202: Any reason to not restore the previous behaviour?

zap/src/main/java/org/zaproxy/zap/eventBus/SimpleEventBus.java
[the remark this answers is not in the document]
rhowe: Sorry, just returning to this after a massive hiatus. How about if we just return a `Stream` here?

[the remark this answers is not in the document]
thc202: Sounds good to me, the `Stream` or the `List` itself (IMO we don't need to be over protective here, it's all internal behaviour).

kingthorin:
@rhowe are you able to address the outstanding comments?

thc202:
Rebased to fix the conflict and target main branch. Will look at addressing the remaining comments.

thc202:
Comments addressed.

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):
Thank you both

Issue #5334, opened 2019-05-02
Bad Addons - Need descriptive message(s)


kingthorin:
I encountered this on a Kali VM this morning. I'm not sure what kind of weirdness I may have done to break them (start an update and then kill the VM while it was writing etc):

```
17730 [ZAP-cfu] WARN org.zaproxy.zap.control.AddOnCollection  - Failed to create add-on for ascanrulesBeta
java.util.zip.ZipException: zip file is empty
	at java.base/java.util.zip.ZipFile$Source.zerror(ZipFile.java:1527)
	at java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1341)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1435)
	at java.base/java.util.zip.ZipFile$Source.<init>(ZipFile.java:1266)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1229)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:727)
	at java.base/java.util.zip.ZipFile$CleanableResource.get(ZipFile.java:845)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:245)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:175)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:189)
	at org.zaproxy.zap.control.AddOn.loadManifestFile(AddOn.java:350)
	at org.zaproxy.zap.control.AddOn.<init>(AddOn.java:428)
	at org.zaproxy.zap.control.AddOnCollection.load(AddOnCollection.java:125)
	at org.zaproxy.zap.control.AddOnCollection.<init>(AddOnCollection.java:53)
	at org.zaproxy.zap.extension.autoupdate.ExtensionAutoUpdate$8.run(ExtensionAutoUpdate.java:958)
17786 [ZAP-cfu] WARN org.zaproxy.zap.control.AddOnCollection  - Failed to create add-on for jxbrowserlinux64
java.util.zip.ZipException: zip END header not found
	at java.base/java.util.zip.ZipFile$Source.zerror(ZipFile.java:1527)
	at java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1428)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1435)
[...]

PR #5569, opened 2019-09-10
Issue 5334 Suppress Stack Trace - Corrupt Addon(s)

kingthorin:
This doesn't address the un-install issue, however, it does address the stack trace. With this change the stack is only included in the log if debug is enabled or running in dev mode (in case one of the devs is
trying to troubleshoot something).

Related to zaproxy/zaproxy#5334

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/control/AddOn.java
+++ b/zap/src/main/java/org/zaproxy/zap/control/AddOn.java
@@ -453,7 +453,12 @@ public class AddOn {
         try {
             return Optional.of(new AddOn(file));
         } catch (AddOn.InvalidAddOnException e) {
-            logger.warn("Invalid add-on: " + file.toString(), e);
+            String logMessage = "Invalid add-on: " + file.toString() + ".";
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/zaproxy/zap/control/AddOn.java
```diff
@@ -453,7 +453,11 @@ public static ValidationResult isValidAddOn(Path file) {
         try {
             return Optional.of(new AddOn(file));
         } catch (AddOn.InvalidAddOnException e) {
-            logger.warn("Invalid add-on: " + file.toString(), e);
+            if (logger.isDebugEnabled() || Constant.isDevMode()) {
+                logger.warn("Invalid add-on: " + file.toString(), e);
+            } else {
+                logger.warn("Invalid add-on: " + file.toString());
```
thc202: Worth including the exception's message in this case. The common message could be extracted to avoid duplication.

kingthorin:
Tweaked

thc202:
Thanks!

PR #5580, opened 2019-09-16
Remove homepage URL from ZAP extensions

thc202:
The URL is redundant (already provided through the Online menu item,
added by Online Menus add-on) and does not provide any specific info
about the extensions.
Deprecate `Constant#ZAP_HOMEPAGE` and `ZAP_EXTENSIONS_PAGE`, not in use
by any core class.


psiinon:
Looks like the build failed due to binary incompatibilities

thc202:
Yes, working on that.

thc202:
The build is now passing with the changes done in #5615.

PR #5586, opened 2019-09-18
install zap python3 version of the library in containers

northdpole:
the containers are based off of ubuntu which still ships python2.7 (to be deprecated in January) as the default python interpreter.

This pull request installs python3 and fetches the python3 version of python-owasp-zap-v2.4
This change allows containers running python3 to be based off of the zap weekly container and the scripts still working.

It also makes it easier to move to python3 as the main interpreter in preparation for removing python2 when it gets deprecated

### DIFF
--- a/docker/Dockerfile-live
+++ b/docker/Dockerfile-live
@@ -20,6 +20,7 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	xterm \
 	net-tools \
 	python-pip \
+	python3-pip \
 	firefox \
 	vim \
 	xvfb \
@@ -27,6 +28,7 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/*  && \
 	pip install --upgrade pip zapcli python-owasp-zap-v2.4 && \
+	pip3 install --upgrade pip zapcli python-owasp-zap-v2.4 && \
 	useradd -d /home/zap -m -s /bin/bash zap && \
 	echo zap:zap | chpasswd && \
 	mkdir /zap  && \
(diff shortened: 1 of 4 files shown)


kingthorin:
@northdpole ?

northdpole:
> @northdpole ?

closed it as I managed to solve the issue from my side another way,
if you still think it's useful I can reopen it

kingthorin:
I'm for re-opening it, definitely a step in the right direction.

thc202:

kingthorin:
https://github.com/zaproxy/zaproxy/blob/develop/docker/CHANGELOG.md

northdpole:
https://github.com/zaproxy/zaproxy/pull/5588  this pr should update the changelog

docker/Dockerfile-live
```diff
@@ -19,6 +19,7 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	openbox \
 	xterm \
 	net-tools \
+	python3-pip \
```
thc202: This could be done in the same order as the stable.

docker/Dockerfile-weekly
```diff
@@ -18,13 +18,15 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	openbox \
 	xterm \
 	net-tools \
-	python-pip \
+	python-pip
```
thc202: The backslash was accidentally removed?

docker/Dockerfile-weekly
```diff
@@ -18,13 +18,15 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
...
+	python3-pip \
 	firefox \
 	xvfb \
 	x11vnc && \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/*
 
+RUN pip3 install --upgrade pip zapcli python-owasp-zap-v2.4
```
thc202: Same here regarding the order.

thc202:
Any reason why it was not added here?

thc202:
Addressed the comments and moved the changelog changes here.

thc202:
Thanks!

PR #5594, opened 2019-09-24
Fix deadlock when adding HTTP views

### LINKED ISSUE #5585: ZAP 2.8.1 hanging on kali
Hi i need help with zaproxy. 
when is start is go the 
22508 [ZAP-BootstrapGUI] INFO org.parosproxy.paros.extension.ExtensionLoader  - Initializing HTTP Panel Syntax Highlighter View Extension point and after that is stop working.

My OS is Linux kali 5.2.0-kali2-amd64 #1 SMP Debian 5.2.9-2kali1 (2019-08-22) x86_64 GNU/Linux

My Java version is 
java -version
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-8u212-b01-1-b01)
OpenJDK 64-Bit Server VM (build 25.212-b01, mixed mode)

this is the info 
https://pastebin.com/HXhjt96E

Untitled1

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Software versions**
 - ZAP: [note that we typically only support the latest version of ZAP]
 - Add-on: Name and version [e.g. Passive Scan Rules Alpha v19. https://github.com/zaproxy/zaproxy/wiki/FAQAddonVersions]
 - OS: [e.g. Windows 10]
 - Java: [e.g. openjdk version "1.8.0_171"]
 - Browser: [e.g. chrome 66, firefox 59, safari 11.1]

See https://github.com/zaproxy/zaproxy/wiki/FAQconfig for the location of the zap.log file for your OS.

**Additional context**

thc202:
Change `HttpPanelManager` to execute the changes in the EDT to avoid
deadlock between bootstrap thread and EDT.
Remove internal synchronisation, no longer needed as the changes will be
done only in the EDT.
Add helper method to `ThreadUtils` that runs a `Runnable` in the EDT
while handling the exceptions, also, tweak/add JavaDocs.

Fix #5585 - ZAP 2.8.1 hanging on kali

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/view/HttpPanelManager.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/HttpPanelManager.java
@@ -29,6 +29,7 @@ import org.zaproxy.zap.extension.httppanel.HttpPanel;
 import org.zaproxy.zap.extension.httppanel.component.HttpPanelComponentInterface;
 import org.zaproxy.zap.extension.httppanel.view.HttpPanelDefaultViewSelector;
 import org.zaproxy.zap.extension.httppanel.view.HttpPanelView;
+import org.zaproxy.zap.utils.ThreadUtils;
 
 public class HttpPanelManager {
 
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/zaproxy/zap/view/HttpPanelManager.java
```diff
@@ -56,68 +61,95 @@ private static synchronized void createInstance() {
     }
 
     public void addRequestPanel(HttpPanel panel) {
-        requestPanels.addPanel(panel);
+        runInEdt(() -> requestPanels.addPanel(panel));
+    }
+
+    private static void runInEdt(Runnable runnable) {
```
psiinon: Is it worth moving this to a public utils type class so that it can be used everywhere?
We need to do it in so many places...

thc202: There's `ThreadUtils` but does not handle the exceptions (I guess we could still use it here).

thc202: Moved the method that handles the exceptions to `ThreadUtils` and changed this class to use it.

Issue #4673, opened 2018-05-16
Dynamic SSL sub-certificate expirations too long


marc-whil:
When I use dynamic ssl certificates for mitm, I'm getting expirations that go out over 8 years, which is causing Android Chrome to complain with NET::ERR_CERT_VALIDITY_TOO_LONG.

I don't see an option to configure the expiration date in the zap options.

The zap root certificate has a proper expiration of 1 year, but the sub-certificates per site have the inappropriately long expiration.  The zap documentation says "Every dynamically generated certificate is valid for 1000 days.", but that doesn't seem to be true.

Can we get an option to configure the duration of sub-certificates?

thc202:
Thanks for raising the issue!

dessant:
This issue affects all recent clients, the maximum validity of certificates issued after March 1, 2018 is limited to 825 days, while zap issues intermediate certificates with a validity of [1000 days](https://github.com/zaproxy/zaproxy/blob/65afe6af651cc789e96a3079edd86dd025ca8da1/zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java#L164).

https://www.trustzone.com/ssl-certificate-validity-is-now-capped-at-a-maximum-of-2-years/

kingthorin:
Have you encountered a UA that rejects Zap's cert due to validity period?

dessant:
Yes, i have tested with the latest version of Firefox on Arch, it errors out with ERR_CERT_VALIDITY_TOO_LONG. I'm also having issues with an Andorid emulator, running Android Pie.

kingthorin:
Does Firefox give you the option to proceed anyway?

I agree it should be fixed, just trying to judge impact.

dessant:
Yes, the error can be bypassed on Firefox. It causes a bigger problem when the entire system is proxied, I couldn't find a workaround for proxying the entire traffic of the OS on Android.

dessant:
Burp has the same issue, but they use the validity of the root certificate for intermediates, so adding a custom certificate is their current workaround.

https://support.portswigger.net/customer/portal/questions/17538500-burp-suite-1-7-37-create-burp-certificate-less-than-39-month

thc202:
@dessant which Firefox version is latest? (for the record)

dessant:
I was using Firefox 69.0.1.

thc202:
Thanks.

kingthorin:
[...]

PR #5598, opened 2019-09-25
Revise site certificate validity period to 825d

kingthorin:
Related to: zaproxy/zaproxy#4673 and per: https://cabforum.org/2017/03/17/ballot-193-825-day-certificate-lifetimes/

We should probably still have a UI component to control this in the options panel(s). For the time being this modification will address the current concerns.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
+++ b/zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
@@ -74,6 +74,18 @@ import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
  */
 public final class SslCertificateServiceImpl implements SslCertificateService {
 
+    /**
+     * Constant used to define the start validity date for site certificates. Used as 30d before
+     * "now"
+     */
+    private static final int SITE_CERTIFICATE_START_ADJUSTMENT = 30;
+    /**
+     * Constant used to define the end validity date for site certificates. 2years minus start
+     * adjustment. Per: https://cabforum.org/2017/03/17/ballot-193-825-day-certificate-lifetimes/
+     */
+    private static final int SITE_CERTIFICATE_END_VALIDITY_PERIOD =
+            825 - SITE_CERTIFICATE_START_ADJUSTMENT;
+
     private X509Certificate caCert = null;
     private PublicKey caPubKey = null;
     private PrivateKey caPrivKey = null;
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -161,7 +162,7 @@ public KeyStore createCertForHost(CertData certData)
                         new X509CertificateHolder(caCert.getEncoded()).getSubject(),
                         BigInteger.valueOf(serial.getAndIncrement()),
                         new Date(System.currentTimeMillis() - Duration.ofDays(30).toMillis()),
-                        new Date(System.currentTimeMillis() + Duration.ofDays(1000).toMillis()),
+                        new Date(System.currentTimeMillis() + Duration.ofDays(825).toMillis()),
```
kingthorin: Would it make more sense for this to be 825-30?

dessant: Yes, it seems a total validity of 825 days is necessary.

dessant:
@kingthorin, thanks for the quick fix!

zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
kingthorin: Tweaked, thanks.

zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -161,7 +163,7 @@ public KeyStore createCertForHost(CertData certData)
                         new X509CertificateHolder(caCert.getEncoded()).getSubject(),
                         BigInteger.valueOf(serial.getAndIncrement()),
                         new Date(System.currentTimeMillis() - Duration.ofDays(30).toMillis()),
-                        new Date(System.currentTimeMillis() + Duration.ofDays(1000).toMillis()),
+                        new Date(System.currentTimeMillis() + Duration.ofDays(825 - 30).toMillis()),
```
kingthorin: I've entered it this way to that it's more clear what's going on (yes 795 might have been more simple but I'm not sure it would be as clear to whoever works on this code next).

dessant: It's possible that the cert could have a longer validity than 825 when the values between the two `System.currentTimeMillis()` differ. The current time could be sourced once and referenced from a variable.

The root certificate currently expires after 365 days, so maybe the intermediate could have a much shorter validity too.

kingthorin: I'll adjust the clock call, that makes sense.

The ultimate plan is for an options UI to be put in place for this at which point users will be able to control the validity period.

zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -17,6 +17,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// ZAP: 2019/09/25 Adjusted site certificates from 1000d to 825d, per
```
thc202: Could be removed (ZAP class).

zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
```diff
@@ -156,12 +158,13 @@ public KeyStore createCertForHost(CertData certData)
...
 
+        long currentTime = System.currentTimeMillis();
         X509v3CertificateBuilder certGen =
                 new JcaX509v3CertificateBuilder(
                         new X509CertificateHolder(caCert.getEncoded()).getSubject(),
                         BigInteger.valueOf(serial.getAndIncrement()),
-                        new Date(System.currentTimeMillis() - Duration.ofDays(30).toMillis()),
-                        new Date(System.currentTimeMillis() + Duration.ofDays(1000).toMillis()),
+                        new Date(currentTime - Duration.ofDays(30).toMillis()),
+                        new Date(currentTime + Duration.ofDays(825 - 30).toMillis()),
```
thc202: Worth extracting a constant (the JavaDoc could reference the decision also).

kingthorin: Will do.

kingthorin:
All addressed I think

thc202:
Thanks!

PR #5599, opened 2019-09-25
Adds option to disable/enable cookies on manual resend

### LINKED ISSUE #1164: Manual "Resend" sets cookies on response
```
If the response to the Resend function sets a cookie, then that cookie appears in the
request pane and will be included in subsequent requests.

To reproduce:
1. Identify a request that results in a set-cookie in the response.
2. Choose to "Resend" this request
3. If the cookie already exists in the request, remove it.
4. Send request
5. The cookie will be set in the Request pane.

Desired behaviour:
The Resend function should not act on set-cookie headers from the server.  Or it should
be optional.


```

Original issue reported on code.google.com by `stephendv` on 2014-04-24 09:18:43

ZainabAlShowely:
Fixes #1164 

Adds to the manual send/resend window an option to disable cookies being set on request after a response setting a cookie. Changes the icon for the 'Use current tracking session' option to a globe.


zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -74,12 +75,9 @@ public HttpPanelSender(HttpPanelRequest requestPanel, HttpPanelResponse response
 
         requestPanel.addOptions(
                 getButtonUseTrackingSessionState(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
+        requestPanel.addOptions(getButtonUseCookies(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
         requestPanel.addOptions(
                 getButtonFollowRedirects(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
-
-        final boolean isSessionTrackingEnabled =
-                Model.getSingleton().getOptionsParam().getConnectionParam().isHttpStateEnabled();
-        getButtonUseTrackingSessionState().setEnabled(isSessionTrackingEnabled);
```
kingthorin: `updateOptions` should be called here so that the session tracking button is disabled if the option to track state is disabled.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -213,6 +211,13 @@ public void cleanup() {
         }
     }
 
+    public void updateOptions() {
+        final boolean isSessionTrackingEnabled =
+                Model.getSingleton().getOptionsParam().getConnectionParam().isHttpStateEnabled();
+
+        useTrackingSessionState.setEnabled(isSessionTrackingEnabled);
```
kingthorin: Could use `getButtonUseTrackingSessionState()` instead.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
```diff
@@ -117,6 +120,15 @@ protected void initialize() {
         getFooterStatusBar().add(getLabelContentLength());
         getFooterStatusBar().addSeparator();
         getFooterStatusBar().add(getLabelTotalLength());
+
+        getWindowPanel()
+                .addComponentListener(
+                        new ComponentAdapter() {
+                            @Override
+                            public void componentShown(ComponentEvent ce) {
+                                sender.updateOptions();
```
kingthorin: re: the earlier comment on `HttpPanelSender` this doesn't seem to be happening for some reason (or at least it wasn't on my Windows dev box).

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -270,7 +291,8 @@ private void checkState() {
      * @since 2.8.0
      */
     public void setUseGlobalState(boolean enableGlobalState) {
-        if (enableGlobalState) {
+        useGlobalState = enableGlobalState;
```
kingthorin: Tweak to `this.useGlobalState = enableGlobalState` for clarity?

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
ZainabAlShowely: I think componentShown might not be appropriate for this case actually, will change.

thc202: Which case is this addressing?

ZainabAlShowely: @thc202 I had thought it might not update if the tracking state option was changed while the message sending panel was open, but that is not the case. It seems sufficient to add a call to the update function on construction of the window to address the initial concern.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -247,7 +253,7 @@ private JToggleButton getButtonUseTrackingSessionState() {
                     new JToggleButton(
                             new ImageIcon(
                                     HttpPanelSender.class.getResource(
-                                            "/resource/icon/fugue/cookie.png"))); // Cookie
+                                            "/resource/icon/fugue/globe-green.png")));
```
thc202: What's the source/license of this icon?

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -259,6 +262,24 @@ private void checkState() {
         }
     }
 
+    private void updateCookieState() {
```
thc202: Might worth merging this method with `checkState` so that we have only one method changing/setting up the state.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -259,6 +262,24 @@ private void checkState() {
...
+        if (!useCookies) {
+            // discard cookies, but keep credentials
+            HttpState state = new HttpState();
+            HttpState proxyState = new HttpState();
+
+            state.setCredentials(AuthScope.ANY, client.getState().getCredentials(AuthScope.ANY));
```
thc202: This should not be needed, we don't set credentials (at least not to one of these states directly).

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -270,7 +291,8 @@ private void checkState() {
      * @since 2.8.0
      */
     public void setUseGlobalState(boolean enableGlobalState) {
-        if (enableGlobalState) {
+        this.useGlobalState = enableGlobalState;
+        if (useGlobalState) {
```
thc202: This could be moved to the merged method mentioned previously.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -280,6 +302,24 @@ public void setUseGlobalState(boolean enableGlobalState) {
...
+    /**
+     * Sets whether or not the requests sent should keep track of cookies.
+     *
+     * @param shouldUseCookies {@code true} if cookies should be used, {@code false} otherwise.
+     * @since TODO Add version
+     */
+    public void setUseCookies(boolean shouldUseCookies) {
+        this.useCookies = shouldUseCookies;
+
+        if (useCookies) {
```
thc202: Same here.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -615,12 +655,17 @@ private HttpMethod runMethod(HttpMessage msg, boolean isFollowRedirect, HttpMeth
             // cant do this for EntityEnclosingMethod methods - it will fail
             method.setFollowRedirects(isFollowRedirect);
         }
+
+        // needs to verify if the state
+        // should store cookies
+        updateCookieState();
```
thc202: Is this really needed? The state/settings should be properly set always?

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -615,12 +655,17 @@ private HttpMethod runMethod(HttpMessage msg, boolean isFollowRedirect, HttpMeth
...
+
         // ZAP: Use custom HttpState if needed
         User forceUser = this.getUser(msg);
         if (forceUser != null) {
             this.executeMethod(method, forceUser.getCorrespondingHttpState());
         } else {
-            this.executeMethod(method, null);
+            this.executeMethod(method, client.getState());
```
thc202: This should not be necessary (this would have the side effect of enabling cookies for the current request, even if the cookies are actually disabled), the client state is already used internally.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1929,6 +1929,7 @@ manReq.name = Manual Request Editor Extension
 manReq.button.send             = Send
 manReq.checkBox.followRedirect = Follow redirect
 manReq.checkBox.useSession     = Use current tracking session
+manReq.checkBox.useCookies     = Update request with cookies from response
```
thc202: `Accept Cookies` instead?

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -178,6 +179,8 @@
     private MultiThreadedHttpConnectionManager httpConnManager = null;
     private MultiThreadedHttpConnectionManager httpConnManagerProxy = null;
     private boolean followRedirect = false;
+    private boolean useCookies = true;
+    private boolean useGlobalState = false;
```
thc202: This can be left unintialised (it's always initialised in the constructor).

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
thc202: This change can be reverted then? (Also the changes in `HttpPanelSender` constructor and the new method`updateOptions`.)

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -259,6 +262,24 @@ private void checkState() {
         }
     }
 
+    private void updateCookieState() {
+        if (!useCookies) {
+            // discard cookies, but keep credentials
+            HttpState state = new HttpState();
```
thc202: btw, probably better to not reset the state and just disable/enable cookies, instead we should provide a method to reset the state (so that we can reset it any time from the UI/programmatically).

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
ZainabAlShowely: From fugue icons, under Creative Commons 3.0 (the other icons seem to come from the same icon set so I assume attribution was made)

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
ZainabAlShowely: I think it isn't needed, as per your comment on not using `client.getState()` it was probably only necessary because the cookies were set for the request and needed to be deleted by resetting the state.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
thc202: Yeah, just checking :)

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/ManualHttpRequestEditorDialog.java
ZainabAlShowely: Yes. These changes are now reverted in the latest commit.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -237,6 +240,7 @@ public HttpSender(ConnectionParam connectionParam, boolean useGlobalState, int i
                         defaultUserAgent);
 
         setUseGlobalState(useGlobalState);
+        setUseCookies(true);
```
kingthorin: There still seems to be a sync issue with opening and closing the dialog and the button state/behaviour. Setting `setUseCookies(useCookies);` here seems closer but still not 100% :cry:

ZainabAlShowely: @kingthorin Weird, will test more and try to see where the problem is!

ZainabAlShowely: @kingthorin should be fixed in the latest commit, the state wasn't being properly relayed from the graphical interface.

kingthorin: Great, I'll test it out later tonight or in the AM.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -237,6 +240,16 @@ public HttpSender(ConnectionParam connectionParam, boolean useGlobalState, int i
...
+    }
+
+    public HttpSender(
```
thc202: It's not necessary to add another constructor (we should move to builder pattern), instead the calling code can make use of the cookie setter after creating the instance.

zap/src/main/java/org/parosproxy/paros/network/HttpSender.java
```diff
@@ -249,16 +262,34 @@ public static SSLConnector getSSLConnector() {
...
+            if (param.isHttpStateEnabled()) {
+                client.setState(param.getHttpState());
+                clientViaProxy.setState(param.getHttpState());
+                setProxyAuth(clientViaProxy);
+                setClientsCookiePolicy(CookiePolicy.BROWSER_COMPATIBILITY);
+            } else {
+                setClientsCookiePolicy(CookiePolicy.IGNORE_COOKIES);
+            }
+        } else {
+            client.setState(new HttpState());
```
thc202: Could call `resetState`.

thc202:
Thank you!

PR #5620, opened 2019-10-09
Ensure -configfile maintains key order

### LINKED ISSUE #5619: The -configfile option can fail with arrays
When array values are defined in a config file, eg:
```
replacer.full_list(0).description=auth1
replacer.full_list(0).enabled=true
replacer.full_list(0).matchtype=REQ_HEADER
replacer.full_list(0).matchstr=Authorization
replacer.full_list(0).regex=false
replacer.full_list(0).replacement=SBSBSB
replacer.full_list(1).description=auth2
replacer.full_list(1).enabled=true
replacer.full_list(1).matchtype=REQ_HEADER
replacer.full_list(1).matchstr=AnotherHeader
replacer.full_list(1).regex=false
replacer.full_list(1).replacement=BLAHBLAH
```
then the key order is not maintained when they are applied. This can break if the keys in later rows are applied before those in earlier ones.

psiinon:
Fixes #5619

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/CommandLine.java
+++ b/zap/src/main/java/org/parosproxy/paros/CommandLine.java
@@ -45,15 +45,18 @@
 // ZAP: 2018/06/29 Add command line to run ZAP in dev mode.
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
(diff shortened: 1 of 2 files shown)


psiinon:
For info implemented the tests first (TDD!) - the config one passed but the configfile one failed ;)

psiinon:
Tweaked to fix new lgtm alerts

PR #5621, opened 2019-10-09
Run tests with JUnit 5

thc202:
Add dependency on JUnit 5 and remove JUnit 4.
Use Mockito library for JUnit 5.

Change tests to:
 - Use new annotations/imports;
 - Inline temporary directories when just needed in few tests;
 - Assert exceptions with `assertThrows`;
 - Explicitly assert that no exception is thrown (in tests that had no
  assertion and where no exception is expected);
 - Make more mocks lenient where needed.

### DIFF
--- a/zap/src/test/java/ch/csnc/extension/httpclient/AliasCertificateUnitTest.java
+++ b/zap/src/test/java/ch/csnc/extension/httpclient/AliasCertificateUnitTest.java
@@ -19,18 +19,18 @@
  */
 package ch.csnc.extension.httpclient;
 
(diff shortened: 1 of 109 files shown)


thc202:
Excluded JUnit 4 introduced by WireMock, updated imports to use Hamcrest's `assertThat`.

psiinon:
Now conflicting

thc202:
Yeah, changed to WIP, need to review latest tests added/changed as well.

thc202:
Conflicts resolved and updated tests added in the meantime.

PR #5656, opened 2019-10-29
Add Base64url to Encoder/Decoder Tool

plonibarploni:
Add an additional field to the encoder/decoder tool to allow encoding/
decoding of base64 encoded strings that conform to the base64url
specification

Signed-off-by: plonibarploni <44826203+plonibarploni@users.noreply.github.com>


zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
```diff
@@ -123,6 +124,18 @@ public String getBase64Decode(String msg) throws IllegalArgumentException, IOExc
         return new String(Base64.decode(msg, Base64.NO_OPTIONS), base64Charset);
     }
 
+    public String getBase64urlEncode(String msg) throws NullPointerException, IOException {
```
preetkaran20: I think better if we can check null in the start of the method by using  ``` Objects.requiresNonNull or null comparison ```  also I think there is no need to explicitly stating Null Pointer Exception in throws.

zap/src/main/java/org/zaproxy/zap/extension/encoder2/EncodeDecodeDialog.java
```diff
@@ -586,6 +612,23 @@ private void updateEncodeDecodeFields() {
...
+            log.error(e.getMessage(), e);
+        }
+
+        try {
+            base64urlDecodeField.setText(
+                    getEncoder().getBase64urlDecode(getInputField().getText()));
+            base64urlDecodeField.setEnabled(base64urlDecodeField.getText().length() > 0);
+        } catch (IOException | IllegalArgumentException e) {
+            base64urlDecodeField.setText(e.getMessage());
+            base64urlDecodeField.setEnabled(false);
```
preetkaran20: can we log the exception here ?

zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
plonibarploni: I was following the style of the existing code. I don't mind changing it, but I think it would be better to change all these instances instead of just this new one.

zap/src/main/java/org/zaproxy/zap/extension/encoder2/EncodeDecodeDialog.java
plonibarploni: As above, the style is the same as the existing code.

zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
preetkaran20: IMHO, I think we should not introduce something if we are planning to refactor that. We will refactor existing later. @kingthorin and @thc202 please suggest.

thc202: I don't think worth refactor/tweak these classes, the plan is to drop them at some point, superseded by the new encoder/decoder add-on (yes, the current implementation is still using the `Encoder` class but hopefully that will be changed...).

zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
```diff
@@ -123,6 +124,18 @@ public String getBase64Decode(String msg) throws IllegalArgumentException, IOExc
...
+        return Base64.encodeBytes(
+                getBytes(msg),
+                base64EncodeOptions == Base64.NO_OPTIONS
```
thc202: This check could be removed (`NO_OPTIONS | URL_SAFE` yields `URL_SAFE`).

zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
thc202: Resolving per above comment.

zap/src/main/java/org/zaproxy/zap/extension/encoder2/EncodeDecodeDialog.java
thc202: Same here.

zap/src/main/java/org/parosproxy/paros/extension/encoder/Encoder.java
thc202: Non essential tweak, resolving.

thc202:
Thanks!

PR #5657, opened 2019-10-29
Allow to raise alerts more easily

thc202:
Add `Alert` builder to allow to create alerts fluently.
Expose the builder to passive/active scan rules to build and raise the
alerts.
Deprecate old way of raising alerts, which required to specify all/many
alert fields even if not needed.
Update passive/active script templates.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -63,6 +63,7 @@
 // ZAP: 2019/03/22 Add bingo with references.
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
+// ZAP: 2019/10/21 Use and expose Alert builder.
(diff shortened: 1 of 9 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -1142,4 +1146,191 @@ public int getWascId() {
...
+
+        /**
+         * Raises the alert with specified data.
+         *
+         * @throws IllegalStateException if the HTTP message was not set.
+         */
+        public void raise() {
+            if (!messageSet) {
+                throw new IllegalStateException(
+                        "A HTTP message must be set before raising the alert.");
```
kingthorin: Should this include source and plugin ID to assist troubleshooting?

thc202: Can add, but the stacktrace will already include the scanner/class.

kingthorin: Okay don’t worry about it.

PR #5662, opened 2019-10-31
Add config for lockbot

kingthorin:
Per: https://github.com/dessant/lock-threads


thc202:
Thanks!

Issue #5303, opened 2019-04-03
API calls for adding and changing alerts


psiinon:
Bit of an omission, and we'll need it for the HUD before too long.
Also requested via https://stackoverflow.com/questions/55488829/how-to-add-zap-alert-through-zap-python-api

PR #5664, opened 2019-11-01
Fix Issue 5303 - Add API call for changing and adding alerts

kingthorin:
AlertAPI.java > Updated with the new constants and functionality to handle updating an alert or adding an alert via the API.
Messages.properties > Updated to include descriptions for the new endpoints and their params.

Fixes zaproxy/zaproxy#5303

Ex (update alert):
image 

Edit: screenshot is out-of-date, there is a description on the name field.


kingthorin:
I intend to do the add functionality, basically the same model in a separate PR.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -442,15 +644,34 @@ private int getRiskId(JSONObject parameters) throws ApiException {
...
         if (riskId < Alert.RISK_INFO || riskId > Alert.RISK_HIGH) {
-            throwInvalidRiskId();
+            throwInvalidAlertElementId(
+                    Constant.messages.getString(
+                            "alert.api.view.alerts.param.riskId.illegal", PARAM_RISK));
         }
         return riskId;
     }
 
+    private RecordHistory getRecordHistory(TableHistory tableHistory, Integer id)
```
kingthorin: I ripped this code from CoreAPI (instead of making that method `public static`, I'm happy to do it another way if applicable. (Or we could move it to some sort of APIUtils class?)

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -578,6 +658,21 @@ private int getRiskId(JSONObject parameters) throws ApiException {
...
+        } catch (HttpMalformedHeaderException | DatabaseException e) {
+            logger.error("Failed to read the history record:", e);
+            throw new ApiException(ApiException.Type.INTERNAL_ERROR, e);
+        }
+        if (recordHistory == null) {
+            throw new ApiException(ApiException.Type.DOES_NOT_EXIST, Integer.toString(id));
+        }
+        return recordHistory;
+    }
+
```
kingthorin: I ripped this code from CoreAPI (instead of making that method `public static`, I'm happy to do it another way if there are suggestions. (Or we could move it to some sort of APIUtils class?)

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
 
                 removeHistoryReferenceAlerts(rootNode);
             }
+        } else if (ACTION_UPDATE_ALERT.equals(name)) {
+            int alertId = ApiUtils.getIntParam(params, PARAM_ALERT_ID);
+            String alertName = ApiUtils.getNonEmptyStringParam(params, PARAM_ALERT_NAME);
+            int riskId = ApiUtils.getIntParam(params, PARAM_ALERT_RISK);
+            int confidenceId = ApiUtils.getIntParam(params, PARAM_ALERT_CONFIDENCE);
+            String desc = ApiUtils.getOptionalStringParam(params, PARAM_ALERT_DESCRIPTION);
```
thc202: This is misleading, the parameter is mandatory. Was it meant to be optional?

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+            String solution = ApiUtils.getOptionalStringParam(params, PARAM_ALERT_SOLUTION);
+            String refs = ApiUtils.getOptionalStringParam(params, PARAM_ALERT_REFS);
+            String evidence = ApiUtils.getOptionalStringParam(params, PARAM_ALERT_EVIDENCE);
+            int cweId = ApiUtils.getIntParam(params, PARAM_ALERT_CWEID);
+            int wascId = ApiUtils.getIntParam(params, PARAM_ALERT_WASCID);
+
+            if (riskId < Alert.RISK_INFO || riskId > Alert.RISK_HIGH) {
+                throwInvalidAlertElementId(
+                        Constant.messages.getString(
+                                "alert.api.action.updateAlert.param.riskId.illegal",
```
thc202: We don't usually i18n the exception messages in the API.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                                PARAM_ALERT_RISK));
+            }
+            if (confidenceId < Alert.CONFIDENCE_FALSE_POSITIVE
```
thc202: Worth extracting a method, could be similar to `getRiskId`.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                logger.error(e.getMessage(), e);
+                throw new ApiException(ApiException.Type.INTERNAL_ERROR, e);
+            }
+
+            if (recAlert == null) {
+                throw new ApiException(ApiException.Type.DOES_NOT_EXIST, PARAM_ALERT_ID);
+            }
+
+            final ExtensionAlert extAlert =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class);
```
thc202: This could be replaced with the instance variable `extension`.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+            }
+            if (confidenceId < Alert.CONFIDENCE_FALSE_POSITIVE
+                    || confidenceId > Alert.CONFIDENCE_USER_CONFIRMED) {
+                throwInvalidAlertElementId(
+                        Constant.messages.getString(
+                                "alert.api.action.addAlert.param.confidenceId.illegal",
+                                PARAM_ALERT_CONFIDENCE));
+            }
+
+            HttpMessage msg = recordHistory.getHttpMessage();
```
thc202: We need to create the `HistoryReference` as well, otherwise this will be interpreted as a new message when the alert is raised.
It could do `HttpMessage msg = getHttpMessage(messageId);` with `getHttpMessage` as:
```Java
private static HttpMessage getHttpMessage(int id) throws ApiException {
    try {
        return new HistoryReference(id, true).getHttpMessage();
    } catch (HttpMalformedHeaderException | DatabaseException e) {
        if (e.getMessage() == null) {
            throw new ApiException(ApiException.Type.DOES_NOT_EXIST, PARAM_ALERT_MESSAGE_ID);
        }
        logger.error("Failed to read the history record:", e);
        throw new ApiException(ApiException.Type.INTERNAL_ERROR, e);
    }
}
```

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+
+            final ExtensionAlert extAlert =
+                    Control.getSingleton().getExtensionLoader().getExtension(ExtensionAlert.class);
+            if (extAlert != null) {
+                Alert newAlert =
+                        new org.parosproxy.paros.core.scanner.Alert(
```
thc202: The package could be removed.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -283,6 +342,154 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
+                newAlert.setDescription(desc);
+                newAlert.setParam(param);
+                newAlert.setAttack(attack);
+                newAlert.setOtherInfo(otherInfo);
+                newAlert.setSolution(solution);
+                newAlert.setReference(refs);
+                newAlert.setEvidence(evidence);
+                newAlert.setCweId(cweId);
+                newAlert.setWascId(wascId);
+                extAlert.alertFound(newAlert, msg.getHistoryRef());
```
thc202: It should return the ID of the alert instead of OK.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -415,12 +622,8 @@ private void processAlerts(
         return new ApiResponseSet<String>("alertsummary", map);
     }
 
-    private static void throwInvalidRiskId() throws ApiException {
-        throw new ApiException(
-                ApiException.Type.ILLEGAL_PARAMETER,
-                "Parameter "
-                        + PARAM_RISK
-                        + " is not a valid risk ID (integer in interval [0, 3]).");
+    private static void throwInvalidAlertElementId(String message) throws ApiException {
```
thc202: Should be renamed to `throwIllegalApiParameter` (just doing that).

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -442,15 +645,34 @@ private int getRiskId(JSONObject parameters) throws ApiException {
         try {
             riskId = Integer.parseInt(riskIdParam);
         } catch (NumberFormatException e) {
-            throwInvalidRiskId();
```
thc202: It's better to keep this method to avoid the duplication.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -35,6 +35,38 @@ alert.label.url				= URL:
...
 
+alert.api.action.updateAlert = Update the alert with the given ID, with the provided details.
+alert.api.action.updateAlert.param.id = The ID of the alert to update.
+alert.api.action.updateAlert.param.name = The name of the alert.
+alert.api.action.updateAlert.param.riskId = The numeric risk representation ('0 - Informational' through '3 - High').
+alert.api.action.updateAlert.param.riskId.illegal = Parameter {0} is not a valid risk ID (integer in interval [0, 3]).
+alert.api.action.updateAlert.param.confidenceId = The numeric confidence representation ('1 - Low' through '3 - High' [user set values '0 - False Positive', and '4 - User Confirmed' are also available]).
+alert.api.action.updateAlert.param.confidenceId.illegal = Parameter {0} is not a valid confidence ID (integer in interval [0, 4]).
+alert.api.action.updateAlert.param.description = The description to be set to the alert.
+alert.api.action.updateAlert.param.param = The name of the parameter the applicable to the alert.
```
thc202: Remove the second "the"?

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
thc202: Yeah, `ApiUtils` would be a good place (if really needed, per comment about the `HistoryReference`).

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin: No just didn’t see another obvious String method.

kingthorin: Your earlier comment addresses this, dunno why I didn’t see that autocomplete option I. Eclipse

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin: I’m not set on this, I just figured since all the descriptions etc were being added I could also tackle this....

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
thc202: OK, for param/attack/evidence/other info/refs might be better if optional?

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
thc202: Right, I'm fine with that.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin: Sure

thc202: CWE/WASC IDs too.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin: What duplication? Both `getRiskId` and the new `getConfidenceId` are going to use `throwIllegalApiParameter` doesn't that reduce duplication?

thc202: It's duplicating the message (`Constant.messages...`).

kingthorin: But that doesn't go away, it just means I need to add a throwInvalidConfidenceId and still have two message keys.... unless we go back to non-i18n messages but I thought we agreed to that in the other comments?

kingthorin: I've reduced the message entries from 6 to 2 in the latest iteration, maybe that addresses your concern?

kingthorin: Spoke in IRC, I think we're on the same page now.

thc202:
Thanks!

PR #5672, opened 2019-11-07
Do not prompt to accept the license on first start

psiinon:
As per discussions on irc, its not really necessary so is just an annoyance for users.


psiinon:
Binary incompatibility, which I dont think is relevant for this PR (or the other one for that matter;) These are really internal methods.

kingthorin:
Agreed, on my cell right now. Will review the other shortly.

thc202:
Thanks!

PR #5677, opened 2019-11-09
Kubernetes Service tests

northdpole:
Allows ZAP to run on kubernetes as a one node service.
Decouples launching zap from orchestrating and provides an example full scan orchestration script with an accompanying Dockerfile and Pod.

Useful for running ZAP as part of a CI where you need a ZAP instance to always be running for proxying Service acceptance tests or end to end tests through it.

Incidentally, with the help of an ingress if it's running on a cluster that supports ingresses, it can be used for basic manual testing with the help of the HUD.

The whole affair is quite quick and dirty and can be certainly optimised/cleaned/documented. If you're interested in merging this I can certainly  do so.


psiinon:
I love the idea behind these and can see them being very useful.
However I do worry about the maintainability.
Could the docker images build on the existing ZAP docker images to make them easier to maintain?
How does zap_remote_full_scan.py differ from zap_full_scan.py ?
In general I'd much rather we extend existing images and scripts rather than adding new ones.

northdpole:
> I love the idea behind these and can see them being very useful.
> However I do worry about the maintainability.
> Could the docker images build on the existing ZAP docker images to make them easier to maintain?
The  existing ZAP  docker images launch ZAP in  the same container and use the API to orchestrate a local ZAP.
This defeats the purpose of running a remote ZAP where several clients can connect to (which is the point of this PR)

> How does zap_remote_full_scan.py differ from zap_full_scan.py ?

zap_remote_full_scan.py doesn't launch a local  ZAP, instead it connects to a remote ZAP instance.
(Try running it outside of Kubernetes or with an invalid --zHost, you should get an IOError)

zap_full_scan.py  launches a local zap instance in the same container.

> In general I'd much rather we extend existing images and scripts rather than adding new ones.

In principle I fully agree, however, this use case fundamentally differs from the normal ZAP use case of "one ZAP instance for a ZAP client"

Here's a complicated suggestion:

If remote ZAP is a direction you'd like to move the project towards, what if we rebuilt the "docker/" part of ZAP to split launching and orchestrating into 2 scripts.

Then there will be one launch_zap script which will be used to launch ZAP (it can be a cleaner, more robust version of the launch_zap script from this PR)  this script just launches a ZAP backend instance and monitors/reports Java heap usage and general performance as well as ZAP heartbeats.

Given that, all the zap_*scan.py scripts can be refactored to only connect to a ZAP instance they didn't launch (with a default being 127.0.0.1:8080) and their only concern is orchestration including shutting down the instance when they're done.

A self-contained image then only needs to launch the first script in the background and the second in the foreground.

This approach allows the work in this PR to be refactored only to 2 different dockerfiles. A zap-k8s that only launches ZAP and a zap-remote that only launches one of the orchestration scripts with a default value for ZAP pointing to wherever the service points to (or it gets it from the environment and expects a configmap to set it as is the case with this PR)

northdpole:
polite bump,  got some free time for documentation or refactoring coming up if this makes it in :-)

omerlh:
I just build a [zap operator](https://github.com/omerlh/zap-operator) that can be used to proxy services on Kubenretes. You just need to create a new CRD object and specify the target service. This was done without any changes to Zap's code...

kingthorin:
@northdpole we've been talking about things like this lately (additional images/implementations, 3rd party add-ons, etc). We appreciate your effort but don't really need something else to maintain or generate. We are quite willing to link to your repo from our automation pages (on zaproxy.org) or something along those lines.

If you could let us know where/what you'd like us to link to we'll figure it out. In the meantime we'd like to close this un-finished PR. Thanks for your interest!

PR #5681, opened 2019-11-11
Remove unused dependencies

thc202:
Remove JDOM and Diff Utils, all (known) add-ons were already updated
accordingly.


psiinon:

thc202:
Addressed.

PR #5690, opened 2019-11-15
fixing resize bug in the splash screen

growlnx:
While zaproxy is loading, the user can change the size of the splash screen window, but the application does not handle it by resizing the image and loading bar.

**before changes:**
image

**after changes:**
image

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
@@ -75,8 +75,10 @@ public class SplashScreen extends JFrame {
 
     public SplashScreen() {
         super();
-
+        this.setUndecorated(true);
+        this.setAlwaysOnTop(true);
         setSize(DisplayUtils.getScaledDimension(420, 430));
+        setResizable(false);
         setLocationRelativeTo(null);
         setTitle(Constant.PROGRAM_NAME);
         setIconImages(DisplayUtils.getZapIconImages());


zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
```diff
@@ -75,8 +75,10 @@
 
     public SplashScreen() {
         super();
-
+        this.setUndecorated(true);
```
thc202: This was not set on purpose, it caused issues in an OS.

zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
```diff
@@ -75,8 +75,10 @@
...
+        this.setAlwaysOnTop(true);
```
thc202: We should not need to set this.

zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
```diff
@@ -75,8 +75,10 @@
...
         setSize(DisplayUtils.getScaledDimension(420, 430));
+        setResizable(false);
```
thc202: This is more a workaround, the progress bar and background image can have a fixed size.

kingthorin:
@growlnx are you able to address thc202's feedback?

zap/src/main/java/org/zaproxy/zap/view/SplashScreen.java
growlnx: Does the splash screen really need to be resizable?

thc202: No, but it's better if it can be (e.g. easier to read the tips and log when maximised).

PR #5691, opened 2019-11-15
Expose Context details to passive scan rules

kingthorin:
- PassiveScanData > A new helper class which exposes the utility methods (getUsers(), and hasContext()) for the HttpMessage currently being passively scanned.
- PluginPassiveScanner > Added new member and getter (getHelper()) for the PassiveScanData (which is set in the init() method).
- PassiveScanThread > Instantiates a PassiveScanData variable with the current HttpMessage and passes it to PluginPassiveScanner.init().

As far as usage goes: Passive scan rules will be able to do things like `getHelper().getUsers()` etc.


thc202:
Thanks!

kingthorin:
Thanks i also have tge TechSet stuff ready, will submit another PR shortly.

PR #5696, opened 2019-11-18
Expose Context techset directly to passive scan rules

kingthorin:
I also considered adding direct support for AuthMethod, AuthzDetectionMethod, DDN, SessionManagementMethod but couldn't think of use-cases. (Also they're available via the extra step
getHelper.getContext.getWhatever...)


kingthorin:
Tested with:
```js
function scan(ps, msg, src) {
     var Tech = Java.type("org.zaproxy.zap.model.Tech");
	print('TechSet includes PHP? ' + ps.getHelper().getTechSet().includes(Tech.PHP));
}
```

```console
TechSet includes PHP? true
```

PR #5697, opened 2019-11-19
Update Session Management default template.js

bscheller-iso:
there could be more infomation about the SessionWrapper. "getHttpMessage()   - returns an HttpMessage" adds no value.

### DIFF
--- a/zap/src/main/dist/scripts/templates/session/Session Management default template.js
+++ b/zap/src/main/dist/scripts/templates/session/Session Management default template.js
@@ -3,7 +3,7 @@
 //
 // The SessionWrapper class provides the following methods:
 //   getSession()       - returns a ScriptBasedSession
-//   getHttpMessage()   - returns an HttmMessage
+//   getHttpMessage()   - returns an HttpMessage
 //   getParam(key)      - returns the parameter associated with the specified required or optional param name
 //
 // The ScriptBasedSession class provides the following methods:


kingthorin:
@bscheller-iso thanks for catching the typo.

We'll need you to fulfill the DCO requirement in order to merge: https://github.com/zaproxy/zaproxy/pull/5697/checks?check_run_id=309914409

thc202:
> there could be more information about the SessionWrapper.

Which information are you looking for? Would the following be more clear about its purpose "The parameter sessionWrapper provides the following methods:" ?

> adds no value.

While redundant it's consistent with the other methods.

bscheller-iso:
The SessionWrapper could be described with some information. 
E.g that he is used to access the actual properties and to manipulate properties that they meet the own requirements.

What is a ScriptBasedSession and how it differs fom an "normal" Session?

Some Information about the call seqence of the script in general would be useful. When the script is called? How can the script be triggered? (I miss this kind of information in several places)
Wich HTTPMessage is exactly returned depending to the call sequence (long living session Vs. an HTTPMessage). The fist one or all? From Spider, AjaxSpider etc or all?

Regarding "extractWebSession":
- I cannot understand the description... what is a script web session? The session of the script? And how it differes from the ScriptBasesSession?
- The name states an extraction of a websession. But the code within the function manipulates some session values.

Issue #5710, opened 2019-11-27
Unable to proxy https in Safari and Chrome on Mac OS Catalina (browsers launched outside ZAP)


alex1704:
Hi,
I generated Root CA Certificate in Dynamic SSL Certificates tab and then added it to trusted certificates on keychains. However when i try to load https on safari or chrome i see error that my connection is not private. I suspect that it might be related to tightening requirements for trusted certificates for MacOS  Catalina - https://support.apple.com/en-us/HT210176.
Firefox is working fine, probably because certificate is added to browser manually instead of relying on one that resides in keychain.
<img width="713" alt="Screenshot 2019-11-27 at 03 28 04" src="https://user-images.githubusercontent.com/1874436/69686615-f8afdd00-10c8-11ea-978d-05afabff71f4.png">

<img width="672" alt="Screenshot 2019-11-27 at 03 29 51" src="https://user-images.githubusercontent.com/1874436/69686599-ecc41b00-10c8-11ea-9947-0ac66ce84629.png">

kingthorin:
Please try the weekly release, it includes some changes that may address this.

https://github.com/zaproxy/zaproxy/wiki/Downloads#zap-weekly

RemakingEden:
I have just tried on the most recent weekly release and I am still getting issues with iOS 13. It seems the certificate is not being trusted fully. Should I start a new bug report or is this one valid as it appears to relate to the same thing?

thc202:
Did you regenerate the Root CA cert?

This issue is fine.

RemakingEden:
@thc202 Thanks for the reply. 

I have tried a few times now, regenerating, removing and re-adding. It does not appear change anything.

kingthorin:
If you’re regenerating the cert the validity dates (timestamps) should be changing.

thc202:
Might be because of:
> TLS server certificates must contain an ExtendedKeyUsage (EKU) extension containing the id-kp-serverAuth OID.

we are not doing that (I think that's the only requirement missing).

kingthorin:
Isn't that this:
https://github.com/zaproxy/zaproxy/blob/ea704dc2fe8e80883cfc0ff9fa67817c11947cc4/zap/src/main/java/org/zaproxy/zap/extension/dynssl/SslCertificateUtils.java#L162-L168

kingthorin:
Oh it's on the root but not on the site certs:
[...]

PR #5714, opened 2019-11-27
Fix OSX id_kp_serverAuth requirement

kingthorin:
SslCertificateServiceImpl > Update createCertForHost(CertData) to include the ExtendedKeyUsage id_kp_serverAuth OID.

Fixes zaproxy/zaproxy#5710 
Related to zaproxy/zaproxy#4673

image

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
+++ b/zap/src/main/java/org/parosproxy/paros/security/SslCertificateServiceImpl.java
@@ -45,9 +45,11 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.bouncycastle.asn1.x500.X500NameBuilder;
 import org.bouncycastle.asn1.x500.style.BCStyle;
 import org.bouncycastle.asn1.x509.BasicConstraints;
+import org.bouncycastle.asn1.x509.ExtendedKeyUsage;
 import org.bouncycastle.asn1.x509.Extension;
 import org.bouncycastle.asn1.x509.GeneralName;
 import org.bouncycastle.asn1.x509.GeneralNames;
+import org.bouncycastle.asn1.x509.KeyPurposeId;
 import org.bouncycastle.asn1.x509.SubjectKeyIdentifier;
 import org.bouncycastle.cert.X509CertificateHolder;
 import org.bouncycastle.cert.X509v3CertificateBuilder;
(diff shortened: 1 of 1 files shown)


thc202:
I was not able to verify that this indeed fixes the issue (no macOS), but the change looks good/makes sense.

kingthorin:
Thanks same here

kingthorin:
Forced push, just corrected a typo in the commit message.

psiinon:
Fraid I dont have time to check, will merge so other people can test the live docker or dev build

RemakingEden:
Not working on iOS i'm afraid. Still getting the warning invalid certificate error. Is there any other information I can provide you? logs etc?

thc202:
Does the invalid cert error provides details about why it's not valid? That would be useful.

RemakingEden:
@thc202 i've just found the issue. It was just me not understanding the changes in iOS 13. Nightmare.

"On iOS 10.3 and onwards, you also need to enable full trust for the root certificate:

Go to Settings > General > About > Certificate Trust Settings.
Under “Enable full trust for root certificates”, turn on trust for the certificate. "

It is probably written somewhere in the docs and i just didn't see it as I am used to how it operated in iOS 12 and below. My bad. If not the quoted line above is from the MITMproxy's doc and could be added?

kingthorin:
@RemakingEden thanks for following up. We can definitely document that somewhere. Can you confirm that it does indeed work with the weekly and the CA cert set as you've outlined?

RemakingEden:
@kingthorin I can confirm that definitely has worked with the latest weekly build with a fresh cert on iOS 13. Thanks for everyones help.

shubhamkumbhalkar:
> @thc202 i've just found the issue. It was just me not understanding the changes in iOS 13. Nightmare.
> 
> "On iOS 10.3 and onwards, you also need to enable full trust for the root certificate:
> 
> Go to Settings > General > About > Certificate Trust Settings.
> Under “Enable full trust for root certificates”, turn on trust for the certificate. "
> 
> It is probably written somewhere in the docs and i just didn't see it as I am used to how it operated in iOS 12 and below. My bad. If not the quoted line above is from the MITMproxy's doc and could be added?
I tried following these steps! Can you tell me which settings we need to update?

kingthorin:
The part between the quotation marks.

PR #5709, opened 2019-11-26
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

PR #5716, opened 2019-11-27
Spelling localization

jsoref:
Split from #5709

I'm not squashing yet, but squashing is fairly trivial (I'll try to avoid leaving each DCO...)


thc202:
It's not necessary to change translated `Messages.properties` files (only main, the others are updated from Crowdin, once restored).

thc202:
Thank you!

Issue #5722, opened 2019-11-29
Add spring security csrf token name to the configuration


MichaelVetter:
The Anti CSRF Tokens Scanner does not detect the csrf token from spring security and reports an alert. The form parameter looks like this in the request:
...&_csrf=db8d189b
See https://docs.spring.io/spring-security/site/docs/5.2.1.RELEASE/reference/htmlsingle/#servlet-csrf-include

When I add the token in the ui, the file config.xml contains this:
```
....
            <token>
                <name>_csrf</name>
                <enabled>true</enabled>
            </token>
        </tokens>
    </anticsrf>
```
Please add the token "_csrf" to the default configuration for the daemon mode.

kingthorin:
https://github.com/zaproxy/zaproxy/blob/90f19b34aca1cf21ca11584cca65b9c7885b79d9/zap/src/main/java/org/zaproxy/zap/extension/anticsrf/AntiCsrfParam.java#L54

MichaelVetter:
Great, thank you. I only tested it with the last (windows) release 2.8.0.

thc202:
That was added in 2.8.0 (#4946), did you use other version prior 2.8.0?

MichaelVetter:
Yes, I used an older version before. Maybe the old configuration was not updated. I will test that again.

thc202:
Old configurations were not being updated so they would miss the new default tokens, #5723 fixes that.

PR #5723, opened 2019-11-29
AntiCSRF Tokens Add New Defaults

kingthorin:
Changed AntiCsrfParam to include new default tokens that are added in versions of zap newer than the config file associated with the original install.

Created method `addMissingTokens()` called in the `parse()` method, which depends on the existing local public method `addToken(String)`. Which in turn ensures uniqueness. (addMissingTokens() call setTokens() to ensure that the config is written.)

Related to zaproxy/zaproxy#5722

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/anticsrf/AntiCsrfParam.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/anticsrf/AntiCsrfParam.java
@@ -20,6 +20,7 @@
 package org.zaproxy.zap.extension.anticsrf;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import org.apache.commons.configuration.ConversionException;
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/zaproxy/zap/extension/anticsrf/AntiCsrfParam.java
```diff
@@ -87,16 +88,24 @@ protected void parse() {
             this.enabledTokensNames = new ArrayList<>(DEFAULT_TOKENS_NAMES.length);
         }
 
-        if (this.tokens.size() == 0) {
-            for (String tokenName : DEFAULT_TOKENS_NAMES) {
-                this.tokens.add(new AntiCsrfParamToken(tokenName));
-                this.enabledTokensNames.add(tokenName);
-            }
-        }
+        addMissingTokens();
```
kingthorin: Should I call `setTokens()` here (or inside `addMissingTokens()`) so that the config is written with the updated tokens?

thc202: Worth doing in `addMissingTokens`.

thc202:
IMO and ideally, we should not add them always, just once on update (e.g. with `Constant#upgradeFrom` or with `VersionedAbstractParam`, which would keep the updates closer to the source).
We have the same problem with other params updated recently (e.g. `HttpSessionsParam`).

kingthorin:
Added `setTokens()` call.

thc202:
Thanks!

PR #5728, opened 2019-12-04
spelling: e.g.

jsoref:
We're starting to use ZAProxy.

image

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/CommandLine.java
+++ b/zap/src/main/java/org/parosproxy/paros/CommandLine.java
@@ -541,7 +541,7 @@ public class CommandLine {
     }
 
     /**
-     * Returns true if ZAP should not make any unsolicited requests, eg check-for-updates etc
+     * Returns true if ZAP should not make any unsolicited requests, e.g. check-for-updates etc
      *
      * @since 2.8.0
      */
--- a/zap/src/main/java/org/parosproxy/paros/CommandLine.java
+++ b/zap/src/main/java/org/parosproxy/paros/CommandLine.java
@@ -541,7 +541,7 @@ public class CommandLine {
     }
 
     /**
-     * Returns true if ZAP should not make any unsolicited requests, e.g. check-for-updates etc
+     * Returns true if ZAP should not make any unsolicited requests, e.g. check-for-updates, etc.
      *
      * @since 2.8.0
      */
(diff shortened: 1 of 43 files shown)


kingthorin:
This should only include changes to the main messages.properties file not all the translated versions. The PR needs a `spotlessApply` to pass the build.

kingthorin:
@jsoref do you have thoughts on use of e.g. versus e.x.? There are probably occurrences of "ex" and "e.x." in the code base/docs as well.

jsoref:
At some point I'll also apply this rule against `ie` which should be `i.e.`.

I think `ex.` is correct (`ex` would be wrong), but unless you have some reason to use `ex.` instead of `e.g.`, I think I'd suggest using the single `e.g.` (`e.g.`, minus the lack of `.`s seems to be quite common in this repo).

kingthorin:
You can run `spotlessApply` via `gradlew spotlessApply` or `gradlew.bat spotlessApply` in the repo root. (Then check for changed files commit and push.)

kingthorin:
Well I mean all 3 (or all 6 given variants) e.g., e.x., i.e. .... doesn't it make most sense to settle on just one?

jsoref:
Fixing `ie` requires considerably more work than fixing `eg`:
```
zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java:        } catch (Exception ie) {
```

kingthorin:
I didn't mean to rope you into more work simply by talking through this.... Feel free to break the PRs in logical places, don't let me change your plan(s).

jsoref:
Hmm, that looks surprisingly reasonable. Thankfully it doesn't appear that any `ie` referred to `IE` (in browser codebases, some inevitably would...).

Current plan: Go home, eat dinner. (I'm reachable on irc...)

zap/src/main/java/org/parosproxy/paros/CommandLine.java
```diff
@@ -541,7 +541,7 @@ public boolean isNoStdOutLog() {
     }
 
     /**
-     * Returns true if ZAP should not make any unsolicited requests, eg check-for-updates etc
+     * Returns true if ZAP should not make any unsolicited requests, e.g. check-for-updates etc
```
thc202: This `etc` does not have a dot, on purpose?

jsoref: Good catch: The rule I was using looked for a character that wasn't a `.` after `etc`, it fails for `there is no character after etc`.

thc202:
Thank you!

PR #5729, opened 2019-12-04
spelling: minutes; 0 is unlimited

jsoref:
It isn't obvious that `min` means `minutes` and not `minimum` given that `max` is present on the line above.
image

There is no good reason for `Unlimited` or `Min(utes)` to be capitalized:
image

Note: the inconsistent use of `ms` vs. `milliseconds`.

This PR should be considered for discussion as opposed to merging.

Offhand, I suspect that `ms` and `min.` might be favored over `milliseconds` and `minutes`. Either way,`min` isn't the right choice.


thc202:
Thanks!

thc202:
> This PR should be considered for discussion as opposed to merging.

It seems I didn't pay attention... no harm in merging, I guess.

> Offhand, I suspect that `ms` and `min.` might be favored over `milliseconds` and `minutes`. Either way,`min` isn't the right choice.

It depends on how much space it occupies in the UI, I checked that and it seemed ok.

thc202:
The help might need to be updated to match the new labels.

PR #5739, opened 2019-12-04
Reword autoupdate message about having current version to make it l10n friendly

### LINKED ISSUE #5738: Improve version message in Manage Add-ons dialogue
**Describe the bug**
For localization purposes it is not acceptable to paste strings together. Just because one language is comfortable using an order of `a` `:` `b` doesn't mean that another language won't need `b` `:` `c`, nor does it mean that another language might not use a character other than `:`.

**To Reproduce**
1. Go to 'Help>Check for Updates'
2. Read the box for `ZAP Core`

**Expected behavior**
1. en*: Don't have a ` ` before `:`.
2. fr/*: Don't force the use of `:` as the separator
3. en*: It's better to put the message before the version.

**Screenshots**
image

**Software versions**
 - ZAP: 2.8.0
 - OS: Windows 10

**Additional context**

jsoref:
Fixes #5738

### DIFF
--- a/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
+++ b/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
@@ -930,7 +930,7 @@ cfu.check.upd.downloaded   = All updates downloaded, see Output tab for details.
 cfu.check.upd.downloading  = Downloading updates. You can close this dialog and the downloads will carry on in the background.
 cfu.check.zap.downloaded   = ZAP downloaded to {0}
 cfu.check.zap.downloading  = Downloading ZAP. You can close this dialog and the download will carry on in the background.
-cfu.check.zap.latest   = ZAP is on the latest version
+cfu.check.zap.latest   = ZAP is up-to-date ({0})
 cfu.check.zap.newer    = There is a more recent version of OWASP ZAP: {0}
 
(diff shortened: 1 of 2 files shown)


jsoref:
The bot is apparently not smart enough to detect that `refresh` functions as an equivalent guard.

thc202:
The analysis is correct, since `refresh` is true by default the NPE would happen when `corePanel` is not `null` and `latestInfo` is `null`.

kingthorin:
@jsoref were you going to fix the null dereference?

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -930,7 +930,7 @@ cfu.check.upd.downloaded   = All updates downloaded, see Output tab for details.
 cfu.check.upd.downloading  = Downloading updates. You can close this dialog and the downloads will carry on in the background.
 cfu.check.zap.downloaded   = ZAP downloaded to {0}
 cfu.check.zap.downloading  = Downloading ZAP. You can close this dialog and the download will carry on in the background.
-cfu.check.zap.latest   = ZAP is on the latest version
+cfu.check.zap.latest   = ZAP is the latest version ({0})
```
thc202: From IRC talk, `ZAP is up-to-date ({0})`.

thc202:
Thanks!

Issue #5744, opened 2019-12-05
Edit>Previous is non-sensical


jsoref:
**Describe the bug**
"Previous" isn't an "Edit" action

**To Reproduce**
1. Click Edit

**Expected behavior**
"Find Previous"?

**Screenshots**
![ZAP Edit menu](https://user-images.githubusercontent.com/2119212/70193129-2948d580-16cc-11ea-9f43-1a9412c6ea50.png)
![Notepad Edit menu](https://user-images.githubusercontent.com/2119212/70193180-5d23fb00-16cc-11ea-8d32-347085db2dce.png)

**Software versions**
 - ZAP: 2.8.0
 - Add-on: 
 - OS: Windows 10
 - Java: 
 - Browser: 

**Additional context**

PR #5745, opened 2019-12-05
Find Next/Find Previous

jsoref:
#5744


jsoref:
Note: I'm naively assuming that "Next" and "Previous" are related to "Find" and not "Search" (and thus that they're in the wrong section).

If they're really "Next search result", then, um... I don't know what the labels should be, but "Next" and "Previous" are wrong.

thc202:
They are part of the search functionality.

kingthorin:
I could merge some variation of this into: https://github.com/zaproxy/zaproxy/pull/5747 if there’s any need/benefit.

jsoref:
ok, so, this isn't the right text. That PR/task is the right direction. `Edit`>`Search`>`Next result` / `Edit`>`Search`>`Previous result` might be better

thc202:
Closing, superseded by #5747.

PR #5747, opened 2019-12-05
Search: Group Search, Next, and Previous

kingthorin:
Add the existing Search, Next, and Previous items to a submenu (so that it's clear that Next and Prev apply to the search). Rename getters to be more clear. Set the hot key for "Prev" as ctrl+shift+g (next was already ctrl+g). Added/renamed resource keys for the submenu vs the search menu
item.

Add the new JMenu (group) via new ExtensionHookMenu convenience method(s) added to facilitate addition of sub-menu (JMenu) objects.

image

I believe this grouping clarifies things, and should fix zaproxy/zaproxy#5744


kingthorin:
Should I also update these: https://github.com/zaproxy/zaproxy/blob/2b63145a4cd9d14187578f95b29463addd935d10/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L1978-L1981

To be `menu.edit.search.item.*`?

zap/src/main/java/org/parosproxy/paros/extension/ExtensionHookMenu.java
```diff
@@ -245,6 +245,16 @@ public void addFileMenuItem(JMenuItem menuItem) {
         addFileMenuItemImpl(menuItem);
     }
 
+    /**
+     * Add a sub-menu to the File menu
+     *
+     * @param menu the sub-menu to add
+     * @since TODO Add version
+     */
+    public void addFileMenuItem(JMenu menu) {
```
thc202: `MenuItem` → `Menu`?

kingthorin: I was worried about naming confusion, it isn't adding the File Menu. Maybe `addFileSubMenu`?

kingthorin:
Tweaked

thc202:
Thanks!

Issue #5748, opened 2019-12-05
New Script dialog has Save button after Cancel button


jsoref:
**Describe the bug**
There should be a consistent button order between Foo and Cancel

**To Reproduce**
1. Go to the Scripts panel
2. Expand Scripts>Templates>Extender
3. Right click a child item
4. Select New Script...

**Expected behavior**

**Screenshots**
![Save button on right](https://user-images.githubusercontent.com/2119212/70202018-412d5300-16e6-11ea-98d1-222090f13219.png)
![Find with extra padding to the right of cancel](https://user-images.githubusercontent.com/2119212/70192603-441a4a80-16ca-11ea-8a43-59b85a5ec8a0.png)

**Software versions**
 - ZAP: 2.8.0
 - Add-on: 
 - OS: Windows 10
 - Java: OpenJDK11U-jre_x64_windows_hotspot_11.0.5_10 jdk-11.0.5+10-jre
 - Browser: 

**Errors from the zap.log file**

**Additional context**
#1427

PR #5754, opened 2019-12-06
[WIP] Normalise position of cancel button

thc202:
Change more dialogues to show the cancel button at the end.

Part of #5748.


thc202:
WIP until we have a firm decision that this is the way to go... some Swing dialogues show the cancel button at the end (e.g. `JFileChooser`) and others (e.g. `JOptionPane`) at the start... (Linux here, since that might matter).

kingthorin:
According to: https://ux.stackexchange.com/a/1077 placement is platform dependent, so I think we have to trust Swing to place them appropriately for system dialogs (file choosers etc). For ZAP dialogs we'd have to either pick an order and stick with it or check the platform and adapt (which seems overly complicated)...

Issue #2016, opened 2015-11-02
Choose a unique port in the GUI if theres a clash


psiinon:
If the default port (8080) or the one selected is not available on start up then show a prompt for a new port and suggest the first free port above the current one.
The dialog should check the selected port is available.
In daemon mode ZAP should quit if the port isnt available, showing a suitable error message

kingthorin:
Do we want to check the binding address, and remind/warn the user if it's not loop local?

Also need to handle binding errors for 0.0.0.0 specification....not sure how that plays out. If you've specified 0.0.0.0 and it can bind to something does it stay happy? (Even if the user can't bind to or use it on localhost:8080 or whatever they "normally" expect...)

PR #5764, opened 2019-12-13
Issue 2016: Handle proxy conflict

kingthorin:
- ProxyServer.java - Add functionality to pick a random port and prompt the user to use it (or try one they want).
- Proxy.java - Save the proxy port if it had to be selected during startup.
- ProxyPortRetryPanel - Added to display a custom yes/no dialog and a portSpinner.
- Messages.properties - Added keys/values to support these changes.
- MainFooterPanel.java - Add left footer to display the primary proxy details.
- MainFrame > Used to hook options update functionality.
- View > Used to update the new footer in postInit.

Fixes zaproxy/zaproxy#2016

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/control/Control.java
+++ b/zap/src/main/java/org/parosproxy/paros/control/Control.java
@@ -79,6 +79,8 @@
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2019/09/30 Reduce View singleton usage and replace null checks with hasView().
 // ZAP: 2019/12/16 Log path of new session.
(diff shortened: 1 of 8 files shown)


kingthorin:
Example:
I had apache configured to listen on 3 ports:
```
Listen localhost:8080
Listen localhost:8081
Listen localhost:8082
```

image

thc202:
The prompt is also being shown for additional proxies (at start up and after closing the Options dialogue, the later for main proxy too).

kingthorin:
I was about to get the startup additional proxies thing to happen, but I'm not able to recreate the other two. Zap won't let me change to a used port or add on an used port.

thc202:
Launch ZAP twice with the same configs, it should reproduce both cases.

thc202:
Thanks!

PR #5766, opened 2019-12-15
CleanCode: AboutDialog

kingthorin:
- Change no-arg constructor to call the other constructor
- Correct the assignment and use of GridBagConstraints so that it's more logical and aligned to the applicable components.
- Replaced the actionlistener with a lambda.
- Removed all instance variables and corresponding getters, instead inlinee the creation of the panels/buttons (in the constructor), they are called/used just once.
- JavaDoc now mentions org.zaproxy.zap.view.AboutPanel.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/view/AboutDialog.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/AboutDialog.java
@@ -32,9 +32,6 @@ import org.parosproxy.paros.extension.AbstractDialog;
 public class AboutDialog extends AbstractDialog {
 
     private static final long serialVersionUID = 1L;
-    private JPanel jPanel = null;
-    private AboutPanel aboutPanel = null;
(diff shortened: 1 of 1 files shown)


thc202:
Looks good, some suggestions if you want to clean up more:
 - Remove commented statement in `initialize`;
 - Change no-arg constructor to call the other constructor and squash the `initialize` method;
 - Rename `jPanel` to `mainPanel`;
 - The JavaDoc comment of `getAboutPanel` should be updated to mention the ZAP class;
 - The `ActionListener` could be replaced with a lambda (`e -> dispose()`).
 - Remove all instance variables and corresponding getters, instead inline the creation of the panels/buttons (in `initialize` or the constructor), they are called/used just once.

kingthorin:
Sure :+1:

thc202:
Thanks!

PR #5769, opened 2019-12-17
Wait for add-ons to be installed before starting proxy

### LINKED ISSUE #5704: Wait for add-ons to be installed
If you install, for example, the alpha scan rules then they are often not installed in time for them to be applied.
The packaged scripts should check to see if any add-ons are being installed and if so wait until then have completed.

psiinon:
Fixes  #5704


thc202:
Thanks!

PR #5774, opened 2019-12-19
Enhancement: add basic pscan tags API

kingthorin:
Per: https://groups.google.com/forum/#!topic/zaproxy-users/C8YFLF0BhkY

- Add two pscan API action endpoints `disableAllTags` and `enableAllTags`, to disable/enable all RegexAutoTagScanners respectively. Some of these can be quite time consuming (especially on large
responses).
- Messages.properties > Added API descriptions.

The disable call should probably be used early in the packaged docker scripts once the python API client is regenerated to include them.


kingthorin:
spotlessApply(ied) 😀

jyothsna-smax:
can we add the python hooks as well.

kingthorin:
I believe It’ll be in the python client API when it’s generated after the next release.

thc202:
For the record, it's possible to call new APIs with the existing client, e.g.:
```Python
zap._request_api(zap.base + 'pscan/action/disableAllTags')
zap._request_api(zap.base + 'pscan/action/enableAllTags')
```

PR #5787, opened 2020-01-06
Set latest version to default config

thc202:
Remove hardcoded version in default configuration file and set the
latest version after copying the default configuration to the home dir.
Extract method to set the latest version and extract a constant for
the version element.
Update the tests to match the new behaviour.


thc202:
Tests updated for the new behaviour.

PR #5794, opened 2020-01-11
Correct MailTo pattern for RegexAutoTagScanner

kingthorin:
config.xml > Correct default.
Constant.java > Add method to look for the incorrect pattern and fix it. (Added a ZAP comment.)

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/Constant.java
+++ b/zap/src/main/java/org/parosproxy/paros/Constant.java
@@ -102,6 +102,8 @@
 // ZAP: 2019/11/07 Removed constants related to accepting the license.
 // ZAP: 2020/01/02 Updated config version and default user agent
 // ZAP: 2020/01/06 Set latest version to default config.
+// ZAP: 2020/01/10 Correct the MailTo autoTagScanner regex pattern when upgrading from 2.8 or
+// earlier.
 package org.parosproxy.paros;
 
 import java.io.File;
@@ -134,6 +136,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.configuration.ConfigurationException;
 import org.apache.commons.configuration.ConversionException;
+import org.apache.commons.configuration.HierarchicalConfiguration;
 import org.apache.commons.configuration.XMLConfiguration;
 import org.apache.log4j.Logger;
 import org.apache.log4j.PropertyConfigurator;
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/Constant.java
```diff
@@ -1092,6 +1095,29 @@ private static void upgradeFrom2_8_0(XMLConfiguration config) {
...
+            String elementBaseKey = autoTagScannersKey + "(" + i + ").";
+            String currentKeyResBodyRegex = elementBaseKey + "resBodyRegex";
+            if (config.getProperty(currentKeyResBodyRegex).equals(badPattern)) {
+                config.setProperty(currentKeyResBodyRegex, goodPattern);
+                break;
+            }
+        }
+
+        try {
+            config.save();
```
kingthorin: I've probably overlooked something, but this seemed necessary to make it stick.

thc202: It shouldn't, the config is already being saved after calling all `upgradeFrom_` methods (when setting the latest version). I also tested without this statement and the tag was updated in the file.

zap/src/main/java/org/parosproxy/paros/Constant.java
```diff
@@ -1092,6 +1095,29 @@ private static void upgradeFrom2_8_0(XMLConfiguration config) {
...
         config.setProperty(
                 ConnectionParam.DEFAULT_USER_AGENT, ConnectionParam.DEFAULT_DEFAULT_USER_AGENT);
+        updatePscanTagMailtoPattern(config);
+    }
+
+    private static void updatePscanTagMailtoPattern(XMLConfiguration config) {
+        String autoTagScannersKey = "pscans.autoTagScanners.scanner";
+        List<HierarchicalConfiguration> tagScanners = config.configurationsAt(autoTagScannersKey);
+        String badPattern = "<.*href\\s*['\"]?mailto:";
+        String goodPattern = "<.*href\\s*=['\"]?mailto:";
```
thc202: Worth adding `\\s*` after the equals.

thc202:
Isn't this for 2.9.0? If so, this should be rebased on `release-2.9.0` (and the PR base changed accordingly).

zap/src/main/java/org/parosproxy/paros/Constant.java
```diff
@@ -1092,6 +1095,29 @@ private static void upgradeFrom2_8_0(XMLConfiguration config) {
...
+
+        for (int i = 0, size = tagScanners.size(); i < size; ++i) {
+            String elementBaseKey = autoTagScannersKey + "(" + i + ").";
+            String currentKeyResBodyRegex = elementBaseKey + "resBodyRegex";
```
thc202: `String currentKeyResBodyRegex = autoTagScannersKey + "(" + i + ").resBodyRegex";`

zap/src/main/java/org/parosproxy/paros/Constant.java
kingthorin: But did you see it in the UI as updated?

zap/src/main/java/org/parosproxy/paros/Constant.java
kingthorin: Okay

kingthorin:
Sure I'll rebase it.

zap/src/main/java/org/parosproxy/paros/Constant.java
thc202: Yes, it was the new value.

kingthorin: Probably because of how I was testing? Inserting the method call at 610'ish so it'd be used in devmode... I'll remove this try block.

thc202: Right, in that case it would not be saved.

kingthorin:
Should be all fixed now.

thc202:
Thank you!

PR #5828, opened 2020-01-25
Add a button on the resend screen to inject a CSRF token

### LINKED ISSUE #111: Enhancement : Provide a button on the resend screen to inject a CSRF token
```
It would be useful to have a button/check box that injects the CSRF token when re-sending
requests.

1) A button or checkbox on the resend screen which would switch on injection of a token.
2) CSRF Token inserted into the request.

Zap version 1.3, Linux Fedora 14.



```

Original issue reported on code.google.com by `darrengarsden` on 2011-06-13 12:04:38

madanalogy:
Resolves #111


zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -274,6 +278,22 @@ private JToggleButton getButtonUseCookies() {
         return useCookies;
     }
 
+    private JToggleButton getButtonUseCsrf() {
+        // TODO: Add resources for icon and tooltip
```
kingthorin: Thus seems to be done

kingthorin:
To address the build failure you can run  the cradle verification spotlessApply task, commit and push the changes (that'll apply the necessary formatting).

To address the DOC requirement you'll need to sign-off the commit(s): 

madanalogy:
> To address the build failure you can run the gradle verification spotlessApply task, commit and push the changes (that'll apply the necessary formatting).
> 
> To address the DCO requirement you'll need to sign-off the commit(s):
> 
> * https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#developer-certificate-of-origin
> * https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

Understood, I will make the necessary changes upon my next push to this PR. Thank you for the help!

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -3,7 +3,7 @@
  *
  * ZAP is an HTTP/HTTPS proxy for assessing web application security.
  *
- * Copyright 2012 The ZAP Development Team
+ * Copyright 2020 The ZAP Development Team
```
thc202: It's not necessary to change it.

kingthorin: spotlessApply does that sometimes, I haven’t figured out the trigger...

thc202: Ah, it was because of the "ZAP" comment (not needed in this class).

madanalogy: Yea it was done by spotlessApply, should I change it back then?

madanalogy:
Implementation complete (I think). As far as possible I tried my best not to touch code already written. I'm not very sure how to go about testing that it works or writing tests for this feature so if there's any advice on that front I would appreciate it very much! :)

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -54,6 +54,7 @@
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2019/12/09 Address deprecation of getHeaders(String) Vector method.
+// ZAP: 2020/01/27 Added injectCsrfToken method
```
kingthorin: You ended up calling it replace*

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1090,4 +1094,27 @@ public void setResponseFromTargetHost(final boolean responseFromTargetHost) {
...
+
+    public static void replaceCsrfToken(
+            HttpMessage msg,
+            AntiCsrfToken antiCsrfToken,
+            String tokenValue,
+            Logger log,
+            Encoder encoder,
+            ExtensionAntiCSRF extAntiCSRF) {
+        log.debug(
+                "regenerateAntiCsrfToken replacing "
```
kingthorin: replace*

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
madanalogy: Thanks for pointing that out

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
kingthorin: Yes please

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
kingthorin: This replace method may only currently be called from regenerate but there’s no way we can be sure that’ll always be the case. 
Also I think it’ll be more confusing than helpful for future debugging/ maintenance of the message isn’t aligned to the local most function name.

madanalogy: So I should move the logging out of the method and have it instead being done right before the method invocation?

kingthorin: Sure it could be moved into the regenerate method ahead of the call

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -345,22 +346,11 @@ private void regenerateAntiCsrfToken(HttpMessage msg, AntiCsrfToken antiCsrfToke
         if (tokenValue != null) {
             // Replace token value - only supported in the body right now
             log.debug(
-                    "regenerateAntiCsrfToken replacing "
+                    "regenerateAntiCsrfToken in AbstractPlugin replacing "
```
thc202: It's not necessary to include the class, the logger is already adding it.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -146,6 +174,33 @@ public void run() {
...
 
+    private void regenerateAntiCsrfToken(HttpMessage msg, AntiCsrfToken antiCsrfToken) {
+        if (antiCsrfToken == null) {
+            return;
+        }
+
+        String tokenValue = null;
+        try {
+            HttpMessage tokenMsg = antiCsrfToken.getMsg().cloneAll();
+            tokenValue = extAntiCSRF.getTokenValue(tokenMsg, antiCsrfToken.getName());
```
thc202: Needs to resend the message before attempting to get the new token value.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -274,6 +329,14 @@ private JToggleButton getButtonUseCookies() {
         return useCookies;
     }
 
+    private JCheckBox getButtonUseCsrf() {
+        if (useCsrf == null) {
+            useCsrf = new JCheckBox();
```
thc202: I'd suggest using a toggle button with an icon, don't know which one though :)

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1090,4 +1094,22 @@ public void setResponseFromTargetHost(final boolean responseFromTargetHost) {
     public String getType() {
         return MESSAGE_TYPE;
     }
+
+    public static void replaceCsrfToken(
```
thc202: - This would be better in `ExtensionAntiCSRF`, this class does not need to know about the anti-CSRF tokens.
 - It can be extracted the whole method (`regenerateAntiCsrfToken`) and also some of the previous logic (starting at `List<AntiCsrfToken> tokens = extAntiCSRF.getTokens(message);`) it's the same for both/all.

thc202:
> I'm not very sure how to go about testing that it works or writing tests for this feature so if there's any advice on that front I would appreciate it very much! :)

[DVWA](https://github.com/ethicalhack3r/DVWA/) uses anti-CSRF tokens in the login form, that would be a good way to test it (e.g. access/submit the login form while proxying through ZAP then manually resend the login request).

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
kingthorin: I think one of my earlier comments may have lead to this. Sorry

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
madanalogy: Would it be okay for me to reuse one of the icons already available in the resources directory? I can pick an appropriate one and put it into my next update to this PR haha

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
madanalogy: > * It can be extracted the whole method (`regenerateAntiCsrfToken`) and also some of the previous logic (starting at `List<AntiCsrfToken> tokens = extAntiCSRF.getTokens(message);`) it's the same for both/all.

Did you mean that instead of just extracting `replaceCsrfToken` I could extract the entire method `regenerateAntiCsrfToken`?

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
kingthorin: Absolutely, or something from the [Fugue collection](https://p.yusukekamiyamane.com/icons/search/fugue/) which we've used a bunch already.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
thc202: Yes.

madanalogy: Just a note, I ran into some issues doing this because `AbstractPlugin` had it's own implementation of `sendAndReceive()` so I implemented a similar logic in the new extracted method

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
madanalogy: > replace*

Sorry I think I missed this, did you mean to replace the word "replacing" with "replace"?

kingthorin: Probably doesn't apply anymore, this was when you'd implemented a replace method not a regenerate......

madanalogy: Gotcha haha, thanks for the clarification :)

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -78,6 +81,7 @@ public HttpPanelSender(HttpPanelRequest requestPanel, HttpPanelResponse response
         requestPanel.addOptions(getButtonUseCookies(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
         requestPanel.addOptions(
                 getButtonFollowRedirects(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
+        requestPanel.addOptions(getButtonUseCsrf(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
```
thc202: Could add only when the extension is enabled.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -274,6 +296,19 @@ private JToggleButton getButtonUseCookies() {
...
 
+    private JToggleButton getButtonUseCsrf() {
+        if (useCsrf == null) {
+            useCsrf =
+                    new JToggleButton(
+                            new ImageIcon(
+                                    HttpPanelSender.class.getResource(
+                                            "/resource/icon/fugue/document-target.png")));
+            useCsrf.setToolTipText(Constant.messages.getString("manReq.checkBox.useCSRF"));
+            useCsrf.setSelected(true);
```
thc202: Should be disabled by default (other options are).

zap/src/main/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRF.java
```diff
@@ -524,4 +528,67 @@ public String generateForm(HttpMessage msg) throws UnsupportedEncodingException
         HistoryReference createHistoryReference(int id)
                 throws DatabaseException, HttpMalformedHeaderException;
     }
+
+    public void injectCsrfToken(HttpMessage message, HttpRequestConfig config) {
```
thc202: - I'd suggest using the "old" name`regenerateAntiCsrfToken`, it's not injecting (just replacing and clear with "anti").
 - The two methods can be merged (and feel free to clean up).

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
thc202: It's better to leave the actual send to the caller, the `HttpSender` might have state (e.g. user, session) which would be lost if the message is sent by the extension. The method could accept a function which sends the `HttpMessage`, e.g.:
```Java
public interface HttpMessageSender {
    void sendMessage(HttpMessage message) throws IOException;
}
```

thc202:
The `encoder` in `AbstractPlugin` can be removed, no longer used.

zap/src/main/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRF.java
thc202: Also worth adding some JavaDoc (and since tag).

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
madanalogy: I did not know that was possible haha, thanks again for pointing me in the right direction :)

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
madanalogy: I was trying to follow the implementation of [`getButtonFollowRedirects()`](https://github.com/zaproxy/zaproxy/pull/5828/files/b7e651c9fab6a40e496ef83f6ca6d997d5ded0dd#diff-1a60d73067c4bf50dcb04b02bd283bddR264) which set the default as enabled. But disabled by default makes more sense especially since the button shouldn't even be there unless the `ExtensionAntiCSRF` is enabled.

madanalogy:
> [DVWA](https://github.com/ethicalhack3r/DVWA/) uses anti-CSRF tokens in the login form, that would be a good way to test it (e.g. access/submit the login form while proxying through ZAP then manually resend the login request).

I tried doing this but it didn't work with manual resend. I set up some debug breakpoints and found that the `getTokens(message)` method in `ExtensionAntiCSRF` was returning an empty list. The message at that point however did contain the token, so I'm wondering if there might be an issue with the method or if I didn't perform a manual resend correctly.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -277,15 +277,11 @@ protected void sendAndReceive(
...
             if (extAntiCSRF != null) {
-                List<AntiCsrfToken> tokens = extAntiCSRF.getTokens(message);
-                AntiCsrfToken antiCsrfToken = null;
-                if (tokens.size() > 0) {
-                    antiCsrfToken = tokens.get(0);
-                }
-
-                if (antiCsrfToken != null) {
-                    regenerateAntiCsrfToken(message, antiCsrfToken);
+                if (!isRegenCsrfCalled) {
```
thc202: It can be sent with the correct parameters instead, e.g.:
```Java
extAntiCSRF.regenerateAntiCsrfToken(message, tokenMsg -> sendAndReceive(tokenMsg, true, false));
```

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -61,23 +62,31 @@
...
 
+        extAntiCSRF =
+                Control.getSingleton().getExtensionLoader().getExtension(ExtensionAntiCSRF.class);
+
         requestPanel.addOptions(
                 getButtonUseTrackingSessionState(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
         requestPanel.addOptions(getButtonUseCookies(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
         requestPanel.addOptions(
                 getButtonFollowRedirects(), HttpPanel.OptionsLocation.AFTER_COMPONENTS);
+        if (extAntiCSRF.isEnabled()) {
```
thc202: `extAntiCSRF != null`, it's null when not enabled.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -91,7 +100,13 @@ public void handleSendMessage(Message aMessage) throws IllegalArgumentException,
         httpMessage.setRequestingUser(null);
         try {
             final ModeRedirectionValidator redirectionValidator = new ModeRedirectionValidator();
+            boolean useAntiCSRF = getButtonUseCsrf().isSelected();
             boolean followRedirects = getButtonFollowRedirects().isSelected();
+
+            if (useAntiCSRF && extAntiCSRF != null) {
```
thc202: Worth swapping the operands and inline the `isSelected` check to avoid creating the button when the extension is disabled.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
```diff
@@ -274,6 +289,19 @@ private JToggleButton getButtonUseCookies() {
...
 
+    private JToggleButton getButtonUseCsrf() {
+        if (useCsrf == null) {
+            useCsrf =
+                    new JToggleButton(
+                            new ImageIcon(
+                                    HttpPanelSender.class.getResource(
+                                            "/resource/icon/fugue/document-target.png")));
+            useCsrf.setToolTipText(Constant.messages.getString("manReq.checkBox.useCSRF"));
+            useCsrf.setSelected(false);
```
thc202: Can be removed, it's not selected by default.

zap/src/main/java/org/parosproxy/paros/network/HttpMessageSender.java
```diff
@@ -0,0 +1,7 @@
+package org.parosproxy.paros.network;
+
+import java.io.IOException;
+
+public interface HttpMessageSender {
```
thc202: I'd keep this in `ExtensionAntiCSRF` for now.

zap/src/main/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRF.java
```diff
@@ -524,4 +525,55 @@ public String generateForm(HttpMessage msg) throws UnsupportedEncodingException
         HistoryReference createHistoryReference(int id)
                 throws DatabaseException, HttpMalformedHeaderException;
     }
+
+    /**
+     * Regenerates the Anti-CSRF Token of a HttpMessage if one exists.
+     *
+     * @param message The {@link HttpMessage} to be checked.
+     * @param httpSender The {@code sendAndReceive} implementation of the caller.
```
thc202: `... to obtain the new anti-csrf token.`?

zap/src/main/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRF.java
```diff
@@ -524,4 +525,55 @@ public String generateForm(HttpMessage msg) throws UnsupportedEncodingException
...
+     * @since 2.9.0
```
thc202: `2.10.0` or `TODO add version`.

zap/src/main/java/org/parosproxy/paros/extension/manualrequest/http/impl/HttpPanelSender.java
thc202: Right, was thinking on the session options.

thc202:
> I tried doing this but it didn't work with manual resend.

It works if you resend the (last) message with the registered token (that's an effect on how the tokens are tracked, this should be changed but out of scope of this PR IMO).

psiinon:
How about this for the button image?
crsf-button

madanalogy:
> How about this for the button image?
> crsf-button

Is this already in the fugue collection or do I need to add it to the resources?

thc202:
You would have to add it.

thc202:
Thanks!

thc202:
Could you update the [help](https://github.com/zaproxy/zap-core-help/blob/0103ce76a6feb3cc502cc7e9b84cd893269c499b/addOns/help/src/main/javahelp/contents/ui/dialogs/man_req.html) to mention the new button?

thc202:
@madanalogy how would you like to be credited (e.g. name, handle)?

madanalogy:
> @madanalogy how would you like to be credited (e.g. name, handle)?
> https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team

Ahmed Bahajjaj (@madanalogy)

Thank you! :)

Issue #5843, opened 2020-02-03
Directly report false positives from inside ZAP 


globeone:
### Issue
At the moment reporting false positives requires numerous steps. New users of ZAP may not know where to report false positives leading to the community missing potential bugs that can be improved and make ZAP better.
Also, the only option is 'delete' of an alert. Meaning that it doesn't show on the report and it's also impossible to create a report on the False Positives found.

### Suggestion for improvement 
Add two context menu items when right clicking on an alert item.

- Mark as False Positive
- Report False Positive to the ZAP project

The Mark as False Positive option could just Flag the option as False Positive and put it at the end of the report as a list of False Postives.

The Report False Positive to the ZAP Project item could include the data needed to evaluate the False Positive and automatically start a GitHub ticket, open a browser and fill the ticket with the useful information needed for developers to fix the issue.

Then end user can then fill the ticket with extra information to further clarify the bug. Or remove any identifying information. 

### Other open source project with similar functionality
 Similar to how "ubuntu-bug" also known as "apport" works 
https://wiki.ubuntu.com/Apport

kingthorin:
>  Also, the only option is 'delete' of an alert. Meaning that it doesn't show on the report and it's also impossible to create a report on the False Positives found.

No you can choose to set any alert as confidence == False Positive, then it's still there and not reported on.

kingthorin:
Context menu for marking alerts False Positive will be in the next weekly release.

globeone:
That's awesome that it got fixed so quickly. We've asked for similar requests from our commercial provider, and they take years to even acknowledge that an issue exists and then come with a half-baked solution that doesn't even work. Kudos to the OWASP ZAP Team!

Issue #1603, opened 2015-06-04
Bulk change alert confidence


zapbot:
```
Provide a way to bulk change alert confidence, such as marking 200 duplicates of an
alert confirmed or false positive.

https://groups.google.com/forum/#!topic/zaproxy-users/gE7htFW6nbQ

```

Original issue reported on code.google.com by `cashwilliams` on 2015-04-16 14:36:30

twright-0x1:
This capability would be _very_ appreciated!

johnccr:
Yes, please add this functionality

gvolluz:
I concur, this would be much appreciated!

PR #5844, opened 2020-02-03
Add context menu item to mark false positives

kingthorin:
Allow multi-select.

See zaproxy/zaproxy#1603, also see zaproxy/zaproxy#5843.

Alert class fix missed fields in `newInstance()` method. (Thanks @thc202)


kingthorin:
JavaDoc fixed.

PR #5848, opened 2020-02-06
Update Python docker scripts to be Python 3 compatible

camswords:
The `/zap/zap-*.py` Python scripts in the Docker image are not Python3 compatible. This is because they still use `dict.has_key(x)` instead of `x in dict`, and `dict.iteritems()` instead of `dict.items()`. Moving to Python3 is important because Python2 is due to EOL in [April 2020](https://www.python.org/psf/press-release/pr20191220/).

I've done my best to test this change locally. I didn't test the API script changes, as I don't have an easy example locally to test, but they are essentially the same as the full script scan changes.

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-02-07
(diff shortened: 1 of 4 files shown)


thc202:
Thanks, could you update the changelog?

I thought these were already compatible but seems we just changed the baseline.

thc202:
Thank you!

Issue #1276, opened 2015-06-04
Break points : allow these to specify request and/or response


zapbot:
```
Right now custom break points apply to _both_ requests and responses.
It would be much better if the user could choose if one applies to a request, a response
or both.
Proposed on https://groups.google.com/d/msg/zaproxy-develop/wR90nefqdG4/XoRDvqMWuHUJ

This should be a relatively straight forward change to make - the relevant code should
all(?) be in: https://code.google.com/p/zaproxy/source/browse/#svn%2Ftrunk%2Fsrc%2Forg%2Fzaproxy%2Fzap%2Fextension%2Fbrk
```

Original issue reported on code.google.com by `psiinon` on 2014-07-23 09:34:47

DiogoMRSilva:
Hi, can you check if it this you want? https://github.com/DiogoMRSilva/zaproxy/tree/breakFileTypeSelector
I Made it some time ago and I believe it may be useful for other people as well. It allows the user to decide if he wants to break on js, css, images etc and change it anytime. If you want I can make a pull request.
![zapBreakSelectorFeaturePrint](https://user-images.githubusercontent.com/6045401/73803035-740fc800-47b7-11ea-82ef-71fbaabc78e1.png)
 @kingthorin

kingthorin:
Seems handy, might need to be tied to some sort of option like "Show Advanced Break Buttons" or something like that so that the toolbar doesn't get too cluttered/overwhelming for newer users.... @psiinon @thc202 what do you think? I'd like to see a PR.

psiinon:
I worry about us overloading the toolsbar(s) so wouldnt want this as standard, but would be fine with it as an option.

DiogoMRSilva:
I only have the break buttons in the break panel so it is not a problem to me. But I agree that it makes the toolbar more cluttered/overwhelming. I will add the option to the options menu and set the default to hidden.

In my opinion the content of that toolbar should be on a "dashboard" tab instead. And the options menu should also be a tab. But this is a discussion for later :)

DiogoMRSilva:
Should I add it to Display or Breakpoints?

psiinon:
I'd say the Break tab as its more specific

DiogoMRSilva:
There is no Break tab in the options only a "Breakpoints". Should I create a new one?

thc202:
[...]

PR #5864, opened 2020-02-14
Add buttons to select the file types ignored on Break

DiogoMRSilva:
Adds buttons to allow the user to select the file types he wants to ignore.
Initial implementation for: https://github.com/zaproxy/zaproxy/issues/1276


zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointManagementInterface.java
```diff
@@ -30,6 +30,12 @@
 
     boolean isBreakAll();
 
+    boolean isBreakOnJavaScript();
```
thc202: These are no longer needed?

DiogoMRSilva: They aren't needed anymore. Should I remover them? What about The API ?

DiogoMRSilva:
@kingthorin  I didn't know webp. I can add it.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointManagementInterface.java
thc202: You mean the ZAP API? I'd expect the respective implementations to provide endpoints to add/remove the ignored rules.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -521,11 +542,19 @@ public void setButtonMode(int mode) {
         private final JToggleButton requestButton;
         private final JToggleButton responseButton;
         private final JToggleButton allButton;
+        private final JToggleButton brkOnJsButton;
```
thc202, line 545: We should be consistent with naming, either `Js` or `JavaScript` (former?).

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -521,11 +542,19 @@ public void setButtonMode(int mode) {
...
+        private final JToggleButton brkOnCSSAndFontsButton;
```
thc202, line 546: `Css`

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -541,6 +570,10 @@ public BreakButtonsUI(String name, BreakPanelToolbarFactory breakToolbarFactory)
             toolBar.add(breakToolbarFactory.getBtnContinue());
             toolBar.add(breakToolbarFactory.getBtnDrop());
             toolBar.add(breakToolbarFactory.getBtnBreakPoint());
+            toolBar.add(brkOnJsButton);
```
thc202, line 573: Might worth adding a separator before this one and before the scope button, as they have opposite behaviour.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -56,18 +62,34 @@
...
     private boolean step = false;
     private boolean stepping = false;
     private boolean drop = false;
     private boolean isBreakRequest = false;
     private boolean isBreakResponse = false;
     private boolean isBreakAll = false;
+    private boolean isBreakOnJavaScript = true;
+    private boolean isBreakOnCSSAndFonts = true;
+    private boolean isBreakOnMultimedia = true;
+    private boolean isOnlyBreakOnScope = false;
```
thc202, line 80: This should be replaced with the use of the existing option (`BreakpointsParam#isInScopeOnly()`), otherwise we might end up with conflicting/misleading state, one set to break and the other unset.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -56,18 +62,34 @@
...
     private BreakPanel breakPanel = null;
 
     private BreakpointsParam breakpointsParams;
     private int mode = 0;
+    private boolean showButtonsState = false;
+
+    private List<BreakpointMessageInterface> ignoreRulesEnable;
+    HttpBreakpointMessage ignoreJavascriptBreakpointMessage;
+    HttpBreakpointMessage ignoreCSSAndFontsBreakpointMessage;
+    HttpBreakpointMessage ignoreMultimediaBreakpointMessage;
```
thc202, line 91: `private`

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -56,18 +62,34 @@
...
+    HttpBreakpointOutOfScope ignoreMessageNotInScope;
```
thc202, line 92: This should not be needed, per previous comment about the existing option.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -446,6 +633,24 @@ public void setButtonMode(int mode) {
         this.mode = mode;
     }
 
+    public void setShowIgnoreFilesButtons(boolean showButtons) {
+        this.showButtonsState = showButtons;
+
+        if (showButtons == false) {
```
thc202, line 639: `!showButtons`

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -469,6 +674,10 @@ public void actionPerformed(ActionEvent e) {
             setBreakAll(false);
             setBreakRequest(false);
             setBreakResponse(false);
+            setBreakOnJavaScript(isBreakOnJavaScript);
+            setBreakOnCSSAndFonts(isBreakOnCSSAndFonts);
+            setBreakOnMultimedia(isBreakOnMultimedia);
+            setOnlyBreakOnScope(isOnlyBreakOnScope);
```
thc202, line 680: These should not be needed.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointMessageHandler2.java
```diff
@@ -32,10 +32,14 @@
 
     protected static final Object SEMAPHORE = new Object();
 
+    private static final String HTTP_TYPE = "HTTP";
```
thc202, line 35: Revert, not used/needed.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
```diff
@@ -70,12 +75,42 @@ public BreakpointsOptionsPanel() {
         panel.add(
                 getCheckBoxAlwaysOnTop(),
                 LayoutHelper.getGBC(0, 2, 2, 1.0, new Insets(2, 2, 2, 2)));
+        panel.add(
+                getCheckBoxShowIgnoreFilesButtons(),
+                LayoutHelper.getGBC(0, 3, 2, 1.0, new Insets(2, 2, 2, 2)));
```
thc202, line 80: Extract a variable for the row.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
```diff
@@ -50,7 +51,11 @@
     private JCheckBox checkBoxConfirmDropMessage = null;
     private JCheckBox checkBoxAlwaysOnTop = null;
     private JCheckBox checkBoxInScopeOnly = null;
+    private JCheckBox checkBoxShowIgnoreFilesButtons = null;
     private JComboBox<String> buttonMode = null;
+    private ZapTextField javascriptUrlRegexField = null;
+    private ZapTextField cssAndFontsUrlRegexField = null;
+    private ZapTextField multimediaUrlRegexField = null;
```
thc202, line 58: Add before `buttonMode`?

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
```diff
@@ -131,7 +197,11 @@ public void initParam(Object obj) {
         // Note param.alwaysOnTop will be null if the user hasn't specified a preference yet
         getCheckBoxAlwaysOnTop().setSelected(param.getAlwaysOnTop() != Boolean.FALSE);
         getCheckBoxInScopeOnly().setSelected(param.isInScopeOnly());
+        getCheckBoxShowIgnoreFilesButtons().setSelected(param.getIsToShowIgnoreFilesButtons());
         getButtonMode().setSelectedIndex(param.getButtonMode() - 1);
+        getJavascriptUrlRegexField().setText(param.getJavascriptUrlRegex());
+        getCssAndFontsUrlRegexField().setText(param.getCssAndFontsUrlRegex());
+        getMultimediaUrlRegexField().setText(param.getMultimediaUrlRegex());
```
thc202, line 204: Should call `discardAllEdits()` after setting each value.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsParam.java
```diff
@@ -44,13 +44,27 @@
     private static final String PARAM_UI_BUTTON_MODE = PARAM_BASE_KEY + ".buttonMode";
     private static final String PARAM_BRK_ALWAYS_ON_TOP = PARAM_BASE_KEY + ".alwaysOnTop";
     private static final String PARAM_BRK_IN_SCOPE_ONLY = PARAM_BASE_KEY + ".inScopeOnly";
+    private static final String SHOW_IGNORE_FILES_BUTTONS =
+            PARAM_BASE_KEY + ".ShowIgnoreFilesButtons";
+    private static final String JAVASCRIPT_URL_REGEX = PARAM_BASE_KEY + ".jsUrlRegex";
+    private static final String CSS_AND_FONTS_URL_REGEX = PARAM_BASE_KEY + ".cssAndFontsUrlRegex";
+    private static final String MULTIMIDIA_URL_REGEX = PARAM_BASE_KEY + ".multimediaUrlRegex";
```
thc202, line 51: Typo in multimedia (same in the default constant).

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsParam.java
```diff
@@ -44,13 +44,27 @@
     private static final String PARAM_UI_BUTTON_MODE = PARAM_BASE_KEY + ".buttonMode";
     private static final String PARAM_BRK_ALWAYS_ON_TOP = PARAM_BASE_KEY + ".alwaysOnTop";
     private static final String PARAM_BRK_IN_SCOPE_ONLY = PARAM_BASE_KEY + ".inScopeOnly";
+    private static final String SHOW_IGNORE_FILES_BUTTONS =
+            PARAM_BASE_KEY + ".ShowIgnoreFilesButtons";
```
thc202, line 48: - Start with lower case.
- Technically this is no longer just about ignore files, scope button is included.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsParam.java
```diff
@@ -125,4 +143,40 @@ public void setInScopeOnly(boolean inScopeOnly) {
         this.inScopeOnly = inScopeOnly;
         getConfig().setProperty(PARAM_BRK_IN_SCOPE_ONLY, inScopeOnly);
     }
+
+    public boolean getIsToShowIgnoreFilesButtons() {
```
thc202, line 147: `isShow...`

zap/src/main/java/org/zaproxy/zap/extension/brk/impl/http/HttpBreakpointManagementDaemonImpl.java
```diff
@@ -29,6 +29,9 @@
 
     private boolean breakRequest;
     private boolean breakResponse;
+    private boolean breakOnJavaScript;
+    private boolean breakOnCSSAndFonts;
+    private boolean breakOnMultimedia;
```
thc202, line 34: These should be removed.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -887,6 +888,10 @@ brk.optionspanel.option.breakmode.dual.label	= Separate Request and Response but
 brk.optionspanel.option.breakmode.simple.label	= Single combined button
 brk.optionspanel.option.confirmDropMessage.label = Confirm drop trapped message
 brk.optionspanel.option.inScopeOnly.label = Only break on messages in scope
+brk.optionspanel.option.ShowBreakFilteringButtons.label = Show buttons to filter the type of requests that ZAP should break on
```
thc202, line 891: - `showBreak`
 - `HTTP requests`

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -887,6 +888,10 @@ brk.optionspanel.option.breakmode.dual.label	= Separate Request and Response but
...
+brk.optionspanel.option.javascriptUrlRegex.label = Javascript url regex:
```
thc202, line 892: - `JavaScript`
 - `URL`

zap/src/main/resources/resource/icon/breakTypes/sources.txt
```diff
@@ -0,0 +1,5 @@
+<div>Icons made by <a href="https://www.flaticon.com/authors/smashicons" title="Smashicons">Smashicons</a> from <a href="https://www.flaticon.com/"                 title="Flaticon">www.flaticon.com</a> is licensed by <a href="http://creativecommons.org/licenses/by/3.0/"                 title="Creative Commons BY 3.0" target="_blank">CC 3.0 BY</a></div>
```
thc202, line 1: - This should be moved to `zap/src/main/dist/license/`.
 - Better have this as plain text (per file extension).
 - To be easier/clear it should be enumerated which icons/files apply to which licenses (and link to the icon set). Was any icon changed?

zap/src/main/resources/resource/icon/breakTypes/sources.txt
```diff
@@ -0,0 +1,5 @@
...
+
+<a target="_blank" href="https://icons8.com/icons/set/javascript">JavaScript icon</a> icon by <a target="_blank" href="https://icons8.com">Icons8</a>
```
thc202, line 3: The terms also require linking from the about dialogue:
https://icons8.com/license
https://icons8.crisp.help/en/article/where-do-i-set-the-link-irwkfh/
`Desktop software — please put the link in the About dialog`

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -56,18 +62,34 @@
...
+    private boolean isBreakOnJavaScript = true;
+    private boolean isBreakOnCSSAndFonts = true;
+    private boolean isBreakOnMultimedia = true;
+    private boolean isOnlyBreakOnScope = false;
 
     private BreakPanel breakPanel = null;
 
     private BreakpointsParam breakpointsParams;
     private int mode = 0;
+    private boolean showButtonsState = false;
```
thc202, line 86: Should be removed, not actually used.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -403,6 +582,14 @@ public void reset() {
...
         }
 
+        setBreakOnJavaScript(isBreakOnJavaScript);
+
+        setBreakOnCSSAndFonts(isBreakOnCSSAndFonts);
+
+        setBreakOnMultimedia(isBreakOnMultimedia);
+
+        setOnlyBreakOnScope(isOnlyBreakOnScope);
+
```
thc202, line 592: Not needed, the state of these buttons does not need to be reset.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -151,6 +214,22 @@ public boolean isBreakAll() {
...
+        return isBreakOnCSSAndFonts;
+    }
+
+    public boolean isBreakOnMultimedia() {
+        return isBreakOnMultimedia;
+    }
+
+    public boolean isOnlyBreakOnScope() {
+        return isOnlyBreakOnScope;
+    }
```
thc202, line 231: Should not be needed.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
```diff
@@ -50,7 +51,11 @@
     private JCheckBox checkBoxConfirmDropMessage = null;
     private JCheckBox checkBoxAlwaysOnTop = null;
     private JCheckBox checkBoxInScopeOnly = null;
+    private JCheckBox checkBoxShowIgnoreFilesButtons = null;
```
thc202, line 54: I wonder if this one would be better in the `Display` panel, this is closely related to "Show break buttons", we can then disable this if the option is set to "Only in the toolbar".

thc202:
Regarding the PR description, this is not related to #1276 nor does it implement it (the discussion was added there but still unrelated to the actual issue, which is to add an option for a breakpoint to break just on the request or just the response).

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
DiogoMRSilva: I think that JavaScript Is better. I will rename the one with Js

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
DiogoMRSilva: Yes it makes sense

zap/src/main/resources/resource/icon/breakTypes/sources.txt
DiogoMRSilva: I may have resized them and I have created a a gray-scale version of each

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
DiogoMRSilva: I can change it. Why do you suggest that?

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
DiogoMRSilva: ok, can you explain why?

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
DiogoMRSilva: I don't agree. I think that if the user only wants the break buttons on the toolbar it doesn't imply that he doesn't want to have the filter buttons on the break tab.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -509,6 +512,24 @@ public void setButtonMode(int mode) {
         responseBreakButtons.setButtonMode(mode);
     }
 
+    public void setShowIgnoreFilesButtons(boolean showButtons) {
```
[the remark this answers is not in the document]
DiogoMRSilva: It needs to be at least protected because it is used in ExtensionBreak

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
DiogoMRSilva: ok. Should I also add isBreakOnJavaScript, isBreakOnCSSAndFonts and isBreakOnMultimedia to the  breakpointsParams, or I should only remove the isOnlyBreakOnScope?

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
thc202: Keep them in order they are shown/used.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
thc202: Memory (discard edits no longer needed) and usability (restore to clean state).

zap/src/main/resources/resource/icon/breakTypes/sources.txt
thc202: OK, we should note that then.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
thc202: When "Only in the toolbar" is selected no button is shown in the Break tab.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
thc202: I don't think they need to be persisted, so fine leaving them here.

> maybe the inScopeOnly param should be removed from the BreakpointsParam instead.

There are other scope related options (e.g. passive scan just in scope) that are persisted, it would be better to preserve the current behaviour.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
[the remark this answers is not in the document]
thc202: Better with package though.

zap/src/main/resources/resource/icon/breakTypes/sources.txt
DiogoMRSilva: I don't remember which are from where but I will get all again from flaticon.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsOptionsPanel.java
DiogoMRSilva: Now when the  "Only in the toolbar" mode is on I deselect the option in the breakpoint options params.

DiogoMRSilva:
Some of the images URLs don't have any file extension. Since I am searching for file extensions to predict the resource type it will fail in those cases. One possible improvement would be to look to the request  header `Accept`. For instance if it has something like ```Accept: image/webp,*/*``` we consider that it is an image that is being requested. What do you think? Should this be a future improvement or something to do now?

kingthorin:
There is a response header method isImage() that checks content type, I don't think it currently includes webp or svg. It could be used and improved, I'm not necessarily sure it needs to be tackled for this PR. As long as the help and info is clear on how i functions I don't think users can complain 😀

DiogoMRSilva:
Looking at the answer to see the content type would be the more reliable way to detect the file type, but I want the user to be able to ignore breaks also in the requests. So I think I cant use that method.

psiinon (changes requested):
This doesnt work if break buttons are just in the main toolbar. Maybe disable in this case and explain in a tooltip?

DiogoMRSilva:
> This doesnt work if break buttons are just in the main toolbar. Maybe disable in this case and explain in a tooltip?

Yes thats more friendly to the users :). Done

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
```diff
@@ -94,6 +95,10 @@
     // ISO Standards compliant format
     private static final String TIME_STAMP_FORMAT_ISO8601 = "yyyy-MM-dd'T'HH:mm:ssZ";
 
+    public static final int TOOL_BAR_ONLY_INDEX = 0;
```
thc202, line 98: Better with an enum.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -95,8 +111,46 @@ public BreakPanelToolbarFactory(BreakpointsParam breakpointsParams, BreakPanel b
...
 
+        setBreakOnJavaScriptAction = new SetBreakOnJavaScriptAction();
+        setBreakOnCssAndFontsAction = new SetBreakOnCssAndFontsAction();
+        setBreakOnMultimediaAction = new SetBreakOnMultimediaAction();
+        setOnlyBreakOnScopeAction = new SetBreakOnlyOnScopeAction();
+
         this.breakpointsParams = breakpointsParams;
         this.breakPanel = breakPanel;
+
+        this.ignoreRulesEnable = new ArrayList<>(0);
```
thc202, line 122: 0 → 3

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakpointsParam.java
```diff
@@ -125,4 +143,40 @@ public void setInScopeOnly(boolean inScopeOnly) {
         this.inScopeOnly = inScopeOnly;
         getConfig().setProperty(PARAM_BRK_IN_SCOPE_ONLY, inScopeOnly);
     }
+
+    public boolean getIsShowIgnoreFilesButtons() {
```
thc202, line 147: `isShow...`

zap/src/test/java/org/zaproxy/zap/extension/brk/BreakpointMessageHandler2UnitTest.java
```diff
@@ -63,6 +66,69 @@ void shouldBeBreakpointIfMessageForceIntercept(boolean request, boolean onlyIfIn
...
+        Message message = mock(Message.class);
+        given(message.isInScope()).willReturn(true);
+        given(breakpointManagementInterface.isBreakRequest()).willReturn(true);
+        given(breakpointManagementInterface.isBreakResponse()).willReturn(true);
+
+        BreakpointMessageInterface skipBreakpoint = mock(BreakpointMessageInterface.class);
+        given(skipBreakpoint.isEnabled()).willReturn(false);
+        given(skipBreakpoint.match(message, request, false)).willReturn(true);
+        List<BreakpointMessageInterface> x = new ArrayList<>();
+        x.add(skipBreakpoint);
```
thc202, line 124: `List<BreakpointMessageInterface> ignoreRules = Arrays.asList(skipBreakpoint);`

thc202:
Could you squash the commits?

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
DiogoMRSilva: I need it to be an int to compare it with the index 
```java
if (brkPanelViewSelect.getSelectedIndex() == TOOL_BAR_ONLY_INDEX) {
     options.getParamSet(BreakpointsParam.class).setShowIgnoreFilesButtons(false);
}
```
Should I change it and add an extra dataStructure to make the translation from enum to int? I think it would make it worst.

DiogoMRSilva:
Thx for the review  :)

> Could you squash the commits?
Yes  I will do It

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
thc202: The enum values can have the corresponding index.

DiogoMRSilva: Thx, I believe that now I did what you were asking.

DiogoMRSilva:
@thc202 I tackled all the changes and squashed the commits

thc202 (approved):

thc202:
Thanks!

psiinon (approved):

PR #5870, opened 2020-02-20
Adding missing check for markdown file in baseline scan

sshniro:


thc202:
Could you apply the same fix in the other scripts and update the changelog?

sshniro:
@thc202 I have added the change to the changelog file.

thc202:
Thank you!

Issue #5542, opened 2019-08-26
Improve dark mode


epicodertime:
It would be nice for appearance for the program 

Updating to make this into a tracker:

- [x] Add FlatLaf to core
- [x] Fix splash screen (Make the tips panel slightly smaller and move it down slightly so that the bottom panel only shows 5 lines [at default font settings/lang]?) [This screenshot is shrunk to 75% for ticket inclusion]

<details>
<summary>Screenshot <sub>(Click the triangle/control to the left to expand)</sub></summary>

![image](https://user-images.githubusercontent.com/7570458/75348496-f3516280-5870-11ea-846a-5f921e0cc89d.png)

</details>

- [x] Fix Quick start panel
- [x] Fix Request / response / break panels
- [x] Fix manual request dialog
- [x] Fix WebSocket message editor
- [x] Fix script console
- [x] Fix About screen
- [x] Fix Java help pages
- [x] Fix Encode/decode/hash (black text)
- [x] Fix Options / Check For Updates (black text)
- [x] Fix Options headings:

<details>
<summary>Screenshot <sub>(Click the triangle/control to the left to expand)</sub></summary>

![image](https://user-images.githubusercontent.com/7570458/75347787-8be6e300-586f-11ea-9717-096fe4e8728b.png)

</details>

- Highlight contrast concerns:
  - [x] https://github.com/zaproxy/zaproxy/issues/5542#issuecomment-591419439 > Options Panel(s) Search Highlighting (Reference: https://github.com/zaproxy/zaproxy/pull/4141/) (Done in: https://github.com/zaproxy/zaproxy/pull/6328)
  - [x] Find Dialog Highlighting (CTRL + F) (Done in: https://github.com/zaproxy/zaproxy/pull/6328)
  - [x] https://github.com/zaproxy/zaproxy/issues/5542#issuecomment-643019236 > Highlighting in request/response panels, as a result of Alert highlighting or Search tab highlighting (it's more 'white' than the normal click & drag highlight) (Done in: https://github.com/zaproxy/zaproxy/pull/6329)
- [x] Active Scan progress dialog: https://github.com/zaproxy/zaproxy/issues/5542#issuecomment-606977117 (Done in: https://github.com/zaproxy/zaproxy/pull/6262)

Add anything else you find above...

thc202:
[...]

PR #5873, opened 2020-02-25
Add FlatLaf L&F

thc202:
Add FlatLaf dependency.
Change ExtensionUiUtils to install the L&F provided by FlatLaf.
Save the class of the look and feel to set it at start up (to not
require installing the L&Fs before start up).

Part of #5542 - Add a dark mode


psiinon:
Love this :)

PR #5876, opened 2020-02-27
Added additional parameter to support base dir and config dir to support github actions

sshniro:


psiinon:
Looks good, its just failing the DCO check as theres no "Signed-off-by:" in a commit

PR #5878, opened 2020-02-27
Set RSyntaxTextArea dark mode if dark LookAndFeel used

psiinon:
Part of https://github.com/zaproxy/zaproxy/issues/5542

Fixes the request / response / break panels and manual request dialog.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/httppanel/view/syntaxhighlight/HttpPanelSyntaxHighlightTextArea.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/httppanel/view/syntaxhighlight/HttpPanelSyntaxHighlightTextArea.java
@@ -21,6 +21,7 @@ package org.zaproxy.zap.extension.httppanel.view.syntaxhighlight;
 
 import java.awt.Color;
 import java.awt.Component;
+import java.io.IOException;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Vector;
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/zaproxy/zap/extension/httppanel/view/syntaxhighlight/HttpPanelSyntaxHighlightTextArea.java
```diff
@@ -129,6 +132,18 @@ public HttpPanelSyntaxHighlightTextArea() {
...
+        if (DisplayUtils.isDarkLookAndFeel()) {
+            try {
+                Theme theme =
+                        Theme.load(
+                                getClass()
+                                        .getResourceAsStream(
+                                                "/org/fife/ui/rsyntaxtextarea/themes/dark.xml"));
+                theme.apply(this);
+            } catch (IOException e) {
+                log.error("Failed to set RSyntaxTextArea dark theme", e);
```
kingthorin: Warn ? ZAP is still going to function fine right? (Just might not look the greatest)

Edit maybe also move the theme application outside the try block? (Or add a conditional) no point trying to apply if there was an exception reading the file...

psiinon: I'm easy on warn vs error.
But I cant see the point of trying to load the theme unless we think we need to :)

kingthorin: Agreed, so keep it inside `if (DisplayUtils.isDarkLookAndFeel())` but outside the try, so that there isn't an attempt to apply something that couldn't be read?

Issue #5899, opened 2020-03-15
Allows Regex duplication in the same Context


khalil2535:
**Describe the bug**
When i add a duplicated domain regex in the "Include in Context" menu it can be allow me to do that while it's shouldn't.

**To Reproduce**
1. Click on your Context multiple times.
2. Go to '1: Include in Context'
3. Added duplicate regex.
4. It's allowed.

**Expected behavior**
I expect to prevent me to do so with a warning or information message.

**Screenshots**
![image](https://user-images.githubusercontent.com/26520106/76695981-7d316600-668e-11ea-9b9d-d50dd38d3fbb.png)

**Software versions**
 - ZAP: all versions

**Errors from the zap.log file**
- No Need

kingthorin:
Fixed in https://github.com/zaproxy/zaproxy/pull/5901

PR #5901, opened 2020-03-17
 prevent adding empty/duplicated regex from UI

khalil2535:
* Solve #5899
* Remove duplicated messages from Messages.properties

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/model/Context.java
+++ b/zap/src/main/java/org/zaproxy/zap/model/Context.java
@@ -327,10 +327,11 @@ public class Context {
     }
 
     /**
-     * Validates that the given regular expressions are not {@code null} nor invalid.
+     * Validates that the given regular expressions are valid and they're not {@code null} nor
+     * empty.
      *
-     * @param regexs the regular expressions to be validated, must not be {@code null}
-     * @throws IllegalArgumentException if one of the regular expressions is {@code null}.
+     * @param regexs the regular expressions to be validated, must not be {@code null} nor empty.
+     * @throws IllegalArgumentException if one of the regular expressions is {@code null} or empty.
      * @throws PatternSyntaxException if one of the regular expressions is invalid.
      */
     private static void validateRegexs(List<String> regexs) {
@@ -340,10 +341,10 @@ public class Context {
     }
 
     /**
-     * Validates that the given regular expression is not {@code null} nor invalid.
+     * Validates that the given regular expression is valid and it's not {@code null} nor empty.
      *
      * @param regex the regular expression to be validated
-     * @throws IllegalArgumentException if the regular expression is {@code null}.
+     * @throws IllegalArgumentException if the regular expression is {@code null} or empty.
      * @throws PatternSyntaxException if the regular expression is invalid.
      */
     private static void validateRegex(String regex) {
@@ -351,33 +352,25 @@ public class Context {
             throw new IllegalArgumentException("The regular expression must not be null.");
         }
         String trimmedRegex = regex.trim();
-        if (!trimmedRegex.isEmpty()) {
-            Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
+        if (trimmedRegex.isEmpty()) {
+            throw new IllegalArgumentException("The regular expression must not be empty.");
         }
+        Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
     }
 
     /**
      * Sets the regular expressions used to include a URL in context.
      *
      * @param includeRegexs the regular expressions
-     * @throws IllegalArgumentException if one of the regular expressions is {@code null}.
+     * @throws IllegalArgumentException if one of the regular expressions is {@code null} or empty.
      * @throws PatternSyntaxException if one of the regular expressions is invalid.
      */
     public void setIncludeInContextRegexs(List<String> includeRegexs) {
         validateRegexs(includeRegexs);
         // Check if they've been changed
-        if (includeInRegexs.size() == includeRegexs.size()) {
-            boolean changed = false;
-            for (int i = 0; i < includeInRegexs.size(); i++) {
-                if (!includeInRegexs.get(i).equals(includeRegexs.get(i))) {
-                    changed = true;
-                    break;
-                }
-            }
-            if (!changed) {
-                // No point reapplying the same regexs
-                return;
-            }
+        if (includeInRegexs.equals(includeRegexs)) {
+            // No point reapplying the same regexs
+            return;
         }
         includeInRegexs.clear();
         includeInPatterns.clear();
@@ -419,18 +412,9 @@ public class Context {
     public void setExcludeFromContextRegexs(List<String> excludeRegexs) {
         validateRegexs(excludeRegexs);
         // Check if they've been changed
-        if (excludeFromRegexs.size() == excludeRegexs.size()) {
-            boolean changed = false;
-            for (int i = 0; i < excludeFromRegexs.size(); i++) {
-                if (!excludeFromRegexs.get(i).equals(excludeRegexs.get(i))) {
-                    changed = true;
-                    break;
-                }
-            }
-            if (!changed) {
-                // No point reapplying the same regexs
-                return;
-            }
+        if (excludeFromRegexs.equals(excludeRegexs)) {
+            // No point reapplying the same regexs
+            return;
         }
 
         excludeFromRegexs.clear();
--- a/zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
@@ -92,6 +92,8 @@ public class MultipleRegexesOptionsPanel extends AbstractMultipleOptionsBaseTabl
             addDialog = new DialogAddRegex(owner);
             addDialog.pack();
         }
+
+        addDialog.setRegexes(this.getRegexes());
         addDialog.setVisible(true);
 
         String regex = addDialog.getRegex();
@@ -106,6 +108,8 @@ public class MultipleRegexesOptionsPanel extends AbstractMultipleOptionsBaseTabl
             modifyDialog = new DialogModifyRegex(owner);
             modifyDialog.pack();
         }
+
+        modifyDialog.setRegexes(this.getRegexes());
         modifyDialog.setRegex(e);
         modifyDialog.setVisible(true);
 
@@ -161,10 +165,18 @@ public class MultipleRegexesOptionsPanel extends AbstractMultipleOptionsBaseTabl
                 Constant.messages.getString("multiple.options.regexes.dialog.regex.invalid.title");
         private static final String TEXT_INVALID_REGEX_DIALOG =
                 Constant.messages.getString("multiple.options.regexes.dialog.regex.invalid.text");
+        private static final String TEXT_ALREADY_EXIST_INVALID_REGEX_DIALOG =
+                Constant.messages.getString(
+                        "multiple.options.regexes.dialog.regex.invalid.alreadyExist.text");
+        private static final String TITLE_ALREADY_EXIST_INVALID_REGEX_DIALOG =
+                Constant.messages.getString(
+                        "multiple.options.regexes.dialog.regex.invalid.alreadyExist.title");
 
         private ZapTextField regexTextField;
 
         protected String regex;
+
+        private List<String> regexes;
 
         public DialogAddRegex(Dialog owner) {
             super(owner, DIALOG_TITLE);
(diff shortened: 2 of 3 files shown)


kingthorin:
- https://wiki.eclipse.org/EGit/User_Guide/Commit#Signing_off_commits

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -212,8 +226,13 @@ protected void init() {
         @Override
         protected boolean validateFields() {
             try {
-                Pattern.compile(getRegexTextField().getText(), Pattern.CASE_INSENSITIVE);
-            } catch (PatternSyntaxException e) {
+                String trimmedRegex = getRegexTextField().getText().trim();
+                if (trimmedRegex.trim().isEmpty()) return false; // to prevent accepting empty regex
```
kingthorin: This is syntactically correct but it would be best to use curly braces for consistency

kingthorin: Also why trim the trimmed value, is there a reason to apply it twice?

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -223,9 +223,6 @@ api.options.addr.dialog.remove.text = Are you sure you want to remove the select
 api.options.addr.dialog.remove.button.cancel = Cancel
 api.options.addr.dialog.remove.button.confirm = Remove
 api.options.addr.dialog.remove.checkbox.label = Do not show this message again
-api.options.addr.table.header.enabled	= Enabled
-api.options.addr.table.header.regex	= Regex
-api.options.addr.table.header.value	= Address
```
kingthorin: Why were these API keys/values removed?

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1106,8 +1103,8 @@ cfu.update.message.resourcesNotSavedAndActiveActions = <html>There are unsaved r
 cfu.update.message.resourcesNotSaved   = <html>The following resources have not been saved:<ul>{0}</ul>The changes might be lost if the update fails. Continue with update?</html>
 cfu.warn.addon.with.extensions.with.missing.requirements = Optional Add-ons:
 cfu.warn.addon.with.missing.requirements = Missing Requirements:
-cfu.warn.addon.with.missing.requirements.addon = A cyclic dependency was detected.
 cfu.warn.addon.with.missing.requirements.addon = Add-on "{0}"
+cfu.warn.addon.with.missing.requirements.addon.cyclic = A cyclic dependency was detected.
```
kingthorin: Why was this changed with no corresponding usage change?

khalil2535:
Done .. (with 2 force pushes 🤦‍♂️)

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -2115,6 +2112,7 @@ multiple.options.regexes.dialog.remove.regex.button.confirm = Remove
 multiple.options.regexes.dialog.remove.regex.checkbox.label = Do not show this message again
 multiple.options.regexes.dialog.regex.invalid.title = Invalid Regular Expression
 multiple.options.regexes.dialog.regex.invalid.text = The provided regular expression is not valid:\n{0}
+multiple.options.regexes.dialog.regex.invalid.alreadyExist.text = regular expression already exists
```
kingthorin: Please turn this into a proper sentence (caps and punctuation).

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
khalil2535: Those keys/values are duplicated..
[line 219](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L219) == [line 226](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L226)
[line 220](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L220) == [line 227](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L227)
[line 221](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L221) == [line 228](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L228)

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
khalil2535: The key is used just once and i see the fisrt key/value is the correct in the case.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
thc202: I'd suggest doing these changes in other PR for clarity.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
thc202: This message is still needed though, the code needs to be changed as well.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
khalil2535: to make sure that it's trimmed 😂
ok i'll fix that

thc202: Better trim in `checkAndEnableConfirmButton()`, not worth enable the button if we are not going to accept it.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
khalil2535: I searched and only place "cfu.warn.addon.with.missing.requirements.addon" is used is here, do you want me to change it [here](https://github.com/zaproxy/zaproxy/blob/cfb89797be118cc9231e81db0c7dc090aec5eafe/zap/src/main/java/org/zaproxy/zap/control/AddOnRunIssuesUtils.java#L232) ?

image

kingthorin: Plus cfu has nothing to do with this PR

khalil2535: Ok

thc202: In the `CYCLIC` case.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
khalil2535: @thc202 did this good suggestion

kingthorin:
I’m trying to review this on a cell phone remotely it looks good to me but could you add a screenshot of the warning dialog when a user tries to add a duplicate entry? (I’m curious if the getLocalizedMessage call returns something useful from the exception we’ve newly thrown. [Hopefully it falls back to the getMessage text if a localized message isn’t found])

khalil2535:
image

kingthorin:
Thanks!

zap/src/main/java/org/zaproxy/zap/model/Context.java
```diff
@@ -343,17 +343,18 @@ private static void validateRegexs(List<String> regexs) {
      * Validates that the given regular expression is not {@code null} nor invalid.
      *
      * @param regex the regular expression to be validated
-     * @throws IllegalArgumentException if the regular expression is {@code null}.
+     * @throws IllegalArgumentException if the regular expression is {@code null} or Empty.
```
thc202: - The description of the method could be tweaked as well. `validateRegexs` should be updated accordingly.
- The empty should be lower case (same in the exception message).

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -212,8 +226,12 @@ protected void init() {
         @Override
         protected boolean validateFields() {
             try {
-                Pattern.compile(getRegexTextField().getText(), Pattern.CASE_INSENSITIVE);
-            } catch (PatternSyntaxException e) {
+                String trimmedRegex = getRegexTextField().getText().trim();
+                if (regexes != null && regexes.contains(trimmedRegex)) {
+                    throw new IllegalArgumentException(TEXT_ALREADY_EXIST_INVALID_REGEX_DIALOG);
```
thc202: - We should not use exceptions for normal/common control flow (when we can avoid it, they are expensive to create). This should show the dialogue and return.
 -  I'd suggest using other title (e.g. `Duplicated Regular Expression`) for the error and show without `The provided regular expression is not valid:`, seems unnecessary.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -212,8 +226,12 @@ protected void init() {
         @Override
         protected boolean validateFields() {
             try {
-                Pattern.compile(getRegexTextField().getText(), Pattern.CASE_INSENSITIVE);
-            } catch (PatternSyntaxException e) {
+                String trimmedRegex = getRegexTextField().getText().trim();
```
thc202: For consistency in what we accept/show we should also trim in `performAction`.

thc202:
This still does not prevent adding duplicates to the `Context`, one can still add them programmatically or through the ZAP API. I'm OK if this is just meant to address the UI but the commit message/PR should say that.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -211,23 +227,23 @@ protected void init() {
...
                         this,
-                        MessageFormat.format(TEXT_INVALID_REGEX_DIALOG, e.getLocalizedMessage()),
-                        TITLE_INVALID_REGEX_DIALOG,
+                        TEXT_ALREADY_EXIST_INVALID_REGEX_DIALOG,
+                        TITLE_ALREADY_EXIST_INVALID_REGEX_DIALOG,
                         JOptionPane.INFORMATION_MESSAGE);
                 getRegexTextField().requestFocusInWindow();
                 return false;
             }
+            Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
```
thc202: This lost the catch and error message, was that on purpose?

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -89,7 +87,7 @@ public void setRegexes(List<String> regexes) {
     @Override
     public String showAddDialogue() {
         if (addDialog == null) {
-            addDialog = new DialogAddRegex(owner);
+            addDialog = new DialogAddRegex(owner, this.getRegexes());
```
thc202: It will have to set the list each time (same for modify), it might change after constructing the dialogue.

thc202:
The modify dialogue needs to override `validateFields` to only do the base validations if the regex is different (to avoid showing the error that's duplicated).

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
khalil2535: do i have to create new object each time or to create `setRegexes` method in the `DialogAddRegex` class ?

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
khalil2535: No.

khalil2535:
> The modify dialogue needs to override `validateFields` to only do the base validations if the regex is different (to avoid showing the error that's duplicated).

It only shows the error when the regex is duplicated and the button "modify" is pressed.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
thc202: I'd go with the setter.

thc202:
> It only shows the error when the regex is duplicated and the button "modify" is pressed.

Exactly, but (IMO) it shouldn't, there's no duplication it's the regex itself. In that case ZAP should just accept it.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
thc202: The new constructors could be reverted, they are no longer needed.

khalil2535: Done @thc202.
Good advice's, I hope doing more development with you in this project.

thc202:
Thanks!

thc202:
@khalil2535 how would you like to be credited (e.g. name, handle)?

khalil2535:
> @khalil2535 how would you like to be credited (e.g. name, handle)?
> https://www.zaproxy.org/docs/desktop/credits/#zap-extended-team

how didn't i see this comment, omg of course.

name : Mahmoud Khalil
handle(if i understand it well): @khalil2535

PR #5912, opened 2020-03-25
Remove hardcoded colour in titled borders

thc202:
Do not specify any colour when creating the titled border, let the look
and feel use the most appropriate one.

Part of #5542.


kingthorin:
Uh wait this says borders but then deals with fonts....

kingthorin:
Okay now I see it’s where the label is part of the border.

PR #5919, opened 2020-04-01
Docker Java 11 Updates

kingthorin:
There doesn't appear to be an OpenJDK Alpine option for the bare image.
We could switch to: https://github.com/docker-library/openjdk/tree/master/11/jdk/slim ?

I've updated the other 3 standard images to OpenJDK 11.

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-04-01
+- Changed live and weekly images to use Java 11.
+
 ### 2020-02-21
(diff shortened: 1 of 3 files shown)


kingthorin:
Wasn't sure if I should label task or enhancement.

thc202:
> There doesn't appear to be an OpenJDK Alpine option for the bare image.

We can also use other Alpine images, from e.g. Azul, AdoptOpenJDK.

> I've updated the other 3 standard images to OpenJDK 11.

I'd leave out the stable image for now (just in case we need to release it again), we can update in two weeks if there are no comments against in the live/weekly images?

kingthorin:
Yup will tackle those tweaks later today.

thc202:
Thank you!

PR #5942, opened 2020-04-20
Allow to configure SOCKS proxy through the UI

### LINKED ISSUE #29: SOCKS (outgoing) proxy support
```
Due to some restrictions, I usually push all of my tool traffic through a server that
everyone knows is 'safe'. I usually use SSH tunnels for this sort of thing (ssh -D
$PORT $HOST), which creates a SOCKS5 proxy listening locally to $HOST. 

That way I can use proxychains to wrap anything I want through my proxy, but ZAP only
accepts http proxies, so I have a polipo instance running acting as a liaison between
my local SOCKS listener and ZAP. It'd be quite nice if ZAP could support SOCKS proxying!
```

Original issue reported on code.google.com by `fitblip` on 2010-12-17 19:07:11

thc202:
Allow to configure the SOCKS proxy in the options panel:
 - host;
 - port;
 - version;
 - if host names should be resolved;
 - authentication (username/password).

Change HTTPS socket factory (`SSLConnector`) and add HTTP socket factory
to not resolve host names if delegating to SOCKS proxy.
Move `Authenticator` to top level class and authenticate to SOCKS proxy
as well.
Add custom `ProxySelector` (and set as default) to log the selections
and failures.
Use system properties over persisted configurations for compatibility
with previous versions.

Fix #29 - SOCKS (outgoing) proxy support


thc202:
Opted that way because the logger instances are mutable (thus not constant), but happy to change.

PR #5944, opened 2020-04-22
Check -config override when determining the locale

### LINKED ISSUE #5934: Allow language to be specified on commandline
In theory this is already possible, eg using options like `-addoninstall help_fr_FR -config 'view.locale=fr_FR'`

However in practice due to the startup order the language will only actually get applied the second time ZAP is started.
This isnt really a problem when using the GUI, but it does mean that the packaged scans cannot be run with a specific language. Thats a shame because if we could do that then people would be able to specify which language they want the alerts in.
This would be especially useful for the baseline action - the language could be added an an optional parameter.

thc202:
Initialise the `Constant` with the `ControlOverrides` so that it can
check if the locale config is being set.

Fix #5934 - Allow language to be specified on commandline


psiinon:
Started up in French - perfect :D

PR #5950, opened 2020-04-25
Issue5897

beldcode:
The code is supposed to integrate a context when using ZAP via Docker with the Python API.


kingthorin:
Was there supposed to be something new here or is this PR an accident?

beldcode:
Sorry, I made a mistake with the pull request.

PR #5955, opened 2020-04-27
Issue5897

beldcode:
The code is supposed to integrate a context when using ZAP via Docker (with Python API).


kingthorin:
Please stop

kingthorin:
If you're unclear on how to use git/GitHub we can probably help you out via [IRC](https://www.zaproxy.org/irc/).

PR #5957, opened 2020-04-27
Fix Issue 5897 - Docker python scripts use Context if available

beldcode:
include context file (zap docker scan)

### DIFF
--- a/docker/zap_common.py
+++ b/docker/zap_common.py
@@ -54,6 +54,8 @@ OLD_ZAP_CLIENT_WARNING = '''A newer version of python_owasp_zap_v2.4
 
 zap_conf_lvls = ["PASS", "IGNORE", "INFO", "WARN", "FAIL"]
 zap_hooks = None
+context_id = None
+context_name = None
 
 def load_custom_hooks(hooks_file=None):
     """ Loads a custom python module which modifies zap scripts behaviour
@@ -376,7 +378,7 @@ def zap_access_target(zap, target):
 @hook(wrap=True)
 def zap_spider(zap, target):
     logging.debug('Spider ' + target)
-    spider_scan_id = zap.spider.scan(target)
+    spider_scan_id = zap.spider.scan(target, contextname=context_name)
     time.sleep(5)
 
     while (int(zap.spider.status(spider_scan_id)) < 100):
@@ -390,7 +392,7 @@ def zap_ajax_spider(zap, target, max_time):
     logging.debug('AjaxSpider ' + target)
     if max_time:
         zap.ajaxSpider.set_option_max_duration(str(max_time))
-    zap.ajaxSpider.scan(target)
+    zap.ajaxSpider.scan(target, contextname=context_name)
     time.sleep(5)
 
     while (zap.ajaxSpider.status == 'running'):
@@ -402,7 +404,7 @@ def zap_ajax_spider(zap, target, max_time):
 @hook(wrap=True)
 def zap_active_scan(zap, target, policy):
     logging.debug('Active Scan ' + target + ' with policy ' + policy)
-    ascan_scan_id = zap.ascan.scan(target, recurse=True, scanpolicyname=policy)
+    ascan_scan_id = zap.ascan.scan(target, recurse=True, scanpolicyname=policy, contextid=context_id)
     try:
         int(ascan_scan_id)
     except ValueError:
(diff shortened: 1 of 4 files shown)


kingthorin:
Please start with a properly up to date base branch.

thc202:
Now ready for reviews :)

kingthorin:
Thank you both!!!

docker/zap_common.py
```diff
@@ -382,23 +387,32 @@ def zap_spider(zap, target):
...
+'''
+In zap_active_scan, we use the global context_id, which is used in zap.ascan.scan to indicate that there is
+a context
+'''
 @hook(wrap=True)
 def zap_active_scan(zap, target, policy):
     logging.debug('Active Scan ' + target + ' with policy ' + policy)
-    ascan_scan_id = zap.ascan.scan(target, recurse=True, scanpolicyname=policy)
+    if context_id is None:
+        ascan_scan_id = zap.ascan.scan(target, recurse=True, scanpolicyname=policy)#, contextid=None)
```
psiinon, line 413: This looks a bit weird. Cant we just use contextid=context_id and drop the test?

docker/zap_common.py
```diff
@@ -382,23 +387,32 @@ def zap_spider(zap, target):
 
 
 @hook(wrap=True)
-def zap_ajax_spider(zap, target, max_time):
+def zap_ajax_spider(zap, target, max_time, context=None):
     logging.debug('AjaxSpider ' + target)
     if max_time:
         zap.ajaxSpider.set_option_max_duration(str(max_time))
-    zap.ajaxSpider.scan(target)
+    if context is not None:
```
psiinon, line 394: Is this test needed or can we pass across context=None?

docker/zap-baseline.py
```diff
@@ -321,10 +321,23 @@ def main(argv):
         time.sleep(2)
 
         # Spider target
-        zap_spider(zap, target)
+        '''
+        Here the zap spider is supposed to only use a context if a context file is given
```
thc202, line 325: I'd suggest removing this comment, it's repeating what the code is doing (same for the others).

docker/zap_common.py
```diff
@@ -510,8 +524,14 @@ def write_report(file_path, report):
...
         f.write(report)
 
+'''
+zap_import_context imports the context and stores a context_id which will be used by zap_active_scan. Therefore the
+context_id must be global so that zap_active_scan has access to the variable.
+'''
+
 @hook(wrap=True)
 def zap_import_context(zap, context_file):
+    global context_id
```
thc202, line 534: The ID is being returned so the caller can pass the ID when calling the other function.

docker/zap-baseline.py
```diff
@@ -321,10 +321,23 @@ def main(argv):
...
+        otherwise the inital spider funcionality without a context is used
+        '''
+        if context_file:
+            zap_spider(zap, target, os.path.splitext(os.path.basename(context_file))[0])
```
thc202, line 329: The name of the context might not be the same as the name of the file, better get the name from ZAP (e.g. last one in the context list), it also avoids processing the name multiple times.

kingthorin: Good catch.

thc202:
The additional parameter in `zap_spider` and `zap_ajax_spider` will break the hooks when using a context, e.g.:
```
TypeError: zap_spider() takes 2 positional arguments but 3 were given
```

I guess we'll need to use the global variable in the meantime to get the context name in the spider functions.

kingthorin:
9c5d70d seems to have included commented code instead of removing it, and a bunch of un-related files got line ending changes? Also the docker file name probably should change.

docker/zap_common.py
beldcode: I think we could use **kwargs since the argument "contextid" can appear or not. This way the test could be dropped..

beldcode: We could use **kwargs in the zap_active_scan function?

beldcode: the context_id variable is initialized in the zap_import_context function.. this function is only called if a context_file is given. So context_id is not always available..

kingthorin:
> Also the docker file name probably should change.

I made a typo that should have said **"shouldn't"**!

docker/zap-baseline.py
beldcode: I'll use ZAP for that! Thx for the hint.

beldcode:
> al

docker/zap-baseline.py
beldcode: @thc202 is there a special reason why to use the last context name from the context list?

thc202: The last one is the one that was imported/added by the script.

docker/zap_common.py
thc202: It's fine if it's `None`, the API client will just ignore that parameter.

thc202:
Rebased to fix the conflicts and bring the latest changes. No functional changes done yet.

thc202:
Addressed the review comments.

thc202 (approved):

thc202:
@beldcode thanks!

Issue #5963, opened 2020-04-30
Add host and path columns to the history table


DiogoMRSilva:
**Is your feature request related to a problem? Please describe.**
When you are looking to the history table is hard to see the path part of the URL, one of the reasons is the difference in the length of the host names. Even with the same host name it requires a certain level of effort to read the path.

**Describe the solution you'd like**
I suggest to create a column for the host name and one for the path with the arguments. These columns can be hidden and only shown if the user desires to.

**Would you like to help implementing this feature?**
I can implement this change

PR #5965, opened 2020-05-01
Add host, path and query columns to the history table

DiogoMRSilva:
resolve zaproxy/zaproxy#5963

I added 4 new columns to the history table: host name, path+query, path and query.
All the new columns are hidden by default.


zap/src/main/java/org/zaproxy/zap/view/table/DefaultHistoryReferencesTableEntry.java
```diff
@@ -120,6 +129,52 @@ public DefaultHistoryReferencesTableEntry(HistoryReference historyReference, Col
...
+        try {
+            hostname1 =
+                    hasColumn(sortedColumns, Column.HOSTNAME)
+                            ? historyReference.getURI().getHost()
+                            : null;
+        } catch (URIException e) {
+            hostname1 = null;
+            e.printStackTrace();
+        }
+        hostname = hostname1;
```
kingthorin, line 141: Since these are basically all the same why not extract a private method and pass the column constant?
Printing the stack trace should be replaced with a logger.

kingthorin: Also is the intermediate storage necessary? (var vs var1)

thc202: That's because they are final instance variables. Although IMO the components should be shown encoded, showing them decoded leads to data loss and misleading values (crude example, the path `/%2F` will be shown as `//` which is a completely different path).

zap/src/main/java/org/parosproxy/paros/extension/history/HistoryTableModel.java
```diff
@@ -44,7 +44,11 @@ public HistoryTableModel() {
                     Column.SIZE_RESPONSE_BODY,
                     Column.HIGHEST_ALERT,
                     Column.NOTE,
-                    Column.TAGS
+                    Column.TAGS,
+                    Column.HOSTNAME,
```
thc202, line 48: I'd expect these to be closer/next to the URL column by default.

zap/src/main/java/org/zaproxy/zap/view/table/AbstractHistoryReferencesTableEntry.java
```diff
@@ -267,6 +303,14 @@ public static Object getPrototypeValue(Column column) {
                 return String.class;
             case CUSTOM:
                 return null;
+            case HOSTNAME:
+                return String.class;
```
thc202, line 307: Can let fall-through.

zap/src/main/java/org/zaproxy/zap/view/table/HistoryReferencesTable.java
```diff
@@ -379,6 +379,11 @@ public void configureTableColumn(final TableModel model, final TableColumnExt co
                     columnExt, hRefModel.getColumnIndex(Column.SIZE_RESPONSE_HEADER), model);
             installSizeBytesRenderer(
                     columnExt, hRefModel.getColumnIndex(Column.SIZE_RESPONSE_BODY), model);
+            installSizeBytesRenderer(columnExt, hRefModel.getColumnIndex(Column.HOSTNAME), model);
```
thc202, line 382: This is not necessary for these columns.

thc202:
Worth noting that this will lead to increase of memory usage even if not visible.

DiogoMRSilva:
> Worth noting that this will lead to increase of memory usage even if not visible.
I added the path and the arguments columns because other people may prefer it to path+args. I can remove those 2 columns.
Do you have other suggestions to reduce the memory usage?

zap/src/main/java/org/zaproxy/zap/view/table/DefaultHistoryReferencesTableEntry.java
DiogoMRSilva: @thc202 I also think that they should be shown encoded, I didn't notice that was happening. @kingthorin the functions I will use now don't throw exception and the code will become more clean.

DiogoMRSilva:
Thanks for the review @thc202 and @kingthorin , I tackled the issues you found.

zap/src/main/java/org/zaproxy/zap/view/table/DefaultHistoryReferencesTableEntry.java
```diff
@@ -87,6 +91,20 @@ public DefaultHistoryReferencesTableEntry(HistoryReference historyReference, Col
...
 
         uri = hasColumn(sortedColumns, Column.URL) ? historyReference.getURI().toString() : null;
+        try {
+            hostname1 =
+                    hasColumn(sortedColumns, Column.HOSTNAME)
+                            ? historyReference.getURI().getHost()
+                            : null;
+        } catch (URIException e) {
+            hostname1 = null;
+            e.printStackTrace();
```
kingthorin, line 101: Should use a logger

zap/src/main/java/org/zaproxy/zap/view/table/DefaultHistoryReferencesTableEntry.java
```diff
@@ -98,6 +101,7 @@ public DefaultHistoryReferencesTableEntry(HistoryReference historyReference, Col
                             : null;
         } catch (URIException e) {
             hostname1 = null;
+            LOGGER.error(e);
             e.printStackTrace();
```
kingthorin, line 105:
```suggestion
```

zap/src/main/java/org/zaproxy/zap/view/table/DefaultHistoryReferencesTableEntry.java
thc202: The host should also be shown encoded (for same reason), e.g.:
`new String(historyReference.getURI().getRawHost())`

thc202:
> I added the path and the arguments columns because other people may prefer it to path+args. I can remove those 2 columns.

I was not suggesting nor asking to revert that, was just noting to revisit later.

> Do you have other suggestions to reduce the memory usage?

The best is to paginate the results but that would require way too many changes to be in scope of this PR. I don't think two more columns would make that much difference overall (there are other bigger culprits, e.g. alerts).

zap/src/main/java/org/zaproxy/zap/view/table/HistoryReferencesTableModel.java
```diff
@@ -58,7 +60,6 @@
         HIGHEST_ALERT,
         NOTE,
         TAGS,
-
```
thc202, line 61: This could be reverted.

DiogoMRSilva:
> > I added the path and the arguments columns because other people may prefer it to path+args. I can remove those 2 columns.
> 
> I was not suggesting nor asking to revert that, was just noting to revisit later.
> 
> > Do you have other suggestions to reduce the memory usage?
> 
> The best is to paginate the results but that would require way too many changes to be in scope of this PR. I don't think two more columns would make that much difference overall (there are other bigger culprits, e.g. alerts).

Can I leave it as it is or I should add the two other columns again?

thc202:
As you prefer :)

DiogoMRSilva:
So, for me it is ready :)

thc202:
Thanks!

PR #5973, opened 2020-05-08
added known csrf token for pfsense

rolfvreijdenberger:
the pfsense opensource security router/firewall distribution uses a known csrf token which has been added to the list.

Signed-off-by: Rolf Vreijdenberger <rolf@vreijdenberger.nl>


thc202:
Thanks!

PR #5974, opened 2020-05-09
Update scanners.md for JWT Scanner addon

preetkaran20:


kingthorin:
https://github.com/SasanLabs/owasp-zap-jwt-addon/blob/32281fdc71d7864f758ec78fccf2e3ea010e1597/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java#L51 Will need to be updated to align with this.

preetkaran20:
> https://github.com/SasanLabs/owasp-zap-jwt-addon/blob/32281fdc71d7864f758ec78fccf2e3ea010e1597/src/main/java/org/zaproxy/zap/extension/jwt/JWTActiveScanner.java#L51 Will need to be updated to align with this.

@kingthorin yes, i didn't updated because i was not sure if this change it correct or not. Will do the change. One question how does a addon appear on the zap market place ?

kingthorin:
When it’s ready to release you’ll just fill out this issue template: https://github.com/zaproxy/zaproxy/issues/new?labels=marketplace&template=Third_party_addon.md

thc202:
How about the other scan rule? The client one being its own rule, which might as well be a passive scan rule as it does not send any request?

Issue #5845, opened 2020-02-04
Alert API Additions


kingthorin:
Create two API endpoints that accept a CSV string of alertIds allowing to change the Confidence (1), or Risk (2), facilitating Bulk changes.

Related to:
zaproxy/zaproxy#1603
zaproxy/zaproxy#5843 (zaproxy/zaproxy#5844)

Bonus implementation detail: handle hyphenated alertId ranges (food for thought).

PR #5975, opened 2020-05-12
Alert API: Bulk Change Confidence or Risk - Fixes 5845

kingthorin:
Added API actions to bulk update Confidence or Risk based on a CSV string of alert IDs. Extracted some code to convenience methods which also lead to minor modification of code within deleteAlert and addAlert processing.

Fixes zaproxy/zaproxy#5845


thc202:
Needs a `spotlessApply`.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -402,6 +392,34 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
             newAlert.setWascId(wascId);
             extension.alertFound(newAlert, msg.getHistoryRef());
             return new ApiResponseElement(name, Integer.toString(newAlert.getAlertId()));
+        }  else if (ACTION_UPDATE_ALERTS_CONFIDENCE.equals(name)) {
+            String alertIds = ApiUtils.getNonEmptyStringParam(params, PARAM_ALERT_IDS);
+            int confidenceId = getConfidenceId(params);
+
+            StringTokenizer tokenizer = new StringTokenizer(alertIds, ",");
+            while (tokenizer.hasMoreTokens()) {
```
thc202, line 400: - I'd suggest collect them first to ensure all IDs are valid before starting processing the alerts (while missing alert will also lead to partial updates it's one less thing).
 - This can also be extracted, something like:
`updateAlerts(alert -> alert.setConfidence(confidenceId));`
and let `updateAlerts` extract/loop the IDs and get/process the alerts.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -63,10 +63,17 @@ alert.api.action.updateAlert.param.references = The reference details for the al
 alert.api.action.updateAlert.param.evidence = The evidence associated with the alert.
 alert.api.action.updateAlert.param.cweId = The CWE identifier associated with the alert.
 alert.api.action.updateAlert.param.wascId = The WASC identifier associated with the alert.
+alert.api.action.updateAlertsConfidence = Update the confidence of the alerts with the given IDs (comma separated values).
```
thc202, line 66: Just `Update the confidence of alerts.`? To avoid repeating the description of the parameter (same for the other endpoint).

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -336,19 +342,7 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
...
-                logger.error(e.getMessage(), e);
-                throw new ApiException(ApiException.Type.INTERNAL_ERROR, e);
-            }
-
-            if (recAlert == null) {
-                throw new ApiException(ApiException.Type.DOES_NOT_EXIST, PARAM_ALERT_ID);
-            }
-
-            Alert updatedAlert = new Alert(recAlert);
+            Alert updatedAlert = getAlertFromDb(alertId, PARAM_ALERT_ID);;
```
thc202, line 345: Double semicolon.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin: For the second one, how would the signature/definition of the method look?

thc202: Right, left out the `params`... should be like:
`void processAlerts(JSONObject params, Consumer<Alert> consumer)`

kingthorin: TY!

thc202:
Thank you!

PR #5997, opened 2020-05-19
Reduce code duplication in Scanner

rh-id:
simplifying duplicate HostProcess for readability

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Scanner.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Scanner.java
@@ -47,6 +47,7 @@
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2019/11/09 Ability to filter to active scan (Issue 5278)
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/Scanner.java
```diff
@@ -325,6 +294,21 @@ public void scan(Target target) {
         }
     }
 
+	private HostProcess getAndInitHostProcess(String hostAndPort, StructuralNode node) {
```
thc202, line 297: `createHostProcess`

Needs a ZAP comment after the license header.

rh-id: Hi, so I need to add `// ZAP: 2020/05/19 Remove HostProcess code duplication` after the `// ZAP: 2019/11/09 Ability to filter to active scan (Issue 5278)` ?

thc202: That works.

thc202:
The building is failing because of the code format (you'll have to run `gradlew spotlessApply` to address it).

thc202:
Thanks!

PR #6006, opened 2020-05-25
Fix "_XSERVTransmkdir: ERROR"

Ferenc-:
Without this patch, when running the docker container as non-root, the following error happens:

_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created.


Ferenc-:
@thc202 Could you please help with review?

thc202:
Thanks!

PR #6015, opened 2020-05-29
Tweak Date

kingthorin:
We're on 2020. (I did this in the Web UI, so the DCO will Likely have to be passed/skipped).


psiinon:
At least we wont have to change this for some time - 2020 is going to last for years! :P

Issue #6020, opened 2020-06-03
Incorrect JavaDoc


panthap2:
This comment is inconsistent: https://github.com/zaproxy/zaproxy/blob/3a1164d46e368d2353b0b664484afa23c746b008/zap/src/main/java/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java#L736

- [x] The return type of the getTxtProxyChainPassword() method is JPasswordField rather than ZapTextField. (Done in https://github.com/zaproxy/zaproxy/pull/6021)

thc202:
Tweaked the title to be generic, there are many cases where the JavaDoc is not correct.

PR #6021, opened 2020-06-03
Minor JavaDoc Correction

kingthorin:
Part of: zaproxy/zaproxy#6020
In follow-up to: https://github.com/zaproxy/zaproxy/pull/2309/

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
@@ -733,7 +733,7 @@ public class OptionsConnectionPanel extends AbstractParamPanel {
     /**
      * This method initializes txtProxyChainPassword
      *
-     * @return org.zaproxy.zap.utils.ZapTextField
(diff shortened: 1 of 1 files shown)


thc202:
Could you leave the issue open for tracking similar problem in other classes?
Quick check, there might be more:
 - `ExtensionEdit`
 - `ExtensionEncoder2` (probably not worth changing?)
 - `MainMenuBar`

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsConnectionPanel.java
```diff
@@ -733,7 +733,7 @@ private ZapTextField getTxtProxyChainUserName() {
     /**
      * This method initializes txtProxyChainPassword
      *
-     * @return org.zaproxy.zap.utils.ZapTextField
+     * @return JPasswordField that will be used to gather the password from the user
```
thc202, line 736: The type can be removed (e.g. `The field that will ...`), that's already provided in the method signature.

thc202:
Thanks!

PR #6026, opened 2020-06-07
JavaDoc Fixes

kingthorin:
Fixes zaproxy/zaproxy#6020


zap/src/main/java/org/parosproxy/paros/extension/edit/ExtensionEdit.java
```diff
@@ -127,7 +128,7 @@ public static KeyStroke getFindDefaultKeyStroke() {
     /**
      * This method initializes popupMenuFind
      *
-     * @return org.parosproxy.paros.extension.ExtensionPopupMenu
+     * @return org.parosproxy.paros.extension.edit.PopupFindMenu
```
thc202, line 131: Better remove the type as done in `getMenuFind()`, it's redundant (given by the method signature) and does not require further changes when the type changes (like happened here and in the other methods).

thc202:
Thanks!

PR #6028, opened 2020-06-08
Fix allow/deny list terminology

psiinon:
This was imported from another project, but we should still fix it.


psiinon:
Also used in the code so I'll need to update this PR...

PR #6044, opened 2020-06-18
Call default filter after custom filters

### LINKED ISSUE #6038: Allow Parsing of Some Non-text Messages by the Spider
 For the GraphQl addOn, I added a spider check for ".graphql" at the end of a URL to check if it is a schema. However, the Github graphql schema ( can be found here https://developer.github.com/v4/public_schema/ ) contains the "Content-type: application/octet-stream" in the response header, so it is not parsed by the spider.

Some non-text messages which pass spider checks of add-ons are allowed to be parsed.

**Additional context**
thc202 suggested on IRC, 

>Maybe query custom filters before the default one(?).The filter is happening in DefaultParseFilter (which uses HttpResponseHeader#isText). 

>We should probably allow to configure what's considered text in the DefaultParseFilter (but we really don't want allow all octet-stream, just if we know how to parse it)

**Would you like to help implementing this feature?**
Yes

ricekot:
- Call the default filter after the custom filters
- Introduce a new filter result "WANTED" to signal
that a parser wants to parse it.

Resolves #6038

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/spider/Spider.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/Spider.java
@@ -210,7 +210,8 @@ public class Spider {
         }
 
         // Add a default parse filter and any custom ones
-        this.addParseFilter(new DefaultParseFilter(spiderParam, extension.getMessages()));
+        controller.setDefaultParseFilter(
+                new DefaultParseFilter(spiderParam, extension.getMessages()));
         for (ParseFilter filter : extension.getCustomParseFilters()) this.addParseFilter(filter);
 
         // Add the scan context, if any
--- a/zap/src/main/java/org/zaproxy/zap/spider/SpiderController.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/SpiderController.java
@@ -63,6 +63,8 @@ public class SpiderController implements SpiderParserListener {
      */
     private LinkedList<ParseFilter> parseFilters;
 
+    private ParseFilter defaultParseFilter;
+
     /** The parsers used by the spider. */
     private LinkedList<SpiderParser> parsers;
 
@@ -226,6 +228,15 @@ public class SpiderController implements SpiderParserListener {
     public void addParseFilter(ParseFilter filter) {
         log.debug("Loading parse filter: " + filter.getClass().getSimpleName());
         parseFilters.add(filter);
+    }
+
+    protected void setDefaultParseFilter(ParseFilter filter) {
+        log.debug("Setting Default filter: " + filter.getClass().getSimpleName());
+        defaultParseFilter = filter;
+    }
+
+    protected ParseFilter getDefaultParseFilter() {
+        return defaultParseFilter;
     }
 
     public void init() {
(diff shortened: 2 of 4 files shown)


ricekot:
I am not very confident about this PR, hence the WIP. Please let me know what additional changes are required.

zap/src/main/java/org/zaproxy/zap/spider/Spider.java
```diff
@@ -209,9 +209,9 @@ private void init() {
             this.addFetchFilter(filter);
         }
 
-        // Add a default parse filter and any custom ones
-        this.addParseFilter(new DefaultParseFilter(spiderParam, extension.getMessages()));
+        // Add custom parse filters and set a default one
         for (ParseFilter filter : extension.getCustomParseFilters()) this.addParseFilter(filter);
+        this.setDefaultParseFilter(new DefaultParseFilter(spiderParam, extension.getMessages()));
```
thc202, line 214: - This could call the controller directly, we don't need to expose in the public API for now.
 - Could be done in the previous line as it no longer needs to be done last.

zap/src/main/java/org/zaproxy/zap/spider/SpiderController.java
```diff
@@ -228,6 +230,15 @@ public void addParseFilter(ParseFilter filter) {
         parseFilters.add(filter);
     }
 
+    public void setDefaultParseFilter(ParseFilter filter) {
```
thc202, line 233: Lets keep these methods internal for now (package accessible).

zap/src/main/java/org/zaproxy/zap/spider/SpiderController.java
```diff
@@ -228,6 +230,15 @@ public void addParseFilter(ParseFilter filter) {
...
+        log.debug("Loading parse filter: " + filter.getClass().getSimpleName());
```
thc202, line 234: This would have to be tweaked to "Setting default filter: ".

zap/src/main/java/org/zaproxy/zap/spider/SpiderTask.java
```diff
@@ -234,22 +234,29 @@ private void runImpl() {
...
-                                    + msg.getRequestHeader().getURI()
-                                    + "] fetched, but will not be parsed due to a ParseFilter rule: "
-                                    + filterResult.getReason());
-                }
-
-                parent.notifyListenersSpiderTaskResult(
-                        new SpiderTaskResult(msg, filterResult.getReason()));
-                return;
+            filterResult = filter.filtered(msg);
+            if (filterResult.isFiltered() || filterResult == FilterResult.WANTED) {
```
thc202, line 240: This could skip some filters when `WANTED`, the suggestion was to call all of them to allow custom restrictions (regardless of the order they were added).

zap/src/main/java/org/zaproxy/zap/spider/filters/ParseFilter.java
```diff
@@ -71,6 +71,9 @@ public FilterResult filtered(HttpMessage responseMessage) {
         /** Indicates that the resource was not filtered. */
         public static final FilterResult NOT_FILTERED = new FilterResult();
 
+        /** Indicates that the resource is wanted by a custom parser. */
```
thc202, line 74: Worth explaining that this skips the default filter, if no other filter filters it out.

thc202:
Thanks!

Issue #4851, opened 2018-07-17
Nashorn deprecated in Java 11


thc202:
Tracker issue for the status of Nashorn JavaScript engine following deprecation, for removal, in Java 11: http://openjdk.java.net/jeps/335

If no one picks the engine we would have to bundle an alternative (e.g. https://github.com/mozilla/rhino).

Pamplemousse:
If the APIs between Nashorn and Rhino has changed, then migrating from one to the other might break some [community scripts](https://github.com/zaproxy/community-scripts).

According to GitHub, 70.8% ([69 scripts](https://github.com/zaproxy/community-scripts/search?l=javascript)) of these are written in Javascript...

thc202:
They are not the same, we already had that problem when moving from Rhino to Nashorn...

kingthorin:
https://medium.com/graalvm/oracle-graalvm-announces-support-for-nashorn-migration-c04810d75c1f

kingthorin:
https://jaxenter.com/nashorn-deprecation-interview-caldwell-146951.html
https://www.infoworld.com/article/3291322/oracle-switch-now-from-nashorn-javascript-engine-to-graalvm.html

thc202:
For the record, GraalVM JavaScript engine has a compatibility mode for Nashorn.

thc202:
The Nashorn engine will be removed in Java 15: https://openjdk.java.net/jeps/372

thc202:
Closing this as done, the GraalVM JavaScript add-on is now available in the marketplace which can be used to run older JavaScript scripts (might require changes in some cases).

PR #6046, opened 2020-06-23
Just warn if ECMAScript engine is not found

thc202:
Warn instead of logging an error, the engine will no longer be available
in Java 15.

Part of #4851 - Nashorn deprecated in Java 11.


thc202:
I thought doing that but wasn't sure of suggesting that yet, we'll have to discuss how (and if) do we want to "migrate" the old scripts using Nashorn. While the GraalVM JavaScript engine has a compatible mode for Nashorn not all scripts will work correctly/the same (e.g. https://medium.com/graalvm/multi-threaded-java-javascript-language-interoperability-in-graalvm-2f19c1f9c37b ).

kingthorin:
Ok, can always be tweaked later.

PR #6051, opened 2020-06-26
Remove Coverity Badge

kingthorin:
Since we have other plans and it hasn't worked in quite a while, the badge may as well go away.


thc202:
Thanks!

kingthorin:
Thanks for taking care of the branch, I'm not sure why the GitHub web UI creates those on upstream instead of origin sometimes. (Might have to do with which repo I was in when I started the changes I guess.)

thc202:
No problem.

PR #6053, opened 2020-06-29
Import HAR files

bunysae:
This is a prerequisite to import HAR files into ZAP.


thc202:
Any reason why this is not being added to the add-on? (It's generally useful but so far (years) only one add-on needs it.)
Adding to core instead of the add-on requires the add-on to wait for a newer core release (even with reflection it would still need the newer core version for importing).

bunysae:
Okay i will move it to the add-on.

Issue #6043, opened 2020-06-18
Active scanner tests the same request twice


DiogoMRSilva:
**Describe the bug**
The active scanner tests some requests twice. I did run the ZAP aScanner on [DVPWA](https://github.com/anxolerd/dvpwa) and I was expecting ZAP to run the AbstractAppPlugin once on each request endpoint but in some endpoints it runs twice. The requests where it runs twice are the ones that have the folder icon on the Sites tree and that have an empty response.

**To Reproduce**
1. Run DVPWA on docker
2. Go to the endpoint http://127.0.0.1:8080/courses/3 and use the functionalities
3. run the scanner on the courses folder
4. check on debugger the tested messages

**Expected behavior**
Run the scan only once on each messages

**Screenshots**
![ascanDuplicated2](https://user-images.githubusercontent.com/6045401/85010264-9ea36180-b157-11ea-82d3-1ad0b3e35a1f.png)

On this case the request `GET http://192.168.56.103:8080/courses/3` is tested twice.

**Would you like to help fix this issue?**
Yes

PR #6057, opened 2020-06-30
Fix bug that makes zap test same request twice

DiogoMRSilva:
resolve zaproxy/zaproxy#6043

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
@@ -93,6 +93,8 @@
 // ZAP: 2019/11/09 Ability to filter to active scan (Issue 5278).
 // ZAP: 2020/09/23 Add functionality for custom error pages handling (Issue 9).
 // ZAP: 2020/10/19 Tweak JavaDoc and init startNodes in the constructor.
+// ZAP: 2020/06/30 Fix bug that makes zap test same request twice (Issue 6043).
+
 package org.parosproxy.paros.core.scanner;
 
 import java.io.IOException;
@@ -101,6 +103,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -114,6 +117,7 @@ import org.parosproxy.paros.model.HistoryReference;
 import org.parosproxy.paros.network.ConnectionParam;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
 import org.parosproxy.paros.network.HttpSender;
 import org.zaproxy.zap.extension.alert.ExtensionAlert;
 import org.zaproxy.zap.extension.ascan.ScanPolicy;
@@ -325,15 +329,25 @@ public class HostProcess implements Runnable {
             }
 
             for (StructuralNode startNode : startNodes) {
+                Map<String, Integer> historyIdsToAdd = new LinkedHashMap<>();
                 traverse(
                         startNode,
                         true,
                         node -> {
                             if (canScanNode(node)) {
-                                messagesIdsToAppScan.add(node.getHistoryReference().getHistoryId());
+                                int nodeHistoryId = node.getHistoryReference().getHistoryId();
+                                if (node.getMethod().equals(HttpRequestHeader.GET)) {
+                                    boolean nodeSeen = historyIdsToAdd.containsKey(nodeHash(node));
+                                    if (!nodeSeen || !isTemporary(node)) {
+                                        historyIdsToAdd.put(nodeHash(node), nodeHistoryId);
+                                    }
+                                } else {
+                                    messagesIdsToAppScan.add(nodeHistoryId);
+                                }
                             }
                         });
 
+                messagesIdsToAppScan.addAll(historyIdsToAdd.values());
                 getAnalyser().start(startNode);
             }
             nodeInScopeCount = messagesIdsToAppScan.size();
@@ -373,6 +387,16 @@ public class HostProcess implements Runnable {
             notifyHostComplete();
             getHttpSender().shutdown();
         }
+    }
+
+    private String nodeHash(StructuralNode node) {
+        String nodeMethod = node.getMethod();
+        String nodeURI = node.getURI().getEscapedURI();
+        return nodeMethod + nodeURI;
+    }
+
+    private boolean isTemporary(StructuralNode node) {
+        return node.getHistoryReference().getHistoryType() == HistoryReference.TYPE_TEMPORARY;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -363,6 +367,45 @@ public void run() {
...
+                    && !hasSamePropertiesAsRequestCreatedToFillSiteMap(sibling)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean hasSamePropertiesAsRequestCreatedToFillSiteMap(StructuralNode node) {
+        return node.getMethod().equals(HttpRequestHeader.GET)
+                && node.getHistoryReference().getHistoryType() == TYPE_TEMPORARY;
```
kingthorin, line 406: Instead of importing and using you can simply reference it statically.
```suggestion
                && node.getHistoryReference().getHistoryType() == HistoryReference.TYPE_TEMPORARY;
```

DiogoMRSilva: thx, I  know but I was using it more than once and lines were getting to big to fit in one line.

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -317,7 +319,7 @@ public void run() {
                         startNode,
                         true,
                         node -> {
-                            if (canScanNode(node)) {
+                            if (canScanNode(node) && bestNodeForUrl(node)) {
```
thc202, line 322: This check will not work properly for all cases, not all nodes are included by default (e.g. out of scope, filters, not recurse) so we can't discard the temporary node beforehand. Taking the example in the issue, no nodes are scanned if you select the node "3" and disable recurse.

It might be safer to de-duplicate once we have all of them.

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -363,6 +365,45 @@ public void run() {
...
+                    && childURI.equals(nodeURI)
+                    && sibling.isLeaf()
+                    && !hasSamePropertiesAsRequestCreatedToFillSiteMap(sibling)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean hasSamePropertiesAsRequestCreatedToFillSiteMap(StructuralNode node) {
```
thc202, line 402: `isTemporary`

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -363,6 +365,45 @@ public void run() {
...
+        return node.getMethod().equals(HttpRequestHeader.GET)
```
thc202, line 403: The type check would be enough, temporary tree nodes are all GET.

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
DiogoMRSilva: I forgot that case. I believe it can be solved by adding the check canScanNode to the sibling
```diff
...
            if (!sibling.isSameAs(node)
+                   && canScanNode(sibling)
                    && sibling.getMethod().equals(nodeMethod)
...
``
What do you think?

Comment (thc202) [reply zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java:322]: That would address most part of it (e.g. out of scope, filters) but still not the above recurse case, the `canScanNode` does not tell if a node is included in the nodes to be scanned, that's done while traversing (to not check more nodes than really needed).

Comment (DiogoMRSilva) [reply zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java:322]: True, I will do it after collecting all history nodes as you suggested.

Comment (thc202) [reply zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java:322]: It can also be de-duplicated as they are added, use a set and resolve the duplication immediately (might be simpler).

Comment (DiogoMRSilva) [reply zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java]: Yes, it's simpler :), I used a dict to keep track of the locations where the requests were stored.

**Comment by DiogoMRSilva** [comment]
> This pull request **introduces 1 alert** when merging [e098336](https://github.com/zaproxy/zaproxy/commit/e098336f1afda474191275a57f37340322ca81f3) into [2ab81f4](https://github.com/zaproxy/zaproxy/commit/2ab81f473288220270dbf13abc3a4a851b482abc) - [view on LGTM.com](https://lgtm.com/projects/g/zaproxy/zaproxy/rev/pr-82f1ca03fc55b29cc57fd9ec0f39687dfd669af4)
> 
> **new alerts:**
> 
>     * 1 for Dereferenced variable may be null

Should I fix this? I am sure that it will not be null because of the previous condition.

**Comment by psiinon** [comment]
Please - even if they are false positives we do our best to reduce LGTM warnings.

File: zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java [hunk 5]
```diff
@@ -313,12 +314,23 @@ public void run() {
...
+                                int nodeHistoryId = node.getHistoryReference().getHistoryId();
+                                if (isTemporary(node) && index == null) {
+                                    indexOfTemporary.put(
+                                            nodeHash(node), messagesIdsToAppScan.size());
+                                    messagesIdsToAppScan.add(nodeHistoryId);
+                                } else if (!isTemporary(node) && index == null) {
+                                    messagesIdsToAppScan.add(nodeHistoryId);
+                                } else if (!isTemporary(node) && index != null) {
+                                    messagesIdsToAppScan.set(index, nodeHistoryId);
+                                }
```
thc202, line 333: The checks can be reorder instead, e.g.:
```Java
if (index == null) {
    if (isTemporary(node)) {
        indexOfTemporary.put(nodeHash(node), messagesIdsToAppScan.size());
    }
    messagesIdsToAppScan.add(nodeHistoryId);
} else if (!isTemporary(node)) {
    messagesIdsToAppScan.set(index, nodeHistoryId);
}
```

DiogoMRSilva: yes, that looks better.

kingthorin (approved):

thc202:
Could you rebase and update the tests to match the new behaviour?

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -313,12 +314,23 @@ public void run() {
...
+                Map<String, Integer> indexOfTemporary = new HashMap<>();
                 traverse(
                         startNode,
                         true,
                         node -> {
                             if (canScanNode(node)) {
-                                messagesIdsToAppScan.add(node.getHistoryReference().getHistoryId());
+                                Integer index = indexOfTemporary.get(nodeHash(node));
+                                int nodeHistoryId = node.getHistoryReference().getHistoryId();
+                                if (index == null) {
```
thc202, line 325: This is not working correctly when the temp node is after the non-temp node, it still scans both of them.

DiogoMRSilva: I incorrectly assumed they always came first. Thx for the catch

thc202 (approved):

thc202:
Thanks!

psiinon (approved):

Issue #5996, opened 2020-05-19
Encoder Add-on Tracker


kingthorin:
- [x] Migrate core components. (Done in: https://github.com/zaproxy/zap-extensions/pull/2412 and https://github.com/zaproxy/zap-extensions/pull/2409)
- [x] Deprecate core components. (Done in: https://github.com/zaproxy/zaproxy/pull/6011)
- [x] Remove core help content. (Done in https://github.com/zaproxy/zap-core-help/pull/313/)
- [x] Hide/un-use core components.
- [x] Remove previously deprecated core code. (After 2.10.0) (Done in https://github.com/zaproxy/zaproxy/pull/6436.)
  - [x] And no longer needed message keys (Messages.properties). (done in: https://github.com/zaproxy/zaproxy/pull/6070)
- [x] Migrate help content. (Done in: https://github.com/zaproxy/zap-extensions/pull/2408)
- [x] Implement other templates/examples? (Maybe ROT13? or ROTn. Done in: https://github.com/zaproxy/zap-extensions/pull/2426)
- [x] In the new add-on remove "Advanced", after 2.10.0 and add-on targeting 2.10.0. (Done in: https://github.com/zaproxy/zap-extensions/pull/2614)

kingthorin:
Are we good with proceeding on the core removal? (Help and Menu Item ... the code can be left for a bit and just unused on the off chance that there's some third party dependency.)

kingthorin:
Re-opened as I realized I never went back and updated templates for other scripting engines.

kingthorin:
Assuming my syntax was correct for these scripts there seems to be a visibility/classloader issue with Kotlin/JRuby.

Went to do the Kotlin template:
```kotlin
import org.zaproxy.addon.encoder.processors.EncodeDecodeResult
import org.zaproxy.addon.encoder.processors.script.EncodeDecodeScriptHelper

/**
 * Process the input value and return the encoded/decoded/hashed etc. value
 *
 * Use helper.newError("Error Description") to provide an error description
 * inside the result view.
 *
 * @param {EncodeDecodeScriptHelper} helper - A helper object with various utility methods.
 *     For more details see https://github.com/zaproxy/zap-extensions/tree/main/addOns/encoder/src/main/java/org/zaproxy/addon/encoder/processors/script/EncodeDecodeScriptHelper.java
 * @param {String} value - The input value
[...]

PR #6070, opened 2020-07-08
Remove core encoder messages.properties entries

kingthorin:
As discussed via IRC on the off chance that a third party addon etc is using these or the core functionality that uses them the core will handle the fact that they're missing.

Part of: zaproxy/zaproxy#5996


thc202:
Thank you!

PR #6073, opened 2020-07-09
LogPanel - Tweaks

kingthorin:
Removed unused variable (and related import) and remove boilerplate javadoc for getScrollLog().


zap/src/main/java/org/parosproxy/paros/extension/history/LogPanel.java
```diff
@@ -87,7 +88,6 @@
 public class LogPanel extends AbstractPanel {
     private static final long serialVersionUID = 1L;
     // ZAP: Added logger.
```
thc202, line 90: Remove this as well.

thc202:
Thanks!

PR #6083, opened 2020-07-21
Add SonarCloud as a Scheduled GitHub Action

kingthorin:
- addOns.gradle.kts > Modified so that jacoco reports include XML output.
- build.gradle.kts > Modified to include sonarqube.
- sonar.yml > Added to run a GitHub action workflow every Saturday in order to update the SonarCloud SAST results and coverage info.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,34 @@
+name: Run Sonar
+
+on: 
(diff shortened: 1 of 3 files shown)


kingthorin:
Example run: https://github.com/kingthorin/zaproxy/actions/runs/176484566

Issue #6098, opened 2020-07-23
Allow custom request headers to be set by spider parser


vladidx:
**Is your feature request related to a problem? Please describe.**
Modern Web applications and API's support different kinds of resource representations and processing logic is often determined by supplied request headers. The current _SpiderParser_ implementation of ZAP doesn't allow to pass custom headers (e.g. "_Accept: application/json, \*/\*_" or "X-Custom: 1234") to the spider controller and therefore limits the spidering flexibility.

**Describe the solution you'd like**
The _SpiderParser_ class, and with that also the _SpiderParserListener_, should offer additional interfaces to pass request headers (e.g. in terms of a string-to-string map) to the spider controller. The custom request headers should be then included into the generated _HttpMessage_ before sending the request to the target.

**Describe alternatives you've considered**
No real alternatives known here. In some cases it is possible to influence the server-side processing logic and resource representation based on special query/path parameters but this is not generally applicable. 

**Additional context**
N/A

**Would you like to help implementing this feature?**
Yes, I have implemented a working PoC and would be happy to open a pull request for further discussion.

kingthorin:
Is this a dev request or a user request?

This could be done via an HttpSender script with the appropriate INITIATOR ID for spider/ajax spider.

psiinon:
An HttpSender script will definitely be able to do this, so well worth considering.

vladidx:
This is more related to a dev request.

I agree, an HttpSender script allows this in a certain way but from my point of view it also increases the complexity:
- The script needs to know from which spider parser the message originates but has only the initiator ID -> there might be cases where headers should only be added/changed if coming from specific spider parsers (as in my case)
- The script could only intercept singular messages. Once cloning  a message, modifying headers, and sending it within the script, the relation to the spider controller is lost to my understanding, i.e. it won't appear in the spider UI and statistics. Moreover, issuing the same URL from the spider parser is currently not possible to trigger the script multiple times; in any case it would be necessary to pass information to the HttpSender script which headers to set for the different messages as the spider parser possesses the knowledge.

thc202:
This is the same use case as the OpenAPI spider parser, which should really be using the spider to send the messages instead of sending them "manually".

lpinilla:
[...]

PR #6104, opened 2020-07-31
#6098: Allowing to pass request headers from a spider parser to the spider task

vladidx:
- Extended SpiderParserListener interface with additional methods
- New methods in SpiderParser including listener notifications
- Extended mechanism to track visited GET/POST requests with headers as part of the SpiderController
- SpiderTask receives request headers and adds them to the created HttpMessage request header
- Adapted SpiderParserTestUtils to reflect header changes

Related to #6098

### DIFF
--- a/.github/workflows/release-weekly-docker.yml
+++ b/.github/workflows/release-weekly-docker.yml
@@ -1,0 +1,23 @@
+name: Release Weekly Docker
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types: 'release-weekly-docker'
+
+jobs:
+  publish:
+    name: Build and Push Docker Image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: echo "::set-env name=REL_DATE::$(date +%F)"
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v1
+        with:
+          username: zapbot
+          password: ${{ secrets.ZAPBOT_DOCKER_TOKEN }}
+          repository: owasp/zap2docker-weekly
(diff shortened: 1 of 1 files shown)


vladidx:
Tooks some attempts but now the checks are green.

An open question from my side would if/how to test the enhancement as existing spiders do not make use of the new feature (yet?).

vladidx:
@thc202 / @psiinon: Any suggestion how to proceed further with this PR?

thc202:
> An open question from my side would if/how to test the enhancement as existing spiders do not make use of the new feature (yet?).

The `SpiderParser` can be tested in isolation, just need to spy and verify that it calls the correct methods of the listener.

> Any suggestion how to proceed further with this PR?

I'd suggest extracting a class for the resource found (containing the source URI, URI, depth, method, body, headers...), it would not require keeping adding more constructors/methods each time we allow to provide more data about the resource and should simplify the handling of GET/POST in the controller. We should have done this before... I'm happy to do that if you don't want to refactor though.

vladidx:
I can try to implement an additional unit test for adding request headers and also refactor the handling of resources detected by spider implementations.

Regarding the refactoring: Is it necessary to retain binary compatibility in the `SpiderParserListener` interface (implemented by SpiderController and TestSpiderController)? If not we could have a single interface method handled by the controller, e.g.
`public resourceFound(SpiderResourceFound resource)`

thc202:
It would be better if we kept binary compatibility, but I guess we can break it in this case. I don't see a reason why the add-ons would be implementing/using the listener directly, and a quick search in GitHub didn't show up anything.

PR #6080, opened 2020-07-17
Enable running zap using podman

NissesSenap:
Podman uses /run/.containerenv instead of .dockerenv
You can find more info here: https://github.com/containers/podman/issues/648

PR #6108, opened 2020-08-03
Add `IS_CONTAINERIZED` env var to containerized check

J12934:
Hi encountered a similar problem to #6080 when running ZAP on a Kubernetes cluster using `containerd`.

I could not find a similar file for containerd, so I've added a more generalized solution, which sets a custom env var in the docker image build, this should then be present in all container runtimes. 

This should allow the container to function correctly in all container runtimes, as it is not bound to a specific implementation.

This should also be able to replace the two existing checks (`os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv')`), as it should work for any container runtime.

### DIFF
--- a/docker/zap_common.py
+++ b/docker/zap_common.py
@@ -244,7 +244,7 @@ def cp_to_docker(cid, file, dir):
 
 
 def running_in_docker():
-    return os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv')
+    return os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv') or os.environ.get("IS_CONTAINERIZED") is not None
 
 
 def add_zap_options(params, zap_options):
--- a/docker/zap_common.py
+++ b/docker/zap_common.py
@@ -244,7 +244,7 @@ def cp_to_docker(cid, file, dir):
 
 
 def running_in_docker():
-    return os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv') or os.environ.get("IS_CONTAINERIZED") is not None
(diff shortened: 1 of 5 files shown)


docker/Dockerfile-bare
```diff
@@ -33,5 +33,6 @@ ENV PATH $JAVA_HOME/bin:/zap/:$PATH
 ENV ZAP_PATH /zap/zap.sh
 ENV HOME /home/zap/
 ENV ZAP_PORT 8080
+ENV IS_CONTAINERIZED true
```
thc202, line 36: This image does not use the Python scripts, does no harm either.

docker/zap_common.py
```diff
@@ -244,7 +244,7 @@ def cp_to_docker(cid, file, dir):
 
 
 def running_in_docker():
-    return os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv')
+    return os.path.exists('/.dockerenv') or os.path.exists('/run/.containerenv') or os.environ.get("IS_CONTAINERIZED") is not None
```
kingthorin, line 247: Should this actually check the value? Can anyone foresee a scenario where someone might purposefully set it to false or some value other than true?

docker/Dockerfile-bare
J12934: Ah right, good catch 👍

docker/zap_common.py
J12934: Good point. I've added an actual check on the value. Just to be sure that its always used consistently.

At the moment there is no real use to setting it to `false` as it would be overwritten by the path checks. Might be possible to extend this in the future when needed.

thc202:
Thanks!

PR #6130, opened 2020-08-19
Test common hooks of packaged scans

thc202:
Test that the hooks are properly loaded, using a file and env var.
Add tests to verify that the common hooks are triggered when and with
the expected values.


psiinon (approved):

thc202:
Rebased to pick #6131 and added a test for a common hook missed earlier.

Issue #6136, opened 2020-08-22
Fuzzer options might not be properly handled leading to exception


peterphung:
![image](https://user-images.githubusercontent.com/51269929/90951801-44d95380-e488-11ea-9398-780733d912f9.png)

PR #6142, opened 2020-08-25
Catch NPE on panel level when saving/validating

thc202:
Move the catch of the NullPointerException from the container panel to
the individual panels to still allow to validate/save other panels that
do not have issues.
Make the error generic now used for save/validate and provide the
details of the exception for easier report.

Related to #6136.
Improvement of #5306.


psiinon (approved):

thc202:
Corrected the date of a ZAP comment.

PR #6137, opened 2020-08-22
Stats: Ignore boundary when counting Content-Types

kingthorin:
- StatsPassiveScanner > Extract the content type and charset if present.
- StatsPassiveScannerUnitTest > UnitTest to assert the new behavior.

Fixes zaproxy/zaproxy#5952

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #6150, opened 2020-08-27
Tag Scanner Unit Test & Maintenance

kingthorin:
- RegexAutoTagScannerUnitTest > Add initial basic unittests.
- StatsPassiveScannerUnitTest > Moved to correct test package location. (In follow-up to zaproxy/zaproxy#6137)

### DIFF
--- a/zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
+++ b/zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
@@ -1,0 +1,119 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2020 The ZAP Development Team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zaproxy.zap.extension.pscan.scanner;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+import net.htmlparser.jericho.Config;
+import net.htmlparser.jericho.Source;
+import org.apache.commons.httpclient.URI;
+import org.apache.commons.httpclient.URIException;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+import org.parosproxy.paros.db.DatabaseException;
+import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.ZAP;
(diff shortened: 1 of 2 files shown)


zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
```diff
@@ -0,0 +1,118 @@
...
+        // Given
+        HttpMessage msg = new HttpMessage(new URI("http://example.com/", true));
+        msg.setResponseBody(BODY.replace("@@body_two@@", "<a href=\"mailto:me@.my.net\">me</a>"));
+        msg.setHistoryRef(mock(HistoryReference.class));
+        // When
+        MAILTO_TAG_SCAN_RULE.scanHttpResponseReceive(msg, -1, new Source(msg.getResponseBody().toString()));
+        // Then
+        verify(listener)
+                .counterInc(
+                        "http://example.com", RegexAutoTagScanner.TAG_STATS_PREFIX + MAILTO_CONFIG);
```
kingthorin, line 101: This seemed like an easier way to go, since the messages don't actually have a `HistoryReference` from which to get the tags. However, tags are applied and counted from the same method 😀

zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.ZAP;
+import org.zaproxy.zap.utils.Stats;
+import org.zaproxy.zap.utils.StatsListener;
+
+public class RegexAutoTagScannerUnitTest {
+
+    private static final String BODY =
+            "<html><head>@@head@@</head><body>@@body_one@@ @@body_two@@</body><html>";
+    private static final String MAILTO_PATTERN = ".*href\\s*=\\s*['\"]?mailto:";
```
kingthorin, line 46: The default patterns appear to be part of the default config file, they aren't code constants etc. So this was necessary.

kingthorin:
This was just a quick setup so that I had something to build on top of in order to see if re2j was a practical alternative. I quickly did some testing with this setup and it doesn't seem to be. However, it can/should also be leveraged in the future for things like: https://github.com/zaproxy/zaproxy/pull/5794

psiinon (approved):

zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
thc202: Not always, the counter might be incremented even when not tagged. The test class has the (mocked) `HistoryReference` set previously so it can be verified through that one (or just check the `parent`). These would be another tests anyway.

zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
thc202: We should test the default regular expressions separately, this class does not depend on them (it would be better to use a test/dummy config and regular expression here).

zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final String MAILTO_CONFIG = "MailTo";
+    private static final RegexAutoTagScanner MAILTO_TAG_SCAN_RULE =
```
thc202, line 48: It would be better to use an instance per test to avoid the need to clean it up more (e.g. when testing the setters).

zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
kingthorin: The mocked history reference doesn't get tags. There's no getter for parent and it can't be spy'd because it doesn't have a zero arg construcutor.

thc202 (approved):

PR #6160, opened 2020-09-04
Properly decode gzip in UI and decode request

### LINKED ISSUE #1351: Decode gzip'ed content in Request tab
```
Hello,
I am testing some application on Android which communicates with server over the XML
SOAP messages. The content is of the request as comming from client is XML encoded
with gzip. The same is on the way back from the server (XML+gzip).

ZAP correctly (in most cases) decodes the responses from server, but the problem is
that requests from client are gzip binary garbage so it is not easy to fuzz or do whatever
else with it. 

Please could you consider adding the support for the gzip encoded requests and not
only the gzip encoded responses?

The request from client looks like this:
POST http://X.XXX.XXX.XX:8085/doLogin HTTP/1.1
Content-Type: application/xml
Content-Encoding: gzip
User-Agent: [android-9eadc32e98dce184]-14100119571190520001
Content-Length: 223
Connection: close
Host: X.XXX.XXX.XX:8085

XXXXXX-Binary-gzip-garbage-XXXXXXXXX


In window of the request payload I would expect to see something like
<MSG>
  <Hdr Id="55035660" Ver="1.038" Dtm="2014-09-08T11:45:13">
    <Sndr AppCd="ESB" AppNm="ESB"/>
    <Rcp AppCd="GCA" AppNm="GCA">
      <GId Id="14090813450758190001" IdTp="RQMSGID"/>
    </Rcp>
  </Hdr>
  <Bd>
    <Usr UsrId="user01">
    <Pass Passwd="password">
  </Bd>
</MSG>



I have tried with ZAP 2.3.1 and also with the weekly build ZAP_D-2014-09-29

Best regards
Michal Ambroz




```

Original issue reported on code.google.com by `michal.ambroz` on 2014-10-02 09:40:37

### LINKED ISSUE #1584: GZip de-compression causes loss of newlines
```
1. Find a URL where the response body contains newlines, and where the server supports
gzip compression.
2. Send a request for the URL *with* the request header "Accept-Encoding: gzip" 
3. Send a request for the URL *without* the request header "Accept-Encoding: gzip"

4. View both requests by clicking on them in the history tab

What is the expected output? 
The two outputs should match, when ZAP is de-compressing the response for display.

What do you see instead?
Currently, the two outputs do not visually match.  The request that was specified *with*
"Accept-Encoding: gzip" is missing newlines that appear in the other request.

Background: 
refer to https://code.google.com/p/zaproxy/issues/detail?id=1346&q=gzip

Please use labels and text to provide additional information.


```

Original issue reported on code.google.com by `colm.p.oflaherty` on 2015-03-30 23:49:59

thc202:
Do not gzip decode reading lines, read byte chunks instead to not lose
the new line characters.
Change request view models (body and header+body) to decode the body,
like done for the response.
Extract common code to `HttpPanelViewModelUtils` and change the models
accordingly.
Normalise behaviour when setting data with null message, by ignoring the
data being set.

Fix #1351 - Decode gzip'ed content in Request tab
Fix #1584 - GZip de-compression causes loss of newlines

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/httppanel/view/impl/models/http/HttpPanelViewModelUtils.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/httppanel/view/impl/models/http/HttpPanelViewModelUtils.java
@@ -19,9 +19,25 @@
  */
 package org.zaproxy.zap.extension.httppanel.view.impl.models.http;
 
+import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.util.regex.Pattern;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+import org.apache.log4j.Logger;
+import org.parosproxy.paros.network.HttpBody;
+import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
 
 public final class HttpPanelViewModelUtils {
+
+    private static final Pattern GZIP_PATTERN =
+            Pattern.compile("\\s*(?:x-)?gzip\\s*", Pattern.CASE_INSENSITIVE);
+
+    private static final Logger logger = Logger.getLogger(HttpPanelViewModelUtils.class);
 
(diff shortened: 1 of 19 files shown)


kingthorin (approved):
Tks

psiinon (approved):

PR #6181, opened 2020-09-17
Fix typo in extension name

thc202:
Extracted from #5719 since it does not require an upgrade from.

The change might affect add-ons using the constant to access the
extension but it's very unlikely that's happening, the add-ons wanting
to add "all" components would do so using the name of the component
itself not the extension.


psiinon (approved):

kingthorin:
Travis seems to need a kick again.

thc202:
Should be good now.

kingthorin (approved):

kingthorin:
Are we good with merging this as-is?

thc202:
Do you mean because the commit does not have the above explanation?

kingthorin:
Just a general question. I feel its all fine, was just trying to be sure. I know we broke it off because it didn’t require upgrade handling but wasn’t clear if there was anything more I guess.

thc202:
OK, it should be good to merge as is.

PR #6182, opened 2020-09-17
Improve error message when starting packaged scans

thc202:
Fail immediately if the spider scans were not started to provide better
error message, show:
`Failed to start the scan, check the log/output for more details.`

instead of, e.g.:
`invalid literal for int() with base 10: 'does_not_exist'`.

when checking the status of the scan.

Correct `suite.py` to exit with expected code when tests fail.

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-09-18
+ - Fail immediately if the spider scans were not started to provide better error message.
+
 ###  2020-08-28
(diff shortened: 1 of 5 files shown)


kingthorin:
Is this expected/normal, I haven't seen this before:
image

thc202:
Interesting, seems some tests failed but not the check.

thc202:
The checks are now showing the expected state when the tests fail (I thought the test script was already doing that...), will fix the hook tests to match the new behaviour.

thc202:
Done.

kingthorin (approved):
LGTM, thanks!

psiinon (approved):

PR #6188, opened 2020-09-21
Correct incremental delay in scheduled executor

### LINKED ISSUE #5701: Fuzzer runs progressively slower
**Describe the bug**
When using the Fuzzer to brute force a string in a request (such as a password) while making use of the timing throttle, the requests get slower and slower over time.

**To Reproduce**
1. Set up a fuzz on a string in a request. Payload is usually a list of strings, but it doesn't matter.
2. Use the timing and throttling controls to set 1-5 threads, and a timing of 250ms.
3. Start the fuzzer.
4. The attack will start off fast but very quickly drop off in speed, far far beyond the manual timing set in the control.

**Expected behavior**
The timing between requests should stay constant at the setting provided by the user.

**Software versions**
ZAP 2.8.1
OS: Linux 64bit
OpenJDK 1.8.0_232
Firefox 70

**Additional context**
The timings get progressively worse (despite the site itself responding in a relatively constant speed) which makes me think that whatever is computing the timing per thread is periodically increasing the indicated timing (250ms) in additive fashion.

thc202:
Check the number of queued tasks not the total number of tasks submitted
when increasing the delay, otherwise it would not work properly when the
tasks are executed/submitted periodically in batches (instead of all at
once).

Fix #5701.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/utils/PausableScheduledThreadPoolExecutor.java
+++ b/zap/src/main/java/org/zaproxy/zap/utils/PausableScheduledThreadPoolExecutor.java
@@ -56,7 +56,7 @@ public class PausableScheduledThreadPoolExecutor extends ScheduledThreadPoolExec
     private long defaultDelayInMs;
 
     private boolean incrementalDefaultDelay;
-    private AtomicInteger taskCount;
+    private AtomicInteger queuedTaskCount;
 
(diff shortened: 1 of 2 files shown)


psiinon (approved):

kingthorin (approved):
Thank you, it was very nice of you to tackle this.

wadim-al:
Thanks, I've just tested the new weekly build and confirm the delay in the fuzzer is now correct.

PR #6193, opened 2020-09-23
Add option to not fix the content length

### LINKED ISSUE #560: Content-Length header always updated on Send - cannot Resend some requests
```
1. Do an Active Scan
2. One of the requests has an invalid Content-Length header
3. Right-Click -> Resend
4. Click "Send"

* You see the request with the invalid C-L header in the Resend dialog (see below)
* C-L is changed and sent as 0 instead of with the invalid value in the UI
* Need a way to NOT update C-L (normally handy, but need a way to disable this feature)


2.0.0 on OSX


Please provide any additional information below.

This is the request in question (from Active Scan):
GET http://site-5.example.tld/ HTTP/1.1
Host: site-5.example.tld
User-Agent: Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.0;)
Pragma: no-cache
Cache-control: no-cache
Content-Type: application/x-www-form-urlencoded
Content-length: |!<>=~=>=<=*(),+-"'/&;

```

Original issue reported on code.google.com by `brectanus@qualys.com` on 2013-03-14 03:25:55

psiinon:
Fixes #560

Adds a 'Fix content length' button to both the Manual Req Editor and Break Panel. If enabled (which it is by default) then it will fix the content length header (as now) If disabled it wont.
Will only appear in the Manual Req Editor for HTTP requests and will be disabled in the Break Panel for non HTTP requests.


zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -860,6 +860,7 @@ brk.add.title                     = Add breakpoint
 brk.alwaysOnTop.message			  = By default ZAP will remain on top of all other windows when a breakpoint is hit.\n\
 Press 'Cancel' to disable this feature.\n\
 This option can changed via Tools/Breakpoints
+brk.checkBox.fixLength            = Fix content length
```
kingthorin, line 863:
```suggestion
brk.checkBox.fixLength            = Fix Content Length
```

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1958,6 +1959,7 @@ log4j.tools.menu.gc         = Run the Garbage Collector
 
 manReq.name = Manual Request Editor Extension
 manReq.button.send             = Send
+manReq.checkBox.fixLength      = Fix content length
```
kingthorin, line 1962:
```suggestion
manReq.checkBox.fixLength      = Fix Content Length
```

kingthorin (changes requested):
Just title caps, otherwise looks good to me.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
thc202: Could it be called "update" rather than "fix"?

psiinon:
Changed to `Update Content Length` :)

psiinon:
Fixed those :)

psiinon:
Ugh, those tests appeared to work locally :/ Investigating..

psiinon:
That looks better ;)

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -356,6 +372,12 @@ public void run() {
 
                             if (isRequest) {
                                 requestPanel.saveData();
+                                Message msg = getMessage();
+                                if (msg instanceof HttpMessage
```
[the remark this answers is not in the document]
thc202: The option is not being shown for the response.

zap/src/test/java/org/zaproxy/zap/extension/httppanel/view/impl/models/http/BodyByteHttpPanelViewModelTest.java
```diff
@@ -122,7 +122,6 @@ void shouldSetDataIntoBody() {
         model.setData(otherBodyContent);
         // Then
         verify(body).setBody(otherBodyContent);
-        verify(header).setContentLength(otherBodyContent.length);
```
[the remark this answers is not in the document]
thc202: Was suggesting e.g. `verify(header, times(0)).setContentLength(anyInt());` and in all places that was checking the header before.

psiinon:
Hopefully fixed all of those :)

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Needs a `spotlessApply`

psiinon:
doh! Done...

PR #6194, opened 2020-09-23
Custom Pages analyzer fallback & clean code

kingthorin:
- In `AbstractPlugin.isPage404` fallback to use `Analyser.isFileExist`. In `Abstractplugin.isFileExist` check `isPage200` first then use `Analyser`.
- Tweak various references to CustomPageType -> CustomPage.Type.

See zaproxy/zaproxy#9.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -580,17 +580,18 @@ public abstract class AbstractPlugin implements Plugin, Comparable<Object> {
     }
 
     /**
-     * Tells whether or not the file exists, based on previous analysis.
+     * Tells whether or not the file exists, based on {@code CustomPage} definition or previous
+     * analysis.
      *
      * @param msg the message that will be checked
      * @return {@code true} if the file exists, {@code false} otherwise
      */
     protected boolean isFileExist(HttpMessage msg) {
-        return parent.getAnalyser().isFileExist(msg);
+        return isPage200(msg);
     }
 
     /**
-     * Tells whether or not the message matches the specific {@code CustomPageType}
+     * Tells whether or not the message matches the specific {@code CustomPage.Type}.
      *
      * @param msg the message that will be checked
      * @param cpType the custom page type to be checked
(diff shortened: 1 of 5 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -1300,6 +1302,9 @@ protected boolean isCustomPage(HttpMessage msg, CustomPage.Type cpType) {
         if (getContext() != null) {
             return getContext().isCustomPageWithFallback(msg, cpType);
         }
+        if (CustomPage.Type.NOTFOUND_404.equals(cpType)) {
+            return !getAnalyser().isFileExist(msg);
+        }
```
kingthorin, line 1307: I actually originally thought hooking analyzer was going to be more involved, but with our most recent changes I believe this is all that's needed.

Also feel free to correct me if the logic is wrong but this is how I was considering it: `isCustomPage` checks if this matches a 404 definition and returns true which is the inverse of `isFileExist` (it returns false in 404 cases, but as far as custom pages are concerned if it is a 404 then we want true).

zap/src/main/java/org/parosproxy/paros/core/scanner/HostProcess.java
```diff
@@ -1289,7 +1289,9 @@ public PluginStats getPluginStats(int pluginId) {
     }
 
     /**
-     * Tells whether or not the message matches the specific {@code CustomPageType}
+     * Tells whether or not the message matches the specific {@code CustomPage.Type}. Falls back to
+     * use {@code Analyser} {@link #isFileExist(HttpMessage)} if the type is {@code
```
kingthorin, line 1293: Should the javadoc here be fore specific about the fallbacks/assumptions since this method uses `Context.isCustomPageWithFallback`?

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -580,17 +580,19 @@ protected void bingo(
...
+     * Tells whether or not the file exists, based on {@code CustomPage} definition or previous
+     * analysis.
      *
      * @param msg the message that will be checked
      * @return {@code true} if the file exists, {@code false} otherwise
      */
     protected boolean isFileExist(HttpMessage msg) {
-        return parent.getAnalyser().isFileExist(msg);
+        boolean isFileExist = isPage200(msg);
+        return isFileExist ? isFileExist : parent.getAnalyser().isFileExist(msg);
```
thc202, line 591: This could be done in the isPage200 and this method call that one.

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6214, opened 2020-10-04
added check for running threads

### LINKED ISSUE #6213: Support for checking if all executor threads paused in the PausableExecutorService
On issuing a pause to the PauseableExecutorService, it holds the new threads from executing but does not pause the already running threads and they are supposed to run to completion.
It will be useful to have an API to expose the information on the number of running threads.

15pratik:
Fixes #6213 
This maintains a counter on the number of threads under execution
the counter is incremented when the runnable is allowed to be submitted when the state is 'unpaused' and the counter is decremented after execution.
added an API to return if there are no threads under execution by checking on the runningCount.


kingthorin:
Needs a `spotlessApply`

thc202:
Closing, the implementation has concurrency issues, also, this would introduce unnecessary overhead to all cases (where it's just needed in one).

Issue #2620, opened 2016-06-24
Allow to add/remove authentication and session management methods


thc202:
Allow add-ons to dynamically add/remove authentication (e.g. Form-based Authentication, HTTP/NTLM Authentication...) and session management (e.g. Cookie-based Session Management, HTTP Authentication Session Management) methods.
The change would allow add-ons to effectively provide authentication and session management methods (currently the methods are loaded when ZAP starts, which would prevent add-ons from being correctly installed/updated) and extract/move the current methods from core (so they can be easily updated).

psiinon:
OK, so it is possible to add session management methods dynamically, but its a bit hacky and doesnt work well with the GUI or the API.

PR #6218, opened 2020-10-05
Added support for plugable SessionManagementMethodTypes

psiinon:
Part of #2620


kingthorin (commented):
Looks good, thanks ⭐

kingthorin (approved):

kingthorin:
I'm not sure what happened with my original review, it should have been approve not comment 🤷‍♂️

thc202:
Some cases that don't seem to be handled:
 - The API does not reflect the types added/removed;
 - The types are not being updated in the contexts when added (nor being removed);
 - When importing a context the data of the type that's not present is lost, this means that even if the user installs the required add-on it has to import the context again (rather than ZAP just use the data already imported);
 - A missing type leads to loss of data in the session, since the UI forces the use of an existing method instead of showing a placeholder informing the user the method is missing;
 - The codebase assumes that the context has always a type but that might no longer be true (e.g. type missing).

psiinon:
Started addressing those points but not finished yet, hence the WIP.

thc202:

psiinon:
Just updated to latest code, previous comments still outstanding.

psiinon:
Superceeded by other PRs

Issue #3813, opened 2017-08-09
Timeout or Termination functionality for Passive Scanning


kingthorin:
Per discussion here: https://github.com/zaproxy/zaproxy/pull/3787

Quote @kingthorin:
> I wonder if we could/should integrate something like this into the core? If we have a way to abandon or terminate a particular plugin (or passive scanning of a specific request/response in general) that could be a way to handle things like the passive scanners that use Hirshberg. 
(Like: _Passive scanning of message x took longer than the configured timeout of 120sec, abandoning assessment of passive plugin y._)

Note: For context, we've had performance problems related to the scanners which depend on the Hirshberg algorithm in the past....

Quote @thc202:
> Other scripts would benefit from this check (e.g. #3531, but if this is going to be added to core then it might not be necessary).
> The timeout could also be dependent on the number of messages to scan?

thc202:
The passive scanners usually "hang" because of regular expressions with big responses. An option to allow to specify the maximum size of the bodies to passive scan should go along with this.

nil0x42:
I'm sad to see that this issue's milestone is now 2.9.0, it makes passive scanning unusable in a lot of contexts.

psiinon:
@nil0x42 PRs are still being accepted for 2.8.0! We're a relatively small team and are focussing on releasing 2.8.0, so all contributions appreciated :)

PR #6226, opened 2020-10-06
Allow to define the maximum body size to passive scan

psiinon:
Partly addresses #3813

Not sure how to add unit tests for this :/

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
@@ -182,6 +182,7 @@ public class PassiveScanThread extends Thread implements ProxyListener, SessionC
                         Source src = new Source(msg.getResponseBody().toString());
                         currentUrl = msg.getRequestHeader().getURI().toString();
                         PassiveScanData passiveScanData = new PassiveScanData(msg);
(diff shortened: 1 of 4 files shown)


zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanThread.java
```diff
@@ -179,6 +179,23 @@ public void run() {
                     try {
                         // Parse the record
                         HttpMessage msg = href.getHttpMessage();
+                        if (this.pscanOptions.getMaxBodySizeInBytesToScan() > 0
```
thc202, line 182: Worth adding `msg.isResponseFromTargetHost()`.

thc202:
- Should the option be explicit in that it applies only to response bodies?
 - Could you update the user guide?

psiinon:
Changed to check both request and response bodies.

kingthorin:
`PausableScheduledThreadPoolExecutorUnitTest` seems to have failed again. I tried to "Re-run" via the web UI but Travis doesn't seem to have started again. You might have to kick it with an amended commit.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thanks

PR #6227, opened 2020-10-07
Update Rust API generation (async / .await)

cak:
Update Rust API:
- to utilize async / .await
   - request v.10
- change parameters to string slices

Rust API PR: https://github.com/zaproxy/zap-api-rust/pull/8

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/RustAPIGenerator.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/RustAPIGenerator.java
@@ -137,7 +137,13 @@ public class RustAPIGenerator extends AbstractAPIGenerator {
         for (ApiParameter parameter : element.getParameters()) {
             String name = parameter.getName();
             String varName = getSafeName(name.toLowerCase(Locale.ROOT));
-            out.write("    params.insert(\"" + name + "\".to_string(), " + varName + ".to_string()" + ");\n");
+            out.write(
+                    "    params.insert(\""
+                            + name
+                            + "\".to_string(), "
+                            + varName
+                            + ".to_string()"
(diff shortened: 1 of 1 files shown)


thc202:

zap/src/main/java/org/zaproxy/zap/extension/api/RustAPIGenerator.java
```diff
@@ -93,8 +93,7 @@ private void generateRustElement(ApiElement element, String component, String ty
             out.write("/**\n");
             out.write(" * " + desc + "\n");
             if (isOptional()) {
-                out.write(" * <p>\n");
-                out.write(" * " + OPTIONAL_MESSAGE + "\n");
+                out.write(OPTIONAL_MESSAGE + "\n");
```
thc202, line 96: Was this changed on purpose? It generates the following now:
```rust
/**
 * Endpoint description.
This component is optional and therefore the API will only work if it is installed
*/
```
I'd expect the ` * ` before the optional message.

thc202:
The code changes look ok to me, though I don't know rust :)

zap/src/main/java/org/zaproxy/zap/extension/api/RustAPIGenerator.java
thc202: Also, the other newline would help separate the optional message from the description.

Leyart: I agree it is missing the " * " in front of the optional message

Leyart (changes requested):
Add " * " again in front of OPTIONAL_MESSAGE

PR #6228, opened 2020-10-07
Update Log4j to 2.x

### LINKED ISSUE #6196: Upgrade the version of log4j

The log4j:1.2.17 jar file is flagged as a potentially vulnerable jar file on a recent Nexus scan.

Upgrade log4j to a more recent version.

**Describe alternatives you've considered**
None.

**Additional context**
None.

**Would you like to help implementing this feature?**
I thought this might be as simple as replacing the `log4j:1.2.17` dependency in `zap.gradle.kts` with the following:

```java
api("org.apache.logging.log4j:log4j-1.2-api:2.13.3")
api("org.apache.logging.log4j:log4j:2.13.3")
```

The above code upgrades log4j to the next major version (2) and uses the `log4j-1.2-api` as an adapter to ensure that usages in ZAP of the log4j 1.2 code do not need to change. I encountered an issue with this approach where [ZapBootstrap](https://github.com/zaproxy/zaproxy/blob/172deaf8bf231b58b535fcd762dc52443d6fb860/zap/src/main/java/org/zaproxy/zap/ZapBootstrap.java#L69) uses the log4j `NullAppender`. This code doesn't compile with the above approach.

If you think this approach is worthy of pursuing, and you can roughly guide me on how to reproduce the issue this solved: `// Nasty hack to prevent warning messages when running from the command line`, then I'm happy to try to create a pull request with this fix/feature.

thc202:
Depend on the new major version and use 1.x bridge.

Configure Log4j 2 to use a single logger context for ZAP, to use the
same log configuration for everything (core, add-ons, dependencies).

Use two Log4j 2 configurations for main code, one during bootstrap and
another for remaining lifecycle:
 - The bootstrap configuration is automatically loaded by Log4j, from
 the file `log4j2.properties`, which just logs errors to std out.
 - The other configuration, bundled as `log4j2-home.properties`, is
 copied to ZAP home as `log4j2.properties` and is loaded once the home
 is known. This configuration is the equivalent Log4j 1 configuration
 also copied/loaded from the ZAP home.

Use a specific configuration for the tests (`log4j2-test.properties`,
log set to off by default), to not require to tweak the log manually
throughout the tests.
Remove usages of Log4j 1 `NullAppender` in the tests, the class does not
exist in the Log4j 1.x bridge, also, superseded by the configuration.

Migrate the following classes to use the new Log4j 2 APIs:
 - `Constant`, to load the configuration;
 - `SplashScreen`, to show the info/error log events;
 - `ExtensionLog4j`, to show and count error log events;
 - `ZAP.UncaughtExceptionLogger`, to log the uncaught exceptions;
 - `ZapBootstrap`, to disable the std out log when specified.

Remove usage of Log4j 1 configuration in `DaemonBootstrap` and
`GuiBootstrap`, no longer needed and would have no effect.
Change `CommandLineBootstrap` to disable the std out log if not already.

Deprecate `ZapOutputWriter`, no longer used nor functional.
Remove `ZAP#JERICHO_LOGGER_PROVIDER`, the logger provider is no longer
needed, Jericho includes a provider for Log4j 2. Exclude it from binary
compatibility checks (was not part of the API).

Close #6196.


zap/src/main/java/org/zaproxy/zap/extension/log4j/ZapOutputWriter.java
```diff
@@ -26,6 +26,7 @@
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.view.ScanStatus;
 
+@Deprecated
```
kingthorin, line 29: Should this indicate version or “see”?

psiinon (approved):

zap/src/main/java/org/zaproxy/zap/ZAP.java
```diff
@@ -188,11 +179,12 @@ public static EventBus getEventBus() {
         return eventBus;
     }
 
-    private static final class UncaughtExceptionLogger implements Thread.UncaughtExceptionHandler {
+    static final class UncaughtExceptionLogger implements Thread.UncaughtExceptionHandler {
 
-        private static final Logger logger = Logger.getLogger(UncaughtExceptionLogger.class);
+        private static final org.apache.logging.log4j.Logger logger =
```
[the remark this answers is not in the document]
thc202: There's already an import of the `Logger` class (from 1.x).

zap/src/main/java/org/zaproxy/zap/extension/log4j/ZapOutputWriter.java
thc202: This is an internal class, I guess we could remove already but just in case deprecated it instead.

kingthorin (approved):

Issue #5898, opened 2020-03-15
Allow to import from file into regular expression panels (e.g. Include in Context)


khalil2535:
**Is your feature request related to a problem? Please describe.**
I want to be able to include more than a domain/domain regex in my context faster, when i need to add the multiple domain regexes now i need to do that manually which consume long time.

**Describe the solution you'd like**
I need to be able to include add the domains i need from a txt file.

**Additional context**
maybe the best way is add a button here which allow the user to add the domain regex.
![image](https://user-images.githubusercontent.com/26520106/76695907-8968f380-668d-11ea-8150-bf455fc87c69.png)

lordoftheseas:
I would like to work on this issue :)

psiinon:
@lordoftheseas great - I've assigned it to you.

StefanTobler:
@shashank-kv any progress on this issue? I am interested in seeing your approach.

ckoidl:
I'd like to pick this one up.

PR #6236, opened 2020-10-11
#5898 added import from file button to regular expression panels

ckoidl:
Added an import button to "Include in Context" which allows adding all regex patterns of a file. There must be one regex per line.


zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
```diff
@@ -64,6 +75,56 @@ public static String getPanelName(Context context) {
...
+                return null;
+            }
+            return file;
+        }
+        return null;
+    }
+
+    private void addImportButton() {
+        String IMPORT_BUTTON_LABEL = Constant.messages.getString("context.include.import.label");
+        JButton importButton = new JButton(IMPORT_BUTTON_LABEL);
```
kingthorin, line 95: There doesn't seem to be any need for this intermediate storage and it isn't a CONSTANT.
```suggestion
        JButton importButton = new JButton(Constant.messages.getString("context.include.import.label"));
```

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
```diff
@@ -64,6 +75,56 @@ public static String getPanelName(Context context) {
...
+                                        .showWarningDialog(
+                                                Constant.messages.getString(
+                                                        "context.include.import.error",
+                                                        e.getLocalizedMessage()));
+                                return;
+                            }
+                            regexesPanel.setRegexes(regexes);
+                        }
+                    }
+                });
```
kingthorin, line 124: The ActionListener creation could be extracted to a new private method (`getImportActionListener` or something like that, create it if null, just return it if not). Then, the addition of the listener could be done with a lambda `importButton.addActionListener(e -> getImportActionListener());`.

Or something like that, hopefully that gives you enough to work with. Feel free to ask questions if I haven't been clear enough.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
```diff
@@ -64,6 +75,56 @@ public static String getPanelName(Context context) {
...
+                        File file = selectImportFile();
+                        if (file != null) {
+                            List<String> regexes = regexesPanel.getRegexes();
+
+                            String line = null;
+                            try (BufferedReader br = new BufferedReader(new FileReader(file))) {
+                                while ((line = br.readLine()) != null) {
+                                    String trimmedRegex = line.trim();
+                                    if (!regexes.contains(trimmedRegex)) {
+                                        Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
```
kingthorin, line 109: Hmmmm I'm not sure about assuming case insensitive. Users can always designate `(?i)` at the start of their pattern if they wish. Unless you've checked and this behavior is assumed elsewhere in context handling.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
```diff
@@ -64,6 +75,56 @@ public static String getPanelName(Context context) {
...
+                                        regexes.add(trimmedRegex);
+                                    }
+                                }
+                            } catch (IOException | PatternSyntaxException e) {
+                                View.getSingleton()
+                                        .showWarningDialog(
+                                                Constant.messages.getString(
+                                                        "context.include.import.error",
+                                                        e.getLocalizedMessage()));
```
kingthorin, line 118: This has bitten us before. I'd suggest including two substitutions. One that is specifically the failed pattern string, and one that's the message or localizedMessage.

Edit: To be more specific, localizedMessage doesn't always exist, or can sometimes be null (I forget precisely if we hit null or simply empty in the past).

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
```diff
@@ -64,6 +75,56 @@ public static String getPanelName(Context context) {
...
+                                return;
+                            }
+                            regexesPanel.setRegexes(regexes);
```
kingthorin, line 121: Note: I haven't tested.

Won't this replace what is already specified? If the user has already set something or imported another file first.

thc202:
For consistency this could be added to `MultipleRegexesOptionsPanel`, to allow to import in all places.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: This is the same behavior as for the add button (MultipleRegexesOptionsPanel.validateFields) and I did the same validation for consistency

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: Existing regexes are loaded in :101 and new ones from file are added (excluding duplicates).

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: While I am not a huge fan of lambdas I can certainly change the code to use it there. I do not fully get where there is a place for getImportActionListener when using a lambda for the ActionListener.
I would extract the actionPerformed method to a new method and then use this as lambda, so no ActionListener object needs to be created. As far as I understand that is the essence of using lambdas.
Could look like this:
```
public void importButtonClicked(ActionEvent event) {
    File file = selectImportFile();
    if (file != null) {
        List<String> regexes = regexesPanel.getRegexes();

        String line = null;
        try (BufferedReader br = new BufferedReader(new FileReader(file))) {
            while ((line = br.readLine()) != null) {
                String trimmedRegex = line.trim();
                if (!regexes.contains(trimmedRegex)) {
                    Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
                    regexes.add(trimmedRegex);
                }
            }
        } catch (IOException | PatternSyntaxException e) {
            View.getSingleton()
                    .showWarningDialog(
                            Constant.messages.getString(
                                    "context.include.import.error",
                                    e.getLocalizedMessage()));
            return;
        }
        regexesPanel.setRegexes(regexes);
    }
}


private void addImportButton() {
    JButton importButton = new JButton(Constant.messages.getString("context.include.import.label"));
    importButton.addActionListener(e -> importButtonClicked(e));
    regexesPanel.addButton(importButton);
}
```

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: Right, changed.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: Thanks for the hint. Changed this to get the error detail (e.getLocalizedMessage) separately. Checking for null before passing it to string replacement to avoid failure. If the message is empty it is not a big deal, in this case just the main error message is displayed without any detail.

ckoidl:
@thc202 I also thought of this, but was not sure if it is desired to have it everywhere. Is it something we want to have? If yes, then I will move it.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
kingthorin: Ya that’s what I was envisioning even if I wasn’t articulating it well 🤪

thc202:
I think we should, the button is not intrusive and the same use case applies to other panels. The next step would be to add an export button, but that's another issue.

zap/src/main/java/org/zaproxy/zap/view/ContextIncludePanel.java
ckoidl: change to lambda (and moved all changes to MultipleRegexesOptionsPanel as per other discussion)

ckoidl:
moved adding of import button to MultipleRegexesOptionsPanel

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
 
         this.owner = owner;
         getTable().setSortOrder(0, SortOrder.ASCENDING);
+        JButton importButton =
+                new JButton(Constant.messages.getString("context.include.import.label"));
+        importButton.addActionListener(e -> importButtonClicked(e));
+        addButton(importButton);
+    }
+
+    public void importButtonClicked(ActionEvent event) {
```
thc202, line 92: Should be at most `protected`, the argument could be removed.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+        File file = selectImportFile();
+        if (file != null) {
+            List<String> regexes = getRegexes();
```
thc202, line 95: This is not a copy, this will lead to partial imports in case of failure (which causes other issues as the changes to the model are not notified to the view).

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+
+            String line = null;
```
thc202, line 97: Can be declared later.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+            try (BufferedReader br = new BufferedReader(new FileReader(file))) {
```
thc202, line 98: Should use a specific charset (e.g. UTF-8).

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+                while ((line = br.readLine()) != null) {
+                    String trimmedRegex = line.trim();
+                    if (!regexes.contains(trimmedRegex)) {
+                        Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
```
thc202, line 102: This should be extracted to use when importing and adding.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+                                        "context.include.import.error", e.getLocalizedMessage()));
+                return;
+            }
+            setRegexes(regexes);
+        }
+    }
+
+    private File selectImportFile() {
+        JFileChooser chooser = new JFileChooser(Constant.getContextsDir());
+        File file = null;
```
thc202, line 119: Can be declared later.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+
+        int rc = chooser.showOpenDialog(View.getSingleton().getMainFrame());
```
thc202, line 121: Should use `owner` as parent.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +83,50 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+                                Constant.messages.getString(
+                                        "context.include.import.error", e.getLocalizedMessage()));
+                return;
+            }
+            setRegexes(regexes);
+        }
+    }
+
+    private File selectImportFile() {
+        JFileChooser chooser = new JFileChooser(Constant.getContextsDir());
```
thc202, line 118: Not sure the contexts dir is the most appropriate, that's used for the whole contexts. Don't specify any?

thc202:
Could you update the PR title and commit message to indicate that this is being added to all panels not just the include in context?

thc202:
Tweaked the title of the issue to be generic.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: If no charset is given, then the system default is used which should be the most appropriate one.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: changed to private

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: created local copy to allow duplication checking

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: done

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: Extracted duplicate checking and valid pattern checking. Using extracted function for adding and importing of regexes.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: done

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: changed parnet to `owner`

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: Removed path.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: On a second thought, while the system default might be the best guess for the file encoding it would still be undefined behavior. So I changed it to use UTF-8.

ckoidl: FileReader with charset is not availble in java 8, but java 8 build is required by the autobuild. I could go with InputStreamReader which is available in both, Java 8 and Java 11, but I checked the zap code for other places where files are read while I found a couple of places where FileReader is used without giving a charset, there was no InputStreamReader solution, so I decided to revert the changes and at least be consistent with th other code.

thc202: Can use NIO, e.g. `Files.newBufferedReader(file.toPath(), StandardCharsets.UTF_8)`.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +82,67 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+                new JButton(Constant.messages.getString("context.include.import.label"));
+        importButton.addActionListener(e -> importButtonClicked());
+        addButton(importButton);
+    }
+
+    /**
+     * Checks if {@code regex} is contained in {@code regexes} and if it is a valid regex pattern.
+     * If {@code regexes} is null, then the duplication check is omitted
+     *
+     * @param regex A regex as string
```
thc202, line 95: `regex` → `trimmedRegex`, or rename the parameter to `regex` (which has more occurrences).

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -76,6 +82,67 @@ public MultipleRegexesOptionsPanel(Dialog owner, RegexesTableModel model) {
...
+    private void importButtonClicked() {
+        File file = selectImportFile();
+        if (file != null) {
+            List<String> regexes = new ArrayList<String>(getRegexes());
+
+            try (BufferedReader br = new BufferedReader(new FileReader(file))) {
+                String line = null;
+                while ((line = br.readLine()) != null) {
+                    String trimmedLine = line.trim();
+                    if (MultipleRegexesOptionsPanel.validateRegexAndCheckForDuplicates(
```
thc202, line 120: - The call does not need to be qualified with the class name.
 - It would be better to add the braces to the if statement, for consistency.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
```diff
@@ -224,17 +291,19 @@ protected void init() {
...
-                        JOptionPane.INFORMATION_MESSAGE);
-                getRegexTextField().requestFocusInWindow();
-                return false;
-            }
             try {
-                Pattern.compile(trimmedRegex, Pattern.CASE_INSENSITIVE);
+                if (MultipleRegexesOptionsPanel.validateRegexAndCheckForDuplicates(
+                        trimmedRegex, regexes)) {
+                    return true;
+                } else {
```
thc202, line 298: The else is not necessary.

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: Isn't it best practice to call static methods in a static way, i.e. by calling it via the class name?
Added braces to if

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: removed else

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: fixed documentation

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
ckoidl: changed to NIO

zap/src/main/java/org/zaproxy/zap/view/MultipleRegexesOptionsPanel.java
thc202: I'm not aware of a best practice that requires qualifying the static method calls in the same class (they are called statically regardless).

ckoidl: You are right, in the same class it is not required, Changed it.

thc202: The call in `validateFields` can be changed as well, same reason.

ckoidl: done

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

thc202:
@ckoidl how would you like to be credited (e.g. name, handle)?

ckoidl:
Please use my name Christian Koidl for credits.

PR #6244, opened 2020-10-15
Add GraphQL to Docker Packaged API Scan

ricekot:
Requires zaproxy/zap-extensions#2573 for synchronous parsing when the API is used.

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-11-16
+ - Update zap-api-scan.py to add support for GraphQL.
+
 ### 2020-10-13
  - Alert_on_Unexpected_Content_Types.js > Added Content-Type application/health+json to the list of expected types.
  
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -17,8 +17,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This script runs a full scan against an API defined by OpenAPI/Swagger or SOAP
-# using ZAP
+# This script runs a full scan against an API defined by OpenAPI/Swagger, SOAP
+# or GraphQL using ZAP.
 #
 # It can either be run 'standalone', in which case depends on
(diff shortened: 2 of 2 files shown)


docker/zap-api-scan.py
```diff
@@ -228,14 +238,14 @@ def main(argv):
...
-        logging.warning('Format must be either \'openapi\' or \'soap\'')
+    if format != 'openapi' and format != 'soap' and format != 'graphql':
+        logging.warning('Format must be either \'openapi\', \'soap\', or \'graphql\'')
         usage()
         sys.exit(3)
 
     if running_in_docker():
         base_dir = '/zap/wrk/'
-        if config_file or generate or report_html or report_xml or report_json or report_md or progress_file or context_file or target_file:
+        if config_file or generate or report_html or report_xml or report_json or report_md or progress_file or context_file or target_file or schema_file:
```
ricekot, line 248: "target_file" and "schema_file" will always be empty at this point in the program. Should they be removed?

psiinon: Just the fact that they have been specified means the user will need to mount a local dir, so this is good.

psiinon:
This will work with the weekly as that includes graphql. To work with the stable release we'll either need to include graphql with that or explicitly install the add-on in this script just in case.
Not a blocker for merging, just a reminder for us all ;)

docker/zap-api-scan.py
thc202: They are not being specified though, they are initialised with the actual values later. Not sure if best Python practice but declaring them later would be more clear.

ricekot:
Updated:
- Use schema only when format is graphql
- Only allow URLs in target when format is graphql
- Declare target_file and schema_file later

docker/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-10-16
```
thc202, line 4: Could you update the date?

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6255, opened 2020-10-21
[WIP] Authenticated scan

Allan-Trindad:
This is my contribution with new files and a new way to get scan authenticated and using new headers.
To have a better understanding of what I've done, I created one blog post explaining all the modifications and reasons.
https://allan-trindad.github.io/blog/2020/10/18/OwaspZap.html

I hope it can be useful.


kingthorin:
Thanks for your contribution.

The LGTM issue and build failures will also need to be addressed.

psiinon:
Oh, very interesting :)
FYI I was going to look at something similar but in a slightly different way: https://groups.google.com/g/zaproxy-users/c/KxtsdtvYQpk/m/Npe98aXIAgAJ
In this case the user just needs to put the auth token in a specific env var for it to be used on all requests.
I'm definitely not saying thats a better approach, just that we need to think about this and hopefully agree the best strategy.
I'd much rather we have a coherant plan rather than just bolting things on. Which is not an implied criticism btw!
I think this is worth a User Group post...

psiinon:
Initial thoughts: 
I dont think we should be including example configs (like Juice Shop or Pet Store) in the ZAP docker images - theres potentially no end to the examples we could include (and would have to maintain). If someones wants to extend our docker images to do so then great :)
I dont think the "DSV, HML, PRD" terminology is very common or easily understood. This looks like its too specific to the authors use case to be generically useful.

Allan-Trindad:
@psiinon thank you for your comments. Of course I'm not going to be mad with your comments.. We are here to improve the tool.

I definitely agree with not importing the swagger, but I didn't know another way to show easily the usage and the main ideia of my modifications.

Allan-Trindad:
@psiinon looks like I don't have permission to post in the google group... =/

psiinon:
Thats weird - the group should be open to everyone.
Have you signed up for it?
FYI I posted to the group here: https://groups.google.com/g/zaproxy-users/c/TS4t11yYvcE/m/v5NCI8H_AgAJ
No replies yet :/

thc202:
Changed to WIP, pending further discussion about what should be provided for authentication.

thc202:
Closing, no activity/feedback in 2 months.

The packaged scans now allow to specify a user (defined in the context) [1] and ZAP allows to read auth headers form env vars. [2]

[1] https://github.com/zaproxy/zaproxy/blob/main/docker/CHANGELOG.md#2020-11-24
[2] https://www.zaproxy.org/docs/desktop/start/features/authentication/#envvars

PR #6276, opened 2020-11-02
Check if Log4j legacy conf backup already exists

thc202:
Avoid logging an error when the backup already exists, which would
happen after running older and newer ZAP versions using the same home.


kingthorin (approved):
Thanks

ricekot (approved):
Thanks!

PR #6278, opened 2020-11-03
Add Alert reference field

psiinon:
Add new alertRef field to alerts.
This is to allow us to tell the difference between different alerts raised by the same rule. By default is will be the plugin id - rules will need to be changed to append a qualifier for each of the alerts they raise.


zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -921,6 +942,49 @@ public void setSource(Source source) {
         this.source = source;
     }
 
+    /**
+     * Gets the alert reference.
+     *
+     * <p>This is a unique identifier for the type of alert raised. A scan rule may raise more that
+     * one type of alert but they should all have different alert references.
```
kingthorin, line 949:
```suggestion
     * one type of alert and they should all have different alert references.
```

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -474,6 +474,7 @@ private void processAlerts(
         Map<String, String> map = new HashMap<>();
         map.put("id", String.valueOf(alert.getAlertId()));
         map.put("pluginId", String.valueOf(alert.getPluginId()));
+        map.put("alertRef", alert.getAlertRef());
```
kingthorin, line 477: Should we also update the core reporting? (Or separate issue/PR?)

psiinon: Doh, forgot about those, will add to this PR...

psiinon: Ha! No need - they now call the alert API endpoints so they dont need to change.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
psiinon: Done.

thc202 (approved):

thc202:
Thank you!

kingthorin:
You decided not to tackle core reports?

psiinon:
You mean the HTML report? Thats really aimed at people and doesnt currently include the plugin id, so yes, wasnt planning on changing that...

kingthorin (approved):

PR #6282, opened 2020-11-05
Allow to instantiate AbstractFrame

thc202:
Do not require to extend the class to use it, it's not needed.


kingthorin (approved):
Seems fine to me. Will that cause any problems for extensions?

thc202:
It's a compatible change, it might cause problems if something is using reflection and checking if it's abstract but unlikely.

ricekot (approved):
Thanks!

thc202:
Actually, this does cause problems with JavaScript scripts that are extending it... ironically this change is to allow them to be used with GraalJS engine.

kingthorin:
So revert this or modify the scripts? Or ______ ?

thc202:
Modify the scripts, which was already the plan (though I was not expecting that to break in the meantime).

PR #6284, opened 2020-11-09
Address exceptions related to parameter parsing

thc202:
Do not allow null values as the older method didn't allow them, which
was causing exceptions in scan rules, also, catch exception thrown when
decoding name/value, for same reason.
Provide a method that does not decode and allows null values.

### DIFF
--- a/zap/src/test/java/org/zaproxy/zap/model/StandardParameterParserUnitTest.java
+++ b/zap/src/test/java/org/zaproxy/zap/model/StandardParameterParserUnitTest.java
@@ -20,9 +20,11 @@
 package org.zaproxy.zap.model;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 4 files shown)


psiinon (approved):

zap/src/test/java/org/zaproxy/zap/model/StandardParameterParserUnitTest.java
```diff
@@ -78,6 +80,114 @@ public void defaultParser() {
...
+        // Then
+        assertThat(parameters, hasSize(2));
+        assertThat(parameters.get(0).getName(), is(equalTo("%x")));
+        assertThat(parameters.get(0).getValue(), is(equalTo("1")));
+        assertThat(parameters.get(1).getName(), is(equalTo("b%25")));
+        assertThat(parameters.get(1).getValue(), is(equalTo("%20")));
+    }
+
+    @Test
+    void shouldHaveEmptyNamesForMissinNamesWhenRawParsing() {
```
ricekot, line 168:
```suggestion
    void shouldHaveEmptyNamesForMissingNamesWhenRawParsing() {
```

kingthorin (commented):
Looks good, just that one typo.

thc202:
Addressed.

kingthorin (approved):
Thanks

PR #6293, opened 2020-11-12
Auth multi-user fix and new API endpoints

psiinon:
The current polling implementation did not handle multiple users polling at the same time. This PR should fix that.

It also adds more API endpoints to make the auth process less opaque and
more controllable.

Still TODO:

- [x] Add test for multi user polling
- [x] new user.getAuthenticatedSession API endpioint


psiinon:
Also looking at setting cookies via the API, but that could be another PR if this gets approved and merged first :)

psiinon:
Have added a setCookie API endpoint - no more changes planned :)

kingthorin:
Wow that's a lot to absorb. I'll try to go through it later today. Not that I disagree with any of it but what's the envisioned use case or model?

psiinon:
The problem with the current implementation is that the poll state (time of last poll, number of poll requests) was global.
This is fine if you are doing one scan at a time but doesnt work if you're running concurrent scans with different users, or are running a scan and doing some manual testing with another user.
The changes to fix this arent actually that big, just moving the state into the User object.
Not sure how likely this is but its a relatively small change and tbh should have been done that way the first time.

The remaining changes are adding new end points - some of which will allow you to make authentication calls and see the requests and responses - this will be _really_ useful for debugging and products that use the ZAP API.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -3160,34 +3160,61 @@ uiutils.desc = Core UI related functionality.
...
-users.api.action.setAuthenticationCredentials.param.contextId = 
-users.api.action.setAuthenticationCredentials.param.userId = 
+users.api.action.setAuthenticationCredentials.param.contextId = The Context Id
+users.api.action.setAuthenticationCredentials.param.userId = The User Id
 users.api.action.setAuthenticationCredentials.param.authCredentialsConfigParams = 
+users.api.action.setAuthenticationState = Sets fields in the authentication state for the user identified by the Context and User Ids.
+users.api.action.setAuthenticationState.param.contextId = The Context Id
+users.api.action.setAuthenticationState.param.userId = The User Id
+users.api.action.setAuthenticationState.param.lastPollResult = Last Poll Result - optional, should be 'true' or 'false'.
+users.api.action.setAuthenticationState.param.lastPollTimeInMs = Last Poll Time in Milliseconds - optional, should be a long on 'NOW' for the current time in Ms.
```
thc202, line 3176: - `on` → `or`?
 - `Ms` → `ms`?

psiinon:
Hopefully fixed all of those and the conflict.

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):

PR #6313, opened 2020-11-20
Manual Request response send

psiinon:
Add a Send button to the manual req editor response panel in tab mode and set the parent for the warning dialogs - otherwise on some OSs these are shown against the main tab and the dialog gets hidden.
Also changed some of the logging - I dont think user gui errors need to be logged...


thc202 (approved):

thc202:
Thanks!

kingthorin (approved):

PR #6322, opened 2020-11-25
Log params at Debug Level

dcuenot:
Hello,

From my point of view, `Params` should be logged at debug level.
Especially because in params, you could have Bearer token (aka sensitive data), and these kind of data should not be logged at info level.

Thanks in advance for your feedback.
Regards

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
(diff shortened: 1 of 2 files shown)


thc202:
I'm fine with the change. Could you update the changelog?

dcuenot:
Changelog updated.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #6327, opened 2020-11-26
Stop using Log4j 1.x where possible

thc202:
Use Log4j 2 classes for logging.
Deprecate public APIs using Log4j 1.x classes.
Remove `ZapOutputWriter` as it's no longer needed (internal and Log4j
1.x specific class).
Suppress breaking changes done to internal classes.


kingthorin (approved):
Thanks! :zap:

psiinon (approved):

PR #6328, opened 2020-11-27
Tweak panel search and find dialog to be friendly w/ dark LaFs

kingthorin:
- HighlighterUtils > Add public getHightlightColor() method which uses DisplayUtils.getHighLightColor() when a dark LaF is in use. (Currently this means a dark orange is used).
  - Have other usages of the public constant highlight color use this method.
- TabbedPaneSearch > Prevent IndexOutOfBoundsException by not adding tab's for which the index is -1 (hidden/disabled).
- FindDialog (CTRL + F) > Set the selection color based on DisplayUtils.getHighlightColor() before selecting the matched string.

Part of zaproxy/zaproxy#5542


kingthorin:
Note: Currently draft, there are other upcoming core highlight changes which this may need to adapt to.

thc202 (approved):

thc202:
Thanks!

psiinon (approved):

PR #6331, opened 2020-11-30
Remove PacScript

thc202:
Remove `PacScript` and related resources/dependencies, it will be moved
to Graal JS add-on (per Nashorn deprecation and removal in newer Java
versions).

Related to #265 and #4851.


psiinon (approved):

thc202:
Change breaks in Java 8.

thc202:
Difference in how Nashorn converts the values in Java 8 and 11, the latter uses integer for integer values while the former uses float values.

This was to reduce the changes done to `PacScript` when moving to GraalJS add-on (which works the same in Java 8 and 11), will just remove it and do the required changes while moving...

kingthorin (approved):

thc202:
Changed to remove the class and related files. WIP until it's in the add-on.

thc202:
Removed WIP.

kingthorin (approved):

PR #6337, opened 2020-12-02
fix/Dockerfile: Use ARG directive instead of ENV

natebwangsut:
Since `DEBIAN_FRONTEND` is only used to do `apt-get` installations, use `ARG` command instead of `ENV` so that the parameter does not persist after the build process has been completed.

This could be helpful for those whos running the zap with `docker run -i` as it implies running container interactively.

Signed-off-by: natebwangsut <nate.bwangsut@gmail.com>

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2020-12-02
+ - Use `ARG` command (for `DEBIAN_FRONTEND`) instead of `ENV` so that the parameter does not persist after the build process has been completed.
+
 ### 2020-11-27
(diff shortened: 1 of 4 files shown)


psiinon (approved):
Seems reasonable to me

thc202:

natebwangsut:
I've updated the `CHANGELOG.md` per requested

thc202:
Please revert the formatting changes and squash.

natebwangsut:
Done.

kingthorin:
The formatting (white space) changes don't appear to have been reverted.

natebwangsut:
Surely I removed the **entire** formatting thing

kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thanks!

Issue #5616, opened 2019-10-07
Nested Data Driven Nodes


crunchcat:
**Describe the bug**
Cannot create nested data-driven nodes. When attempted, ZAP creates duplicate entries in the tree for each entry in the Structural Modifiers table.

**To Reproduce**
1. Visit a page with nested data-driven nodes to populate the tree in the sidebar.
2. Right click on parent data-driven node > Flag as Context > ... Data-Driven Node. Confirm.
3. Right click on child data-driven node > Flag as Context > ... Data-Driven Node. Confirm.
4. Notice the tree for each node has been duplicated into two trees containing one data-driven node each.
    For the tree containing the parent data-driven node, requests directly to the child DDN will each be listed as separate endpoints and requests to children of the child DDN are missing. The tree containing the child data-driven node is duplicated for each value of the parent DDN but requests to it's children are displayed properly. 

**Expected behavior**
Data driven nodes should be aware of each other and able to be nested. The site tree should be updated to properly display each of the data-driven nodes without duplicate entries for each value of either node.

**Screenshots**
I can't easily attach screenshots but here's a representation of the tree sidebar after nested DDNs are configured for the context. In this example, the app returns hostnames and uptime for multiple user devices, DDN1 is a user ID and DDN2 is a device ID.

- Context
    - Example.com
- Sites
    - https://example.com
        - users
            - \<\<DDN1\>\>
                - devices
                    - GET:device1
                    - GET:device2
            - user1
                - devices
                    - \<\<DDN2\>\>
                        - GET:hostname
                        - GET:uptime
            - user2
                - devices
                    - \<\<DDN2\>\>
                        - GET:hostname
                        - GET:uptime

An example URL would be `https://example.com/users/user1/devices/device1/hostname`

**Software versions**
 - ZAP: 2.8.0
 - Add-on: `[[id=accessControl, version=5.0.0],
[id=alertFilters, version=9.0.0], [id=ascanrules,
version=33.0.0], [id=bruteforce, version=8.0.0], [id=diff,
version=9.0.0], [id=directorylistv1, version=3.0.0],
[id=directorylistv2_3, version=3.0.0], [id=fuzz,
version=11.0.0], [id=gettingStarted, version=10.0.0],
[id=help, version=9.0.0], [id=hud, version=0.7.0],
[id=importurls, version=6.0.0], [id=invoke, version=9.0.0],
[id=jxbrowser, version=14.0.0], [id=jxbrowsermacos,
[...]

PR #6341, opened 2020-12-03
Refactor DataDrivenNode functionality

Jman420:
The goal of this PR is to provide support for multiple data driven nodes in a URI such that URIs such as http://example.org/path/to/ddn/123/another/ddn/987/ and http://example.org/path/to/ddn/456/another/ddn/789/ can be replaced with http://example.org/path/to/ddn/�node1�/another/ddn/�node2�/ to treat both URIs as a single endpoint.

Changes to accomplish goal:

Separate concepts of StructuralParameters and DataDrivenNodes
Separate Context UI Panels for StructuralParameters and DataDrivenNodes
Introduce a Tree Structure for representing DataDrivenNodes
Update StandardParameterParser logic to work with DataDrivenNodes Tree Structure

Supporting Help Docs PR : https://github.com/zaproxy/zap-core-help/pull/334


kingthorin:
Related: https://github.com/zaproxy/zaproxy/issues/5616

Jman420:
@zaproxy/team-zaproxy Any comments or suggestions on these changes?

I'm not sure how to fix the japicmp build errors... I did indeed remove a bunch of code files due to my refactoring, so I would expect there to be differences between the two builds.  Any guidance there would be helpful.

Thanks!

thc202:
The breaking changes need to be suppressed in `zap/zap.gradle.kts`, under the `japicmp` task. Run the task to generate the report and know what needs to be suppressed. (Better rebase first on the `main` branch, the file was changed.)

I'd say that this is breaking too much though, add-ons might be using some of the methods/classes being removed. Ideally the old classes/methods should be deprecated first and removed only after some releases.

Some loose comments:
 - Needs tests;
 - Commit History should be tidied up, hard to follow.

Jman420:
@thc202 Ok, I can bring the classes I removed back... I assume that would include the UI classes I removed too, although that will add some complexity to the UI since there will be two places that Data Driven Nodes & Structural Parameters are handled.

How does the team go about marking classes as deprecated?  Is there an attribute to use?

I'll take a stab at cleaning up the commit history... I'm not very familiar with rewriting commit histories.

thc202:
They don't need to be used just don't remove them, you can add the annotation `@Deprecated` and the JavaDoc tag `@deprecated` (e.g. `HttpMessageList` class).

Jman420:
> They don't need to be used just don't remove them, you can add the annotation `@Deprecated` and the JavaDoc tag `@deprecated` (e.g. `HttpMessageList` class).

Got ya.  Thanks!  Took a bit of a break from this over the holidays.  Plan on diving back in again next week.

Jman420:
Moving PR to main branch and cleaning up commit history

thc202:
That can be done in this pull request.

PR #6348, opened 2020-12-07
Do not update Content-Length in byte models

thc202:
The Content-Length is now updated optionally based on the UI option.
Extract method that finds the header limit and correct line feed case.


psiinon (approved):

kingthorin (approved):
Thanks

PR #6350, opened 2020-12-07
Add node for URLs ending with a slash

### LINKED ISSUE #6297: Site tree doesnt correctly represent URLs ending in a /
**Describe the bug**
In the History tab I added a breakpoint for a particular URL. In the Add Breakpoint window, the prefilled String does not match the URL.

**To Reproduce**
1. Run Python's HTTP server via "python3 -m http.server"
2. In Firefox browse to http://127.0.0.1:8000/search/?q=test (The page will give a 404 error, but that doesn't matter.)
3. Go to the History tab in ZAP, select this request and right-click on "Break...". Now it looks like this:

add_breakpoint

Now this breakpoint will not that work. The correct String should be `http://192.168.56.1:8000/search/\?q\=test` instead of `http://192.168.56.1:8000/\?q\=test`.

**Software versions**
 - ZAP: 2.9.0
 - Arch Linux
 - Browser: Firefox 82.0.3

psiinon:
Fix #6297

This allows us to differentiate between https://www.example.com/a and https://www.example.com/a/

The main impact is to the active scanner as this works off the sites tree - if you select the new "/" node then it will have no children and therefore not recurse as most people will want.
So now if you specify a URL with a trailing slash then the slash will be ignored if you specify 'recurse'.

Will require a corresponding change to the Quick Start add-on.

Initially WIP as I need to add some specific tests to make sure we treat nodes with and without a slash differently.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/model/SiteNode.java
+++ b/zap/src/main/java/org/parosproxy/paros/model/SiteNode.java
@@ -319,10 +319,14 @@ public class SiteNode extends DefaultMutableTreeNode {
         } else if (this.getParent().isRoot()) {
             hierarchicNodeName = this.getNodeName();
         } else {
-            String name =
-                    this.getParent().getHierarchicNodeName(specialNodesAsRegex)
-                            + "/"
-                            + this.getCleanNodeName(specialNodesAsRegex);
+            String nodeName = this.getCleanNodeName(specialNodesAsRegex);
+            String name;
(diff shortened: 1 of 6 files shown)


psiinon:
Added another test so I think its ready to review - lmk if you can think of any others I should add...

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #6356, opened 2020-12-11
Add target to ajaxSpider.scan_as_user call

vismiktor:
Without this parameter ajaxSpider starts to crawl first included in a context URL and not a target which was set. It isn't an expected behavior. I had API URL configured in the context additionally to an app URL and ajaxSpider crawled only API URL.

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


thc202:

vismiktor:
> Could you update the changelog?

Done

thc202 (approved):

thc202:
Thanks!

kingthorin (approved):
Thanks

PR #6371, opened 2020-12-21
Adding support for handling manipulation of multiple NameValuePairs

preetkaran20:


zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppVariantPlugin.java
```diff
@@ -0,0 +1,76 @@
...
+import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.ascan.VariantFactory;
+
+/**
+ * {@code AbstractAppVariantPlugin} is the abstract base class which is used to run per variant to
+ * modify multiple name value pairs of the {@code HttpMessage} per variant.
+ *
+ * @author KSASAN preetkaran20@gmail.com
+ */
+public abstract class AbstractAppVariantPlugin extends AbstractAppPlugin {
```
preetkaran20, line 16: @thc202 @kingthorin was planning to have another base class for usecases where scanrule wants to modify multiple name value pairs which is a use case for FileUpload scan rule. WDYT ?

kingthorin: Wouldn’t file upload use the multipart variant which can already handle manipulation of various elements? (File name, disposition, content type)

kingthorin:
This also seems to be a binary incompatible change.

Why not introduce a new variant in zap-extensions common lib?

kingthorin:

preetkaran20:
> To address the DCO requirement you'll need to sign-off the commit(s):
> 
> * https://github.com/zaproxy/zaproxy/blob/develop/CONTRIBUTING.md#developer-certificate-of-origin
> * https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff

yeah yeah, this i know 👍

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppVariantPlugin.java
preetkaran20: No no, Variant handles the manipulation but AbstractAppParamPlugin calls the scan method for one field at a time hence this new class gives the access to entire variant so that child can manipulate fields as per there need.

kingthorin: Ah okay, I see what you're after.

preetkaran20:
> Why not introduce a new variant in zap-extensions common lib?

Actually this new class can be parent of AbstractAppParamPlugin and it looks very logical. Also same code will get copied at different places if we add in common lib.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppVariantPlugin.java
thc202: Why not change `AbstractAppParamPlugin` to just expose all parameters if that's what's needed?
e.g. `scan(HttpMessage, List<NameValuePair>)`
Those interested in all parameters just need to override the method.

Worth noting that variants do not support changing multiple parameters at the same time.

preetkaran20: @thc202 yeah that also can be done but i think it is ok to expose Variant because there are some extra methods/utilities present in variant which can be utilised  like setparameter/setEscapeParameter etc methods  else we need to expose them from parent.

Oh variants don't support changing multiple parameters ? because of start index and end index of value getting changed after setparameter is invoked ?
How should be proceed on this ? shall we re-initialise the variant with new message after setting paramter ?

thc202: The `NameValuePair`s are already being exposed through `scan(HttpMessage, NameValuePair)` (and there are scan rules that make use of it), the only difference is that we would be exposing all of them at once. Those methods would not be irrelevant, they are used by scan rules that scan just one parameter at the same time.

The variants don't keep track of what changes were already done to a given message, the parameters would have to be set at the same time for properly construct the payload.

preetkaran20: no no, my point is that for the method that will expose list of name value pairs how are we going to call `setParameter/setEscapedParameter` methods?  we need to create 2 new methods which takes original nameValuePair as input and hence for someone who is implementing abstractAppParamPlugin it is tough to understand which method to invoke. I think new class creation is a clearly demarkation between the 2 usecases.

-----

```java
    private String setParameter(HttpMessage msg, NameValuePair originalPair, String value) {

        StringBuilder newBodyBuilder = new StringBuilder(msg.getRequestBody().toString());
        int idx = originalPair.getPosition() - 1;
        MultipartFormParameter mpPart = multiPartParams.get(idx);
        if (LOGGER.isDebugEnabled()) {
            LOGGER.debug(
                    "i: "
                            + idx
                            + " pos: "
                            + originalPair.getPosition()
                            + " S: "
                            + mpPart.getStart()
                            + " E: "
                            + mpPart.getEnd());
        }
        newBodyBuilder.replace(mpPart.getStart(), mpPart.getEnd(), value);

        String newBody = newBodyBuilder.toString();
        msg.getRequestBody().setBody(newBody);
        return newBody;
    }
```
What i was refering is that, the above method takes the start and end index from the already set HttpMessage in variant and if we set one value in the httpmessage, start and end index will not point to correct locations if we add another value to the httpmessage and hence it will break the httpmessage.

thc202: I'm not sure it is, e.g. a list allows to add one element or multiple at the same time. If you have multiple parameters you call the method that accepts multiple, if you have one parameter you call the method that accepts one.

Yes, that's what I said in the previous comment, variants don't currently keep track of what changes were done to a given message, the parameters need to be set at the same time.

preetkaran20: So my point was that:
AbstractAppParamPlugin currently has following methods
```java
    protected String setParameter(HttpMessage message, String param, String value) {
        return variant.setParameter(message, originalPair, param, value);
    }

    protected String setEscapedParameter(HttpMessage message, String param, String value) {
        return variant.setEscapedParameter(message, originalPair, param, value);
    }
```

Now we need to add 2 new methods:
```java
 protected String setParameter(HttpMessage message, String param, String value) {
        return variant.setParameter(message, originalPair, param, value);
    }

    protected String setEscapedParameter(HttpMessage message, String param, String value) {
        return variant.setEscapedParameter(message, originalPair, param, value);
    }
 protected String setParameter(HttpMessage message, String param, String value, NameValuePair nameValuePair) {
        return variant.setParameter(message, nameValuePair, param, value);
    }

    protected String setEscapedParameter(HttpMessage message, String param, String value, NameValuePair nameValuePair) {
        return variant.setEscapedParameter(message, nameValuePair, param, value);
    }
```

I was thinking there is some difficulty in understanding and implementation where if child has called the original methods then it will cause some issues.

why to cause all this confusion.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -55,10 +55,11 @@
     // insensitive & DOTALL, and hit "test")
 
     private List<NameValuePair> params = Collections.emptyList();
-    private List<MultipartFormParameter> multiPartParams = new ArrayList<MultipartFormParameter>();
+    private List<MultipartFormParameter> multiPartParams;
 
     @Override
     public void setMessage(HttpMessage msg) {
+        multiPartParams = new ArrayList<MultipartFormParameter>();
```
preetkaran20, line 62: @thc202  @kingthorin my initial thought is to call setMessage again after setting the message for parsing the message again. This can be done from the setParameter api or by exposing a new api which can handle multiple updates and hence the new Api will not impact the performance of existing setparameter api.
WDYT ?

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppVariantPlugin.java
thc202: The proposed methods do not support changing multiple parameters. Probably better to discuss how the variant will support changing multiple parameters instead, that's what actually will define how the plugin will expose that.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: It's not necessary to re-parse the message to be able to set multiple parameters.

preetkaran20: sure will think more on this.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -235,6 +236,15 @@ private String setParameter(HttpMessage msg, NameValuePair originalPair, String
                             + mpPart.getEnd());
         }
         newBodyBuilder.replace(mpPart.getStart(), mpPart.getEnd(), value);
+        int offset = value.length() - mpPart.getEnd() + mpPart.getStart() - 1;
```
preetkaran20, line 239: need to validate... not corect loginc if value is 0 length

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -235,6 +236,15 @@ private String setParameter(HttpMessage msg, NameValuePair originalPair, String
                             + mpPart.getEnd());
         }
         newBodyBuilder.replace(mpPart.getStart(), mpPart.getEnd(), value);
+        int offset = value.length() - mpPart.getEnd() + mpPart.getStart();
```
preetkaran20, line 239: need to write unittest cases

thc202:
The state of the `NameValuePair` should remain the same between `setParameter` calls, otherwise that would cause unexpected injections when setting the parameters in following/other messages.

preetkaran20:
> The state of the `NameValuePair` should remain the same between `setParameter` calls, otherwise that would cause unexpected injections when setting the parameters in following/other messages.

oh ok, so what should be the solution ?  shall we return deep cloned value from `getParamList()` method of variant ? or shall we make `NameValuePair` as immutable ?  or it is fine to go ahead with a note over javadoc ?

preetkaran20:
@thc202 @kingthorin can you guys please review this PR ?

thanks,
Karan

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppVariantPlugin.java
preetkaran20: > There's still no proposal on how to change the parameters at the same time [#6371 (comment)](https://github.com/zaproxy/zaproxy/pull/6371#discussion_r555697929).

@thc202  Now this PR handles changing multiple parameters. Please look at PR: https://github.com/zaproxy/zaproxy/pull/6371/files#diff-9dfdfe7aed90f3b927955bcfb81a1c9ae34e745c7b5b4e04d79b5b0130cfb12aR241

preetkaran20:
> There's still no proposal on how to change the parameters at the same time [#6371 (comment)](https://github.com/zaproxy/zaproxy/pull/6371#discussion_r555697929).

Now this PR handles changing multiple parameters. Please look at: https://github.com/zaproxy/zaproxy/pull/6371/files#diff-9dfdfe7aed90f3b927955bcfb81a1c9ae34e745c7b5b4e04d79b5b0130cfb12aR241

thc202:
Not at the same time, causing the issues mentioned in https://github.com/zaproxy/zaproxy/pull/6371#issuecomment-787463798

preetkaran20:
> Not at the same time, causing the issues mentioned in [#6371 (comment)](https://github.com/zaproxy/zaproxy/pull/6371#issuecomment-787463798)

We are only using the position from name value pair and in case name value pair is modified it (position is modified) other implementations of AppPlugin will get wrong information and they will get impacted. Is this you are pointed to ?
if so
I think the simple way is we can return the deep cloned list of nameValuePairs from `getParamList()` method.

thc202:
The problem would happen in the same scan rule when trying to modify other message(s) (e.g. sending several payloads).

preetkaran20:
> The problem would happen in the same scan rule when trying to modify other message(s) (e.g. sending several payloads).

In same scan rule, it should be the responsibility of the implementer 

> The problem would happen in the same scan rule when trying to modify other message(s) (e.g. sending several payloads).

@thc202 are you saying that the same scan rule is sending multiple messages then if at once it changes the position variable in name value pair then it will impact subsequent messages right ? 
there are 2 ways to think the same,
1. we can add the javadoc over the method so that developer takes care of this scenario.
2. we can also add an extra layer of protection such that when the name value pair is again provided then we check in the list of name value pairs and if those contains the provided name value pair then only proceed with setParameter method.

Even a thought that comes to my mind is we are just overthinking the usecase as there are many things which can be done with the existing implementation like in case AppParamPlugin's scan method which takes nameValuePair as input can be changed then it will impact other messages/scan rules. 

Thoughts ?

thc202:
> In same scan rule, it should be the responsibility of the implementer

Of doing what? I don't think the scan rules can do anything to prevent/workaround the issue.

> ... then it will impact subsequent messages right ?

Right.

> Thoughts ?

As mentioned above, my suggestion is to allow to replace multiple parameters at the same time. It's clearer that way what's being done and it's easier to reason about the outcomes (i.e. no side effects for having set parameters before).

preetkaran20:
> > In same scan rule, it should be the responsibility of the implementer
> 
> Of doing what? I don't think the scan rules can do anything to prevent/workaround the issue.
@thc202 
I think there is some confusion, so how name value pair will change? The writer of scan rule can only update the name value pair. I am not getting the point. can you explain when and how name value pair will change ?

thc202:
It's the implementation in the PR that changes the name value pairs, when a parameter is set into the message. The scan rules can't do anything to prevent that.

preetkaran20:
> It's the implementation in the PR that changes the name value pairs, when a parameter is set into the message. The scan rules can't do anything to prevent that.

in this PR ? can you point to it ?

thc202:
It's the code you referenced earlier https://github.com/zaproxy/zaproxy/pull/6371/files#diff-9dfdfe7aed90f3b927955bcfb81a1c9ae34e745c7b5b4e04d79b5b0130cfb12aR240-R248

preetkaran20:
> It's the code you referenced earlier https://github.com/zaproxy/zaproxy/pull/6371/files#diff-9dfdfe7aed90f3b927955bcfb81a1c9ae34e745c7b5b4e04d79b5b0130cfb12aR240-R248

oh ... i think i got it. You are saying that the variant is kind of singleton and hence the state of variant is changed causing the next message to get impacted because of that. right ? I think the implementer is responsible to reset variant with the original message as being done in AppParamPlugin.
So steps can be like this:
1. Set the original message in variant
2. clone the original message
3. Modify cloned message using setParameter
4. repeat step 2.

preetkaran20:
> That's the internal representation of the name value pair (we are not talking just about multipart but all variants). The cloning of the message does not workaround the issue, shifting the positions of the name value pairs leads to different injections points (even though the message might not have been injected with a previous parameter).

I am also talking about all the variants and we are not shifting name value pairs as per my understanding. we are changing the start and end value only and that is reseted by setMessage.

preetkaran20:
> > That's the internal representation of the name value pair (we are not talking just about multipart but all variants). The cloning of the message does not workaround the issue, shifting the positions of the name value pairs leads to different injections points (even though the message might not have been injected with a previous parameter).
> 
> I am also talking about all the variants and we are not shifting name value pairs as per my understanding. we are changing the start and end value only and that is resetted by setMessage api.

are you saying that changing the multipart variant's state can impact existing implementations or something else so don't change the variant state and instead change the message directly as being done currently ?

thc202:
> So steps can be like this:

The scan rules shouldn't need to do all that, this also leads back to https://github.com/zaproxy/zaproxy/pull/6371#discussion_r555699503

It's not about being multithreaded but changing its state (which has implications on how the parameters can be set).

preetkaran20:
> > So steps can be like this:
> 
> The scan rules shouldn't need to do all that, this also leads back to [#6371 (comment)](https://github.com/zaproxy/zaproxy/pull/6371#discussion_r555699503)
> 
> It's not about being multithreaded but changing its state (which has implications on how the parameters can be set).

So suggested way is to pass multiple name value pairs into setParameter method of variant with multiple values something like

```java  public void setParameters(HttpMessage msg, List<NameValuePair> pairs, List<String> values)```
or  
```java public void setParameters(HttpMessage msg, MultiValuedMap<NameValuePair, String> pairValues)```

And pass the list of nameValuePairs to child implementation so that they can do the entire modifications and then call setParameters.

**Comment by preetkaran20** [comment]
@thc202  @kingthorin 
As out discussion here. I have created a new PR: https://github.com/zaproxy/zaproxy/pull/6590.
Please review.

**Comment by preetkaran20** [comment]
@thc202 can you please review it.

**Comment by preetkaran20** [comment]
Hi @thc202 ,

If we use the same class (AbstractAppParamPlugin) instead of a new class then AbstractAppParamPlugin will look something like this:
```java
abstract class AbstractAppParamPlugin {
 public abstract void scan(HttpMessage msg, String param, String value);

public void scan(HttpMessage msg, List<NameValuePair> nameValuePairs) {
   // This is the new method which will invoke the above method in default implementation
 // and in cases like FileUploadScanRule this method will be overridden by the child.
}

}
```

In this there is one issue which is even if the child implementation has overridden this method still they need to implement the scan method/provide empty implementation. is this fine ? or you have some other suggestion ?

thanks,
Karan

preetkaran20:
> Hi @thc202 ,
> 
> If we use the same class (AbstractAppParamPlugin) instead of a new class then AbstractAppParamPlugin will look something like this:
> 
> ```java
> abstract class AbstractAppParamPlugin {
>  public abstract void scan(HttpMessage msg, String param, String value);
> 
> public void scan(HttpMessage msg, List<NameValuePair> nameValuePairs) {
>    // This is the new method which will invoke the above method in default implementation
>  // and in cases like FileUploadScanRule this method will be overridden by the child.
> }
> 
> }
> ```
> 
> In this there is one issue which is even if the child implementation has overridden second method (With multiple NameValuePairs) still they need to implement the scan method/provide empty implementation. is this fine ? or you have some other suggestion ?
> 
> thanks,
> Karan

@thc202 any updates on the query ?

thanks,
Karan

preetkaran20:
Hi @thc202,

Any updates on the above queries ?

thanks,
Karan

thc202:
The existing method does not need to be abstract anymore.

The proposed API forces all parameters to be set un/escaped, though some of them might not be in the expected form. It also requires setting all of them even if you just want to change a couple of them.

preetkaran20:
> The existing method does not need to be abstract anymore.
> 
> The proposed API forces all parameters to be set un/escaped, though some of them might not be in the expected form. It also requires setting all of them even if you just want to change a couple of them.


ok got it.

For proposed API comment, yes it doesn't allow some of the parameters to be escaped or unescaped. I think i will create a java class along with nameValuePair, param, value we have another enum stating escaped or unescaped and let the implementer handles it. However Api doesn't require setting all the namespaces, which ever namespace/param/value is provided will set.

thc202:
Yes, I edited my comment some minutes ago to strikethrough that part.

Could you elaborate more? How is the user expected to use it?

btw, IMO the other PR can be closed, the changes can be done in this pull request.

preetkaran20:
> Yes, I edited my comment some minutes ago to strikethrough that part.
> 
> Could you elaborate more? How is the user expected to use it?
> 
> btw, IMO the other PR can be closed, the changes can be done in this pull request.

May be something like:
```java
enum Operation {
UNESCAPED,
ESCAPED
}
Class AppParameter {
private NameValuePair nameValuePair;
private String param;
private String value;
private Operation operation;
}

public void setParameters(List<AppParameter> param) {
// Implementers required to handle the value setting using Operator
}

Operation can be a boolean too.

```

preetkaran20:
> > Yes, I edited my comment some minutes ago to strikethrough that part.
> > Could you elaborate more? How is the user expected to use it?
> > btw, IMO the other PR can be closed, the changes can be done in this pull request.
> > @thc202
> > May be something like:
> 
> ```java
> enum Operation {
> UNESCAPED,
> ESCAPED
> }
> class AppParameter {
> private NameValuePair nameValuePair;
> private String param;
> private String value;
> private Operation operation;
> 
> public AppParameter(NameValuePair nameValuePair, String param, String value, Operation operation) {
> }
> 
> }
> 
> public void setParameters(List<AppParameter> param) {
> // Implementers required to handle the value setting using Operator
> }
> 
> Operation can be a boolean too.
> ```
> 
> This also will require modification in variant to accept List of AppParameter.
@thc202 
Is this what you were asking ?

preetkaran20:
> > > Yes, I edited my comment some minutes ago to strikethrough that part.
> > > Could you elaborate more? How is the user expected to use it?
> > > btw, IMO the other PR can be closed, the changes can be done in this pull request.
> > > @thc202
> > > May be something like:
> > 
> > 
> > ```java
> > enum Operation {
> > UNESCAPED,
> > ESCAPED
> > }
> > class AppParameter {
> > private NameValuePair nameValuePair;
> > private String param;
> > private String value;
> > private Operation operation;
> > 
> > public AppParameter(NameValuePair nameValuePair, String param, String value, Operation operation) {
> > }
> > 
> > }
> > 
> > public void setParameters(List<AppParameter> param) {
> > // Implementers required to handle the value setting using Operator
> > }
> > 
> > Operation can be a boolean too.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This also will require modification in variant to accept List of AppParameter.

@thc202 
Please let me know if you have any concerns? otherwise I will start implementation.

thanks,
Karan

thc202:
> How is the user expected to use it?

Was left unanswered, I guess the user will have to "manually" create each `AppParameter`s? What happens when the user specifies the same parameter more than once?

preetkaran20:
> > How is the user expected to use it?
> 
> Was left unanswered, I guess the user will have to "manually" create each `AppParameter`s? What happens when the user specifies the same parameter more than once?

Yes, The user has to create the AppParameter.  
hmm, if the user mentions a parameter multiple times then it can be an issue. 
We can make the input as `Set<AppParameter>` instead of `List` and add the equals and hashcode in AppParameter to rely on `ParamName` to remove duplicates.
Or using `Map` with Key pointing to ParamName.

What do you think?

preetkaran20:
> > > How is the user expected to use it?
> > 
> > 
> > Was left unanswered, I guess the user will have to "manually" create each `AppParameter`s? What happens when the user specifies the same parameter more than once?
> 
> @thc202
> Yes, The user has to create the AppParameter.
> hmm, if the user mentions a parameter multiple times then it can be an issue.
> We can make the input as `Set<AppParameter>` instead of `List` and add the equals and hashcode in AppParameter to rely on `ParamName` to remove duplicates.
> Or using `Map` with Key pointing to ParamName.
> 
> What do you think?


@thc202 any updates on this?

thc202:
I'd suggest a builder like pattern to specify the names/values for each parameter, that should make it easier to use for the user, while avoiding usage errors.

preetkaran20:
> I'd suggest a builder like pattern to specify the names/values for each parameter, that should make it easier to use for the user, while avoiding usage errors.

Sure, makes sense.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class AppParametersBuilder {
+
+    private List<AppParameter> appParameters = new ArrayList<>();
+    private Set<Integer> nameValuePairPositions = new HashSet<>();
+    private final Logger logger = LogManager.getLogger(this.getClass());
+
+    public AppParametersBuilder addAppParameter(
+            NameValuePair nameValuePair, String param, String value, AppParamValueType operation) {
+        if (!nameValuePairPositions.contains(nameValuePair.getPosition())) {
```
preetkaran20, line 25: @thc202 I have made the condition on the position, considering it is unique and also param can be null too.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -47,6 +47,17 @@ public String setParameter(
...
             HttpMessage msg, NameValuePair originalPair, String param, String value);
 
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param message the message that will be changed
+     * @param appParams list of name of the parameter
+     * @return the parameter values set to HttpMessage
+     */
+    default List<String> setParameters(HttpMessage message, List<AppParameter> appParams) {
```
preetkaran20, line 57: The return types are quite conflicting, like for multipart it is returning body of HttMessage where as for cookie it returns something else and for some variants it is returning the passed values.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -209,33 +211,66 @@ public void setMessage(HttpMessage msg) {
...
     public String setEscapedParameter(
             HttpMessage msg, NameValuePair originalPair, String name, String value) {
-        return setParameter(msg, originalPair, value);
+        return setParameter(msg, Arrays.asList(originalPair), Arrays.asList(value));
     }
 
-    private String setParameter(HttpMessage msg, NameValuePair originalPair, String value) {
+    @Override
+    public List<String> setParameters(HttpMessage msg, List<AppParameter> appParameters) {
+        return Arrays.asList(
```
preetkaran20, line 225: As for single parameter setParameter method is returning HTTP message's body so returned the same(by adding to list). @thc202

preetkaran20:
Hi @thc202,

Please review the PR.

thanks,
Karan

preetkaran20:
Hi @thc202 and @kingthorin,

Please review the PR.

thanks,
Karan

preetkaran20:
> Hi @thc202 and @kingthorin,
> 
> Please review the PR.
> 
> thanks,
> Karan

Hi @thc202,

Please review the PR.

thanks,
Karan

kingthorin:
> Please review the PR. Also 'zapjapicmp' is failing, can you please help in resolving this.

That means you introduced breaking changes to the public API. Run the task locally and check the report for more details.

preetkaran20:
> > Please review the PR. Also 'zapjapicmp' is failing, can you please help in resolving this.
> 
> That means you introduced breaking changes to the public API. Run the task locally and check the report for more details.

It is actually failing because I introduced a new Default method in Variant interface. METHOD_NEW_DEFAULT is the rule that is failing and I think we need to ignore it same as how we ignored METHOD_ABSTRACT_NOW_DEFAULT. @kingthorin @thc202 please suggest

image

preetkaran20:
@thc202 @kingthorin please review the PR.

thanks,
Karan

preetkaran20:
Hi @thc202 ,

can you please review the PR?

thanks,
Karan

preetkaran20:
> Hi @thc202 ,
> 
> can you please review the PR?
> 
> thanks,
> Karan

Hi @thc202,

Can you please review the PR?

preetkaran20:
> > Hi @thc202 ,
> > can you please review the PR?
> > thanks,
> > Karan
> 
> Hi @thc202,
> 
> Can you please review the PR?

Hi @thc202 ,

Can you please review the PR?

preetkaran20:
> > > Hi @thc202 ,
> > > can you please review the PR?
> > > thanks,
> > > Karan
> > 
> > 
> > Hi @thc202,
> > Can you please review the PR?
> 
> Hi @thc202 ,
> 
> Can you please review the PR?

Hi @thc202,

A reminder 😆 

thanks,
Karan

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParamValueType.java
```diff
@@ -0,0 +1,13 @@
+package org.parosproxy.paros.core.scanner;
+
+/**
+ * {@code AppParamValueType} represents the type of AppParameter's value. This is useful for the
+ * use-cases like say there is a parameter which is a URL then caller can itself encode/escape it so
+ * that it can be directly used to modify {@code HttpMessage} or {@code Variant} has to escape it.
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public enum AppParamValueType {
```
thc202, line 10: Better move to `AppParameter`, this also applies to the name not just the value.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
```diff
@@ -0,0 +1,41 @@
+package org.parosproxy.paros.core.scanner;
+
+/**
+ * {@code AppParameter} class wraps the parameters which are used to modify the {@code HttpMessage}.
+ * It is specifically used for updating multiple parameters of {@code HttpMessage}
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public class AppParameter {
```
thc202, line 9: `AppParameter` is misleading, not just about the target app. `InputVector` would be closer to what it is.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
```diff
@@ -0,0 +1,41 @@
...
+    private NameValuePair nameValuePair;
+    private String param;
+    private String value;
+    private AppParamValueType appParamValueType;
+
+    AppParameter(
+            NameValuePair nameValuePair,
```
thc202, line 16: I'm not sure this needs to be passed around, the variants will have them already (this just needs to provide the position).

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
```diff
@@ -0,0 +1,41 @@
...
+        this.param = param;
+        this.value = value;
+        this.appParamValueType = appParamValueType;
+    }
+
+    public NameValuePair getNameValuePair() {
+        return nameValuePair;
+    }
+
+    public String getParam() {
```
thc202, line 30: `getName()`

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public class AppParametersBuilder {
+
+    private List<AppParameter> appParameters = new ArrayList<>();
+    private Set<Integer> nameValuePairPositions = new HashSet<>();
+    private final Logger logger = LogManager.getLogger(this.getClass());
+
+    public AppParametersBuilder addAppParameter(
```
thc202, line 23: It should be possible to set the name/value separately.

zap/src/main/java/org/parosproxy/paros/core/scanner/MultipartFormParameter.java
```diff
@@ -44,6 +44,16 @@ public MultipartFormParameter(
         this.type = type;
     }
 
+    public MultipartFormParameter(MultipartFormParameter multipartFormParameter) {
```
thc202, line 47: The changes to this class should be reverted, the variant should adjust the position based on the data injected.

zap/src/test/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -228,7 +228,293 @@ void shouldInjectParamValueModificationInFileNameParam() {
     }
 
     @Test
-    void shouldInjectParamValueModificationInFileContentTypeParam() {
+    public void shouldInjectSmallerParamValueModificationInFileNameParam() {
```
thc202, line 231: Remove the public modifiers.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
thc202: Sounds good.

thc202: But it should allow to replace, that way the scan rules can reuse the builder to craft another message with slightly changes.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
thc202: Fine returning void.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: Well, that allows to provide evidence of what was injected for instance. Although for multiple parameters it would be harder to provide a more precise evidence I guess.

thc202:
> It is actually failing because I introduced a new Default method in Variant interface.

Just suppress it. (That shouldn't be suppressed generally as that's a breaking change, or potential, needs to be decided on a case by case basis).

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -218,4 +219,15 @@ protected String setParameter(HttpMessage message, String param, String value) {
...
     }
+
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param message the message that will be changed
+     * @param appParams list of the parameters
+     * @return the parameter values which are added to the HttpMesage
+     */
+    protected List<String> setParameters(HttpMessage message, List<AppParameter> appParams) {
```
thc202, line 230: Should also provide a method that returns the builder and both methods should mention each other, to make it clear how the API is supposed to be used.

thc202:
I'd suggest moving the new classes to ZAP package.

zap/src/main/java/org/parosproxy/paros/core/scanner/MultipartFormParameter.java
preetkaran20: yeah sure.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
preetkaran20: hmm sure.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
preetkaran20: ok, mention meaning java doc mention?

thc202: Right.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParamValueType.java
preetkaran20: ok, `with the comment this also applies to name`, you are suggesting for javadoc?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
preetkaran20: I think that these method names should align with earlier Api's parameter names. Else it will confuse users. WDYT?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
preetkaran20: Didn't get you, Can you please explain more?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParamValueType.java
thc202: Was about the name of the enum (though JavaDoc should be updated accordingly), e.g. `PayloadFormat`?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
preetkaran20: shouldn't we align with the existing APIs because for a single call we are passing NameValuePair? Also, I think there is no need for the caller to know these internal things?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
thc202: The main consumers of this API would be the variants, which know about the name (`NameValuePair#getName()`).

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParamValueType.java
preetkaran20: ok

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
thc202: The name can be an input vector as well and should be possible to change it without changing the value.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
preetkaran20: hmm right as a consumer will be using the builder. Make sense.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
preetkaran20: ok, so you are suggesting the addition of one more method with just name. Right?

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
thc202: The only user of the constructor is the builder and the consumers are the variants, for the builder it doesn't matter if it passes the position or the whole `NameValuePair`, the variants just need to know about the position.

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParametersBuilder.java
thc202: Yeah, one method for the value and another for the name (I guess we could still provide this one for the cases both name/value are changed).

zap/src/main/java/org/parosproxy/paros/core/scanner/AppParameter.java
preetkaran20: hmm right, i think same reason like getName api.

preetkaran20:
Hi @thc202,

Can you please review it?

thanks,
Karan

preetkaran20:
Hi @thc202,

Please review the changes as suggested.

thanks,
Karan

zap/zap.gradle.kts
```diff
@@ -163,6 +164,7 @@ val japicmp by tasks.registering(JapicmpTask::class) {
         reportName = "japi.html"
         isAddDefaultRules = true
         addRule(JApiChangeStatus.MODIFIED, AcceptMethodAbstractNowDefaultRule::class.java)
+        addRule(JApiChangeStatus.MODIFIED, AcceptMethodNewDefaultRule::class.java)
```
thc202, line 167: Please, suppress the incompatibility instead.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -209,33 +212,78 @@ public void setMessage(HttpMessage msg) {
...
+                    new MultipartFormParameter(
+                            mpPart.getName(),
+                            mpPart.getValue(),
+                            originalStart,
+                            mpPart.getEnd() + offset,
+                            mpPart.getPosition(),
+                            mpPart.getType());
+            multiPartParamsClone.set(idx, mpPart);
+            // As we have updated one value and this can impact other multipart params
+            // Hence this logic is to update the other multipart params
```
thc202, line 271: It just needs to keep the offset and sum to existing positions. No need to copy the parameters again.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
```diff
@@ -0,0 +1,67 @@
...
+ * {@code InputVector} class wraps the parameters which are used to modify the {@code HttpMessage}.
+ * It is specifically used for updating multiple parameters of {@code HttpMessage}
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public class InputVector {
+    private int position;
+    private String name;
+    private String value;
+    private PayloadFormat payloadFormat;
```
thc202, line 32: `final`

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
```diff
@@ -0,0 +1,67 @@
...
+    public String getValue() {
+        return value;
+    }
+
+    public PayloadFormat getPayloadFormat() {
+        return payloadFormat;
+    }
+
+    /**
+     * {@code PayloadFormat} represents the type of InputVectors's name and value. This is useful
```
thc202, line 58: `represents the format of the payload`.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
```diff
@@ -0,0 +1,67 @@
...
+     * for the use-cases like say there is a parameter which is a URL then caller can itself
+     * encode/escape it so that it can be directly used to modify {@code HttpMessage} or {@code
+     * Variant} has to escape it.
+     */
+    public static enum PayloadFormat {
```
thc202, line 63: - Remove static.
 - Normally this would be declared above, before the instance variables.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+ * {@code InputVectorBuilder} provides an easy way to build {@code InputVectors} and validates the
+ * provided inputs.
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public class InputVectorBuilder {
+
+    private Map<Integer, InputVector> inputVectorsMap = new HashMap<>();
+
+    public InputVectorBuilder addAppParameter(
```
thc202, line 40: setNameAndValue

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+            NameValuePair nameValuePair, String param, String value, PayloadFormat payloadFormat) {
+        inputVectorsMap.put(
+                nameValuePair.getPosition(),
+                new InputVector(nameValuePair.getPosition(), param, value, payloadFormat));
+        return this;
+    }
+
+    public InputVectorBuilder addAppParameterWithoutParam(
```
thc202, line 48: setValue

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+            NameValuePair nameValuePair, String value, PayloadFormat payloadFormat) {
+        return this.addAppParameter(nameValuePair, null, value, payloadFormat);
+    }
+
+    public InputVectorBuilder addAppParameterWithoutValue(
```
thc202, line 53: setName

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+            NameValuePair nameValuePair, String param, String value, PayloadFormat payloadFormat) {
+        inputVectorsMap.put(
+                nameValuePair.getPosition(),
+                new InputVector(nameValuePair.getPosition(), param, value, payloadFormat));
+        return this;
+    }
+
+    public InputVectorBuilder addAppParameterWithoutParam(
+            NameValuePair nameValuePair, String value, PayloadFormat payloadFormat) {
+        return this.addAppParameter(nameValuePair, null, value, payloadFormat);
```
thc202, line 50: `null` has special meaning (e.g. remove the IV), it should use the original values of the `NameValuePair`.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public InputVectorBuilder addAppParameterWithoutValue(
+            NameValuePair nameValuePair, String param, PayloadFormat payloadFormat) {
+        return this.addAppParameter(nameValuePair, param, null, payloadFormat);
+    }
+
+    public List<InputVector> build() {
+        return Collections.unmodifiableList(new ArrayList<>(inputVectorsMap.values()));
```
thc202, line 59: If it's creating a new list it does not need to return an unmodifiable one.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -218,4 +221,20 @@ protected String setParameter(HttpMessage message, String param, String value) {
     protected String setEscapedParameter(HttpMessage message, String param, String value) {
         return variant.setEscapedParameter(message, originalPair, param, value);
     }
+
+    /** @return {@code InputVectorBuilder} which is used to build the {@code InputVector} */
```
thc202, line 225: Add a see to the other method.

thc202:
Please, update also the PR title and tidy up the commit history.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
thc202, line 32: It needs to have the format for name and value, since they can be set raw/escaped separately.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
thc202, line 40: Worth noting/doc the caveat that calling these methods replaces the IV (so previous changes are lost).

zap/zap.gradle.kts
preetkaran20: didn't get you, can you please elaborate?

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: in the for loop? actually, the setter changes are not there hence i cannot set the values inside the multipartformparameter hence need to create a new instance with newer values.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
preetkaran20: yeah .... should not have done this mistake. thanks

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
preetkaran20: so are you suggesting to have 2 payloadformat parameters?

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: can you please elaborate?

preetkaran20:
> Please, update also the PR title and tidy up the commit history.

I have updated the PR title and also tidy up the commits. However not able to do more. May be this time please merge using squash option in UI. Next time will take care of this.

zap/zap.gradle.kts
preetkaran20: got it, please review now.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: Calculate the difference of the values injected to the ones already present then you can just offset (i.e. sum) the start/end of the following parameters when replacing.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -218,4 +221,24 @@ protected String setParameter(HttpMessage message, String param, String value) {
     protected String setEscapedParameter(HttpMessage message, String param, String value) {
         return variant.setEscapedParameter(message, originalPair, param, value);
     }
+
+    /**
+     * @return {@code InputVectorBuilder} which is used to build the {@code InputVector} which is
+     *     used by {@link #setParameters(HttpMessage, List)}
+     * @see #setParameters(HttpMessage, List)
```
thc202, line 228: Since you added a link the see is not needed.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -47,6 +48,14 @@ public String setParameter(
     public String setEscapedParameter(
             HttpMessage msg, NameValuePair originalPair, String param, String value);
 
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param message the message that will be changed
+     * @param appParams list of name of the parameter
+     */
+    default void setParameters(HttpMessage message, List<InputVector> appParams) {};
```
thc202, line 57: - Add since tag, same for the new methods in the implementation.
- `appParams` needs to be renamed.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public enum PayloadFormat {
+        REQUIRES_ESCAPING,
+        ALREADY_ESCAPED
+    }
+
+    private final int position;
+    private final String name;
+    private final String value;
+    private final PayloadFormat namePayloadFormat;
```
thc202, line 43: Move up to keep it closer to the variable that it applies to (same in other places, e.g. parameters).

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+ * provided inputs.
+ *
+ * @author preetkaran20@gmail.com KSASAN
+ */
+public class InputVectorBuilder {
+
+    private Map<Integer, InputVector> inputVectorsMap = new HashMap<>();
+
+    /**
+     * Note: calling this method with same NameValuePair will overwrite the earlier name and value
```
thc202, line 40: This applies to all three methods.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+     * Note: calling this method with same NameValuePair will overwrite the earlier value with the
+     * newer one.
+     *
+     * @param nameValuePair
+     * @param value
+     * @param valuePayloadFormat
+     * @return {@code InputVectorBuilder}
+     */
+    public InputVectorBuilder setValue(
+            NameValuePair nameValuePair, String value, PayloadFormat valuePayloadFormat) {
```
thc202, line 77: `payloadFormat`

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+        return this.setNameAndValue(
+                nameValuePair,
+                nameValuePair.getName(),
+                value,
+                PayloadFormat.ALREADY_ESCAPED,
```
thc202, line 82: This needs to be reviewed, I think some of the variants create the `NameValuePair`s with unescaped name/value.

zap/zap.gradle.kts
```diff
@@ -154,7 +154,8 @@ val japicmp by tasks.registering(JapicmpTask::class) {
     fieldExcludes = listOf()
 
     classExcludes = listOf(
-        "org.zaproxy.zap.extension.custompages.ContextCustomPagePanel\$CustomPagesMultipleOptionsPanel")
+        "org.zaproxy.zap.extension.custompages.ContextCustomPagePanel\$CustomPagesMultipleOptionsPanel",
+        "org.parosproxy.paros.core.scanner.Variant#setParameters(HttpMessage, List)")
```
thc202, line 158: I'd expect this to be in the method excludes.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
preetkaran20: but every method is set*.

preetkaran20: will add since tag

thc202: Sorry, is that referring to which comment?

thc202: Ok, I see the edit, as mentioned earlier they might not be "app" parameters.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: but to do that we don't have setters, we need to create again the object.
I think we should add the setters as that will make it easier to modify the values.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
preetkaran20: oh ok

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
preetkaran20: ok will change the variable name

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
preetkaran20: Does it matters? I think we can go with AlreadyEscaped parameter as default as this Api is not used till now.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
preetkaran20: oh sorry for the confusion.

zap/zap.gradle.kts
preetkaran20: I am not able to exclude this method with following string: `org.parosproxy.paros.core.scanner.Variant#setParameters(org.parosproxy.paros.network.HttpMessage,java.util.List)`

can you please help with this?

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
thc202: I meant it should be this description for all three, in all cases the name/value is replaced.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
thc202: Same for the name method.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
thc202: It does because it might change the meaning of the name/value, but not a "blocker" (we can tweak as needed later).

zap/zap.gradle.kts
preetkaran20: got it.please ignore.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: @thc202 can you please help with this.

zap/zap.gradle.kts
thc202: There's a classpath issue that prevents the check from working when added to the method excludes, add just the class. Not worth keeping this waiting for that.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: are you suggesting to add a Map of position vs offset? in the start i will have offset as 0 then once i start modifying, i find the offset in the map and then while replacing use the offset.

cannot we create a setter with package private visibility?

zap/zap.gradle.kts
preetkaran20: I fixed it for now.

thc202: It's preferable not to add that in this PR, it might break other plugins.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: I tried map thing and it is making the logic complex. we can easily do this with setters. @thc202

zap/zap.gradle.kts
preetkaran20: oh ... ok

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -47,6 +48,15 @@ public String setParameter(
...
             HttpMessage msg, NameValuePair originalPair, String param, String value);
 
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param message the message that will be changed
+     * @param inputVectors list of name of the parameter
+     * @since TODO add version
+     */
+    default void setParameters(HttpMessage message, List<InputVector> inputVectors) {};
```
thc202, line 58: Can be `2.11.0`, the methods added in `AbstractAppParamPlugin` should have the tag as well. Also worth adding to the new classes.

preetkaran20: ok

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -209,33 +212,85 @@ public void setMessage(HttpMessage msg) {
...
+        return setParameter(msg, Arrays.asList(originalPair.getPosition()), Arrays.asList(value));
     }
 
-    private String setParameter(HttpMessage msg, NameValuePair originalPair, String value) {
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param msg the message that will be changed
+     * @param inputVectors list of name of the parameter
+     * @since TODO add version
```
thc202, line 229: Needs update.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: It just needs an int for the offset, I'll get to you later.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: done

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: This will need #6690.

thc202: With the changes in place this can now be:
```diff
-        List<MultipartFormParameter> multiPartParamsClone =
-                new ArrayList<MultipartFormParameter>(this.multiPartParams);
+        int offset = 0;
         for (int index = 0; index < nameValuePairPositions.size(); index++) {
             int originalPosition = nameValuePairPositions.get(index);
             String value = values.get(index);
             int idx = originalPosition - 1;
 
-            MultipartFormParameter mpPart = multiPartParamsClone.get(idx);
+            MultipartFormParameter mpPart = multiPartParams.get(idx);
             if (LOGGER.isDebugEnabled()) {
                 LOGGER.debug(
                         "i: "
@@ -268,34 +269,8 @@ public class VariantMultipartFormParameters implements Variant {
                                 + " E: "
                                 + mpPart.getEnd());
             }
-            newBodyBuilder.replace(mpPart.getStart(), mpPart.getEnd(), value);
-            int originalStart = mpPart.getStart();
-            int offset = value.length() - mpPart.getEnd() + mpPart.getStart();
-            mpPart =
-                    new MultipartFormParameter(
-                            mpPart.getName(),
-                            mpPart.getValue(),
-                            originalStart,
-                            mpPart.getEnd() + offset,
-                            mpPart.getPosition(),
-                            mpPart.getType());
-            multiPartParamsClone.set(idx, mpPart);
-            // As we have updated one value and this can impact other multipart params
-            // Hence this logic is to update the other multipart params
-            for (idx = 0; idx < multiPartParamsClone.size(); idx++) {
-                mpPart = multiPartParamsClone.get(idx);
-                if (mpPart.getStart() > originalStart) {
-                    mpPart =
-                            new MultipartFormParameter(
-                                    mpPart.getName(),
-                                    mpPart.getValue(),
-                                    mpPart.getStart() + offset,
-                                    mpPart.getEnd() + offset,
-                                    mpPart.getPosition(),
-                                    mpPart.getType());
-                    multiPartParamsClone.set(idx, mpPart);
-                }
-            }
+            newBodyBuilder.replace(mpPart.getStart() + offset, mpPart.getEnd() + offset, value);
+            offset += value.length() - (mpPart.getEnd() - mpPart.getStart());
```

thc202: Note that some tests will need to update the position of the `NameValuePair`s.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -47,6 +48,15 @@ public String setParameter(
...
             HttpMessage msg, NameValuePair originalPair, String param, String value);
 
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param message the message that will be changed
+     * @param inputVectors list of name of the parameter
+     * @since 2.11.0
+     */
+    default void setParameters(HttpMessage message, List<InputVector> inputVectors) {};
```
thc202, line 58: Remove semicolon.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -209,33 +212,85 @@ public void setMessage(HttpMessage msg) {
...
+     * @param msg the message that will be changed
+     * @param inputVectors list of name of the parameter
+     * @since 2.11.0
+     */
+    @Override
+    public void setParameters(HttpMessage msg, List<InputVector> inputVectors) {
+        this.setParameter(
+                msg,
+                inputVectors.stream()
+                        .map(inputVector -> inputVector.getPosition())
```
thc202, line 236: Use method references here and following.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -209,33 +212,85 @@ public void setMessage(HttpMessage msg) {
...
+                        .collect(Collectors.toList()),
+                inputVectors.stream()
+                        .map(inputVector -> inputVector.getValue())
+                        .collect(Collectors.toList()));
+    };
```
thc202, line 241: Remove semicolon.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+        return this.setNameAndValue(
+                nameValuePair,
+                param,
+                nameValuePair.getValue(),
+                payloadFormat,
+                PayloadFormat.ALREADY_ESCAPED);
+    }
+
+    public List<InputVector> build() {
+        return new ArrayList<>(inputVectorsMap.values());
```
thc202, line 107: Better if we return a consistent order, most of the variants would benefit of that, e.g.:
```Java
return inputVectorsMap.values().stream().sorted(Comparator.comparingInt(InputVector::getPosition)).collect(Collectors.toList());
```

zap/src/test/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -227,6 +229,305 @@ void shouldInjectParamValueModificationInFileNameParam() {
                 message.getRequestBody().toString(), equalTo(newMsg.getRequestBody().toString()));
     }
 
+    @Test
+    void shouldInjectSmallerParamValueModificationInFileNameParam() {
```
thc202, line 233: Can be parameterized, it's just testing different values (same for some of the following tests).

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
preetkaran20: as we are using HashMap with key position, hence it becomes somehow consistent order but will do.

zap/src/test/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
preetkaran20: yeah

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVectorBuilder.java
preetkaran20: done

thc202: `HashMap` is not a `SortedMap` though.

preetkaran20: yeah, may be if size is more than 12, then i think hashcode will not give increasing order.

zap/src/test/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
thc202: I meant JUnit test parameterization (e.g. `VariantJSONQueryUnitTest`). The tests `shouldInjectParamValueMultipleTimesModificationsWith*` can also be parameterized.

preetkaran20: oh ok. I have done the changes as suggested. Please have a look.

preetkaran20:
Hi @thc202,

I have done the changes as suggested, please review the PR.

thanks,
Karan

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
thc202: Parameters and methods should be updated accordingly for consistency.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
thc202: This one was not addressed yet.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -215,33 +218,61 @@ public void setMessage(HttpMessage msg) {
...
     }
 
-    private String setParameter(HttpMessage msg, NameValuePair originalPair, String value) {
+    /**
+     * Sets the parameters into the given {@code message}.
+     *
+     * @param msg the message that will be changed
+     * @param inputVectors list of name of the parameter
+     * @since 2.11.0
+     */
```
thc202, line 236: Maybe remove this one, is exactly the same as the interface.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
```diff
@@ -215,33 +218,61 @@ public void setMessage(HttpMessage msg) {
...
             HttpMessage msg, NameValuePair originalPair, String name, String value) {
-        return setParameter(msg, originalPair, value);
+        return setParameter(msg, Arrays.asList(originalPair.getPosition()), Arrays.asList(value));
     }
 
     @Override
     public String setEscapedParameter(
             HttpMessage msg, NameValuePair originalPair, String name, String value) {
-        return setParameter(msg, originalPair, value);
+        return setParameter(msg, Arrays.asList(originalPair.getPosition()), Arrays.asList(value));
```
thc202, line 227: Could use `Collections.singletonList` instead.

zap/src/test/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParametersUnitTest.java
```diff
@@ -206,13 +213,13 @@ void shouldInjectParamValueModificationInFileParam() {
                 message.getRequestBody().toString(), equalTo(newMsg.getRequestBody().toString()));
     }
 
-    @Test
-    void shouldInjectParamValueModificationInFileNameParam() {
+    @ParameterizedTest
+    @ValueSource(strings = {"injected", "inj", "injectedFileName", ""})
+    void shouldInjectEmptyParamValueModificationInFileNameParam(String newValue) {
```
thc202, line 218: Remove `Empty`.

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: Oh missed your comment. Will address

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: thanks for mentioning it.

zap/src/main/java/org/zaproxy/zap/core/scanner/InputVector.java
preetkaran20: sure

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantMultipartFormParameters.java
preetkaran20: I think this code is working because of sorting by position. Actually, I didn't do this earlier and used ArrayList to store the offset because of the unordering thing.

preetkaran20:
Hi @thc202,

Please review.

thanks,
Karan

preetkaran20:
Hi @thc202,

Please review the PR.

thanks,
Karan

thc202:
Rebased and tidy up commit history. Also moved the `scan` method back to its initial location, tidy up the JavaDoc, and removed unnecessary message parameter.

thc202 (approved):

thc202:
Thank you!

preetkaran20:
> Rebased and tidy up commit history. Also moved the `scan` method back to its initial location, tidy up the JavaDoc, and removed unnecessary message parameter.

Hi @thc202 , How did you rebase and squashed it? can you please let me know? I tried doing squash but due to merge commits, i faced issues.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -97,11 +99,22 @@ public void scan() {
...
+     * parameters in a {@code HttpMessage}.
+     *
+     * <p>By default this method calls {@link #scan(HttpMessage, NameValuePair)} for each {@code
+     * NameValuePair}.
+     *
+     * @param nameValuePairs list of parameters handled by the current variant
+     * @since 2.11.0
+     * @see #setParameters(HttpMessage, List)
+     */
+    protected void scan(List<NameValuePair> nameValuePairs) {
```
preetkaran20, line 114: hmm yeah, we didn't need the message. thanks.

preetkaran20:
> > Rebased and tidy up commit history. Also moved the `scan` method back to its initial location, tidy up the JavaDoc, and removed unnecessary message parameter.
> 
> Hi @thc202 , How did you rebase and squashed it? can you please let me know? I tried doing squash but due to merge commits, i faced issues.

Also please review https://github.com/SasanLabs/owasp-zap-fileupload-addon/pull/1/files

kingthorin (approved):

PR #6373, opened 2020-12-23
Include prod version of webswing if the req token is supplied

psiinon:
Should I update all images like this or shall we just test with the weekly?
Works ok here...


kingthorin (approved):

kingthorin:
Leave the weekly for a week or two then do the others?

thc202 (approved):

Issue #6374, opened 2020-12-23
GLib-GObject errors with zap2docker/stable


orionCook:
**Describe the bug**
When I run this command 
```
docker exec -i owasp zap-api-scan.py -t https://${TARGET_HOST}/openapi.json -f openapi \
> -z "-configfile /zap/wrk/scan_config.txt" \
> -c warnings_config.txt \
> -r report.html \
> -n fuse_api.context \
> -J report.json
```
The scan exits with a 3, but still prints out like it ran the full scan.  From what I can see, it's an issue with the firefox client installed on the image. I have attached the stderr and stdout output in the docker-output.txt file attached

[docker-output.txt](https://github.com/zaproxy/zaproxy/files/5737255/docker-output.txt)

**To Reproduce**
1. Standup a server with an openapi.json endpoint and target that
2. Update the TARGET_HOST and WORKING_DIR variables in the attached script
3. Run the commands in scratch_1.txt against the server
[scratch_1.txt](https://github.com/zaproxy/zaproxy/files/5737311/scratch_1.txt)

**Expected behavior**
I'd expect the scans to work as they did in owasp/zap2docker-stable:2.9.0

**Software versions**
all standard software in the owasp/zap2docker-stable docker image

**Errors from the zap.log file**
[zap.log](https://github.com/zaproxy/zaproxy/files/5737297/zap.log)

**Would you like to help fix this issue?**
I would love to, but doubt I'll have the time :(

thc202:
That's "normal" output when running Firefox (https://github.com/mozilla/geckodriver/issues/347#issuecomment-261186848).

thc202:
Closing per previous comment.

orionCook:
I know it's not an issue with firefox, but I still can't run the latest stable docker build successfully. I've had to roll back to 2.9.0 to get a non 3 exit code

thc202:
Sorry overlooked that remark. Based on the data provided I don't see the reason why the script would exit with 3, could you run the script with `-d` flag?

orionCook:
> 
> 
> Sorry overlooked that remark. Based on the data provided I don't see the reason why the script would exit with 3, could you run the script with `-d` flag?

Sure thing. I'll post the results in a bit

orionCook:
[owasp_debug_output.txt](https://github.com/zaproxy/zaproxy/files/5776481/owasp_debug_output.txt)
[...]

PR #6378, opened 2021-01-04
Log more details on unknown value in JSON variant

thc202:
Log the name of the field and the char.

From logs in #6374.


kingthorin (approved):
LGTM

psiinon (approved):

PR #6406, opened 2021-01-19
Fix: #6405. Lint files with black+isort.

ioggstream:
## This PR

- lints files with black and isort, thus simplifying further contributions.
- fixes tox.ini indentation according to PEP8

Fixes: #6405

## Note

* .py file edits are done via black+isort, no manual modifications.
* linting is a preliminary activity to further PRs (eg. better argument management)

PR #6410, opened 2021-01-19
Refactoring attempt

ioggstream:
## This PR

will be useful to discuss some refactoring proposal, including:

- use `%r` or `%s` with logging instead of string concatenation for code readability;
- use startswith with a tuple

have your say.

### DIFF
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -66,6 +66,7 @@ from zap_common import *
 class NoUrlsException(Exception):
     pass
 
+SUPPORTED_FORMATS = ["openapi", "soap", "graphql"]
 
 config_dict = {}
 config_msg = {}
@@ -241,8 +242,8 @@ def main(argv):
     if len(target) == 0:
         usage()
         sys.exit(3)
-    if format != 'openapi' and format != 'soap' and format != 'graphql':
-        logging.warning('Format must be either \'openapi\', \'soap\', or \'graphql\'')
+    if format not in SUPPORTED_FORMATS:
+        logging.warning('Format must be one of %r', SUPPORTED_FORMATS)
         usage()
         sys.exit(3)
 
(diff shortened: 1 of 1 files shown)


docker/zap-api-scan.py
```diff
@@ -241,8 +242,8 @@ def main(argv):
     if len(target) == 0:
         usage()
         sys.exit(3)
-    if format != 'openapi' and format != 'soap' and format != 'graphql':
-        logging.warning('Format must be either \'openapi\', \'soap\', or \'graphql\'')
+    if format not in SUPPORTED_FORMATS:
```
ioggstream, line 245: here and log levels can be validated by switching from `getopt` -> `ArgumentParser`. In python2, Argument parser was not in the PSL, while in python3 you have it already and you don't need to install it anymore.

docker/zap-api-scan.py
```diff
@@ -295,7 +296,7 @@ def main(argv):
     if port == 0:
         port = get_free_port()
 
-    logging.debug('Using port: ' + str(port))
+    logging.debug('Using port: %s', port)
```
ioggstream, line 299: the logger takes care calling __str__ or __repr__ depending on `%s` or `%r`

docker/zap-api-scan.py
```diff
@@ -518,7 +519,7 @@ def main(argv):
                     f.write('# Only the rule identifiers are used - the names are just for info\n')
                     f.write('# You can add your own messages to each rule by appending them after a tab on each line.\n')
                     for key, rule in sorted(all_dict.items()):
-                        f.write(key + '\tWARN\t(' + rule + ')\n')
+                        f.write(f'{key}\tWARN\t({rule})\n')
```
ioggstream, line 522: python 3.6 will be out of support from 2021-12. If we want to retain it, we can't use f-strings. Have your say.

thc202:
It might not worth doing major "cosmetic" refactors as per https://github.com/zaproxy/zaproxy/pull/6406#issuecomment-774077795 but if you still think worth doing to make other changes you plan easier, I'm all for it.

kalijali (approved):

kalijali (approved):

docker/zap-api-scan.py
k3v3n: Any reason why not using f-string syntax ?

ioggstream: I used to create strings only if logs are actually printed. Probably in this case it is ok using f-string.

PR #6414, opened 2021-01-21
Correct proxy scripts' return value evaluation

### LINKED ISSUE #6381: Proxy script return value misinterpreted in 2.10.0
**Describe the bug**
Hi,
Not sure if it's related to the version, but since the update to 2.10.0, I have this issue.

If I enable a proxy script, I have the error "Secure Connection Failed" in the browser when I try to connect to any url.
- it was working fine before the update (not 100% sure when it start to fail)
- I'm using dynamic SSL certificates, and SSL is working fine if I use proxy without any script enabled
- I tested with Jython and  Zest script, even with basic provided templates (so it's not related to the script himself)
- I tried to regenerate SSL certs, but doesn't help.
- no error in logs

From the browser side, it looks like the SSL is not validated when script is enabled.

Any advice to solve this issue ? 

Thanks

**To Reproduce**
1. Setup dynamic SSL certificates and browser proxy to connect through zap
2. Create new proxy script (jython or zest) and choose any default template
3. Enable the script
4. Connect to any url from browser

**Expected behavior**
Connection should be done and the script should be executed.

**Software versions**
 - ZAP: 2.10.0
 - Browser: Tried with Firefox & Chrome
 - OS : Kali linux

No error in zap log.

sure

thc202:
Use return value as forward message instead of drop.
Update tests accordingly.

Fix #6381.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/script/ProxyListenerScript.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/script/ProxyListenerScript.java
@@ -65,9 +65,9 @@ public class ProxyListenerScript implements ProxyListener {
         for (CachedScript<ProxyScript> cachedScript : scripts.getCachedScripts()) {
             ProxyScript script = cachedScript.getScript();
             try {
-                boolean dropMessage =
+                boolean forwardMessage =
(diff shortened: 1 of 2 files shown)


kingthorin (approved):
TY

ricekot (approved):

PR #6425, opened 2021-01-28
Tidy up tests' setup

thc202:
Split initialisation of `Constant#messages` from the `Constant`, not all
tests need both, and was causing older data from the `Constant` to be
used in other tests leading to failures.
Adjusts tests accordingly.
Remove now unnecessary mocks in auth tests.


psiinon (approved):

kingthorin (approved):
Thank you!

PR #6449, opened 2021-02-10
[BugFix] Added target to spider scanning

copethomas:
When scanning a URL using the spider with a user inside a context, the ZAP API throws a error as it is expecting a URL to be provided. This small change passes through that URL to the ZAP API, fixing the issue.

### DIFF
--- a/docker/zap_common.py
+++ b/docker/zap_common.py
@@ -403,7 +403,7 @@ def raise_scan_not_started():
 def zap_spider(zap, target):
     if scan_user:
         logging.debug('Spider %s as user %s', target, scan_user['name'])
-        spider_scan_id = zap.spider.scan_as_user(context_id, scan_user['id'])
+        spider_scan_id = zap.spider.scan_as_user(context_id, scan_user['id'],target)
     else:
         logging.debug('Spider %s', target)
         spider_scan_id = zap.spider.scan(target, contextname=context_name)


kingthorin:
What problem are you trying to solve?

psiinon:
I think the fix is good but the tests are failing.
Let us know if you need some help with those.

thc202:
Is this change really needed? The scripts access the target beforehand, so passing the URL when starting the spider is redundant, maybe the target is not in scope (in which case it would fail either way).

copethomas:
Hello, I have put together https://github.com/zaproxy/zaproxy/issues/6458 to better explain the issue and why I believe this fix is required (@thc202 ). Would you be able to review and if you think it's OK, I will sign off the commits and update the tests to reflect the change.

psiinon:
I can reproduce this. I've got a context defined with authentication and a test user. If I run a baseline scan just specifying the context then it fails saying the url is out of scope but if I also include the user then is works.
So yes, this does need fixing.
I've not had a chance to test this fix but I'll try to do that tomorrow...

thc202:
The fix would be in ZAP then, not the script? ZAP should not require the URL if there's already nodes in scope (which they should be).

copethomas:
Is this a issue with the port being define then or a bug with ZAP?

thc202:
Did you fix the regular expression of your context? Does it still have the same problem?

thc202:
@psiinon this change would break when scanning as user since the target would no longer be in scope.

thc202:
Any news on this one? (Or the issue raised.)

PR #6464, opened 2021-02-24
Fixed parsing of URI upon changing Request method to/from CONNECT in Manual Request Editor

### LINKED ISSUE #6427: Manual Request Editor CONNECT Http Method Broken
I'm not sure what CONNECT is even for, but when you switch a request that works to CONNECT, it eats all of the URL except for the protocol like so:

CONNECT https HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT 6.3; WOW64; rv:39.0) Gecko/20100101 Firefox/39.0
Pragma: no-cache
Cache-Control: no-cache
Content-Length: 78
Accept: text/html
Content-Type: application/x-www-form-urlencoded
Host: localhost:8443

And then when you hit SEND, you get this error:
1186267 [Thread-15] ERROR org.parosproxy.paros.extension.manualrequest.http.impl.HttpPanelSender - host parameter is null
java.lang.IllegalArgumentException: host parameter is null
	at org.apache.commons.httpclient.HttpConnection.<init>(HttpConnection.java:220) ~[zap-D-2021-01-20.jar:?]
	at org.apache.commons.httpclient.HttpConnection.<init>(HttpConnection.java:168) ~[zap-D-2021-01-20.jar:?]
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$HttpConnectionWithReference.<init>(MultiThreadedHttpConnectionManager.java:1145) ~[commons-httpclient-3.1.jar:?]
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager$ConnectionPool.createConnection(MultiThreadedHttpConnectionManager.java:762) ~[commons-httpclient-3.1.jar:?]
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager.doGetConnection(MultiThreadedHttpConnectionManager.java:476) ~[commons-httpclient-3.1.jar:?]
	at org.apache.commons.httpclient.MultiThreadedHttpConnectionManager.getConnectionWithTimeout(MultiThreadedHttpConnectionManager.java:416) ~[commons-httpclient-3.1.jar:?]
	at org.apache.commons.httpclient.HttpMethodDirector.executeMethod(HttpMethodDirector.java:189) ~[zap-D-2021-01-20.jar:?]
	at org.apache.commons.httpclient.HttpClient.executeMethod(HttpClient.java:397) ~[commons-httpclient-3.1.jar:?]
	at org.parosproxy.paros.network.HttpSender.executeMethod(HttpSender.java:429) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.network.HttpSender.runMethod(HttpSender.java:671) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.network.HttpSender.send(HttpSender.java:626) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.network.HttpSender.sendAuthenticated(HttpSender.java:601) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.network.HttpSender.sendAndReceiveImpl(HttpSender.java:1033) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.network.HttpSender.sendAndReceive(HttpSender.java:993) ~[zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.extension.manualrequest.http.impl.HttpPanelSender.handleSendMessage(HttpPanelSender.java:120) [zap-D-2021-01-20.jar:D-2021-01-20]
	at org.parosproxy.paros.extension.manualrequest.ManualRequestEditorDialog$2.run(ManualRequestEditorDialog.java:239) [zap-D-2021-01-20.jar:D-2021-01-20]
	at java.lang.Thread.run(Thread.java:834) [?:?]

(And I also notice the Response pane doesn't change when there was an error.) Seems like if you hit SEND, it should first clear out the response pane, then try to SEND, and if successful, show the response, and if not, maybe display some kind of error message?

And if you 'accidentally' select CONNECT, and then switch to another HTTP Method, the rest of the URL is still lost, so the manual request is now broken until you close it, find the original request in the history, and reopen it from that request, which is annoying.  (Maybe you should have a 'restore original request' button? Although I realize we are somewhat out of room for new buttons.)

So, please fix CONNECT, or drop it if it's not a useful HTTP method to test.

pranavsaxena17:
Changed the method `mutateHttpMethod` to correctly parse the URI for CONNECT. Also fixed the handling of CONNECT while sending a manual request.Fixes #6427

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
+++ b/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
@@ -58,6 +58,7 @@
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
 // ZAP: 2020/12/09 Handle content encodings in request/response bodies.
 // ZAP: 2021/04/01 Detect WebSocket upgrade messages having multiple Connection directives
+// ZAP: 2021/05/11 Fixed conversion of Request Method to/from CONNECT
 package org.parosproxy.paros.network;
 
 import java.net.HttpCookie;
@@ -980,15 +981,11 @@ public class HttpMessage implements Message {
     }
 
     public void mutateHttpMethod(String method) {
-        // String header = reqPanel.getTxtHeader().getText();
-        String header = getRequestHeader().toString();
         try {
-            HttpRequestHeader hrh = new HttpRequestHeader(header);
-
-            URI uri = hrh.getURI();
-            // String body = reqPanel.getTxtBody().getText();
+            URI uri = getRequestHeader().getURI();
             String body = getRequestBody().toString();
-            String prevMethod = hrh.getMethod();
+            String prevMethod = getRequestHeader().getMethod();
+
             if (prevMethod.equalsIgnoreCase(method)) {
                 return;
             }
@@ -1016,7 +1013,6 @@ public class HttpMessage implements Message {
                     }
                     uri.setQuery(sb.toString());
                 }
-                hrh.setURI(uri);
                 // Clear the body
                 body = "";
 
@@ -1040,12 +1036,28 @@ public class HttpMessage implements Message {
                     }
                     body = sb.toString();
                     uri.setQuery(null);
-                    hrh.setURI(uri);
                 }
             }
-            hrh.setMethod(method);
-
-            getRequestHeader().setMessage(hrh.toString());
+            if (prevMethod.equalsIgnoreCase(HttpRequestHeader.CONNECT)) {
+                String scheme;
+                if (getRequestHeader().getHostPort() == 443) {
+                    scheme = "https://";
+                } else {
+                    scheme = "http://";
+                }
+                uri = new URI(scheme + uri, true);
+            } else if (method.equals(HttpRequestHeader.CONNECT)) {
+                uri = URI.fromAuthority(uri.getAuthority());
+            }
+            getRequestHeader()
+                    .setMessage(
+                            method
+                                    + " "
+                                    + uri
+                                    + " "
+                                    + getRequestHeader().getVersion()
+                                    + "\r\n"
+                                    + getRequestHeader().getHeadersAsString());
             getRequestBody().setBody(body);
         } catch (HttpMalformedHeaderException e) {
             // Ignore?
--- a/zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
@@ -201,6 +201,62 @@ public class HttpMessageUnitTest {
     }
 
     @Test
+    public void shouldCorrectlyMutateFromConnectHttpMethodWhenGenericPort() throws Exception {
+        // Given
+        HttpMessage message =
+                new HttpMessage(
+                        new HttpRequestHeader(
+                                "CONNECT "
+                                        + "www.example.com:9000 HTTP/1.1\r\nHost: www.example.com:9000"));
+        // When
+        message.mutateHttpMethod(HttpRequestHeader.GET);
+        // Then
+        assertThat(message.getRequestHeader().getMethod(), is(HttpRequestHeader.GET));
+        assertThat(
+                message.getRequestHeader().getURI().toString(), is("http://www.example.com:9000"));
+        assertThat(
+                message.getRequestHeader().getHeader(HttpRequestHeader.HOST),
+                is("www.example.com:9000"));
+    }
+
+    @Test
+    public void shouldCorrectlyMutateFromConnectHttpMethodWhenHttpsPort() throws Exception {
+        // Given
+        HttpMessage message =
+                new HttpMessage(
+                        new HttpRequestHeader(
+                                "CONNECT "
+                                        + "www.example.com:443 HTTP/1.1\r\nHost: www.example.com:443"));
+        // When
+        message.mutateHttpMethod(HttpRequestHeader.GET);
+        // Then
+        assertThat(message.getRequestHeader().getMethod(), is(HttpRequestHeader.GET));
+        assertThat(
+                message.getRequestHeader().getURI().toString(), is("https://www.example.com:443"));
+        assertThat(
+                message.getRequestHeader().getHeader(HttpRequestHeader.HOST),
+                is("www.example.com:443"));
+    }
+
+    @Test
+    public void shouldCorrectlyMutateToConnectHttpMethod() throws Exception {
+        // Given
+        HttpMessage message =
+                new HttpMessage(
+                        new HttpRequestHeader(
+                                "GET "
+                                        + "http://www.example.com:9000/path HTTP/1.1\r\nHost: www.example.com:9000"));
+        // When
+        message.mutateHttpMethod(HttpRequestHeader.CONNECT);
+        // Then
+        assertThat(message.getRequestHeader().getMethod(), is(HttpRequestHeader.CONNECT));
+        assertThat(message.getRequestHeader().getURI().toString(), is("www.example.com:9000"));
+        assertThat(
+                message.getRequestHeader().getHeader(HttpRequestHeader.HOST),
+                is("www.example.com:9000"));
+    }
+
+    @Test
     void shouldNotSetContentEncodingsWhenSettingHttpRequestBody() {
         // Given
         HttpRequestHeader header = mock(HttpRequestHeader.class);


thc202:
Please revert the changes done to the models, that's not were the issue should be fixed.

pranavsaxena17:
Where should I make changes for correctly handling a CONNECT request then? The message on which `mutateHttpMethod` is called has no information about the URI, as it's lost because that message is generated by using the `parse` method, which parses only the authority and discards the rest of the URI.

thc202:
That's expected since the CONNECT method does not require the whole URI, when changing back to GET (or other method for that matter) one needs to add information as needed.

pranavsaxena17:
@thc202 I reverted the changes made to the models and fixed some handling. Does this need more changes?

thc202:
At least the following:
 - Tests that show that the issue is fixed;
 - Update commit/PR title;
 - Add ZAP comment.

pranavsaxena17:
@thc202 I made the required changes( I assume that by "ZAP comment" you mean a comment in the code explaining my change). Is this now ready for review?

kingthorin:
By "ZAP Comment" he meant you changed a Paros file so way up at the top it should have a comment explaining what was changed and referencing the issue, see lines 22-59 in src/main/java/org/parosproxy/paros/network/HttpMessage.java

pranavsaxena17:
@thc202 @kingthorin I added the ZAP comment. Does this need anything else?

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -983,11 +984,19 @@ public void mutateHttpMethod(String method) {
...
+            URI uri = getRequestHeader().getURI();
+            String hostName = getRequestHeader().getHeader("Host");
+            String scheme = uri.getScheme();
+
+            // If the previous method is CONNECT, then the message URI has
+            // no scheme
+            if (prevMethod.equals(HttpRequestHeader.CONNECT)) {
+                scheme = uri.toString();
+            }
+            // String body = reqPanel.getTxtBody().getText();
```
psiinon, line 998: Can delete this

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1046,6 +1046,22 @@ public void mutateHttpMethod(String method) {
...
             getRequestHeader().setMessage(hrh.toString());
             getRequestBody().setBody(body);
+
+            // CONNECT method does not require the entire URI and parses only the authority as the
+            // URI, hence add relevant information
+            if (method.equals(HttpRequestHeader.CONNECT)
+                    || prevMethod.equals(HttpRequestHeader.CONNECT)) {
+                String scheme;
+                if (uri.getScheme() != null) {
+                    scheme = uri.getScheme();
```
ricekot, line 1056: why change the URI at all in this case?

ricekot (changes requested):
Changing to CONNECT works, but changing from CONNECT to any other method does not seem to be working for me. 
https://user-images.githubusercontent.com/16446369/112192180-c1096480-8c2c-11eb-9467-0c7931f8c9d2.mp4
Also, all query string parameters are lost when changing method to CONNECT - they should be preserved.

thc202:
> Also, all query string parameters are lost when changing method to CONNECT - they should be preserved.

The CONNECT method just uses the authority.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
pranavsaxena17: When a CONNECT request is parsed, all information except the scheme is lost. Hence if the original request is CONNECT we have to craft the URI again

thc202: Note that the scheme should be lost too, the authority is just the host and port

ricekot: the following comment says the opposite of what you're saying here? (talking about line [1058](https://github.com/zaproxy/zaproxy/pull/6464/files#diff-03602e8a54ab912dc5998078c7a4dbde859050d3425c594f50498e810416618cR1058))

pranavsaxena17:
> Changing to CONNECT works, but changing from CONNECT to any other method does not seem to be working for me.
> 
> ~Also, all query string parameters are lost when changing method to CONNECT - they should be preserved.~ [#6464 (comment)](https://github.com/zaproxy/zaproxy/pull/6464#issuecomment-805105284)

Umm this seems to be working for me though...

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -979,15 +980,14 @@ public void setNote(String note) {
...
 
     public void mutateHttpMethod(String method) {
-        // String header = reqPanel.getTxtHeader().getText();
         String header = getRequestHeader().toString();
         try {
             HttpRequestHeader hrh = new HttpRequestHeader(header);
-
-            URI uri = hrh.getURI();
-            // String body = reqPanel.getTxtBody().getText();
+            String prevMethod = getRequestHeader().getMethod();
```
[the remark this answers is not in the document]
pranavsaxena17: But how? Both the `getRequestHeader` and the `getMethod` functions just return already initialized data members of their repective classes, and don't modify anything. Or am I missing something?

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1046,6 +1046,22 @@ public void mutateHttpMethod(String method) {
 
             getRequestHeader().setMessage(hrh.toString());
             getRequestBody().setBody(body);
+
+            // CONNECT method does not require the entire URI and parses only the authority as the
+            // URI, hence add relevant information
+            if (method.equals(HttpRequestHeader.CONNECT)
```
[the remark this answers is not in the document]
pranavsaxena17: If the line 1052 and the subsequent logic is moved between 1044 and 1045, then woudn't we be back to square one?(as the `setMessage` call on line 1047 would again mess up the URI and make it equal to only the scheme of the request)

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
pranavsaxena17: @thc202 When changing from  CONNECT to anyother method, should I append a `http` by default or just let it be the host and port only

ricekot:
Seems like it fails when the Host also has a port in it. For example, try it with the following request:
```
GET http://www.example.com:20 HTTP/1.1
Host: www.example.com:20
```

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
thc202: Defaulting to `http` seems reasonable, it could also check the port, if it's 443 use `https` otherwise `http`.

pranavsaxena17: @thc202 But the `setURI` method of the `HttpRequestHeader` class automatically sets the scheme to `http` if it is null, and thus a URI for a request header cannot be set without the scheme. So should I make changes in `HttpRequestHeader` too?(The URI doesn't seem to have a `setScheme` method to set the scheme afterwords)

thc202: Yes, the `setURI` would have to take into account the method, like `parse` already does. The scheme would be set to the default value (which is `http`).

pranavsaxena17: @thc202 The `URI` object cannot be intialized without a scheme. Thus whenever the method is changed to CONNECT, if I set the scheme to `http` by default as you mention, then the request in the Manual Request Editor will inevitably be shown with the scheme attached;i.e there seems to be no way to not show the scheme in the editor for CONNECT requests without making changes to the models themself. What should I do now?

thc202: It can, `new URI("example.com:443", true)` works as expected. You can also see CONNECT requests in the Request panel properly (i.e. without scheme).

(The default to `http` was when changing from CONNECT to other method.)

pranavsaxena17: @thc202  Well, it seems that initializing `URI` with the default uri without scheme `(www.any_domain_name.org:9000)` throws an incorrect scheme `URIException` whereas it is correctly initialized for `(www.example.com:9000`. This seems to be an effect of the apache commons library.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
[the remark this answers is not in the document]
ricekot: Yeah, but because Java uses references, if you accidentally modify `prevMethod` or `uri` you will also modify the original message.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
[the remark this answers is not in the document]
ricekot: Exactly, that's why the suggestion to use `hrh` instead of `getRequestHeader()`.

pranavsaxena17:
@ricekot @thc202 I made the required changes.(Note that conversion related to the default URI in the request editor is still buggy, [here](https://github.com/zaproxy/zaproxy/pull/6464#discussion_r600562620) is why).Kindly review

zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
```diff
@@ -464,9 +463,8 @@ private void parse(boolean isSecure) throws URIException, HttpMalformedHeaderExc
         }
 
         if (mMethod.equalsIgnoreCase(CONNECT)) {
-            parseHostName(sUri);
-            mUri = parseURI(mHostName);
-
+            parseHostNameAndPort(getHeader(HOST));
+            mUri = parseURI(getHeader(HOST));
```
thc202, line 467: The changes done to this class can be reverted per latest changes from upstream.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -980,15 +981,13 @@ public void setNote(String note) {
...
         String header = getRequestHeader().toString();
         try {
             HttpRequestHeader hrh = new HttpRequestHeader(header);
-
+            String prevMethod = hrh.getMethod();
             URI uri = hrh.getURI();
-            // String body = reqPanel.getTxtBody().getText();
             String body = getRequestBody().toString();
-            String prevMethod = hrh.getMethod();
+
```
thc202, line 990: Why these changes?

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1044,9 +1043,18 @@ public void mutateHttpMethod(String method) {
...
+                if (getRequestHeader().getHostPort() == 443) {
+                    connectURI = new URI("https://" + getRequestHeader().getHeader("Host"), true);
+                } else {
+                    connectURI = new URI("http://" + getRequestHeader().getHeader("Host"), true);
+                }
+                hrh.setURI(connectURI);
+            }
             getRequestHeader().setMessage(hrh.toString());
             getRequestBody().setBody(body);
+
```
thc202, line 1057: Revert.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1044,9 +1043,18 @@ public void mutateHttpMethod(String method) {
                 }
             }
             hrh.setMethod(method);
-
+            if (prevMethod.equalsIgnoreCase(HttpRequestHeader.CONNECT)) {
+                URI connectURI;
+                if (getRequestHeader().getHostPort() == 443) {
+                    connectURI = new URI("https://" + getRequestHeader().getHeader("Host"), true);
+                } else {
+                    connectURI = new URI("http://" + getRequestHeader().getHeader("Host"), true);
```
thc202, line 1051: - Since only the scheme is different declare a variable with the scheme and create the URI once.
- It would be better to use the host name and port, rather than the host header (which might not be present).

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
thc202: That should be addressed in upstream.

pranavsaxena17:
Made the required changes. Kindly review

thc202:
Thank you!

psiinon (approved):

ricekot (approved):

ricekot:
Yeah, it does!

PR #6495, opened 2021-03-28
Modified Open Session Dialog

### LINKED ISSUE #701: Give more session information in Open session dialog in Details mode
```
1. click "Open Session ..."
2. clik "Details"

What is the expected output?
I expect more information about sessions.
Currently the Size column shows 166 bytes for each and every session regardless of
the length of capture or size of the data file or script file.

Windows 7, ZAP 2.1.0

I would like to see following information before opening selected session
1) session data size,
2) number of requests/responses
3) start time of the session
4) end time of the session or duration. 

Information can be provided just for a session that is selected, not necessary for
all sessions at once.


```

Original issue reported on code.google.com by `jan.ruzicka@comtechmobile.com` on 2013-06-26 15:05:42

pranavsaxena17:
The Open Session dialog now displays the selected session's information(namely session start and end time, number of requests and the session data size) within the dialog.

Fixes #701

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
+++ b/zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
@@ -54,16 +54,35 @@
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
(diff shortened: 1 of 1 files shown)


thc202:
Some quick comments:
 - The window is no longer a dialogue but a frame, they behave differently by default. (It should be kept as a dialogue IMO.)
 - The session is left open.
 - The UI strings are not internationalised.

PR #6508, opened 2021-04-01
Detect WebSocket upgrade messages having multiple Connection directives

pf-msi:
`HttpMessage.isWebSocketUpgrade()` does not recognize WebSocket upgrade message when the _Connection_ header has multiple values. 

Connection header is allowed to have a comma-separated list of HTTP headers, i.e. `Connection: keep-alive, Upgrade`, and not just a single `Connection: Upgrade` keyword. 

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection 

> any comma-separated list of HTTP headers

or https://tools.ietf.org/html/rfc6455: 

> The request MUST contain a |Connection| header field whose value MUST **include** the Upgrade token. 

Also, request messages could be considered WebSocket upgrade messages also, similarly to `isEventStream()` method checking both request and response headers.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
+++ b/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
@@ -57,6 +57,7 @@
 // ZAP: 2020/07/31 Tidy up parameter methods
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
 // ZAP: 2020/12/09 Handle content encodings in request/response bodies.
+// ZAP: 2021/04/01 Detect WebSocket upgrade messages having multiple Connection directives
 package org.parosproxy.paros.network;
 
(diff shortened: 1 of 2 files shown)


thc202:
Is there are use case for the request check?

pf-msi:
It's not a "must have", but I've been working on a custom active script for validation of a Cross-Site WebSocket Hijacking vulnerability and I found this method could be useful if it was applied for request messages also. I can check request headers on the script side, without this method, but since it's there in HttpMethod I thought I could use it as well.

thc202:
Great, just checking if you were already relying on that (or planning to).

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #6519, opened 2021-04-06
Adding WS config into main

rsoreq:


thc202:
Closing, seems accidental (if it was on purpose it's better to reach in the dev mailing list for a proposal first).

PR #6526, opened 2021-04-08
Export Menu Tweaks

kingthorin:
* PopupMenuExportMessage
  - Set/name logger as constant.
  - Use line feed `\n` when writing text file content.

* PopupMenuExportResponse
  - Set/name logger as constant.
  - Use line feed `\n` when writing text file content.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
@@ -34,6 +34,7 @@
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2019/11/05 Use WritableFileChooser for saves.
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
+// ZAP: 2021/04/08 Name logger (LOG) fullcaps as constant, use LF as EOL for text file content
 package org.parosproxy.paros.extension.history;
 
 import java.io.BufferedWriter;
@@ -57,9 +58,9 @@ public class PopupMenuExportMessage extends JMenuItem {
 
     private static final long serialVersionUID = 1L;
 
-    private static final Logger log = LogManager.getLogger(PopupMenuExportMessage.class);
+    private static final Logger LOG = LogManager.getLogger(PopupMenuExportMessage.class);
 
-    private static final String CRLF = "\r\n";
+    private static final String EOL = "\n";
     private ExtensionHistory extension = null;
 
     public PopupMenuExportMessage() {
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
```diff
@@ -34,6 +34,7 @@
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2019/11/05 Use WritableFileChooser for saves.
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
+// ZAP: 2021/04/08 Set logger as constant, use system's specific EOL for text file content
```
thc202, line 37: Tweak to say that the variable was renamed? It was already a constant.

zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
```diff
@@ -57,9 +58,9 @@
 
     private static final long serialVersionUID = 1L;
 
-    private static final Logger log = LogManager.getLogger(PopupMenuExportMessage.class);
+    private static final Logger LOG = LogManager.getLogger(PopupMenuExportMessage.class);
 
-    private static final String CRLF = "\r\n";
+    private static final String EOL = System.lineSeparator();
```
thc202, line 63: IMO it's better to use the same line endings always for consistency, I'd suggest changing the other export instead.

zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportResponse.java
```diff
@@ -52,7 +53,7 @@
     private static final String NEWLINE = System.lineSeparator();
 
     // ZAP: Added logger
-    private static Logger log = LogManager.getLogger(PopupMenuExportResponse.class);
+    private static Logger LOG = LogManager.getLogger(PopupMenuExportResponse.class);
```
thc202, line 56: `final`

thc202:
These can be moved to the new add-on...

zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
kingthorin: You always use `\n`?

kingthorin: I set it that way, lemme know if it was the wrong interpretation.

kingthorin:
Tweaked

zap/src/main/java/org/parosproxy/paros/extension/history/PopupMenuExportMessage.java
thc202: Sounds good, worth noting that it will have mixed line endings (HTTP header/body separators) but any decent text editor should cope with that.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6562, opened 2021-04-24
Fixed regex generation for host site URLs included in context

pranavsaxena17:
Upon including host site URLs such as `https://www.google.com/?gws_rd=ssl` in the context using the Include in Context menu option, the regex in the Session Properties dialog is incorrectly generated as `https://www.google.com/https://www.google.com/\?gws_rd\=ssl.*` instead of `https://www.google.com/\?gws_rd\=ssl.*`. There are also a few other related cases(with/without the trailing slash etc.) which were also incorrectly handled. This PR fixes these cases, and includes tests to verify the regex generation in all possible cases.


thc202:
Better use `www.example.com`, what's in the PR description should be added to the commit.

pranavsaxena17:
@thc202 Could you hop on IRC for a minute ?

pranavsaxena17:
@thc202 Kindly review

zap/src/test/java/org/zaproxy/zap/model/SessionStructureUnitTest.java
```diff
@@ -61,6 +62,14 @@ public void setUp() throws Exception {
...
         Control.initSingletonForTesting(model);
         msg = new HttpMessage();
+        sitesNode = mock(StructuralSiteNode.class);
+        given(sitesNode.isRoot()).willReturn(true);
+        hostNode = mock(StructuralSiteNode.class);
+        given(hostNode.getParent()).willReturn(sitesNode);
+        given(hostNode.isLeaf()).willReturn(true);
+        URI uri = new URI("https://www.example.com", true);
+        given(hostNode.getURI()).willReturn(uri);
+        given(hostNode.getName()).willReturn("https://www.example.com");
```
thc202, line 72: The new tests could be done in a nested class, since this is not required for all of them.

pranavsaxena17:
Tweaked. Kindly review

thc202 (approved):

PR #6564, opened 2021-04-24
Fix scrolling in Sites tab with FlatLaf L&F

### LINKED ISSUE #6520: Can not scroll past 'contexts' in 'sites' tab with FlatLaf L&F
**Describe the bug**
Can not use mouse wheel to scroll past 'Context' in 'Sites' tab.

**To Reproduce**
1. Go to 'Sites'
2. Browse around a bunch with the proxy on so the Sites populate
3. Scroll to the top of the list with both sites and contexts expanded
4. Use mouse wheel to scroll down

**Expected behavior**
The list scrolls normally to the bottom.

**Observed behavior**
The scrolling stops at the end of Contexts. Neither can you use arrow keys to go down from the bottom most entry in contexts.
However dragging the scrollbar manually with the mouse past that point allows normal scrolling. Just not when any context is in view.

**Screenshots**
N/A

**Software versions**
 - ZAP: 2.10
 - Add-on: N/A
 - OS: Ubuntu 20.04 (5.4.0-70-generic, amd64)
 - Java: openjdk 11.0.10 2021-01-19
 - Browser: N/A

None

**Additional context**
None

I don't do java.

thc202:
Check the scrollable unit increment of the Sites tree if the Contexts
tree reports no more increments, it might be that the view is almost in
the Sites tree but not yet in it, which would cause the Contexts tree to
report no more increments while it might still have more for the Sites
tree.

Fix #6520.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/view/ContextsSitesPanel.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/ContextsSitesPanel.java
@@ -110,7 +110,12 @@ public class ContextsSitesPanel extends JPanel {
         public int getScrollableUnitIncrement(
                 Rectangle visibleRect, int orientation, int direction) {
             if (visibleRect.getY() < sitesTree.getBounds().getY()) {
-                return contextsTree.getScrollableUnitIncrement(visibleRect, orientation, direction);
+                int unitIncrement =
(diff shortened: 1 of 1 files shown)


kingthorin (approved):

ricekot (approved):

ricekot:
Thank you!

PR #6581, opened 2021-05-04
improve: ci workflow add concurrency grouping

kingthorin:
To cancel in-progress builds when a force push is done and a ci workflow is already running.

Ref:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency
- https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
       - main
   pull_request:
 
+concurrency: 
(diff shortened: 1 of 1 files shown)


thc202 (approved):

thc202:
Thank you!

ricekot (approved):

PR #6583, opened 2021-05-04
Update PHP Client API Generator

ricekot:
Add `@property Zap $zap` above generated class to explicitly specify magic property `$zap` (and to pass PHPStan tests :^).


ricekot:
Updated to use `property Zap $zap` instead of `property mixed $zap`.

thc202 (approved):

psiinon (approved):

PR #6586, opened 2021-05-06
improve: custompages: Address SAST (sonar) findings

kingthorin:


kingthorin:
In ContextCustomPagePanel the constructor CustomPagesMultipleOptionsPanel(CustomPageTableModel model, int contextId)` has an un-used parameter (contextId). However, I couldn't see how to deprecate it and call the new one since the call to the super constructor needs to be first. I guess I could introduce a breaking change and a jacoco exclusion? (I'd need tips on that as well, I don't recall the location/syntax.)

psiinon (approved):

thc202 (approved):

thc202:
Needs a Spotless apply.

kingthorin:
Build passed, just waiting on LGTM

PR #6590, opened 2021-05-07
AbstractAppVariantPlugin for the usecases where ScanRule needs to manipulate multiple namevalue pairs.

preetkaran20:
This is the new PR with the earlier changes as discussed in https://github.com/zaproxy/zaproxy/pull/6371


preetkaran20:
Hi @thc202  @kingthorin 
please review.

preetkaran20:
Hi @thc202  @kingthorin 
please review.

thc202:
Why a new PR? The discussions were still ongoing (e.g. the need for a new class; the API to set multiple parameters), not that we can't continue in this one but the context is lost.

preetkaran20:
> Why a new PR? The discussions were still ongoing (e.g. the need for a new class; the API to set multiple parameters), not that we can't continue in this one but the context is lost.

I have reopened the earlier one too. Lets discuss on any one. Actually i sometimes face issues with handling multiple branches. Will try to handle from next time.

PR #6599, opened 2021-05-14
Dynamically switch to user using ID

Frettarix:
By using a dynamic way you can use the same method for all your Docker files (just ease of use from my side), but by using the ID you can have security software monitoring on your platform whether it's a high privileged account. 

Blocking HP accounts will be as easy as blocking and ID below 1000. Although to enable this all users must be using an ID.


kingthorin:

Frettarix:
Will have to redo the commit, can't sign currently due to git errors.

PR #6604, opened 2021-05-17
Dynamic Docker USER

Frettarix:
By using the ID instead of the name (dynamically) certain systems can perform better root-user checks.

(Followed all the signing steps this time)


Frettarix:
To elaborate, often platform teams use the ID's of the user to identify whether the user is a root account or a user account. This method is used for instance on some Kubernetes platforms to ensure that root access on containers is blocked.

The change itself doesn't impact any project as it's still using the same user, just logging into the user via the docker USER command by using the User's ID instead of the User's name.

Feel free to contact if you have any questions. ;)

thc202:
Is this no longer wanted?

Frettarix:
I didn't get the idea it was being picked up. You can still implement it if you wish but for me it won't matter much. It may help others with their platforms though.

PR #6612, opened 2021-05-19
Parameterize test in HttpMethodBaseUnitTest

eas5:
Problem:
(1) A test method with many individual assertions stops being executed on the first failed assertion, which prevents the remaining ones' execution.
(2) Classes containing test methods with repeated test code steps, typically, parts that set up test fixtures, generate additional effort in case of test maintenance.

Solution:
Parameterized tests make it possible to run a test multiple times, with different arguments, as individual and independent tests. This way, we could make 3 original tests become 5 independent ones, contributing to test maintenance efforts when needed. In this refactoring, no original assertion value was changed.

### DIFF
--- a/zap/src/test/java/org/apache/commons/httpclient/HttpMethodBaseUnitTest.java
+++ b/zap/src/test/java/org/apache/commons/httpclient/HttpMethodBaseUnitTest.java
@@ -20,41 +20,30 @@
 package org.apache.commons.httpclient;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 import java.util.List;
-import org.junit.jupiter.api.Test;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 class HttpMethodBaseUnitTest {
 
-    @Test
-    void testParseCookieHeaderEmpty() {
-        List<Cookie> cookies = HttpMethodBase.parseCookieHeader("example.com", "");
-        assertThat(cookies.size(), is(0));
+    @ParameterizedTest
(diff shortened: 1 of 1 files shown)


zap/src/test/java/org/apache/commons/httpclient/HttpMethodBaseUnitTest.java
```diff
@@ -20,41 +20,29 @@
...
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 import java.util.List;
-import org.junit.jupiter.api.Test;
+import java.util.stream.*;
+import org.junit.jupiter.params.*;
+import org.junit.jupiter.params.provider.*;
```
kingthorin, line 29: Please use specific imports, not wildcard.

eas5: Ok. Uploading a review. Thank you.

eas5:
Thank you @thc202. Commits are now squashed into a single one.

kingthorin (approved):

zap/src/test/java/org/apache/commons/httpclient/HttpMethodBaseUnitTest.java
```diff
@@ -20,41 +20,30 @@
...
 
 class HttpMethodBaseUnitTest {
 
-    @Test
-    void testParseCookieHeaderEmpty() {
-        List<Cookie> cookies = HttpMethodBase.parseCookieHeader("example.com", "");
-        assertThat(cookies.size(), is(0));
+    @ParameterizedTest
+    @MethodSource("cookieHeaderProvider")
+    public void testParseCookieHeader(String cookieHeaderValue, int numberOfCookies) {
```
thc202, line 36: Remove `public` modifier.

eas5: Done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #6621, opened 2021-05-23
Add workflow to update marketplace with weekly

thc202:
The workflow will send a repository dispatch to the marketplace when the
weekly is published.

### DIFF
--- a/buildSrc/src/main/java/org/zaproxy/zap/tasks/HandleWeeklyRelease.java
+++ b/buildSrc/src/main/java/org/zaproxy/zap/tasks/HandleWeeklyRelease.java
@@ -1,0 +1,111 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2021 The ZAP Development Team
+ *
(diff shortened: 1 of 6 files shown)


psiinon (approved):

buildSrc/src/main/java/org/zaproxy/zap/tasks/HandleWeeklyRelease.java
```diff
@@ -0,0 +1,111 @@
...
+        getLogger().info("File downloaded to: " + release);
+        return release;
+    }
+
+    private static String extractFileName(String url) {
+        int idx = url.lastIndexOf("/");
+        if (idx == -1) {
+            throw new IllegalArgumentException(
+                    "The provided URL does not have a file name: " + url);
+        }
```
kingthorin, line 104: Wouldn't this also be true if the index was the end (length) of the string? (https://example.org/)

thc202: Yes, this is just a quick check that something is there (e.g. it wouldn't work properly with an empty path either).

kingthorin: okay

kingthorin (approved):

Issue #6461, opened 2021-02-19
Automation framework tracker


psiinon:
Implement a new automation framework

- [ ] Implement initial framework
- [ ] Ajax Spider integration (https://github.com/zaproxy/zap-extensions/pull/2725)
- [ ] New report add-on
- [ ] Report integration
- [ ] Support initial sanity check validation
- [ ] Use for baseline (if hooks not used)
- [ ] Use for full scan (if hooks not used)
- [ ] OpenAPI integration
- [ ] SOAP integration
- [ ] GraphQL integration
- [ ] Use for API scan  (if hooks not used)
- [ ] New Github action which uses this directly
- [ ] Script support (to replace scan hooks)

psiinon:
Updated to break down authentication support

so87:
When you write the documentation, could you please give one example for standard website and an API? showing an example website, and how you would make an automation framework config file on it to do various types of scanning would be incredibly helpful.

for example, this page ends up creating more questions for me than it answers: https://www.zaproxy.org/docs/desktop/addons/automation-framework/

kingthorin:
![image](https://user-images.githubusercontent.com/7570458/141163513-2b648be8-c32d-4a5e-950b-dffd6e39d035.png)

https://www.zaproxy.org/blog/2021-06-15-baseline-scan-changes/

I'm not saying the docs couldn't be better, I'm sure they always can, but there's still lots available to you.

psiinon:
@so87  What questions? If you dont ask them then we cant answer them :P

so87:
first of all thanks for the help. I know i've been commenting alot and with several issues. i appreciate your time.

"-autogenconf <filename> Generate template automation file using the current configuration."
What configuration would this be referring to? A config file(rules/alerts), a context file, or you are supposed to put other automation flags on the command line and it would generate a automation file based off that? If the latter, why would you ever do that?

Also, are there any examples for this automation framework?

Are you expected to?
1. generate the min automation framework config file
2. fill out yaml according to your app
Or is there a better way?

"https://www.zaproxy.org/docs/desktop/addons/automation-framework/job-addons/"
- ok we can install an addon, but how can we use it? Can i see an example for replacer and formhandler?

[...]

PR #6624, opened 2021-05-25
Use the auto framework for simple baseline scans

psiinon:
Still todo:

- [x] Fix xml report https://github.com/zaproxy/zap-extensions/pull/2967
- [x] Implement json report template https://github.com/zaproxy/zap-extensions/pull/2952
- [x] Return right result code
- [x] Remove temp add-on copy lines

Part of #6461

Depends on: 

- [x] https://github.com/zaproxy/zap-extensions/pull/2944 
- [x] https://github.com/zaproxy/zap-extensions/pull/2949 
- [x] https://github.com/zaproxy/zap-extensions/pull/2970

### DIFF
--- a/docker/zap-baseline.py
+++ b/docker/zap-baseline.py
@@ -43,12 +43,15 @@
 # at the end of each line.
 
 import getopt
+import json
 import logging
 import os
 import os.path
 import sys
 import time
+import yaml
 from datetime import datetime
+from shutil import copyfile
 from zapv2 import ZAPv2
 from zap_common import *
 
@@ -85,7 +88,7 @@ def usage():
     print('    -a                include the alpha passive scan rules as well')
     print('    -d                show debug messages')
     print('    -P                specify listen port')
-    print('    -D                delay in seconds to wait for passive scanning ')
+    print('    -D secs           delay in seconds to wait for passive scanning ')
     print('    -i                default rules not in the config file to INFO')
     print('    -I                do not return failure on warning')
     print('    -j                use the Ajax spider in addition to the traditional one')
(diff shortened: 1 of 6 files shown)


psiinon:
I think the package scan test failures are ok as the images dont have the yaml package installed as thats added by this PR.

psiinon:
Changed to currently only us the AF is `--auto` is used - we can then ask people to try it out and switch the AF to be on by default if we dont get any complaints.

Ready for review now.

docker/zap-baseline.py
```diff
@@ -85,21 +88,59 @@ def usage():
...
     print('    -n context_file   context file which will be loaded prior to spidering the target')
     print('    -p progress_file  progress file which specifies issues that are being addressed')
     print('    -s                short output format - dont show PASSes or example URLs')
-    print('    -T                max time in minutes to wait for ZAP to start and the passive scan to run')
+    print('    -T mins           max time in minutes to wait for ZAP to start and the passive scan to run')
     print('    -U user           username to use for authenticated scans - must be defined in the given context file')
     print('    -z zap_options    ZAP command line options e.g. -z "-config aaa=bbb -config ccc=ddd"')
     print('    --hook            path to python file that define your custom hooks')
+    print('    --auto            use the automation framework if supported for the given parameters (this will become the default soon)')
+    print('    --autooff         do not use the automation framework even if supported for the given parameters')
```
ricekot, line 104: I feel that only one of the two options is needed - the automation framework is not used when `--auto` isn't specified, which is the same as `--autooff`, but I may be wrong.

docker/zap-baseline.py
```diff
@@ -271,23 +330,119 @@ def main(argv):
                     in_progress_issues[issue["id"]] = issue
 
     if running_in_docker():
```
ricekot, line 332: IMO we should also allow using the automation framework when we are not `running_in_docker()` (and start a container ourselves, as was done before).

docker/zap-baseline.py
```diff
@@ -271,23 +330,119 @@ def main(argv):
...
+                    # Write the yaml file to the mapped directory, if there is one
+                    copyfile(yaml_file, '/zap/wrk/zap.yaml')
+ 
+            # Run ZAP inline with the yaml file
+            try:
+                params = ['-autorun', yaml_file]
+    
+                add_zap_options(params, zap_options)
+    
+                out = run_zap_inline(port, params)
```
ricekot, line 392: Could also use the existing `start_zap()` function and read from `zap.out`?

docker/zap-baseline.py
psiinon: This is true at an given time, but the plan is we have a phased approach.
To start with its off, so you will need to use `--auto` to enable. Strictly speaking we wont need `--autooff`.
Then we will switch to being on by default, then you will need `--autooff` to disable it, at which point we can remove `--auto`.
However I added both at the start so that if someone tries it out and finds a problem then they can seet `--autooff` right now and know that their jobs will keep working no matter when we switch over. Then they can retest at their leisure.

docker/zap-baseline.py
psiinon: No, because thats starts ZAP in the background and we wont know when to stop it. For the AF its better to run inline.

docker/zap-baseline.py
psiinon: We do allow the AF to be used outside of docker, but then you wont be using the packaged scans :)

psiinon:
Tests fixed, thanks to @thc202 :D

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #6625, opened 2021-05-25
Deprecate core/view/homeDirectory API endpoint

ricekot:
Also change the default value of OptionsParam#userDirectory from null
to the user's home directory.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/model/OptionsParam.java
+++ b/zap/src/main/java/org/parosproxy/paros/model/OptionsParam.java
@@ -42,6 +42,7 @@
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
+// ZAP: 2021/05/25 Change the default value of userDirectory from null to the user's home directory.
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/model/OptionsParam.java
```diff
@@ -92,7 +93,8 @@
     private Vector<AbstractParam> paramSetList = new Vector<>();
     private Map<Class<? extends AbstractParam>, AbstractParam> abstractParamsMap = new HashMap<>();
     private boolean gui = true;
-    private File userDirectory = null;
+    // ZAP: Change the default value of userDirectory from null.
```
kingthorin, line 96: Move this to L44/45?

ricekot: +1 for moving to L44/45. The deprecation and code changes are part of another class, would it be okay to include those here?

thc202: Not necessary.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thanks

Issue #6656, opened 2021-06-19
Charset UTF-8 is not always considered


seelmann:
**Describe the bug**
Sending an HTTP request with `Content-Type: application/json;charset=UTF-8` doesn't encode the body as UTF-8.

**To Reproduce**
1. Send a request containing only ASCII characters in the JSON body

    Open Tools -> "Manual Request Editor", enter the following header and body, then click "Send"
    
    ```
    POST https://petstore3.swagger.io/api/v3/pet HTTP/1.1
    Content-Type: application/json;charset=UTF-8
    
    {"id": 10, "name": "doggie", "photoUrls": ["string"]}
    ```
    
    Response: HTTP 200

2. Send a request containing non-ASCII characters in the JSON body 

    Open Tools -> "Manual Request Editor", enter the following header and body (note that the "o" was replaced with a degrees symbol "°"), then click "Send"
    
    ```
    POST https://petstore3.swagger.io/api/v3/pet HTTP/1.1
    Content-Type: application/json;charset=UTF-8
    
    {"id": 10, "name": "d°ggie", "photoUrls": ["string"]}
    ```
    
    Response: HTTP 400
    
    Inspecting the request via the "History" tab shows the following request was sent, the degress symbol was not correctly encoded:
    
    ```
    POST https://petstore3.swagger.io/api/v3/pet HTTP/1.1
    Content-Type: application/json;charset=UTF-8
    Content-Length: 53
    Host: petstore3.swagger.io
    
    {"id": 10, "name": "d�ggie", "photoUrls": ["string"]}
    ```

3. And now the weird part

    From "History" tab select the failed request, right click, select "Open/Resend with Request Editor...". In the opened "Manual Request Editor" dialog enter the header and body from 2. again, then click "Send"
    
    ```
    POST https://petstore3.swagger.io/api/v3/pet HTTP/1.1
    Content-Type: application/json;charset=UTF-8
    
    {"id": 10, "name": "d°ggie", "photoUrls": ["string"]}
    ```
    
    Response: HTTP 200
    
    Inspecting the request via the "History" tab shows that this time the degress symbol was not correctly encoded. Aslo note the changed Content-Length from 53 to 54 due to the correct encoding.
    
    ```
    POST https://petstore3.swagger.io/api/v3/pet HTTP/1.1
    Content-Type: application/json;charset=UTF-8
    Content-Length: 54
    Host: petstore3.swagger.io
    
[...]

PR #6657, opened 2021-06-21
Use Content-Type charset in HTTP body text views

thc202:
Set the request/response body through the `HttpMessage` to take into
account the charset defined in the Content-Type header.

Part of #6656.


psiinon (approved):

ricekot (approved):

ricekot:
Thank you!

Issue #6665, opened 2021-07-03
blank window with tiling window manager


knassar702:
hello everyone , im currently using [DWM](https://dwm.suckless.org/) this a tiling window manager when run zaproxy starts with blank screen

![screenshot-FZsLNH8sAbly3](https://user-images.githubusercontent.com/45688522/124339007-5660b100-dbab-11eb-8ecb-870281c8477d.png)

for fix that add this line to `zap.sh` file

```
export _JAVA_AWT_WM_NONREPARENTING=1
```
* https://github.com/swaywm/sway/issues/595

best regards
@knassar702

kingthorin:
Thanks this is covered in the reference our FAQ refers to: https://www.zaproxy.org/faq/why-am-i-getting-blank-zap-windows-on-linux/

knassar702:
hi @kingthorin , why don't we just add this to `zap.sh` file ?

psiinon:
@knassar702 add what to the zap.sg file? That FAQ links to a page containing a set of suggestions. Which one(s) do you think we should be addins?

psiinon:
Oh, the export you mentioned above ;) I dont have a problem with that, as long as it doesnt cause any other problems. Submit a PR?

knassar702:
@psiinon 
Done :D

PR #6670, opened 2021-07-05
fix blank window in linux

knassar702:
for: #6665


kingthorin:
What platforms/guis has this been tested on?

knassar702:
> What platforms/guis has this been tested on?
#### Operating system : 
* [Void Linux](https://voidlinux.org)
* [Arch Linux](https://archlinux.org)
* [Ubuntu](https://ubuntu.com/)

#### Window Managers :
* [i3-gaps](https://github.com/Airblader/i3)
* [DWM](https://dwm.suckless.org)
* [BSPWM](https://github.com/baskerville/bspwm)

kingthorin:
Wasn’t tested on a non-tiling manager?

knassar702:
yes , with [MATE](https://mate-desktop.org/) and [XFCE](https://www.xfce.org/)

kingthorin:
Great!

kingthorin (approved):

thc202:
IMHO it's not correct saying that all WMs are non-reparenting (if you look at the code there are differences in how both are handled), it would be better if it was set only when really needed.

kingthorin:
@knassar702 are you able to add the logic @thc202 mentioned?

knassar702:
hi @kingthorin, I think this does not make an issue on another Desktop ENV or Window manager, as I said before I test it on different Linux distros and WMS & Desktop Environments and i didn't get any problem !

Issue #6516, opened 2021-04-05
False-positive: 10053 Apache Range Header DoS (CVE-2011-3192) for nginx


rugk:
**Describe the bug**
Apache Range Header DoS (CVE-2011-3192) is reported on a Docker file with an nginx server, just because some header "HTTP/1.1 206 Partial Content" is returned.
This is a false-positive.

**To Reproduce**
Run the ZAP stable Docker image (full scan) against our project PrivateBin (docker image `privatebin/nginx-fpm-alpine`).
Here via GitHub:
https://github.com/PrivateBin/docker-nginx-fpm-alpine/actions/runs/716506686

See also attached HTML report there.

**Expected behavior**
10053 should not be reported.

**Screenshots**
<details>

Medium (Medium) | Apache Range Header DoS (CVE-2011-3192)
-- | --
Description | The  byterange filter in earlier versions of the Apache HTTP Server allows  remote attackers to cause a denial of service (memory and CPU  exhaustion) via a Range request header that identifies multiple  overlapping ranges. This issue was exploited in the wild in August 2011.
 
URL | http://localhost:8080/js/showdown-1.9.1.js
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/robots.txt
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/css/bootstrap/bootstrap-theme-3.4.1.css
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/css/bootstrap/privatebin.css?1.3.4
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/js/privatebin.js?1.3.4
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/img/safari-pinned-tab.svg?1.3.4
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/js/kjua-0.6.0.js
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/css/bootstrap/bootstrap-3.4.1.css
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/js/prettify.js?1.3.4
Method | GET
Evidence | HTTP/1.1 206 Partial Content
URL | http://localhost:8080/js/rawinflate-0.3.js
Method | GET
[...]

PR #6687, opened 2021-07-13
update: scanners.md deprecate Apache Range Header DoS

kingthorin:
Fixes zaproxy/zaproxy#6516


thc202 (approved):

psiinon (approved):

psiinon:
We should also update https://www.zaproxy.org/docs/alerts/10053/ eg as per https://www.zaproxy.org/docs/alerts/10016/ - I think this has to be done manually?

Issue #6691, opened 2021-07-16
Do not add zero Content-Length by default in GET requests


Sravan-Apps:
**Describe the bug**
ZAP Automated scan (Spider and Active Scan) sends Request header  `Content-Length: 0` in case of GET requests. This can trigger some servers or WAF's and block the ZAP scan.

**To Reproduce**
1. Start zap using `java -jar zap-2.10.0.jar`
2. Use netcat to listen on tcp port 7777 `nc -4lkp 7777` from a terminal
3. Now in ZAP Click on `Automated scan` icon. In `Url to Attack` enter `http://127.0.0.1:7777`  and click `Attack` button. 
4. Go to terminal window where netcat is listening and see the request sent from ZAP.

**Expected behavior**
`Content-Length: 0` Header will not be added for GET requests.
```
GET / HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:82.0) Gecko/20100101 Firefox/82.0
Pragma: no-cache
Cache-Control: no-cache
Host: 127.0.0.1:7777
```

Per RFC we should not add Content-Length header for GET,HEAD,TRACE methods. https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
```
A user agent SHOULD NOT send a Content-Length header field when the request message does not contain
a payload body and the method semantics do not anticipate such a body.
```

Content-Length is valid only  for POST,PUT, PATCH, DELETE methods.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

**Screenshots**
```
$ nc -4lkp 7777
GET / HTTP/1.1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:82.0) Gecko/20100101 Firefox/82.0
Pragma: no-cache
Cache-Control: no-cache
Content-Length: 0
Host: 127.0.0.1:7777
```

**Software versions**
 - ZAP:  2.10.0
 - Add-on: quickstart and spider
 - OS: Debian 10
 - Java: openjdk 11.0.11
 - Browser: NA

**Errors from the zap.log file**
NA

**Additional context**
After looking at code I found that this is happening due to code [here](https://github.com/zaproxy/zaproxy/blob/v2.10.0/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java#L245).

From git history I found that the code is there since `v1.2.0` and this was probably carried over from Paros. 

[...]

PR #6692, opened 2021-07-16
Avoid setting Request Header Content-length:0 in GET requests 

Sravan-Apps:
Per RFC we should not add Content-Length header for GET,HEAD,TRACE methods. 
https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2

Content-Length is valid only  for POST,PUT, PATCH, DELETE methods.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Method

This addresses https://github.com/zaproxy/zaproxy/issues/6691 and is similar to https://github.com/zaproxy/zaproxy/pull/4593

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
+++ b/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
@@ -60,6 +60,7 @@
 // isSpecificType(Pattern).
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
 // ZAP: 2021/05/10 Use authority for CONNECT requests.
+// ZAP: 2021/07/16 Issue 6691: Do not add zero Content-Length by default in GET requests
 // ZAP: 2021/07/19 Include SVG in isImage().
 package org.parosproxy.paros.network;
 
@@ -244,11 +245,6 @@ public class HttpRequestHeader extends HttpHeader {
         }
 
         setHeader(ACCEPT_ENCODING, null);
-
-        // ZAP: changed from method to version
-        if (version.equalsIgnoreCase(HTTP11)) {
-            setContentLength(0);
-        }
     }
 
     /**
--- a/zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
@@ -209,6 +209,16 @@ class HttpRequestHeaderUnitTest {
         assertThat(header.getHeaderValues(HttpHeader.COOKIE), hasSize(1));
     }
 
+    @Test
+    void shouldNotHaveContentLengthHeaderByDefault() throws Exception {
+        // Given / When
+        URI uri = new URI("http://example.com", true);
+        HttpRequestHeader header =
+                new HttpRequestHeader(HttpRequestHeader.GET, uri, HttpHeader.HTTP11);
+        // Then
+        assertThat(header.getHeaderValues(HttpHeader.CONTENT_LENGTH), is(empty()));
+    }
+
     private static Stream<Arguments> falseTestCssUrls() {
         return falseTestUrls("css");
     }


zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
```diff
@@ -243,8 +244,12 @@ public HttpRequestHeader(String method, URI uri, String version)
...
-        // ZAP: changed from method to version
-        if (version.equalsIgnoreCase(HTTP11)) {
+    // Content-Length is valid only  for POST,PUT, PATCH, DELETE methods
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
+    // Per RFC we should not add this header for other methods
+    // https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
+    if (method.equalsIgnoreCase(POST) || method.equalsIgnoreCase(PUT) || 
+        method.equalsIgnoreCase(PATCH) || method.equalsIgnoreCase(DELETE)) {
             setContentLength(0);
         }
```
thc202, line 254: I'd suggest just remove it, the components sending bodies already add/update the Content-Length when needed (if not they should).

thc202:
- It should have tests.
 - The DCO check is failing.

zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
Sravan-Apps: Thanks for the suggestion.. Done!

Sravan-Apps:
>     * It should have tests.
> 
>     * The DCO check is failing.

Added Unit Test case.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,6 +204,17 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
         assertThat(header.getHeaderValues(HttpHeader.COOKIE), hasSize(1));
     }
 
+    @Test
+    void shouldNotHaveContentLengthHeaderIfGetRequest() throws Exception {
```
thc202, line 208: `shouldNotHaveContentLengthHeaderByDefault`, it's the same behaviour regardless of the method.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,6 +204,17 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+        // Given
```
thc202, line 209: `// Given / When`

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,6 +204,17 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+        HttpRequestHeader header =
+                new HttpRequestHeader("GET / HTTP/1.1\r\nHost: example.com\r\n\r\n");
```
thc202, line 211: Should use the constructor `HttpRequestHeader(String, URI, String)`.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,6 +204,17 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+        //When
+        boolean empty = header.getHeaderValues(HttpHeader.CONTENT_LENGTH).isEmpty();
+        // Then
+        assertThat(empty, is(equalTo(true)));
```
thc202, line 215:
```Java
// Then
assertThat(header.getHeaderValues(HttpHeader.CONTENT_LENGTH), is(empty()));
```

thc202:
You don't need to reference me in the commit message, in fact, it would be better if you don't (I end up being spammed with notifications each time those commits are pushed).

The changes could be amended.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
Sravan-Apps: done

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
Sravan-Apps: done

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
Sravan-Apps: done

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
Sravan-Apps: done

kingthorin (commented):
Seems good to me, just needs conflicts addressed. Let us know if you need help.

thc202:
Fixed up the commits.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

Issue #6697, opened 2021-07-20
PII Disclosure false positive in SVG image


mig5:
**Describe the bug**

In https://github.com/zaproxy/zaproxy/issues/6288 and the corresponding PR https://github.com/zaproxy/zap-extensions/pull/2825 , you added rules to skip PII Disclosure false positives on CSS files.

I have just experienced the same issue, but in an SVG image.

**To Reproduce**

Run a scan against 	https://mydex.org/resources/papers/AchievingTransformationAtScale/mydex-ecosystem.svg (I am using owasp/zap2docker-live Docker image)

**Expected behavior**

The string '5798064304462212' and '4749835958005446' ought not to be detected as credit cards in the above SVG because the real values in the SVG are actually this:

```
747.5798064304462 212.30854248687663
```

```
871.4749835958005 446.74815452755905
```

Given that part of the matching string begins in a new set of digits beyond whitespace, along with decimal points, I don't think that this should be considered a match (although I understand that spaces are common in credit card number representation, but decimal points are not).

It seems to me that, like CSS, it seems unlikely that an SVG image would contain a credit card, but this might be disputable.

**Screenshots**

![Screenshot_2021-07-20_09-46-21](https://user-images.githubusercontent.com/99173/126241592-43ed5993-f527-4f1a-8cb2-0953cf15d5ea.png)

**Software versions**
 - ZAP: (Docker) owasp/zap2docker-live:latest Digest:sha256:ab4ca81362d9003efa9db59674b2bb4b401268bc6084730c028a63531b7fc046
 - OS: Ubuntu 18.04

**Would you like to help fix this issue?**

Sorry, I can't help fix this issue.

kingthorin:
Related: https://github.com/zaproxy/zaproxy/issues/6639

thc202:
Worth noting that this is fixed only with weekly/2.11.0.

kingthorin:
It should be >80% or as-is since on responses the isImage check just looks for “image” in content-type. The only cases missed will be SVGs served with a non-standard content-type

thc202:
True, was focusing just on the request part.

PR #6698, opened 2021-07-20
improve: HttpRequestHeader: now includes svg in isImage()

kingthorin:
- HttpRequestHeader > Updated image pattern to include svg, added zap comment.
- HttpRequestHeaderUnitTest > Added tests for isImage method.

Related to zaproxy/zaproxy#6697

### DIFF
--- a/zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
@@ -26,13 +26,18 @@ import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.TreeSet;
+import java.util.stream.Stream;
 import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 /** Unit test for {@link HttpRequestHeader}. */
 class HttpRequestHeaderUnitTest {
@@ -204,20 +209,12 @@ class HttpRequestHeaderUnitTest {
         assertThat(header.getHeaderValues(HttpHeader.COOKIE), hasSize(1));
     }
 
+    private static Stream<Arguments> falseTestCssUrls() {
+        return falseTestUrls("css");
+    }
+
     @ParameterizedTest
-    @ValueSource(
-            strings = {
-                "http://example.org/css/file.ext", // In directory path
-                "http://ericsson.com/", // In domain name
-                "https://example.css", // In domain extension (TLD)
-                "https://example.css/dir/file.ext", // In domain extension (TLD)
-                "https://example.org/dir/file?foo=bar&thing=css", // In parameter value
-                "http://example.org/css/file.ext?foo=bar&type=.css", // In parameter value including
-                // period
-                "http://example.org/css/file.ext?foo=bar&thing=styles.css", // In parameter value,
-                // plausible filename
-                "https://example.org/dir/file?foo=bar&css=file.ext" // In parameter name
-            })
+    @MethodSource("falseTestCssUrls")
     void isCssShouldReturnFalseWhenUrlDoesNotIndicateCss(String url) {
         // Given
         HttpRequestHeader reqHeader = createRequestHeader(url);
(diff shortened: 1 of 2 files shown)


psiinon (approved):

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -240,6 +240,48 @@ void isCssShouldReturnTrueWhenUrlIndicatesCss(String url) {
...
+                "https://example.org/dir/file?foo=bar&thing=image", // In parameter value
+                "https://example.org/dir/file?foo=bar&thing=gif", // In parameter value
+                "http://example.org/image/file.ext?foo=bar&type=.image", // In parameter value
+                // including period
+                "http://example.org/image/file.ext?foo=bar&type=.gif", // In parameter value
+                // including period
+                "http://example.org/image/file.ext?foo=bar&thing=image.gif", // In parameter value,
+                // plausible filename
+                "https://example.org/dir/file?foo=bar&image=file.ext", // In parameter name
+                "https://example.org/dir/file?foo=bar&image=file.gif" // In parameter name and value
```
thc202, line 260: Since the data is the same for CSS it would be better if these were extracted into a method (that allows to specify the extension) to avoid duplicating them (same for following test).

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -240,6 +240,48 @@ void isCssShouldReturnTrueWhenUrlIndicatesCss(String url) {
...
+    @ParameterizedTest
+    @ValueSource(
+            strings = {
+                "http://example.org/example.svg", // In path
+                "http://example.org/assets/images/example.gif", // In deeper path
+                "http://example.org/images/example.png?foo=bar", // In path, ignoring params
+                "http://example.org/images/example.jpg?foo=bar&thing=.jpg", // In path, ignoring
+                // params
+            })
+    void isImageShouldReturnTrueWhenUrlIndicatesImage(String url) {
```
thc202, line 278: Ideally this should check all extensions not just some in some places.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
kingthorin: I dont disagree but it isn't just the extensions, some are paths etc. Also I believe I came up with one or two additional test values for isImage

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
kingthorin: I'll see what i can come up with.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
thc202: I noticed but "image" is not relevant (at least for the code, fine if you want to test the name itself, can be configured in that case).

kingthorin: I'd mainly included that because the repose isImage uses that as it's pattern

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,42 +209,140 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+    @ParameterizedTest
+    @MethodSource("trueTestImageUrls")
+    void isImageShouldReturnTrueWhenUrlIndicatesImage(String url) {
+        // Given
+        HttpRequestHeader reqHeader = createRequestHeader(url);
+        // When / Then
+        assertTrue(reqHeader.isImage());
+    }
+
+    private static Stream<Arguments> falseTestUrls(String pathPart, String extension) {
```
kingthorin, line 267: I wasn't sure on the name of `pathPart`, it's often the path but not always. I had bothered to include it as the response header check use those strings. Maybe it isn't worth it? Obviously separate code 🤷

thc202: I don't think it's necessary since it's not relevant for the code being tested.

kingthorin: Okay I’ll simplify it and remove those parts

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,42 +209,117 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+    @MethodSource("trueTestImageUrls")
+    void isImageShouldReturnTrueWhenUrlIndicatesImage(String url) {
+        // Given
+        HttpRequestHeader reqHeader = createRequestHeader(url);
+        // When / Then
+        assertTrue(reqHeader.isImage());
+    }
+
+    private static Stream<Arguments> falseTestUrls(String extension) {
+        List<Arguments> urls = new ArrayList<Arguments>();
```
thc202, line 268: Remove type argument (same for other method).

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,42 +209,117 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+        urls.add(arguments("http://example.org/" + extension + "/file.ext")); // In directory path
```
thc202, line 269: This is from existing code but it's better if the comments are added in the previous line to avoid the formatting breaks.

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -204,42 +209,117 @@ void shouldReplaceAnyCookieHeaderWhenSettingCookieParams() {
...
+        urls.add(
+                arguments(
+                        "https://example.org/dir/file?foo=bar&"
+                                + extension
+                                + "=file."
+                                + extension)); // In parameter name and value
+        return urls.stream();
+    }
+
+    private static Stream<Arguments> trueTestUrls(String[] exts) {
```
thc202, line 302: `String...` and avoid explicit array creation.

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

PR #6728, opened 2021-08-07
FIX HTML RENDER Error for Browser View addon

ErvisTusha:


thc202:
That needs to be done conditionally based on the Java version, the minimum is Java 8.

kingthorin:
@ErvisTusha do you intend to finish this?

kingthorin:
@ErvisTusha do you intend to finish this?

thc202:
Maybe we should document this in the error message or help, with JVM option it's now easier to set the args needed. (This would also have to use the correct location of JavaFX.)

thc202:
Was this closed on purpose?

PR #6730, opened 2021-08-09
i18n(ES) change Lenguage to Idioma

joebordes:
Idioma seems better suited but this called my attention because Lenguage is misspelled, it should be Lenguaje


thc202:
Translations need to be done through Crowdin: https://crowdin.com/project/owasp-zap
The file will be overwritten in the next sync.

joebordes:
done

thc202:
Thank you!

zap/src/main/dist/lang/Messages_es_ES.properties
```diff
@@ -2199,7 +2199,7 @@ target.allInScope=Todo lo que se encuentre en el alcance
 target.empty=Vac\u00edo
 
 technologies.db=DB
-technologies.lang=Lenguage
+technologies.lang=Idioma
```
thc202, line 2202: Just a note, I don't think Idioma fits here, this is for programming languages.

PR #6746, opened 2021-08-11
Update localized resources

zapbot:
Update resources from Crowdin.

PR #6750, opened 2021-08-11
Normalise line endings of bat file

thc202:
Normalise the line endings of `gradlew.bat`.

---
Extracted from #6746.


thc202:
The file has been this way since when it was introduced (checked the history), not sure how it was missed when all files were normalised but it was. They were already crlf when checked out but not stored that way in Git, I think. I checked zap-extensions and it was ok.

kingthorin:
Okay, thanks

kingthorin (approved):

ricekot (approved):

PR #6753, opened 2021-08-13
Update Host header in place

thc202:
Keep the Host header in the same place as it was found when updating its
value to match the host of the connection.


kingthorin (approved):

kingthorin:
Thanks!

thc202:
Tweaked the tests to make clear the expected host header.

kingthorin (approved):

ricekot (approved):
Thank you!

PR #6763, opened 2021-08-18
Register IDs for Community Scripts active scan rules

amankumarchagti:


kingthorin:
Re: The DCO check we can set it to pass if you can just confirm that you agree with https://www.eclipse.org/legal/DCO.php

psiinon:
FYI I'm guiding @amankumarchagti on this - next task will be to use these in the scan rules :D

amankumarchagti:
I agree to DCO

kingthorin:
Setting DCO pass.
image

kingthorin (approved):

thc202 (approved):

PR #6776, opened 2021-08-26
Use only zapbot credentials in release workflows

thc202:
Do not persist default credentials and use only zapbot's ones.


ricekot (approved):

ricekot:
Thank you!

PR #6792, opened 2021-09-07
improve: ZapNumberSpinner: highlight value on focus

kingthorin:
When a number spinner gets focus (ex: is tabbed to) select the text so that the value can be easily set by typing.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
+++ b/zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
@@ -19,14 +19,18 @@
  */
 package org.zaproxy.zap.utils;
 
+import java.awt.event.FocusAdapter;
+import java.awt.event.FocusEvent;
 import java.text.AttributedCharacterIterator;
 import java.text.FieldPosition;
 import java.text.Format;
 import java.text.NumberFormat;
 import java.text.ParseException;
 import java.text.ParsePosition;
+import javax.swing.JFormattedTextField;
 import javax.swing.JSpinner;
 import javax.swing.SpinnerNumberModel;
+import javax.swing.SwingUtilities;
 import javax.swing.text.DefaultFormatterFactory;
 import javax.swing.text.NumberFormatter;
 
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
```diff
@@ -55,6 +59,19 @@ public ZapNumberSpinner(int minValue, int defaultValue, int maxValue) {
                 .getTextField()
                 .setFormatterFactory(
                         new DefaultFormatterFactory(new ZapNumberFormatter(minValue, maxValue)));
+        DefaultEditor editor = (DefaultEditor) getEditor();
+        editor.getTextField()
```
thc202, line 63: Better create a variable with the above text field and use the same for the two tweaks.

zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
```diff
@@ -55,6 +59,19 @@ public ZapNumberSpinner(int minValue, int defaultValue, int maxValue) {
...
+                .addFocusListener(
+                        new FocusAdapter() {
+                            @Override
+                            public void focusGained(final FocusEvent e) {
```
thc202, line 67: Remove final.

zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
```diff
@@ -55,6 +59,19 @@ public ZapNumberSpinner(int minValue, int defaultValue, int maxValue) {
...
+                                SwingUtilities.invokeLater(
+                                        () -> {
+                                            JTextField tf = (JTextField) e.getSource();
+                                            tf.selectAll();
```
thc202, line 71: This could use the new variable (which should make the lambda one statement).

kingthorin: Okay I switched it to:
```java
                                SwingUtilities.invokeLater(
                                        () -> {
                                            textField.selectAll();
                                        });
```
But then sonar wants me to use a method reference. If I switch it to `JTextField::selectAll` then eclipse says that params aren't acceptable for invoke later... 🤷

kingthorin: Will just leave it as a lambda for now.

thc202: `textField::selectAll`

zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
thc202: With above I was referring to `((NumberEditor) getEditor()).getTextField()`.

kingthorin:
Fixed

zap/src/main/java/org/zaproxy/zap/utils/ZapNumberSpinner.java
```diff
@@ -51,10 +56,17 @@ public ZapNumberSpinner(int minValue, int defaultValue, int maxValue) {
             this.defaultValue = defaultValue;
         }
         setModel(new SpinnerNumberModel(this.defaultValue, minValue, maxValue, 1));
-        ((NumberEditor) getEditor())
-                .getTextField()
+        JTextField textField = ((NumberEditor) getEditor()).getTextField();
```
thc202, line 59: Could be `JFormattedTextField` and avoid the cast.

kingthorin: DOH

kingthorin:
Ok, tweaked that too 👍 

Thanks for the reviews 🏅

thc202 (approved):

thc202:
Thank you!

ricekot (approved):

PR #6801, opened 2021-09-13
Allow code to set the ZAP exit status

psiinon:
This is principly for the automation framework when running in cmdline mode, but any other add-on can use it as well if needed.

One question - should we prevent code from (re)setting the exit value to zero?

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/control/Control.java
+++ b/zap/src/main/java/org/parosproxy/paros/control/Control.java
@@ -84,6 +84,7 @@
 // ZAP: 2020/11/23 Allow to initialise the singleton with an ExtensionLoader for tests.
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
 // ZAP: 2021/05/14 Remove empty statement.
+// ZAP: 2021/09/13 Added setExitStatus.
 package org.parosproxy.paros.control;
 
 import java.awt.Desktop;
@@ -125,6 +126,7 @@ public class Control extends AbstractControl implements SessionListener {
     private MenuToolsControl menuToolsControl = null;
     private SessionListener lastCallback = null;
     private Mode mode = null;
+    private int exitStatus = 0;
 
     private Control(Model model, View view) {
         super(model, view);
@@ -313,7 +315,7 @@ public class Control extends AbstractControl implements SessionListener {
                                 } catch (Throwable e) {
                                     log.error("An error occurred while shutting down:", e);
                                 } finally {
-                                    System.exit(0);
+                                    System.exit(exitStatus);
                                 }
                             }
                         },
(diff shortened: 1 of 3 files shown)


thc202:
That makes sense.

It could accumulate log messages too, and log just before exit (rather than when set) to show as late as possible.

thc202 (approved):

thc202:
Thank you!

zap/src/main/java/org/parosproxy/paros/control/Control.java
```diff
@@ -330,6 +332,29 @@ public void run() {
...
+     * is run in cmdline mode. Any add-on can set an exit status so attempts to reset the status to
+     * zero will be rejected.
+     *
+     * @param exitValue the non zero value ZAP will exit it with
+     * @param logMessage the message that will be logged at info level
+     * @since 2.11.0
+     */
+    public void setExitStatus(int exitStatus, String logMessage) {
+        if (exitStatus == 0) {
+            log.error("Not setting the exit status to zero - culprit: {}", logMessage);
```
kingthorin, line 347: `Now`?

The condition is equal zero so ??

psiinon: Its an error message saying that we're _not_ going to set the status to zero, as thats the default and something else might have already set it to something non zero and we dont want to lose that.

kingthorin: But we're checking that what's being passed is zero, so zero is being set from outside (non-default) but still zero?

Actually I guess it's kind of ambiguous, how does java handle that? Local scope first? Since both the passed var and class var are the same name.

psiinon: It will test the parameter, ie prevent add-ons from resetting the exit status to zero. In order to test the class var you'd need to use this.exitStatus

kingthorin: Ok, I don't see why the caller can't or shouldn't set zero, but fair enough.

kingthorin (approved):

kingthorin:
Are we ignoring the LGTM issue?

thc202:
I am, it's failing in other PRs (e.g. zap-extensions).

kingthorin:
Seems it worked after being kicked. One minor issue to address.

thc202:
You kicked it?? :/

zap/src/main/java/org/parosproxy/paros/control/Control.java
thc202: It's to prevent add-ons from overriding the failure status set by others (0 is the default anyway).

kingthorin:
No I assumed one of you did. Maybe it was something wonky on the service?

thc202:
Seems so.

psiinon:
Hopefully fixed :)

PR #6808, opened 2021-09-14
Add container detection for Docker and Flatpak 

psiinon:
The plan is to disable things like browser launch which we know wont work in containers, but this could also be used in telemetry.
Similar changes will be needed for other containers like Flathub and Snap (although we've not been able to update that for 2.10).

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2021-09-15
+ - Added /zap/container file to make it easier to detect if we are running in a container like docker.
+
 ### 2021-08-11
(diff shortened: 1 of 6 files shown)


kingthorin (approved):
LGTM

psiinon:
For info the Flathub repo is https://github.com/flathub/org.zaproxy.ZAP - we'll probably need to add and install a file there but we'll also need to change the core to look for that file...

docker/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2021-09-14
```
thc202, line 4: Could be updated.

psiinon:
Addressed those comments and added support for flatpak

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #6819, opened 2021-09-21
Enable AF support for the baseline -u config_url option

psiinon:

### DIFF
--- a/docker/integration_tests/baseline_tests.sh
+++ b/docker/integration_tests/baseline_tests.sh
@@ -59,8 +59,8 @@ fi
 rm ~/.ZAP_D/config.xml
 
 echo
-echo "Baseline test 3 (vs example.com with INFO/WARN/FAIL and OUTOFSCOPE set)"
-/zap/zap-baseline.py -t https://www.example.com/ --auto -c configs/baseline3.conf > /zap/wrk/output/baseline3.out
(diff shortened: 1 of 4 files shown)


psiinon:
I think we should ignore th LGTM warning as we'll probably use it in the future...

thc202 (approved):

thc202:
Sounds good to me.

docker/integration_tests/baseline_tests.sh
```diff
@@ -82,7 +82,30 @@ fi
 rm ~/.ZAP_D/config.xml
 
 echo
-echo "TEST Baseline 4 (new vs old) - we expect _some_ differences and this will not fail the whole script"
+echo "Baseline test 4 (vs example.com with INFO/WARN/FAIL and OUTOFSCOPE set via URL)"
+/zap/zap-baseline.py -t https://www.example.com/ --auto -u https://raw.githubusercontent.com/zaproxy/zaproxy/main/docker/integration_tests/configs/baseline3.conf > /zap/wrk/output/baseline4.out
```
ricekot, line 86: We don't need to specify `--auto` now right?

psiinon: Good point, let me try that - it should work and if so I can remove that...

psiinon: Yep, all worked, so removed them :)

ricekot (approved):
Thanks!

PR #6828, opened 2021-09-24
Fix concurrency issue when downloading add-ons

thc202:
Use thread-safe collections to manage the current and completed
downloads, as new downloads might be added while processing the existing
ones which would lead to `ConcurrentModificationException`.

---
Reported in OWASP Slack.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #6834, opened 2021-09-29
Adding Spring as a JAVA Technology

sgerlach:
Signed-off-by: Scott Gerlach <scott.gerlach@stackhawk.com>


thc202 (approved):

sgerlach:
This chicken and egg problem is difficult. I can't add a tech set unless it's in Core but I am not supposed to add TechSets to core without a test that uses it. Is there an example of how to make the rule add the tech set from the rule. Not to mention how difficult it is to work on Core and Extensions at the same time because dependency.

psiinon (approved):

psiinon:
So most of the time extensions just cant use anything thats been added to the core after the last release, because they then wont be able to be used in the current version. Right now its a bit different as we are about to release 2.11.0 so we have been taking snapshots.
But for tech we _should_ be able to add it on the fly, but I havnt tried that in a while...

thc202:
> but I am not supposed to add TechSets to core without a test that uses it. 

You can add them, as long as that's going to be used in a scan rule at some point.

PR #6835, opened 2021-09-30
Active Scan stats

psiinon:

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/PluginStats.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/PluginStats.java
@@ -19,6 +19,8 @@
  */
 package org.parosproxy.paros.core.scanner;
 
+import org.zaproxy.zap.utils.Stats;
+
 /**
  * The stats of a {@link Plugin}, when the {@code Plugin} was started, how many messages were sent,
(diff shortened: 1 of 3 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/PluginStats.java
```diff
@@ -38,11 +42,11 @@
     /**
      * Constructs a {@code PluginStats}.
      *
-     * @param pluginName the name of the plugin.
      * @see #start()
      */
-    PluginStats(String pluginName) {
-        this.pluginName = pluginName == null ? "" : pluginName;
+    public PluginStats(Plugin plugin) {
```
thc202, line 47: Keep package? External code should not need to create instances.

psiinon: Thats weird, I didnt mean to change that :/

psiinon: Fixed.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #6836, opened 2021-09-30
Passive Scan stats

psiinon:
Record the number of alerts each rule raises and use the ID for the times - the stats with the names will be removed in a future release.


psiinon:
Changed to WIP as a realised we should also record the max pscan queue size as a highwater mark...

psiinon:
Done.

thc202 (approved):

kingthorin (approved):

PR #6838, opened 2021-09-30
Scripts stats

psiinon:
Some add-ons will need to be changed in order to record their script calls. Those that are not using the script cache anyway.
All script failures _should_ be recorded, as long as they use the standard methods for handling failures.


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #6841, opened 2021-09-30
Use assertThrows in ApiUtilsUnitTest

mureinik:
Many unit tests in the `ApiUtilsUnitTest` class follow the following pattern:
```java
try {
    // Given
    // Set up some faulty params

    // When
    ApiUtils.getXyzParam(params, name);
} catch (ApiException e) {
    // Then
    assertThat(e.getType(), some_logic)));
}
```

If `getXyzParam` throws an exception, this pattern properly asserts its details, but if the method fails to throw an exception at all
(e.g., due to a bug), no asserting will be called and the test will pass.
E.g., if `getIntParam` is replaced with a single line of `return -1;`, all these tests will pass, although they clearly shouldn't.

A stronger pattern would be to use JUnit Jupiter's `assertThrows`, as the `shouldThrowNullPointerExceptionWhenGettingBooleanFromNullParams` test does.

This patch fixes the other tests and aligns them to a similar implementation as `shouldThrowNullPointerExceptionWhenGettingBooleanFromNullParams`.

### DIFF
--- a/zap/src/test/java/org/zaproxy/zap/utils/ApiUtilsUnitTest.java
+++ b/zap/src/test/java/org/zaproxy/zap/utils/ApiUtilsUnitTest.java
@@ -34,7 +34,7 @@ class ApiUtilsUnitTest {
     private static final String HOST = "example.com";
 
     @Test
-    void shouldThrowNullPointerExceptionWhenGettingIntFromNullParams() throws Exception {
+    void shouldThrowNullPointerExceptionWhenGettingIntFromNullParams() {
         // Given
         JSONObject params = null;
         // When / Then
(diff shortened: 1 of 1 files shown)


kingthorin (approved):
Seems reasonable to me

psiinon (approved):

mureinik:
The CI failed because of the spotless plugin (I had a couple of lines that were too long).

Fixed and pushed an updated patch.

kingthorin:
For future reference running the spotlessApply gradle task will format things as expected.

mureinik:
> For future reference running the spotlessApply gradle task will format things as expected.

Yup, that's exactly what I did :-)

kingthorin (approved):

thc202:
Thank you!

kingthorin:
Thanks!

PR #6853, opened 2021-10-05
Make base_dir in docker scripts configurable

va1entin:
When running ZAP in Docker, the path `/zap/wrk` is hard-coded as `base_dir`. When using CLI options to e.g. create a report, this leads to issues in container environments that deny access to that path.

I encountered the problem in the zap2docker stable image running on OpenShift 4, which does not allow root access within containers. Adding an environment variable to be able change the path and keeping `/zap/wrk` as default makes it work again. I am using /tmp as base_dir for example.
There were a few other issues I had to work around running ZAP on OpenShift but this is the only to require a code change so far. I'll gladly supply a little how-to, if anyone's interested 😃

Looking forward to your feedback!


thc202:
The changelog should be updated.

psiinon:
I'm in favour of the change :)
However these instances should also probably be changed:

- https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L111
- https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L293
- https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L401-L402
- https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L505

docker/zap-api-scan.py
```diff
@@ -392,7 +392,7 @@ def main(argv):
 
         if context_file:
             # handle the context file, cant use base_dir as it might not have been set up
```
psiinon, line 394: Did you notice this comment? Might need to check to see if it has been set up...

va1entin:
> 
> 
> The changelog should be updated.

Done 👍 

> 
> 
> I'm in favour of the change :) However these instances should also probably be changed:
> 
>     * https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L111
> 
>     * https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L293
> 
>     * https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L401-L402
> 
>     * https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L505

Thanks for the hints and feedback! I have updated the instances `/zap/wrk` that I could find in my branch which is based on `develop`. Your links point to `main` - is `develop` actually the correct target branch for this?

psiinon:
We use `main` now :)

docker/zap-api-scan.py
va1entin: I have added a check and logging to deal with that part. It might be a better solution to just set `base_dir ` up here though: https://github.com/zaproxy/zaproxy/blob/main/docker/zap-baseline.py#L171

What's your opinion on that?

va1entin:
With the Automation Framework prominently featured in 2.11, one can adapt the automation plan used by the Python scripts and configure a different target directory for reports.

The Python scripts (actually only baseline as of now) write their automation plan to `/home/zap/zap.yaml` (or `/zap/wrk/zap.yaml` if it exists) upon execution. That plan file can then be adapted to generate a report by adding the following lines

```
- type: report
    parameters:
      reportDir: /my/directory
      template: traditional-html
```
For other classic report formats use: traditional-md, traditional-xml, traditional-json as `template`. Omitting `template` entirely results in a modern HTML report.

See also: https://www.zaproxy.org/docs/desktop/addons/report-generation/automation/

The custom plan can then be executed with `zap-x.sh`

```
/zap/zap-x.sh -cmd -autorun <path to plan>
```

PR #6916, opened 2021-11-08
Allow to require add-ons when starting

thc202:
Some add-ons might be required for ZAP to be actually functional, those
add-ons will be defined in the code and automatically set as required.
When not found ZAP will fail to start, exiting with appropriate error
message.
The extensions of required add-ons will be always enabled.
The required add-ons can't be uninstalled, only updated.


kingthorin (approved):
Seems reasonable to me.

zap/src/main/java/org/parosproxy/paros/control/Control.java
```diff
@@ -139,6 +143,24 @@ private Control() {
...
+                        model.getOptionsParam().getCheckForUpdatesParam().getAddonDirectories());
+        if (overrides != null) {
+            AddOnCollection addOnCollection = addOnLoader.getAddOnCollection();
+            overrides
+                    .getRequiredAddOns()
+                    .forEach(
+                            id -> {
+                                AddOn addOn = addOnCollection.getAddOn(id);
+                                if (addOn == null) {
+                                    String message = "The required add-on was not found: " + id;
```
psiinon, line 157: Maybe add: "If you are building ZAP from the source code then you will need to install the add-on from zap-extensions". Could we have a new task in zap-extensions, like "install-essential"? Or would that be too tricky? ... I guess we could still give the instructions like: "In zap-extensions call 'gradle ...' with the right add-on name...

thc202: I'd prefer update the dev docs instead of hardcode instructions.

No, it's easy to add such task, but we'll need to maintain the list of required add-ons in two places.

psiinon: Could we link to the dev docs then? I agree they are better than hardcoded instructions

thc202: OK.

thc202: Tweaked the message.

psiinon (approved):

kingthorin (approved):

thc202:
Renamed from "required" to "mandatory".

kingthorin (approved):
Still looks good to me.

psiinon (approved):
Me too :)

PR #6924, opened 2021-11-13
Address build deprecations

thc202:
Use properties to set the main classes and if the JaCoCo report should
be generated.


kingthorin (approved):
TY!

psiinon (approved):

Issue #6930, opened 2021-11-16
Load scripts via "zap-baseline.py" script  for CI/CD


Zilborg:
**Is your feature request related to a problem? Please describe.**
For authentication and ANTI-CSRF and etc. scripts loading and check in CI/CD. 

**Describe the solution you'd like**
Add option in `zap-baseline.py` script for loading.

**Would you like to help implementing this feature?**
I've implemented this in pull request and it working for my case.

psiinon:
This is already possible, albeit not as easy as it could be.
I dont think we should be extending the packaged scans for these sort of things - the Automation Framework is how we should support them going forwards: https://www.zaproxy.org/docs/automate/automation-framework/

Zilborg:
In this point (https://www.zaproxy.org/docs/automate/automation-framework/) I don't see authentication and Anti-CSRF options.
Via my wrap I can export scripts from ZAP Desktop to container via baseline and include check from different roles in CI/CD and it does not mean deploy ZAP API like stand alone server.

psiinon:
The AF is being actively developed - initial auth support is here: https://www.zaproxy.org/docs/desktop/addons/automation-framework/environment/
In time it will support all of the auth and ant-csrf features that ZAP supports :)

kingthorin:
Per above and the discussion on the associated PR this seem addressed.

PR #6931, opened 2021-11-16
Added script loading parameter for baseline

Zilborg:
I used it for authentication and csrf scripts.
Issue: #6930

### DIFF
--- a/docker/zap-baseline.py
+++ b/docker/zap-baseline.py
@@ -103,6 +103,7 @@ def usage():
     print('    --hook            path to python file that define your custom hooks')
     print('    --auto            use the automation framework if supported for the given parameters (this is now the default)')
     print('    --autooff         do not use the automation framework even if supported for the given parameters')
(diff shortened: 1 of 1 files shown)


psiinon:
This can already be acheived via the script options. The options support all scriptin g engines, unlike this PR which just supports Zest and Nashorn.
As per https://github.com/zaproxy/zaproxy/issues/6930#issuecomment-970083672 I dont think we should be extending the packaged scripts as they have proved difficult to support and maintain.

psiinon:
Thanks for the PR btw, but we are looking to migrate the packaged scan scripts to use the Authentication Framework not extend them.

psiinon:
To do the same thing via the existing options should be something like: `-z "-config dirs=/zap/wrk/scripts"`

Zilborg:
> To do the same thing via the existing options should be something like: `-z "-config dirs=/zap/wrk/scripts"`

Oh, I see.
Thanks for your time!

Issue #6938, opened 2021-11-17
Snapshot Session As... -- Returns unhelpful message: 'Error saving session snapshot'


davewichers:
I'm using ZAP 2.11.0, and when I invoke Snapshot Session As..., an error dialog comes up and says only: 'Error saving session snapshot'.

This isn't very helpful. Seems like the error dialog should explain why, somehow. Or an error message should appear in the Output tab, but I can't find anything useful.

Any clue what the problem might be? And also why it doesn't provide more info? (Can we improve that?)

I don't have the session Persisted. Do you have to do that first before you can use this feature? If so, then the error dialog should explain that. I have no clue if that is the cause of the problem.

UPDATE: I successfully invoked Persist Session... first, and then tried Snapshot Session As..., and now it works. So if you have to do A before you can use B, then can the error dialog be updated to explain that?  I have no idea if there are ANY other features like this in ZAP, but if there are, all of them should be updated to be more explanatory like this. (Thx)

kingthorin:
Thanks for following up

davewichers:
You're welcome. I know we can improve the error message, but as an alternative, is there a way to make it work without requiring Persist Session be enabled first?  Might be slower, but it should be possible shouldn't it?

kingthorin:
I’ll have to look at the code, I was thinking more that snapshot should be disabled if the session hasn’t already been persisted.

thc202:
It will not be slower, it's just how it was designed, creating snapshots for temporary sessions might not be the best or expected behaviour.

Agreed, disable and inform with tooltip why.

PR #6984, opened 2021-12-16
feat: Snapshot enable/disable and tooltip

kingthorin:
- The snapshot menu item is now only enabled when the session has previously been persisted. It is disabled if the session has not been persisted. Additionally a tooltip with explanation is shown when the menu item is disabled

image

Fixes zaproxy/zaproxy#6938


kingthorin:
Set to WIP. It was pointed out via IRC that API actions which cause this to get out of sync.

zap/src/main/java/org/parosproxy/paros/view/MainMenuBar.java
```diff
@@ -322,11 +328,13 @@ private JMenuItem getMenuFileSaveAs() {
                         public void actionPerformed(java.awt.event.ActionEvent e) {
                             if (Model.getSingleton().getSession().isNewState()) {
                                 getMenuFileControl().saveAsSession();
+                                toggleSnapshotState(true);
                             } else {
                                 View.getSingleton()
                                         .showWarningDialog(
                                                 Constant.messages.getString(
                                                         "menu.file.sessionExists.error"));
+                                toggleSnapshotState(false);
```
thc202, line 337: If it's not new state it means that we can snapshot, right?

kingthorin: Yup I think I set this one backwards.

kingthorin: Actually I think I had this right. If it's in new state the save happens then the button is enabled, else (there was an error, ensure it is still disabled). I guess technically I should recheck the save state after the attempted save?

kingthorin:
Added ZAP comment to paros class.

zap/src/main/java/org/parosproxy/paros/view/MainMenuBar.java
thc202: It's true that this menu shouldn't be enabled in the first place but the snapshot shouldn't be disabled either (it wasn't before these changes).

kingthorin: Okay. So revert the two changes in this method? Im unclear where this is going…

thc202: I think it just needs to change to true (since it's the same for both branches move down).

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

davewichers:
Thanks for this fix! Much appreciated.

PR #7005, opened 2022-01-04
Process triggers and notifications also on exception during ActiveScans sendAndReceive.

denniskniep:
Signed-off-by: Dennis Kniep <kniepdennis@gmail.com>


denniskniep:
ToBeDiscuessed: 
* process triggers and notifications only on specific ExceptionTypes: `SocketTimeoutException`, `IOException` ?
* Set the exception/error message as response body?

kingthorin:
I'd be good with the exceptions being made into the response body (that's done elsewhere IIRC). I wonder if this should be tied to an option though and default off (to maintain closest to current behavior)?

denniskniep:
Discussion from IRC:

- [ ]  have a visual indication - just be an icon for the ones with errors (like the spider has with the red one)
- [ ] provide messages with errors separately, at least for the API/programmatically (way easier to inspect them that way rather than go through thousands of requests)
- [ ] add exceptions to the response body

denniskniep:
@thc202 can you provide more Context regarding
> provide messages with errors separately, at least for the API/programmaticall

Is it only separating for API (/ascan/view/messagesIds/)

thc202:
For the code and the ZAP API, in the `ActiveScan` expose a method that returns a list with the IDs of the messages with errors (similar to 'SpiderScan' which exposes successful and error messages separately). In the ZAP API another endpoint to get those requests.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -309,17 +311,21 @@ protected void sendAndReceive(
...
+        try {
+            if (isFollowRedirect) {
+                parent.getHttpSender()
+                        .sendAndReceive(message, getParent().getRedirectRequestConfig());
+            } else {
+                parent.getHttpSender().sendAndReceive(message, false);
+            }
+        } finally {
+            // ZAP: Notify parent
+            parent.notifyNewMessage(this, message);
```
thc202, line 323: Might be better just notify on `IOException`s.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -309,16 +312,24 @@ protected void sendAndReceive(
...
-            parent.getHttpSender().sendAndReceive(message, false);
+        try {
+            if (isFollowRedirect) {
+                parent.getHttpSender()
+                        .sendAndReceive(message, getParent().getRedirectRequestConfig());
+            } else {
+                parent.getHttpSender().sendAndReceive(message, false);
+            }
+        } catch (IOException e) {
+            // TODO: How to update HistoryRef with error Response?
```
denniskniep, line 323: @thc202 can you give me any hint here?
How can I update the HistoryRef with error Response? In case HttpMessage has already a HistoryRef and is attached to the Alert

denniskniep: Discussed in IRC: No workarounds here. Problem is releated to oast
I guess the way how histRef ist set to HttpMessage is like the following:
.alertFound(alert, null) is triggered by oast here before exception is catched:

https://github.com/zaproxy/zap-extensions/blob/7016bdf3607f31f42ceb3206c86b437131b0452b/addOns/oast/src/main/java/org/zaproxy/addon/oast/ExtensionOast.java#L268

And then a HistRef is created
https://github.com/zaproxy/zaproxy/blob/f68b04e580fac25e2bacecec4bf7365cdc03e5fc/zap/src/main/java/org/zaproxy/zap/extension/alert/ExtensionAlert.java#L189

Which sets automatically the histref member on the Httpmessage

Resulting in that the Histref is set to the message not including the error message which is later set

> we should change oast to allow to manipulate the alert/message before caching it

denniskniep:
Anything else to do here before merge ?

zap/src/main/java/org/zaproxy/zap/extension/callback/ui/DefaultCustomColumnHistoryReferencesTableModel.java
```diff
@@ -25,7 +25,10 @@
 import org.zaproxy.zap.view.table.DefaultHistoryReferencesTableEntry;
 import org.zaproxy.zap.view.table.DefaultHistoryReferencesTableModel;
 
-/** @deprecated (2.11.0) Superseded by the OAST add-on. */
```
psiinon, line 28: Ditto

zap/src/main/java/org/zaproxy/zap/extension/callback/ui/CustomColumn.java
```diff
@@ -19,7 +19,7 @@
  */
 package org.zaproxy.zap.extension.callback.ui;
 
-/** @deprecated (2.11.0) Superseded by the OAST add-on. */
+/** @deprecated (2.11.0) {@link org.zaproxy.zap.view.table.CustomColumn}. */
```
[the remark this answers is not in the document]
denniskniep: This was deprecated once due to moving Oast addon from ZAP to ZAP-Extension.

But I used the original class again. Therefore its there again but within an other package. I thought this info could be useful to know if someone stumbles upon that deprecation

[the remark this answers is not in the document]
psiinon: ok, find by me :)

PR #7047, opened 2022-01-24
Show source of CONNECT requests as proxy

thc202:
Add the proxy CONNECT type to the history type info, to not show them
as "Undefined".


psiinon (approved):

ricekot (approved):
Thanks!

PR #7080, opened 2022-02-09
Deprecate and disable proxy related code

thc202:
No longer manage the main nor additional proxies.
Remove proxy info in the main footer.
Do not parse `-host` and `-port`.
Add listener of break serialise requirement for the proxies in
the add-on.
Suppress deprecation warnings where needed.


thc202:
Rebased to fix the conflict.

kingthorin (approved):

PR #7085, opened 2022-02-12
API - Do not require Authentication Realm and use default value

andrzejressel:
Despite realm being required field in documentation it's not enforced.

That's good, because user may not care about specific realm - unfortunately setting null realm cause hard to debug issue:

- NTCredentials constructor will throw on null realm: https://github.com/zaproxy/zaproxy/blob/8ac4b10d3d9d816957985694c2b7cf127d1364d3/zap/src/main/java/org/zaproxy/zap/authentication/HttpAuthenticationMethodType.java#L168-L173
- This exception will not be printed when using importurls endpoint in daemon mode: https://github.com/zaproxy/zap-extensions/blob/main/addOns/importurls/src/main/java/org/zaproxy/zap/extension/importurls/ExtensionImportUrls.java#L100

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/authentication/HttpAuthenticationMethodType.java
+++ b/zap/src/main/java/org/zaproxy/zap/authentication/HttpAuthenticationMethodType.java
@@ -382,8 +382,8 @@ public class HttpAuthenticationMethodType extends AuthenticationMethodType {
     public ApiDynamicActionImplementor getSetMethodForContextApiAction() {
         return new ApiDynamicActionImplementor(
                 API_METHOD_NAME,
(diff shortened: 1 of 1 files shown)


thc202:
That add-on does not need to be changed, was superseded.

kingthorin:
Exim does already handle this situation by logging the exception: https://github.com/zaproxy/zap-extensions/blob/c9b0b4bff44ebb6261be455810cb3360ce56f6c4/addOns/exim/src/main/java/org/zaproxy/addon/exim/urls/UrlsImporter.java#L102-L107 (as well as updating the output panel).

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #7101, opened 2022-02-23
Workaround issues when displaying big messages

### LINKED ISSUE #5791: Big messages cause GUI to freeze
**Describe the bug**
After doing scans and collecting info on site, it detected 'Application Error Disclosure' (6).  Upon expanding and clicking on a link (javascript file, can't tell how large), it's instantly freezes and nothing else can be done (it may recover after a few minutes wait).  In other parts of app, I've seen the "Very large response body (xxx bytes) - switch views..." warning, but not in this situation.  I have to do a 'killall java' to exit the app and restart.  

On occasion, it will recover after a few minutes, but the app moves slow so I just restart it. When it does recover, the response is unintelligible (see screenshot) and navigating around is slow. 

**To Reproduce**
1. Have JS file (size?) in Alert, Application Error Disclosure 
2. Click on link.
3. App freezes instantly or moves slow.

**Expected behavior**
To see the Request and Response data of the alert.

**Screenshots**
##Click on the 'GET:https://xx/build/js/app.js?v=...
AED-js
##Screen freezes, switching desktop and back shows gray app
goesGray
##If it does recover, Response window looks like
if-it-does-recover

**Software versions**
 - ZAP: 2.8.1
 - OS: BlackArch 5.4.6 x86_64
 - Java: openjdk 13.0.1
 - Hardware:  Happens on both VirtualBox and HD install on laptop.

Will update, don't have currently.

**Additional context**

thc202:
Split long lines and disable line wrap.

Fix #5791.


kingthorin:
Thanks is exciting. Will test it out this evening.

thc202:
Thank you, the values might need tweaking but they seemed to work ok here.

kingthorin (approved):

psiinon (approved):
Wow - this makes such a huge difference!
Only comment - do we still need the "Large Response" option?
We could always keep it but default to -1 ?

kingthorin:
That's a good point, even with Large Response I haven't encountered a message that wasn't displayed awfully fast. I'd be happy for it to be -1 by default.

I just did a bit of quick testing with -1. Even displaying the text representation of a 6.7MB PDF was pretty snappy. A few HTML pages around 1.5MB displayed fine too.

thc202:
Was going to remove it :)

kingthorin:
Okay

thc202:
Applied the same workarounds to the request and removed the large request/response views.

thc202:
Binary compatibility check is not happy…

thc202:
Binary compatibility restored, ready for review.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -3289,8 +3293,6 @@ view.options.label.fontName					= Font Name:
 view.options.label.fontSize         		= Font Size:
 view.options.label.generalFont              = General Font
 view.options.label.workPanelsFont           = Work Panels Font
-view.options.label.largeRequestSize			= Large request view min size:
-view.options.label.largeResponseSize		= Large response view min size:
```
kingthorin, line 3293: What about the label for the view switching dropdown?

kingthorin: OH I guess that stays until we actually remove the functionality.

kingthorin (approved):

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
thc202: Right, that's in case some add-on is using the views (and thus the strings). Alternatively, I could hardcode the strings and remove them from the file.

kingthorin: Naw, that’s just extra work. Thanks for tackling this. Seems to work wonderfully.

psiinon (approved):

PR #7110, opened 2022-03-01
Use Medium For Default Policy Strength / Threshold

ricekot:
Set Default Alert Threshold and Attack Strength to "Medium" when they
are set to "Default".


thc202:
IMO this should be considered an error (and thus throw). In the UI it does not allow to choose "default" for default, we shouldn't in the ZAP API as well.

zap/src/main/java/org/zaproxy/zap/extension/ascan/ScanPolicy.java
```diff
@@ -114,23 +121,30 @@ public void save() throws ConfigurationException {
...
-        if (name.isEmpty()) {
-            return defaultValue;
+    private AlertThreshold getAlertThresholdFromConfig() {
+        String alertThreshold = conf.getString("scanner.level", "");
+        if (alertThreshold.isEmpty()
+                || !EnumUtils.isValidEnum(AlertThreshold.class, alertThreshold)
+                || AlertThreshold.OFF.name().equals(alertThreshold)
+                || AlertThreshold.DEFAULT.name().equals(alertThreshold)) {
+            logger.warn(
+                    "Found illegal value {} for alert threshold, using default (MEDIUM) instead.",
```
kingthorin, line 131: The user can set "default" to any of the regular values, so this is a bit misleading. Perhaps:
```suggestion
                    "Found illegal value {} for alert threshold, using MEDIUM instead.",
```

kingthorin: Something similar for Str too?

ricekot: Thanks, tweaked.

kingthorin (approved):
Thanks

thc202 (approved):

thc202:
Thank you!

PR #7112, opened 2022-03-02
Allow Default threshold OFF

psiinon:
Undoing part of the change done in https://github.com/zaproxy/zaproxy/pull/7110 ;)
But also allowing the option to be set in the GUI.
Its a perfectly valid and very useful option as it allows people to explicitly enable specific rules.


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

Issue #7079, opened 2022-02-09
Get False Positive count using alertCountsByRisk  api


d-apurva:
**Is your feature request related to a problem? Please describe.**
I have automated the process of scanning api and ui using zap api's and Jenkins. The automation is working fine end to end and I am creating a Github issue also attaching the reports generated by ZAP.
In Github issue description I am providing count of Low, High, Medium and Informational alerts using alertCountsByRisk api.
In some scenarios I am marking some of the alerts to False Positive. So now I can see the count of all alerts except for False Positive.
And I have automated script such a way that if all alerts count are zero, Github issue will be closed with a comment that there is no vulnerability, so closing the issue
Now I want to provide the information about False Positive count also.

**Describe the solution you'd like**
Now if I run alertCountsByRisk  api, I get output as
`{"High":0,"Low":0,"Medium":0,"Informational":2}`
But I want output as
`{"High":0,"Low":0,"Medium":0,"Informational”:0, "FalsePositive":2}`

psiinon:
This is where the API endpoint is defined: https://github.com/zaproxy/zaproxy/blob/1f69b4fafbadb341dc95dc56ba1f5f3407ac589d/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java#L75

abhinav-koppula:
Hi,
I am interested in working on this issue. Can you please assign this to me?

d-apurva:
@abhinav-koppula Do we have any update on this?

LakiG:
Hi,
In case the issue is not being worked on, I would be interested in it.

d-apurva:
@LakiG as there is no update from @abhinav-koppula .
So can you please help with this.

LakiG:
Sure
@zap team, is it okay if I can work on this?

thc202:
It would be preferable to finish the other issue first.

thc202:
Reassigned.

PR #7118, opened 2022-03-06
alert: Add false positive count in alertCountsByRisk api endpoint

LakiG:
alertCountsByRisk will first check if the alert is False Positive and add it to the False Positive counter, else it will add to Risk counter.

Fixes https://github.com/zaproxy/zaproxy/issues/7079

Signed-off-by: LakiG <garg.lakshya@gmail.com>

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
@@ -297,7 +297,8 @@ public class AlertAPI extends ApiImplementor {
             boolean recurse = this.getParam(params, PARAM_RECURSE, false);
 
             // 0 (RISK_INFO) -> 3 (RISK_HIGH)
-            int[] counts = new int[] {0, 0, 0, 0};
+            int[] riskCounts = new int[] {0, 0, 0, 0};
+            int falsePositiveCount = 0;
 
             AlertTreeModel model = extension.getTreeModel();
             AlertNode root = (AlertNode) model.getRoot();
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -297,7 +297,9 @@ public JSON toJSON() {
             boolean recurse = this.getParam(params, PARAM_RECURSE, false);
 
             // 0 (RISK_INFO) -> 3 (RISK_HIGH)
-            int[] counts = new int[] {0, 0, 0, 0};
+            int[] riskCounts = new int[] {0, 0, 0, 0};
+            // 0 (CONFIDENCE_FALSE_POSITIVE)
+            int[] confidenceCounts = new int[] {0};
```
thc202, line 302: It could just be `int` and named `falsePositiveCount`. Since the endpoint is about the risks I don't think it's going to return all confidences.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -308,14 +310,21 @@ public JSON toJSON() {
 
                 ApiResponseList alertList = filterAlertInstances(child, url, recurse);
                 if (alertList.getItems().size() > 0) {
-                    counts[alert.getRisk()] += 1;
+                    if (alert.getConfidence() == 0) {
```
thc202, line 313: Use the constant.

thc202:
Worth mentioning the change in the [release notes](https://github.com/zaproxy/zap-core-help/blob/24059345c52da4a129d62d7275ed4d69e2b993fe/addOns/help/src/main/javahelp/contents/releases/2.12.0.html#L40) (around the endpoints sections), this might cause existing workflows to break since the number of the expected alerts might no longer be the same.

thc202 (approved):

thc202:
Thank you!

PR #7131, opened 2022-03-09
Skip empty javascript URIs when canonicalizing

thc202:
Do not try to canonicalize empty javascript URIs, it just leads to
unnecessary log warnings when pages are crawled.
Parameterize related test, add the empty javascript URI and a tel URI
for another example scheme that's ignored.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/spider/URLCanonicalizer.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/URLCanonicalizer.java
@@ -103,6 +103,9 @@ public final class URLCanonicalizer {
      * @return the canonical url
      */
     public static String getCanonicalURL(String url, String baseURL) {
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

kingthorin:
I see we also skip mailto links should we be adding `tel:`?

I know it could become a never-ending battle but I feel like `tel:` might be getting popular enough.

thc202:
We do ignore `tel:` as it doesn't have an authority. I can add it to the test to be clear.

kingthorin:
Sorry, I didn't browse the whole file. I just saw what was included in the diff :wink:

thc202:
No worries, tweaked the test.

kingthorin (approved):

ricekot (approved):
Thank you!

Issue #7174, opened 2022-03-28
Don't log parsing exception when request with json content type has empty body


DiogoMRSilva:
### Is your feature request related to a problem? Please describe.

When a website has many requests with json content type and empty body, the logs are full of json parsing errors. Since this is not a ZAP problem I would like to clean the logs.

### Describe the solution you'd like

On `parseContent` of `VariantJSONQuery.java` return 0 parameters if the request body is empty.

### Describe alternatives you've considered

No alternatives

### Would you like to help fix this issue?

- [X] Yes

PR #7175, opened 2022-03-28
Remove parsing error when json request has empty body

DiogoMRSilva:
Closes https://github.com/zaproxy/zaproxy/issues/7174

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
@@ -73,6 +73,9 @@ public class VariantJSONQuery extends VariantAbstractRPCQuery {
 
     @Override
     public void parseContent(String content) {
+        if (content.isEmpty()) {
+            return;
+        }
         sr = new SimpleStringReader(content);
         parseObject();
     }
--- a/zap/src/test/java/org/parosproxy/paros/core/scanner/VariantJSONQueryUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/core/scanner/VariantJSONQueryUnitTest.java
@@ -85,6 +85,25 @@ class VariantJSONQueryUnitTest {
         assertLogEvent(Level.WARN, "EOF reached while reading JSON field name");
     }
 
+    @Test
+    void shouldNotThrowExceptionIfResponseIsEmpty() throws HttpMalformedHeaderException {
+        // Given
+        withLoggerAppender();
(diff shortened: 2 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
```diff
@@ -74,6 +74,9 @@ public boolean isValidContentType(String contentType) {
     @Override
     public void parseContent(String content) {
         sr = new SimpleStringReader(content);
+        if (content == null || content.length() == 0 || content.strip().length() == 0) {
```
kingthorin, line 77: isEmpty()

kingthorin:
Looks like `strip` is a Java 11 feature?

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
thc202: And is the strip really needed?

thc202: `content` can't be null here.

DiogoMRSilva: will fix it

DiogoMRSilva:
> Looks like `strip` is a Java 11 feature?

yes I think so, will remove it

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantJSONQuery.java
DiogoMRSilva: replaced the whole check with `isBlank()` and did put the `sr` instantiating after the check.

thc202: That method is also Java 11.

kingthorin: To stay Java 8 compatible you'll need to use apache lang or lang3 `StringUtils.isBlank(content)`

thc202: Better the latter.

DiogoMRSilva: So going to leave it with only the `isEmpty` check.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #7177, opened 2022-03-29
Follow unencoded redirects

DiogoMRSilva:
If a redirect location is not encoded zap will throw an exception and not follow the redirect
image
```
HTTP/1.1 302 Found
Date: Tue, 29 Mar 2022 18:47:19 GMT
Server: Apache/2.4.38 (Debian)
Location: ./error.php?msg=Error, need to provide a query to sear
Content-Length: 0
Content-Type: text/html; charset=UTF-8
```

there is an example on the `WackoPicko` `/error?msg=...` endpoint.
The browsers follow this redirects, so I think zap should also support them.

Could add a test but need to change the method permission to access it directly

### DIFF
--- a/zap/src/main/java/org/apache/commons/httpclient/HttpMethodDirector.java
+++ b/zap/src/main/java/org/apache/commons/httpclient/HttpMethodDirector.java
@@ -686,7 +686,11 @@ public class HttpMethodDirector {
 			);
 			
             String charset = method.getParams().getUriCharset();
-            redirectUri = new URI(location, true, charset);
+            try {
+                redirectUri =  new URI(location, true, charset);
+            } catch (URIException ex) {
+                redirectUri =  new URI(location, false, charset);
+            }
 			
             if (redirectUri.isRelativeURI()) {
 				if (this.params.isParameterTrue(HttpClientParams.REJECT_RELATIVE_REDIRECT)) {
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

thc202:
The accessibility does not need to change, refer to tests in `HttpSenderUnitTest`.
The fix is not complete, redirects followed automatically (e.g. `sendAndReceive(HttpMessage)`) will have the same problem.

DiogoMRSilva:
> The accessibility does not need to change, refer to tests in HttpSenderUnitTest.

The changed method is private, do you suggest to do a test on the public method that ends up using it or with something like reflection?

> The fix is not complete, redirects followed automatically (e.g. sendAndReceive(HttpMessage)) will have the same problem.

On `sendAndReceive` the `escaped` parameter on the `URI` constructor is already set to false
```
                String location = temp.getResponseHeader().getHeader(HttpHeader.LOCATION);
                URI baseUri = temp.getRequestHeader().getURI();
                URI newLocation = new URI(baseUri, location, false);
```
Not sure if the place I changed should also be replaced by a single function call with the `escaped` set to false.

> Needs a ZAP comment.

Sry forgot to check.

thc202:
The former, it's the same as other tests in that class.

That's not the redirect followed automatically though, they happen when `sendAuthenticated` is called.
That excerpt can be replaced with a call to `extractRedirectLocation`.

No worries.

thc202:
Don't worry about adding tests, these will be superseded soon. For the automatic redirects, it should be tweaked in `HttpMethodDirector#processRedirectResponse`.

its0x08 (approved):

thc202:
Would be possible to allow to push to the branch? I'd like to address the remaining comments.

DiogoMRSilva:
> Would be possible to allow to push to the branch? I'd like to address the remaining comments.

Yes ofc

thc202:
OK, I can wait then.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thanks

PR #7190, opened 2022-04-07
Populate Find dialog with selected text

psiinon:
In most IDEs and editor apps if you invoke the Find option with any text selected then that text is put in the Find dialog.
Not with ZAP, or at least not without this change :)
If works whether you invoke the Find option via menus or the hotkey.


kingthorin (approved):

zap/src/main/java/org/parosproxy/paros/view/FindDialog.java
```diff
@@ -78,6 +79,10 @@
     /** @param lastInvoker The lastInvoker to set. */
     public void setLastInvoker(JTextComponent lastInvoker) {
         this.lastInvoker = lastInvoker;
+        String selectedText = lastInvoker.getSelectedText();
```
thc202, line 82: The `lastInvoker` might be null.

psiinon: Good point - fixed.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

kingthorin:
Thanks. These small usability tweaks can and do actually make a big difference to user experience :grinning:

PR #7202, opened 2022-04-11
Remove -nouseragent and deprecate related methods

thc202:
Remove the undocumented/Paros command line option `-nouseragent`,
didn't do anything anymore (the user agent was already an empty string
which would have no effect later when attempting to change it).
Deprecate `HttpSender#setUserAgent` and `getUserAgent()`, the user
agent was not being set but appended, also, better alternatives exist
to actually set/change it programmatically (e.g. `HttpSenderListener`).
Remove related instance variable and private method no longer needed.


kingthorin (approved):

psiinon:
Looks like it needs a spotlessApply :)

thc202:
Done :)

psiinon:
Still good :)

PR #7205, opened 2022-04-12
chore: Remove use of "and/or" in user facing strings

kingthorin:
Going along with my old writing rant: http://andore.ddns.net

Remaining occurrences are in the license header `(?<!it )and\/or` or other code comments.


kingthorin:
While I tackled this I also noticed that the API uses "scanners" in a bunch of places it should probably use "scan rules", both in endpoint/param description(s) and in actual endpoints. Think that's work fixing? (I guess the endpoints would have to be deprecated/replaced or maybe again just do the user facing strings?)

thc202:
The referenced link does not work.

kingthorin:
Tweaked :) Apparently it's a lame http link :)

ricekot:
The commit message could also be updated.

thc202:
I'm fine with these changes (IMO languages/English are not boolean arithmetic though), as long as we are consistent throughout. The dev guidelines should be updated.

Don't think worth changing the endpoints.

thc202 (approved):

kingthorin:
I agree it's not boolean logic. But, in like 98% of cases the "or" is really fine, and the "and/or" just looks rushed or something...

kingthorin:
I will tweak the commit message like @ricekot suggested. Just need to finish my work day and do a few family things.

kingthorin:
Tweaked commit message.

psiinon:
I wonder how many of these I added :P 
In any casse I think its good to be consistant and I'm ok with us not using this term :)

psiinon (approved):

PR #7206, opened 2022-04-12
ascan, pscan: Update enable/disable API actions

ricekot:
Update the ascan enableScanners API action to throw ApiExceptions when
the provided scan rule IDs are invalid.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Consumer;
 import java.util.regex.PatternSyntaxException;
 import net.sf.json.JSON;
 import net.sf.json.JSONException;
@@ -294,7 +295,7 @@ public class ActiveScanAPI extends ApiImplementor {
     public ApiResponse handleApiAction(String name, JSONObject params) throws ApiException {
         log.debug("handleApiAction " + name + " " + params.toString());
         ScanPolicy policy;
-        int policyId;
+        int categoryId;
 
         User user = null;
         Context context = null;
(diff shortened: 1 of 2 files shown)


kingthorin (approved):

ricekot:
Updated the commit message: scan IDs -> scan rule IDs.

thc202:
For consistency `setEnabledPolicies` should be updated as well (`PassiveScanAPI`also). Worth refactoring to reuse `getScannerFromParamId` which is now doing the same.

thc202:
The release notes should be updated to mention the potential breaking change.

zap/src/main/java/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -779,55 +773,53 @@ private ActiveScan getActiveScan(JSONObject params) throws ApiException {
...
+            throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, e.getMessage(), e);
         }
     }
 
-    private static boolean hasPolicyWithId(int policyId) {
-        return Arrays.asList(Category.getAllNames()).contains(Category.getName(policyId));
+    private static void verifyCategoryId(int categoryId) throws ApiException {
+        if (categoryId < 0) {
+            throw new ApiException(
+                    ApiException.Type.ILLEGAL_PARAMETER, Integer.toString(categoryId));
```
thc202, line 806: Including the `categoyId` might be misleading, it might not be the original value but the internal value to detect that the parsing failed. Ideally it should be the name of the parameter.

zap/src/main/java/org/zaproxy/zap/extension/ascan/ActiveScanAPI.java
```diff
@@ -779,55 +773,53 @@ private ActiveScan getActiveScan(JSONObject params) throws ApiException {
...
-        final int id = getParam(params, PARAM_ID, -1);
-        if (id == -1) {
-            throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, PARAM_ID);
-        }
-        if (!hasPolicyWithId(id)) {
-            throw new ApiException(ApiException.Type.DOES_NOT_EXIST, PARAM_ID);
+    private void updateRulesOfCategoryInPolicy(
+            int categoryId, ScanPolicy policy, Consumer<Plugin> consumer)
+            throws ApiException, ConfigurationException {
+        verifyCategoryId(categoryId);
```
thc202, line 816: Could be removed if called by `setEnabledCategories`, other calls to `updateRulesOfCategoryInPolicy` are calling `verifyCategoryId` beforehand.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #7243, opened 2022-04-25
Deprecate and disable ExtensionState

thc202:
The extension was not working as intended, the `SessionChangedListener`
was not being called (not added in the `hook` method) and was just
adding a menu separator.
The state will be managed by the network add-on.


kingthorin (approved):

thc202:
The menu separator was being added between the (Edit >) Find… and Search… menu items, I can add it in other extension/class to keep the same aspect/behaviour but (IMO) does not seem to need a separator there.

kingthorin:
Let's leave it out for now. We can always redo something later if anyone asks. I agree it seems unnecessary.

psiinon (approved):

PR #7256, opened 2022-04-29
Deprecate HttpSender#setAllowCircularRedirects

thc202:
The circular redirects have been allowed by default since 2.5.0, the
option is not exposed in the UI. One can use a redirect validator to
prevent the circular redirects.


thc202:
Earlier I said there were no usages of the method in `zap-extensions` but after searching again there's one, in the Fuzzer add-on, I'm going to replace that with a `HttpRedirectionValidator`.

kingthorin (approved):

thc202:
The Fuzzer add-on is always setting it to false… and the user can't change it, I'm thinking on changing that to true, to keep the same behaviour as core. Thoughts?

kingthorin:
Consistent behavior seems best to me.

psiinon (approved):

Issue #1364, opened 2015-06-04
Heavy lag when loading a session file


zapbot:
```
What steps will reproduce the problem?
1.Create a session, and save it. Mine is almost 1gb in size
2. Close ZAP
3. Restart ZAP and load the session

What is the expected output? What do you see instead?
I expect the session to be loaded. The Loading Session File message box appears, and
stays for a long/undefined period. The session is not loaded.

What version of the product are you using? On what operating system?
ZAP 2.3.1 on Windows 7 sp1

Please provide any additional information below.
Unable to include file as it contains corporate information.

```

Original issue reported on code.google.com by `daniel.billing.nvm` on 2014-10-14 12:11:51

cmlh:
Also occurs on Kali Linux 2.0 VM with the 2.4.2 upstream release.

Is zaproxy/zaproxy#1650 a duplicate of this issue?

psiinon:
Yes, treating #1650 as a dup of this unless we get any more feedback on it

thc202:
ZAP should also report the loading progress.

davewichers:
I added a new issue, #2582, indicating that ZAP should report loading progress.

hackerlawyer:
I have a similar problem. The session file loads in the background but the loading session file message box never disappears, even after attempting to close the box, making it impossible to do any work. I have tried uninstalling and reinstalling and the same problem persists. On Kali VM Virtualbox 6.0. I haven't been able to do any work for the past 3 days.

Jman420:
I'm having similar issues with loading sessions that have been produced by ZAP Docker containers.  I do not see any errors in my zap.log file (see attached).  I've only waited for about 20mins to load a ~850mb session file, but will allow ZAP to try to open the session for a few hours now.
[zap.log](https://github.com/zaproxy/zaproxy/files/5271346/zap.log)

Jman420:
Just to update.  I left ZAP loading the previous session for a few hours without any success.  It seems like the persisted session that I have is not able to be re-loaded into a ZAP instance for continued analysis.

najum98:
Same issue here, I dont know how to get rid of it, I created issue on Github but it got closed stating that its a duplicate. What's the solution?

thc202:
The issue is not yet fixed, there's currently no workaround.

najum98:
> The issue is not yet fixed, there's currently no workaround.

[...]

PR #7269, opened 2022-05-04
Fix get all alert tags

userdelroot:
This fix includes the loading of saved session files, as well as minor fixes to help with this routine.
Reference: https://github.com/zaproxy/zaproxy/issues/1364

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/db/paros/ParosTableAlertTag.java
+++ b/zap/src/main/java/org/zaproxy/zap/db/paros/ParosTableAlertTag.java
@@ -40,6 +40,7 @@ public class ParosTableAlertTag extends ParosAbstractTable implements TableAlert
     private PreparedStatement psInsertOrUpdate;
     private PreparedStatement psGetAllTags;
     private PreparedStatement psGetTagsByAlertId;
+    private PreparedStatement psGetAllRecords;
     private PreparedStatement psDeleteByTagId;
     private PreparedStatement psDeleteByAlertIdTagKey;
     private PreparedStatement psDeleteAllTagsForAlert;
@@ -76,6 +77,8 @@ public class ParosTableAlertTag extends ParosAbstractTable implements TableAlert
                     conn.prepareStatement("SELECT * FROM alert_tag WHERE alert_id = ?");
             psGetAllTags =
                     conn.prepareStatement("SELECT DISTINCT key, value FROM alert_tag ORDER BY key");
+            psGetAllRecords =
+                    conn.prepareStatement("SELECT * FROM alert_tag ORDER BY alert_id ASC");
             psDeleteByTagId = conn.prepareStatement("DELETE FROM alert_tag WHERE tag_id = ?");
             psDeleteByAlertIdTagKey =
                     conn.prepareStatement("DELETE FROM alert_tag WHERE alert_id = ? AND key = ?");
@@ -202,7 +205,7 @@ public class ParosTableAlertTag extends ParosAbstractTable implements TableAlert
     public synchronized List<RecordAlertTag> getAllRecords() throws DatabaseException {
         try {
             List<RecordAlertTag> result = new ArrayList<>();
-            try (ResultSet rs = psGetAllTags.executeQuery()) {
+            try (ResultSet rs = psGetAllRecords.executeQuery()) {
                 RecordAlertTag rat;
                 while ((rat = build(rs)) != null) {
                     result.add(rat);
--- a/zap/src/main/dist/db/hsqldb.properties
+++ b/zap/src/main/dist/db/hsqldb.properties
@@ -185,6 +185,7 @@ alerttag.ps.readbyalertidtagkey = SELECT * FROM alert_tag WHERE alert_id = ? AND
 alerttag.ps.insertorupdate = MERGE INTO alert_tag AS tag USING (VALUES(?, ?, ?)) AS v(aid, key, val) ON tag.alert_id = v.aid AND tag.key = v.key WHEN MATCHED AND tag.value <> v.val THEN UPDATE SET tag.value = v.val WHEN NOT MATCHED THEN INSERT (alert_id, key, value) VALUES (v.aid, v.key, v.val)
 alerttag.ps.gettagsbyalertid = SELECT * FROM alert_tag WHERE alert_id = ?
 alerttag.ps.getalltags = SELECT DISTINCT key, value FROM alert_tag ORDER BY key
+alerttag.ps.getallrecords = SELECT * FROM alert_tag ORDER BY alert_id ASC
 alerttag.ps.deletebyalertidtagkey = DELETE FROM alert_tag WHERE tag_id = ? AND key = ?
 alerttag.ps.deletebytagid = DELETE FROM alert_tag WHERE tag_id = ?
 alerttag.ps.deletealltagsforalert = DELETE FROM alert_tag WHERE alert_id = ?
--- a/zap/src/main/java/org/zaproxy/zap/db/sql/SqlTableAlertTag.java
+++ b/zap/src/main/java/org/zaproxy/zap/db/sql/SqlTableAlertTag.java
@@ -126,11 +126,12 @@ public class SqlTableAlertTag extends SqlAbstractTable implements TableAlertTag 
 
     @Override
     public List<RecordAlertTag> getAllRecords() throws DatabaseException {
-        SqlPreparedStatementWrapper psGetAllTags = null;
+        SqlPreparedStatementWrapper psGetAllRecords = null;
         try {
-            psGetAllTags = DbSQL.getSingleton().getPreparedStatement("alerttag.ps.getalltags");
+            psGetAllRecords =
+                    DbSQL.getSingleton().getPreparedStatement("alerttag.ps.getallrecords");
             List<RecordAlertTag> result = new ArrayList<>();
-            try (ResultSet rs = psGetAllTags.getPs().executeQuery()) {
+            try (ResultSet rs = psGetAllRecords.getPs().executeQuery()) {
                 RecordAlertTag rat;
                 while ((rat = build(rs)) != null) {
                     result.add(rat);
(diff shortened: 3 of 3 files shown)


kingthorin:

kingthorin:
Also it's a fairly small change set. So review shouldn't be too bad. I'd suggest just squash or fix-up all the commits into one.

zap/src/main/java/org/zaproxy/zap/extension/alert/ExtensionAlert.java
```diff
@@ -601,13 +603,30 @@ private void refreshAlert(Session session) throws DatabaseException {
...
         final ExtensionHistory extensionHistory =
                 Control.getSingleton().getExtensionLoader().getExtension(ExtensionHistory.class);
 
-        for (int i = 0; i < v.size(); i++) {
-            int alertId = v.get(i);
+        // Grab all the alerts.
+        List<RecordAlertTag> rat = tableAlertTag.getAllRecords();
+
+        // Session alert list.
+        Vector<Integer> v = tableAlert.getAlertList();
```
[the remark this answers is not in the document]
kingthorin: Also it was probably already somewhat like this but in a effort to be better it would be nice to have meaningful var names vs `v`, `l`, `x`, etc.

kingthorin (approved):

kingthorin:
Obviously that shouldn't have been "approve". I don't have time to redo all the comments at the moment.

zap/src/main/java/org/zaproxy/zap/extension/alert/ExtensionAlert.java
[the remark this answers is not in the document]
userdelroot: That vector was already in the previous code. However, I do agree, I should of corrected it.

zap/src/main/java/org/zaproxy/zap/db/paros/ParosTableAlertTag.java
```diff
@@ -201,11 +201,16 @@ public synchronized Map<String, String> getAllTags() throws DatabaseException {
...
         try {
+
             List<RecordAlertTag> result = new ArrayList<>();
-            try (ResultSet rs = psGetAllTags.executeQuery()) {
-                RecordAlertTag rat;
-                while ((rat = build(rs)) != null) {
-                    result.add(rat);
+            try (PreparedStatement psReadScan =
+                    getConnection()
+                            .prepareStatement("SELECT * FROM alert_tag ORDER BY alert_id ASC")) {
```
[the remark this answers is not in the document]
userdelroot: The initial key tag "alerttag.ps.getalltags" did not contain all columns. This would cause the initial code base to fail with a sqlexception tag_id doesn't exist. 
I added that key tag "alerttag.ps.getalltags2" to include all the columns.

zap/src/main/java/org/zaproxy/zap/extension/alert/ExtensionAlert.java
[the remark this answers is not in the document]
kingthorin: Well if sync isn't needed they a list should be fine.

zap/src/main/java/org/zaproxy/zap/db/paros/ParosTableAlertTag.java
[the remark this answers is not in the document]
userdelroot: You can't use that there. One is for hsqldb and the other is based on paros db. 
I had to update both for it to work if someone uses hsqldb.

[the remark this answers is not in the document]
kingthorin: I wasn't questioning the need for different SQL. I was just wondering why it's literal here instead of getting what was added to hsqldb.properties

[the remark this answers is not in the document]
kingthorin: Oh okay, got ya.

userdelroot:
DCO is complete.

userdelroot:
What else needs done to get this merged and into a release?

kingthorin:
Address the review comments. Squash/fixup the commits. Address anything else that comes up for other reviewers.

userdelroot:
All those are done, unless i missed something.

kingthorin:
The vector and variable naming hasn't been addressed, and there's still 6 commits here.

zap/src/main/java/org/parosproxy/paros/network/HttpBody.java
```diff
@@ -557,7 +558,8 @@ public void setCharset(String charsetName) {
             setCharsetImpl(null);
             return;
         }
-
+        charsetName = StringUtils.remove(charsetName, ",text/html");
+        charsetName.toUpperCase();
```
psiinon, line 562: toUpperCase() just returns upper case, it doesnt change the string its invoked on. 
Best to just add this to the line above?

thc202: IMHO I don't think this should be added, the existing code copes with wrong charsets and we shouldn't be adding these one offs to workaround broken servers.

thc202: If this is actually causing problems better provide more information about the actual issue.

zap/src/main/java/org/parosproxy/paros/model/Session.java
```diff
@@ -287,6 +289,8 @@ public void run() {
 
     protected void open(String fileName) throws DatabaseException, IOException, Exception {
 
+        final CountDownLatch latch = new CountDownLatch(2);
```
thc202, line 292: Please, provide information about the issues that you are trying to fix. This is deadlocking in daemon and command line modes.

zap/src/main/java/org/parosproxy/paros/network/HttpBody.java
userdelroot: The wrong charset got through "utf-8,text/html"  was trying to be passed to that. Which was causing an exception and was potentially causing the code to stop processing while loading a saved session.

zap/src/main/java/org/parosproxy/paros/model/Session.java
userdelroot: In my tests this didn't cause a deadlock. There 2 InvokeAndWait under the open() method in the Session.java.
These were not loading, potentially because of the exception from the setcharset. 
InvokeLater allows both areas to run while waiting for them to finish prior to moving onto loading more of the session.

zap/src/main/java/org/parosproxy/paros/network/HttpBody.java
thc202: Please provide a thread dump when that hang happens.

zap/src/main/java/org/parosproxy/paros/model/Session.java
thc202: I understand what the code is doing not what is trying to fix. If this is being done as workaround for the other issue then it shouldn't be needed if the other issue is fixed.

To reproduce the deadlock you just need to load a session in daemon and command line modes (you can also check the code, the count down only happens when there's view).

zap/src/main/java/org/parosproxy/paros/network/HttpBody.java
thc202: Which exception was thrown?

zap/src/main/java/org/parosproxy/paros/model/Session.java
userdelroot: Oh, ic what happened. I will fix this.

zap/src/main/java/org/parosproxy/paros/network/HttpBody.java
thc202: That means it's one of the exceptions being caught. In which case did the entire loading fail?

userdelroot: At a certain point it was causing the entire session to fail to load. and it was handing in the open method during the InvokeAndWait. 
The InvokeLater as well as the additional code changes addressed all of this. 
If you want I can test it again to see if it loads now without this.

thc202: I'd suggest trying without the changes in `HttpBody` and `Session`, if it still hangs obtain a thread dump which should give more details about the issue.

userdelroot: Is there an easy way to do this ? I will most likely kill this pull request and start fresh from the upstream repo, if you are ok with that.

userdelroot: Yea, sorry about that. Been awhile since I used github and or Java..

kingthorin: No worries.

userdelroot: Getting a unit test failure now, `[org.parosproxy.paros.core.scanner.UtilUnitTest](file:///zap/zaproxy/zap/build/reports/tests/test/classes/org.parosproxy.paros.core.scanner.UtilUnitTest.html)`
However, I did not change anything in that code. Did something change recently?

thc202: No, that's a timing issue, bound to happen sometimes.

userdelroot:
New commits are done with only the needed code to fix the loading of a saved session.

thc202:
Could you try loading your session with latest `upstream/main`? The DB was tweaked (#7271) to make the tags query faster, which hopefully does not require further code changes.

userdelroot:
Using the upstream/main the session loads much faster.

userdelroot:
The getAllRecords, in the upstream/main will still fail as it isn't getting all of the columns. 
The code changes above fix this. [Commit](https://github.com/zaproxy/zaproxy/pull/7269/commits/fe7571b2f3c86d0ade3c8b235175604b6a415b14) could probably be left out of the merge.

thc202:
Thanks for trying that.

Agreed, that needs to be fixed.

thc202:
Commit removed and squashed the other fix.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #7299, opened 2022-05-24
Spider: find table/td background, video src/poster, and img srcset

Leyart:
Covers the following cases:

/test/html/body/table/background.found
/test/html/body/table/td/background.found
/test/html/body/video/src.found
/test/html/body/video/poster.found
/test/html/body/img/srcset1x.found
/test/html/body/img/srcset2x.found

Introduces a custom URL processor for URLs in attributes that require special handling (see srcset)

As always, commits will be squashed prior to final merge

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
@@ -39,7 +39,7 @@ import org.zaproxy.zap.spider.URLCanonicalizer;
  */
 public class SpiderHtmlParser extends SpiderParser {
 
-    /** The Constant URL_PATTERN defining the pattern for a meta url. */
+    /** The Constant URL_PATTERN defining the pattern for a meta URL. */
     static final Pattern URL_PATTERN =
             Pattern.compile("url\\s*=\\s*[\"']?([^;'\"]+)", Pattern.CASE_INSENSITIVE);
 
@@ -47,6 +47,17 @@ public class SpiderHtmlParser extends SpiderParser {
             Pattern.compile(
                     "(?:http(?:s?):)?//[^\\x00-\\x1f\"'\\s<>#()\\[\\]{}]+",
                     Pattern.CASE_INSENSITIVE);
+
+    private static final Pattern SRCSET_PATTERN =
+            Pattern.compile("[^\"'=\\s,]+\\.[^\\s,]+", Pattern.CASE_INSENSITIVE);
+    /**
+     * Functional interface that allows custom handling of URLs retrieved in attributes, to be
+     * manipulated before the URL is sent to a {@link SpiderParser#processURL} call.
+     */
+    @FunctionalInterface
+    private interface CustomUrlProcessor {
+        void process(HttpMessage message, int depth, String localURL, String baseURL);
+    }
 
     /** The params. */
     private SpiderParam params;
@@ -74,7 +85,7 @@ public class SpiderHtmlParser extends SpiderParser {
             source = new Source(message.getResponseBody().toString());
         }
 
-        // Get the context (base url)
+        // Get the context (base URL)
         String baseURL = message.getRequestHeader().getURI().toString();
 
         // Try to see if there's any BASE tag that could change the base URL
@@ -120,12 +131,27 @@ public class SpiderHtmlParser extends SpiderParser {
     }
 
     /**
+     * Implements the CustomUrlProcessor to handle the special format of a srcset attribute element,
+     * see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset the srcset
+     * attribute contains one or more strings separated by commas, indicating possible image sources
+     * for the user agent to use.
+     */
+    private void srcSetProcessor(HttpMessage message, int depth, String localURL, String baseURL) {
+        Matcher results = SRCSET_PATTERN.matcher(localURL);
+        while (results.find()) {
+            if (!results.group().isEmpty()) {
+                processURL(message, depth, results.group(), baseURL);
+            }
+        }
+    }
+
+    /**
      * Parses the HTML Jericho source for the elements that contain references to other resources.
      *
      * @param message the message
      * @param source the source
      * @param depth the depth
-     * @param baseURL the base url
+     * @param baseURL the base URL
      * @return {@code true} if at least one URL was found, {@code false} otherwise.
      */
     private boolean parseSource(HttpMessage message, Source source, int depth, String baseURL) {
@@ -208,6 +234,30 @@ public class SpiderHtmlParser extends SpiderParser {
             resourcesfound |= processAttributeElement(message, depth, baseURL, el, "src");
         }
 
+        // Process Table elements
+        elements = source.getAllElements(HTMLElementName.TABLE);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "background");
+        }
+
+        // Process TD elements
+        elements = source.getAllElements(HTMLElementName.TD);
+        for (Element src : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, src, "background");
+        }
+
+        // Process Video elements
+        elements = source.getAllElements(HTMLElementName.VIDEO);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "src");
+            List<Element> videoSourceElements = el.getAllElements(HTMLElementName.SOURCE);
+            for (Element sourceElement : videoSourceElements) {
+                resourcesfound |=
+                        processAttributeElement(message, depth, baseURL, sourceElement, "src");
+            }
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "poster");
+        }
+
         // Process Img elements
         elements = source.getAllElements(HTMLElementName.IMG);
         for (Element el : elements) {
@@ -215,6 +265,9 @@ public class SpiderHtmlParser extends SpiderParser {
             resourcesfound |= processAttributeElement(message, depth, baseURL, el, "longdesc");
             resourcesfound |= processAttributeElement(message, depth, baseURL, el, "lowsrc");
             resourcesfound |= processAttributeElement(message, depth, baseURL, el, "dynsrc");
+            resourcesfound |=
+                    processAttributeElement(
+                            message, depth, baseURL, el, "srcset", this::srcSetProcessor);
         }
 
         // Process META elements
@@ -260,20 +313,44 @@ public class SpiderHtmlParser extends SpiderParser {
      *
      * @param message the message
      * @param depth the depth
-     * @param baseURL the base url
+     * @param baseURL the base URL
      * @param element the element
      * @param attributeName the attribute name
      * @return {@code true} if a URL was processed, {@code false} otherwise.
      */
     private boolean processAttributeElement(
             HttpMessage message, int depth, String baseURL, Element element, String attributeName) {
+        return processAttributeElement(message, depth, baseURL, element, attributeName, null);
+    }
+
+    /**
+     * Processes the attribute with the given name of a Jericho element, for an URL. If an URL is
+     * found, notifies the listeners.
+     *
+     * @param message the message
+     * @param depth the depth
+     * @param baseURL the base URL
+     * @param element the element
+     * @param attributeName the attribute name
+     * @param customUrlProcessor functional interface for custom manipulation of urls
+     * @return {@code true} if a URL was processed, {@code false} otherwise.
+     */
+    private boolean processAttributeElement(
+            HttpMessage message,
+            int depth,
+            String baseURL,
+            Element element,
+            String attributeName,
+            CustomUrlProcessor customUrlProcessor) {
         // The URL as written in the attribute (can be relative or absolute)
         String localURL = element.getAttributeValue(attributeName);
         if (localURL == null) {
             return false;
         }
 
-        if (!attributeName.equalsIgnoreCase("ping")) {
+        if (customUrlProcessor != null) {
+            customUrlProcessor.process(message, depth, localURL, baseURL);
+        } else if (!attributeName.equalsIgnoreCase("ping")) {
             processURL(message, depth, localURL, baseURL);
         } else {
             for (String pingURL : localURL.split("\\s")) {
--- a/zap/src/test/resources/org/zaproxy/zap/spider/parser/html/TableElementsSpiderHtmlParser.html
+++ b/zap/src/test/resources/org/zaproxy/zap/spider/parser/html/TableElementsSpiderHtmlParser.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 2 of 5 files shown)


zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -47,6 +47,37 @@ public class SpiderHtmlParser extends SpiderParser {
...
+        @Override
+        public void process(HttpMessage message, int depth, String localURL, String baseURL) {
+            // Regexp to split entries for srcsets: [^"\'=\s,]+\.([^\s,]+)
+            // 1) Matches a single character not present in [^"\'=\s,] multiple times
+            // 2) Matches a .
+            // 3) Matches a single character not present in [^\s,] multiple times
+            Matcher results = Pattern.compile("[^\"'=\\s,]+\\.([^\\s,]+)").matcher(localURL);
+            while (results.find()) {
+                if (!results.group().isEmpty()) {
+                    processURL(message, depth, results.group(), baseURL);
```
kingthorin, line 76: results.group is the entire match right? So the group inside the regex can be marked as non-capturing? (I believe that has a slight performance benefit.)

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -267,13 +329,37 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+        return processAttributeElement(message, depth, baseURL, element, attributeName, null);
+    }
+
+    /**
+     * Processes the attribute with the given name of a Jericho element, for an URL. If an URL is
+     * found, notifies the listeners.
+     *
+     * @param message the message
+     * @param depth the depth
+     * @param baseURL the base url
```
kingthorin, line 341: I’d suggest using all caps URL in all textual/comment occurrences

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
Leyart: Yes, I had it as a capturing group because I wanted initially to check also for a set of possible filetypes, but I deemed it not necessary. Fixing it accordingly

kingthorin: 👍

Leyart:
Documentation updated in https://github.com/zaproxy/zap-core-help/pull/449

kingthorin (approved):

Leyart:
@psiinon With another review here, I could continue my quest for 100% ;)

psiinon:
@Leyart On holiday until next week, but I'll look at it on Monday :)

psiinon:
@kingthorin thanks! Got this on my first day off so it got lost...

psiinon (changes requested):
Looks good but its not finding `/test/html/body/table/td/background.found` for me :/
Tested vs https://security-crawl-maze.app/ and I've manually checked that the URL isnt in the Site Tree...

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -208,13 +239,44 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
             resourcesfound |= processAttributeElement(message, depth, baseURL, el, "src");
         }
 
+        // Process Table elements
+        elements = source.getAllElements(HTMLElementName.TABLE);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "background");
+            Source tableSource = new Source(el.getContent().toString());
```
psiinon, line 246: Why not just use do `source.getAllElements(HTMLElementName.TD)` ?
We dont care is a 'TD' element is not in a table, we just want to know if it has a `background` element 😄

psiinon: (at the same level ypou process the 'TABLE' elements of course ;)

zap/src/test/resources/org/zaproxy/zap/spider/parser/html/TableElementsSpiderHtmlParser.html
```diff
@@ -0,0 +1,39 @@
...
+
+<!--  Ignored: -->
+<table></table>
+<table>
+    <tr>
+        <td></td>
+    </tr>
+</table>
+<table background="mailto:background@example.com"></table>
+<td background="td.without.table@example.com"></td>
```
thc202, line 34: Why is this under the head?

zap/src/test/resources/org/zaproxy/zap/spider/parser/html/VideoElementsSpiderHtmlParser.html
```diff
@@ -0,0 +1,40 @@
...
+<body>
+
+<video src="//video.example.com/base/scheme">
+<video src="http://video.example.com:8000/b">
+<video src="https://video.example.com/c?a=b#fragment">
+<video src="video/relative">
+<video src="">
+<video src="/video/absolute">
+<video src="ftp://video.example.com/">
+<video src="http://video.with.poster.com/" poster="poster/relative">
```
thc202, line 17: Should use `example` domains, there's `.org` if you want other than `.com`.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
Leyart: Because I was not sure what we were aiming at and I was trying to be as compliant as possible with the HTML spec :) 
If I can just search for TD elements, so be it

Leyart:
> Looks good but its not finding `/test/html/body/table/td/background.found` for me :/ Tested vs https://security-crawl-maze.app/ and I've manually checked that the URL isnt in the Site Tree...

That's really odd because I put tests and I also test manually all the entries against crawlmaze.
Just rerun the dev build from the PR and it is definitely finding the entry for me:

Screen Shot 2022-06-12 at 23 30 17

Would you mind double checking again? Otherwise I would really not know what to look for

zap/src/test/resources/org/zaproxy/zap/spider/parser/html/TableElementsSpiderHtmlParser.html
Leyart: Because it is a mistake, thanks for spotting :)

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -39,14 +39,45 @@
...
     static final Pattern URL_PATTERN =
             Pattern.compile("url\\s*=\\s*[\"']?([^;'\"]+)", Pattern.CASE_INSENSITIVE);
 
     private static final Pattern PLAIN_COMMENTS_URL_PATTERN =
             Pattern.compile(
                     "(?:http(?:s?):)?//[^\\x00-\\x1f\"'\\s<>#()\\[\\]{}]+",
                     Pattern.CASE_INSENSITIVE);
+    /**
+     * Functional interface that allows custom handling of URLs retrieved in attributes, to be
+     * manipulated before the URL is sent to a {@link SpiderParser:processURL} call.
```
thc202, line 52: Should use `#` instead of `:`.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -39,14 +39,45 @@
...
+     */
+    @FunctionalInterface
+    private interface CustomUrlProcessor {
+        void process(HttpMessage message, int depth, String localURL, String baseURL);
```
thc202, line 56: Allow to pass the spider tasks and make `SrcSetProcessor` a class method.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -39,14 +39,45 @@
...
+     */
+    private class SrcSetProcessor implements CustomUrlProcessor {
+
+        @Override
+        public void process(HttpMessage message, int depth, String localURL, String baseURL) {
+            // Regexp to split entries for srcsets: [^"\'=\s,]+\.([^\s,]+)
+            // 1) Matches a single character not present in [^"\'=\s,] multiple times
+            // 2) Matches a .
+            // 3) Matches a single character not present in [^\s,] multiple times
+            Matcher results = Pattern.compile("[^\"'=\\s,]+\\.[^\\s,]+").matcher(localURL);
```
thc202, line 73: Extract a constant for the pattern, it's not necessary to compile it each time.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -39,14 +39,45 @@
...
+     * for the user agent to use.
+     */
+    private class SrcSetProcessor implements CustomUrlProcessor {
+
+        @Override
+        public void process(HttpMessage message, int depth, String localURL, String baseURL) {
+            // Regexp to split entries for srcsets: [^"\'=\s,]+\.([^\s,]+)
+            // 1) Matches a single character not present in [^"\'=\s,] multiple times
+            // 2) Matches a .
+            // 3) Matches a single character not present in [^\s,] multiple times
```
thc202, line 72: This is redundant.

psiinon:
For some reason its still not finding `/test/html/body/table/td/background.found` for me, but I'm happy for it to be merged anyway as it fixes 5 of the failing cases and if the 6th is also fixed in our automated tests then we'll know its just a problem with my set up.

thc202:
You might have to increase the crawl depth.

psiinon:
Ha! Good point :)

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
thc202: Sorry, called it wrong, I meant the spider parser not task, i.e. `void process(SpiderHtmlParser, HttpMessage, int, String, String);`. `SrcSetProcessor` can then be a class method and use a method reference in the `processAttributeElement` call, instead of creating a new instance each time.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -208,13 +238,44 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+            resourcesfound |= processAttributeElement(message, depth, baseURL, src, "background");
+        }
+
+        // Process Video elements
+        elements = source.getAllElements(HTMLElementName.VIDEO);
+        for (Element el : elements) {
+            boolean srcFound = processAttributeElement(message, depth, baseURL, el, "src");
+            resourcesfound |= srcFound;
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "poster");
+            if (!srcFound) {
```
thc202, line 259: It could it anyway, we want to find the most URLs.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -208,13 +238,44 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+                Source videoSource = new Source(el.getContent().toString());
```
thc202, line 260: Why parse it again?

psiinon:
> You might have to increase the crawl depth.

Yep, that did it :D 
Note to self - need to increase this in the daily test...

Leyart:
Thanks for checking :) I still need to address thc202 comments, I will do that in the following days. Really looking forward to have 100% coverage with the spider :)

psiinon:
Me too!

thc202:
@Leyart do you have an eta? (No rush, just checking.)

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
Leyart: There is actually no need to add the SpiderHtmlParser, just the method reference is sufficient. 
Changed the code to reflect that

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
Leyart: By spec, src is optional. If not defined, the source of the video is provided by the source elements within the video block body. But I agree that we want to find the most URLs.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
Leyart: It is not the same thing: here we are getting to body of the video tag to parse all the source elements

Leyart:
@thc202 Addressed the last comments. Let me know if it is all good and I will squash the commits into one

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
```diff
@@ -208,13 +234,41 @@ private boolean parseSource(HttpMessage message, Source source, int depth, Strin
...
+        elements = source.getAllElements(HTMLElementName.TD);
+        for (Element src : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, src, "background");
+        }
+
+        // Process Video elements
+        elements = source.getAllElements(HTMLElementName.VIDEO);
+        for (Element el : elements) {
+            resourcesfound |= processAttributeElement(message, depth, baseURL, el, "src");
+            Source videoSource = new Source(el.getContent().toString());
```
kingthorin, line 253: I think the earlier point here was that you should be able to `el.getAllElements(HTMLElementName.SOURCE)` without having to reparse into a new `Source` object.

http://jericho.htmlparser.net/docs/javadoc/net/htmlparser/jericho/Element.html (See "Methods inherited from class Segment")

Leyart: Thanks, really useful, amended :)

kingthorin: Thanks for the quick tweak :grinning:

kingthorin (approved):

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderHtmlParser.java
thc202: Sounds good.

thc202:
Squash away :)

Leyart:
> Squash away :)

Done

thc202 (approved):

thc202:
Thank you!

psiinon:
Good work!

PR #7302, opened 2022-05-27
Clarify History Tags vs Alert Tags

kingthorin:
As discussed on the weekly team call and IRC.

- AbstractPlugin
    - Add overridden methods from super class.
- Alert
    - Add ZAP note.
    - Add addTag(String)
    - Add addTag(String, String)
    - Add removeTag(String)
    - Add removeTag(String, String)
- AlertUnitTest
    - Add tests asserting the new behavior.
- AntiCsrfDetectScanner
    - Updated to use the new method.
- PluginPassiveScanner
    - Add overridden methods from super class.
    - Deprecate addTag(String)
    - Add addHistoryTag(String)
- RegexAutoTagScanner
    - Use addHistoryTag(String)
- ScriptspassiveScanner
    - Deprecate addTag(String)
    - Add addHistoryTag(String)
- PassiveScanTaskHelper
    - Rename addTag(String) to addHistoryTag(String)
- PassiveScanThread
    - Use renamed method from helper.
- ScriptsPassiveScannerUnitTest
    - Updated to use new methods.
- PluginPassiveScannerUnitTest
    - Updated to use new methods.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
@@ -63,10 +63,12 @@
 // ZAP: 2022/02/03 Removed SUSPICIOUS, WARNING, MSG_RELIABILITY, setRiskReliability(int, int) and
 // getReliability()
 // ZAP: 2022/02/25 Remove code deprecated in 2.5.0
+// ZAP: 2022/05/26 Add addTag and removeTag methods
 package org.parosproxy.paros.core.scanner;
 
 import java.net.URL;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import javax.swing.ImageIcon;
 import org.apache.commons.httpclient.URI;
@@ -1084,6 +1086,59 @@ public class Alert implements Comparable<Alert> {
         /** @since 2.11.0 */
         public Builder setTags(Map<String, String> tags) {
             this.tags = tags;
+            return this;
+        }
+
+        /**
+         * Adds an Alert tag with the given key (tag/name) to the existing collection for this
+         * Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder addTag(String tag) {
+            addTag(tag, "");
+            return this;
+        }
+
+        /**
+         * Adds an Alert tag with the given key (tag/name) and value to the existing collection for
+         * this Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder addTag(String tag, String value) {
+            if (this.tags == null) {
+                this.tags = new HashMap<>();
+            }
+            this.tags.put(tag, value);
+            return this;
+        }
+
+        /**
+         * Removes an Alert tag with the given key (tag/name) from the existing collection for this
+         * Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder removeTag(String tag) {
+            if (this.tags == null) {
+                return this;
+            }
+            this.tags.remove(tag);
+            return this;
+        }
+
+        /**
+         * Removes an Alert tag with the given key (tag/name) and value from the existing collection
+         * for this Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder removeTag(String tag, String value) {
+            if (this.tags == null) {
+                return this;
+            }
+            this.tags.remove(tag, value);
             return this;
         }
 
--- a/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTaskHelper.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTaskHelper.java
@@ -169,7 +169,7 @@ public class PassiveScanTaskHelper {
      *
      * @param tag the name of the tag.
      */
-    public void addTag(HistoryReference href, String tag) {
+    public void addHistoryTag(HistoryReference href, String tag) {
         if (shutDown) {
             return;
         }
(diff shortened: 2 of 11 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1081,12 +1082,48 @@ public Builder setAlertRef(String alertRef) {
...
+        /**
+         * Removes a tag with the given key (tag/name) and value from the existing collection for
+         * this Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder removeAlertTag(String tag, String value) {
+            this.tags.remove(tag, value);
+            return this;
+        }
```
kingthorin, line 1125: I figured if I was introducing an add method then it made sense to also introduce a remove method.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1081,12 +1082,48 @@ public Builder setAlertRef(String alertRef) {
             return this;
         }
 
-        /** @since 2.11.0 */
+        /** @since 2.11.0 @Deprecated 2.12.0 Replaced by {@link #setAlertTags(Map)} */
```
thc202, line 1085: I understood that the only methods that were going to be renamed were the history tag ones. `Alert#getTags` and `#setTags` do not have `Alert`.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1081,12 +1082,48 @@ public Builder setAlertRef(String alertRef) {
...
             return this;
         }
 
+        /**
+         * Adds a tag to the existing collection for this Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder addAlertTag(String tag, String value) {
+            this.tags.put(tag, value);
```
thc202, line 1102: This will lead to a `NullPointerException`, would be good to have some tests.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: I may have lost track of where the conversation landed last week. That’s fine it simplifies the PR.

kingthorin:
Done & done

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1087,6 +1089,59 @@ public Builder setTags(Map<String, String> tags) {
...
+            return this;
+        }
+
+        /**
+         * Adds an Alert tag with the given key (tag/name) and value to the existing collection for
+         * this Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder addTag(String tag, String value) {
```
ricekot, line 1109: This method (and the others) could be synchronized to avoid potential race conditions. The `HashMap` could be replaced with a `ConcurrentHashMap`.

thc202: In which case will the concurrent access happen? The builders should not be being shared.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1087,6 +1089,59 @@ public Builder setTags(Map<String, String> tags) {
             return this;
         }
 
+        /**
+         * Adds an Alert tag with the given key (tag/name) to the existing collection for this
+         * Alert/Builder.
+         *
+         * @since 2.12.0
+         */
+        public Builder addTag(String tag) {
```
thc202, line 1098: These new methods need to be exposed in the `AlertBuilder`s, otherwise the chaining will not work properly (to raise the alert).

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTaskHelper.java
```diff
@@ -168,8 +168,20 @@ public void raiseAlert(HistoryReference href, Alert alert) {
      * Adds the given tag to the specified message.
      *
      * @param tag the name of the tag.
+     * @deprecated 2.12.0 Replaced by {@link #addHistoryTag(HistoryReference, String)}
```
thc202, line 171: This is a new class, the "old" method can be renamed.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: K

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTaskHelper.java
kingthorin: Okay, thanks

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
ricekot: I don't have an exact case in mind but just being safe. Maybe builders accessed from scripts or multithreaded passive scan rules?

thc202: They would still have their own builder in those cases.

ricekot: Damn, my bad, that makes sense. Thanks for the clarification.

kingthorin:
Done & done

thc202:
There's a typo in the commit message, `User` → `Use`?

kingthorin:
Tweaked.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #7315, opened 2022-06-03
Tidy up HttpSender

thc202:
Remove commented code and make listeners comparator final, the current
HttpSender implementation will be extracted into its own class.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #7316, opened 2022-06-03
Allow to replace HttpSender implementation

thc202:
Add two interfaces to allow to replace the implementation:
 - `HttpSenderContext`, to keep the state of each `HttpSender`
 instance;
 - `HttpSenderImpl`, creates contexts and sends the messages. It also
 keeps the listeners as they are shared between all `HttpSender`s.

Extract the current implementation into `HttpSenderContextParos` and
`HttpSenderParos`.
An add-on should now be able to set its own implementation.


kingthorin (approved):

thc202:
Fixed a class cast exception when using the old internal client implementation while having a newer `HttpSender` implementation set.

kingthorin (approved):

ricekot (approved):

kingthorin:
Conflicting.

thc202:
Conflict addressed.

kingthorin:
Thanks

PR #7325, opened 2022-06-08
Fix some scaling issues

### LINKED ISSUE #541: Break panel icons do not resize
```
I'd rather open it up for comment to see if anyone else agrees first. 

One: If you're continually using them, you'd want something big you can easily click.
There's a bit of precision clicking going or a little extra care which is required
if you're clicking constantly at a high rate.

As a bonus, i've seen quite a few people starting off with zap miss them completely
because of their size and not know how to intercept. While yes, rtfm, ui that allows
you to easily infer how to use it is a good thing.

I do think burp got the right scale + having names on this one. What do people think
comparing the two?
```

Original issue reported on code.google.com by `contact@markcunningham.ie` on 2013-02-27 09:59:47

psiinon:
There will be others, but these are the ones I've spotted so far.
Fixes #541

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
@@ -35,6 +35,7 @@ import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.view.TabbedPanel;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.extension.brk.impl.http.HttpBreakpointMessage;
+import org.zaproxy.zap.utils.DisplayUtils;
 import org.zaproxy.zap.utils.Stats;
 import org.zaproxy.zap.view.TabbedPanel2;
 import org.zaproxy.zap.view.ZapToggleButton;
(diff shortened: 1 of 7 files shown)


kingthorin:
So should this be "fix" for 541 or are you planning to make 541 more generic and leave it open?

kingthorin (approved):

psiinon:
Part of 541 is a request to make the icons scalable independant of the font scaling.
This doesnt do that which is why I didnt mark it as fixing it, but open to suggestions ...

kingthorin:
Let's break out another issue to scaling unrelated to Break buttons? Maybe a second one for separating options for image scaling vs text?

psiinon:
I'll raise one for that..

psiinon:
Doh! Only just realised getScaledSize(int) does what I need so no need for getScaledInteger(int) ;)

kingthorin (approved):

ricekot (approved):

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -175,15 +176,17 @@ private void setActiveIcon(boolean active) {
             if (active) {
                 parent.setIconAt(
                         parent.indexOfComponent(breakPanel),
-                        new ImageIcon(
-                                BreakPanelToolbarFactory.class.getResource(
-                                        "/resource/icon/16/101.png"))); // Red X
+                        DisplayUtils.getScaledIcon(
+                                new ImageIcon(
+                                        BreakPanelToolbarFactory.class.getResource(
+                                                "/resource/icon/16/101.png")))); // Red X
```
thc202, line 182: Worth extracting a method (passing just the resource path).

thc202:
Thank you!

PR #7328, opened 2022-06-08
Fix NullPointerException when obtaining host name

thc202:
Do not attempt to lower case the scheme if the URI does not have one
(i.e. an authority).


ricekot (approved):

ricekot:
Thank you!

PR #7332, opened 2022-06-09
Fix PatternSyntaxException in Analyser

thc202:
Quote the query component that's going to be used in the regular
expression, it might have reserved characters (e.g. `*`).


psiinon:
Worth making this protected and adding a unit test?

psiinon (approved):

PR #7351, opened 2022-06-19
Do not process location URL when processing links

thc202:
Remove the process URL call with the location since that's already
being done early and would lead to warnings if the location was not
present.


ricekot (approved):

psiinon (approved):

kingthorin:
We'll need to remember to move this over to the add-on.

PR #7359, opened 2022-06-23
Deprecate Tab interface

thc202:
The interface is not in effective use, does not provide behaviour nor
is it used as a tag interface.
Remove it's usage from the codebase and suppress the binary
incompatible changes.


kingthorin (approved):

kingthorin:
There are more suppressions than changed files?

psiinon (approved):

kingthorin:
Okay, yup that makes sense. Thanks.

Issue #7352, opened 2022-06-21
pscan could be less chatty/loggy


kingthorin:
I deleted a bunch of stuff from the Sites Tree while there was still a passive scan queue. 

1. The passive scanner probably doesn't need to log the full trace on these.
2. Could the queue be modified when things are deleted from the Sites Tree?

Versions:
Java: 11.0.13
ZAP: Dev

```console
68347 [ZAP-PassiveScan-4] ERROR org.zaproxy.zap.extension.pscan.PassiveScanTask - Parser failed on record 387 from History table
org.parosproxy.paros.network.HttpMalformedHeaderException: No history reference for id 387 type=2
	at org.parosproxy.paros.model.HistoryReference.getHttpMessage(HistoryReference.java:454) ~[main/:?]
	at org.zaproxy.zap.extension.pscan.PassiveScanTask.run(PassiveScanTask.java:101) ~[main/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
	at java.lang.Thread.run(Thread.java:829) ~[?:?]
68348 [ZAP-PassiveScan-1] ERROR org.zaproxy.zap.extension.pscan.PassiveScanTask - Parser failed on record 388 from History table
org.parosproxy.paros.network.HttpMalformedHeaderException: No history reference for id 388 type=2
	at org.parosproxy.paros.model.HistoryReference.getHttpMessage(HistoryReference.java:454) ~[main/:?]
	at org.zaproxy.zap.extension.pscan.PassiveScanTask.run(PassiveScanTask.java:101) ~[main/:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
[...]

PR #7364, opened 2022-06-26
pscan: Reduce logging

kingthorin:
Reduce logging in circumstances where the message/history reference may have been deleted between being enqueued and being passively scanned.

Fixes zaproxy/zaproxy#7352

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
@@ -23,8 +23,12 @@ import net.htmlparser.jericho.Source;
 import org.apache.commons.httpclient.URI;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.parosproxy.paros.db.DatabaseException;
+import org.parosproxy.paros.db.RecordHistory;
 import org.parosproxy.paros.model.HistoryReference;
+import org.parosproxy.paros.model.Model;
 import org.parosproxy.paros.network.HttpHeader;
+import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
 import org.zaproxy.zap.utils.Stats;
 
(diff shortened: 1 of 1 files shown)


psiinon (approved):

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
```diff
@@ -204,16 +204,18 @@ public void run() {
                 }
             } else {
                 logger.error(
-                        "Parser failed on record {} from History table", href.getHistoryId(), e);
+                        "Parser failed on record {} from History table (perhaps it was deleted between when it was enqueued and now)",
```
thc202, line 207: IMO it could just check if the record still exists or not, if it doesn't not worth log anything.

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
```diff
@@ -204,16 +204,18 @@ public void run() {
...
+                        href.getHistoryId());
+                logger.debug(e);
```
thc202, line 209: This will just log the exception's message not the stacktrace. Was that on purpose?

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
```diff
@@ -204,16 +204,18 @@ public void run() {
...
                     msg = href.getHttpMessage();
                     logger.error(
                             "Parser failed on record {} from History table, msg {}",
                             href.getHistoryId(),
-                            msg,
-                            e);
-                    logger.error("Req Header {}", msg.getRequestHeader(), e);
+                            msg);
+                    logger.debug(e);
+                    logger.debug("Req Header {}", msg.getRequestHeader());
```
thc202, line 218: Does this work as intended?

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
kingthorin: Nope, I can adjust it

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
kingthorin: This part may be unnecessary in the context of the original issue. I was going for consistency. Will test more.

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
kingthorin: You mean here or just before processing it?

thc202: Here.

kingthorin: I think I'm overthinking this, what call should I be making to see if the record still exists?

thc202: `TableHistory#read` it returns null if it no longer exists.

kingthorin:
Tweaked

kingthorin:
Tweaked

thc202 (approved):

thc202:
Thank you!

Issue #7336, opened 2022-06-12
Add "trashbin" for recent items deleted by mistake


aerosayan:
### Is your feature request related to a problem? Please describe.

Problem : When rightclicking on a request, the "Delete" option is right next to the "Note", and "Break" option. Thus it's easy to hit it. Accidentally deleting a critical request is problematic. Even in persistent sessions, the requests which get deleted seems impossible to recover.

### Describe the solution you'd like

1. "Trashbin" tab where we can review the deleted items, and recover anything that has been accidentally deleted.
2. Move "Delete" option away from other commonly used options.

### Describe alternatives you've considered

none.

### Additional context

PS : I have a few other requests. For example, being able to highlight requests with different colors would be helpful. Should we make separate tabs for each request, or include all in a single one?

### Would you like to help fix this issue?

- [ ] Yes

thc202:
Have you checked Neonmarker add-on?

aerosayan:
Since implementing this new requested feature will be time consuming, a temporary (but very effective) fix may be to create two empty dummy/inactive region (of 1 cm height) above and below the "Delete" option. This would improve visibility of the "Delete" option, and will *significantly* reduce accidental clicks on it.

psiinon:
Or we could add a "Are you sure you want to delete this?" dialog with a "Do not show me this agin" checkbox option...

aerosayan:
@psiinon it's a good trade off. Yeah.

Although I've seen that people freak out when such a prompt props up by accident and they click enter without even thinking. It might be necessary to set the default action be "No", such that people don't click enter on "Yes" by mistake. Either they would have to use the mouse to click "Yes", or use the arrow keys to focus on "No" and then click enter.

Thanks

kingthorin:
There is already a confirmation dialog, that isn't ignoreable on the Sites tree "Delete" menu item:
https://github.com/zaproxy/zaproxy/blob/393fed828496f8eb1cb4d7b1b0f690945171c8d8/zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuPurgeSites.java#L60-L69

The same could/should be one on the History table version:
https://github.com/zaproxy/zaproxy/blob/8ac4b10d3d9d816957985694c2b7cf127d1364d3/zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuPurgeHistory.java

thc202:
That's being done to, but when deleting more than one.

kingthorin:
[...]

PR #7365, opened 2022-06-26
Consistently prompt when deleting history/sites items

kingthorin:
- Make the history delete context menu consistent with the Sites Tree delete context menu. (Prompt the user to ensure they're confident about removing the items.) In both cases allow the user to suppress the confirmation dialog.

Related to zaproxy/zaproxy#7336


thc202:
Please, add the "don't warn again" option.

kingthorin:
Sure I can look into it. The subject edit doesn’t seem accurate, unless you’re suggesting that I make the Sites Tree handling like the original History handling instead of what I had actually done.

thc202:
Sorry, updated to match the actual change.

kingthorin:
Can you think of a ZAP dialog that uses the "don't warn again" option, that isn't tied to an options panel?

kingthorin:
Thank you!

kingthorin:
Okay I've implemented a checkbox for the History variant, if you're happy with that I'll go ahead and do something similar for the Sites Tree.

kingthorin:
Tweaked and added Sites Tree handling as well, as discussed via IRC.

kingthorin:
Kicked to resolve the stuck checks.
```bash
git commit --amend --no-edit
git push --force
```

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -881,13 +885,41 @@ public void eventReceived(Event event) {
...
-            int result =
-                    getView()
-                            .showConfirmDialog(
-                                    Constant.messages.getString("history.purge.warning"));
-            if (result != JOptionPane.YES_OPTION) {
-                return;
+
+        if (hasView() && !hrefs.isEmpty()) {
+            FileConfiguration config = Model.getSingleton().getOptionsParam().getConfig();
+            boolean hideDialog = config.getBoolean(REMOVE_CONFIRMATION_KEY, false);
```
thc202, line 891: e.g. `confirmRemoval` and does not need to negate in the if.

zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuPurgeSites.java
```diff
@@ -59,16 +62,44 @@ public boolean isButtonEnabledForSiteNode(SiteNode sn) {
 
     @Override
     public void performHistoryReferenceActions(List<HistoryReference> hrefs) {
-        if (hrefs.size() > 0) {
-            int result =
-                    View.getSingleton()
-                            .showConfirmDialog(Constant.messages.getString("sites.purge.warning"));
-            if (result != JOptionPane.YES_OPTION) {
-                return;
+        if (!hrefs.isEmpty()) {
```
thc202, line 65: Check empty and return.

thc202:
In the Sites tree the delete shortcut key does not show the "don't show this again" option.

kingthorin:
Will track that down.

kingthorin:
Shortcut key now implements the same functionality.

zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuPurgeSites.java
thc202: This was not changed.

kingthorin: Doh, I think I only did it in one class 🤦‍♂️

kingthorin:
Fixed

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -881,13 +885,43 @@ public void eventReceived(Event event) {
...
-                            .showConfirmDialog(
-                                    Constant.messages.getString("history.purge.warning"));
-            if (result != JOptionPane.YES_OPTION) {
-                return;
+        if (hrefs.isEmpty()) {
+            return;
+        }
+        if (hasView()) {
+            FileConfiguration config = Model.getSingleton().getOptionsParam().getConfig();
+            boolean confirmRemoval = config.getBoolean(REMOVE_CONFIRMATION_KEY, false);
```
thc202, line 893: I think this was meant to be true by default, otherwise there's no prompt ever. (Same in the other dialogues.)

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -881,13 +885,43 @@ public void eventReceived(Event event) {
...
+                                null);
+                if (result != JOptionPane.YES_OPTION) {
+                    return;
+                }
+                Model.getSingleton()
+                        .getOptionsParam()
+                        .getConfig()
+                        .setProperty(
+                                REMOVE_CONFIRMATION_KEY,
+                                removeWithoutConfirmationCheckBox.isSelected());
```
thc202, line 924: This one needs to be negated.

kingthorin:
Hmmm I’m sure I tested at the time. Will go through it again this evening.

thc202:
Same here (regarding the testing).

kingthorin:
Okay, unless you’ve already tackled it I’m happy to tonight.

thc202:
I didn't, only tweaked while testing the Clear History PR.

kingthorin:
Okay, I got it then.

Issue #7178, opened 2022-03-29
automation: Add save-as


tkisason:
Edit: **See https://github.com/zaproxy/zaproxy/issues/7178#issuecomment-1118991263**

---

### Describe the bug

This is an issue in the ZAP GUI. When the user wants to create a new automation job, he adds the automation tab, adds the parameters, fills in the variables, etc. After the first save if an user alters the job and wants to save the new plan the interface is bugged and he cannot save the job, the file picker window doesn't show up. 

### Steps to reproduce the behavior

1. Create a new automation job in the ZAP GUI
2. Edit the automation job, add parameters, etc.
3. Save the plan (a file picker opens up and the user selects where to save the new yaml file)
4. Run the plan to test it out
5. Alter the job
6. Try to save the altered plan, the save option doesn't work work and the file picker doesn't open. 

### Expected behavior

on 6. the file picker should appear when the user clicks "Save Plan"

### Software versions

OWASP ZAP
Version: 2.11.1
---
Installed Add-ons: [[id=alertFilters, version=13.0.0],
[id=ascanrules, version=43.0.0], [id=automation,
version=0.9.0], [id=bruteforce, version=11.0.0],
[id=callhome, version=0.0.3], [id=commonlib, version=1.6.0],
[id=diff, version=11.0.0], [id=directorylistv1,
version=5.0.0], [id=domxss, version=12.0.0], [id=encoder,
version=0.6.0], [id=formhandler, version=4.0.0], [id=fuzz,
version=13.5.0], [id=gettingStarted, version=13.0.0],
[id=graaljs, version=0.2.0], [id=graphql, version=0.7.0],
[id=help, version=14.0.0], [id=hud, version=0.13.0],
[id=importurls, version=8.0.0], [id=invoke, version=11.0.0],
[id=network, version=0.0.1], [id=oast, version=0.6.0],
[id=onlineMenu, version=9.0.0], [id=openapi,
version=24.0.0], [id=pscanrules, version=37.0.0],
[id=quickstart, version=32.0.0], [id=replacer,
version=9.0.0], [id=reports, version=0.10.0], [id=retest,
version=0.2.0], [id=retire, version=0.9.0], [id=reveal,
version=4.0.0], [id=saverawmessage, version=6.0.0],
[id=savexmlmessage, version=0.2.0], [id=scripts,
version=29.0.0], [id=selenium, version=15.5.1], [id=soap,
[...]

PR #7377, opened 2022-07-03
Adding save-as.png icon file for save-as button in automation toolbar, regarding issue #7178 

Siddharth3330:
Added
- A save-as.png file (at zap/src/main/resources/resource/icon/16) is added to be used as icon for new save-as button in automation toolbar.


kingthorin:
This shouldn’t be necessary, add-ons can contain and use their own resources.

Siddharth3330:
yes I got it, Thanks!

PR #7380, opened 2022-07-11
Correct start time of passive scan rules

thc202:
Use a new start time for each scan rule instead of using the start time
of the scan task otherwise it would check/report the cumulative time
taken of previous scan rules too.


ricekot (approved):
Thank you!

PR #7383, opened 2022-07-17
fix: Change `[` to `[[` on lines 51 and 54

MichaelSodeke:
/usr/share/zaproxy/zap/src/main/dist/zap.sh: line 51: [: -ge: unary operator expected
/usr/share/zaproxy/zap/src/main/dist/zap.sh: line 54: [: -ge: unary operator expected

Signed-off-by: Cosmonaut <olubemiga@gmail.com>


psiinon:
Why do you think these changes are required?
Do you have a situation where they fail?
And if so, why not change all of the other `if` statements?

PR #7388, opened 2022-07-23
Include architecture in support info

thc202:
Include the architecture which is useful to debug/reproduce issues.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #7401, opened 2022-08-03
Add option to choose the icon size

### LINKED ISSUE #7326: Allow icons to be scaled independently of the text
As per #541 buttons are scaled when the default font is scaled.
Having an option to scale them independently would help some people.
I think keeping the current behaviour as the default would be best.

Siddharth3330:
Trying to work on #7326 but unbale to approach it properly. I added a bar in UI which would take the icon size input from user, but I am unable to make use of that value.
Also not sure whether I am storing the input value correctly or not.
Looking for suggestions to approach the issue.    

Resolves #7326. 

Signed-off-by: Siddharth3330 <siddharthsingh58000@gmail.com>


Siddharth3330:
Screenshot (189)-1

thc202:
That should be stored in `OptionsParamView` (check how other options are initialised/saved).

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
```diff
@@ -111,6 +111,7 @@ public int getValue() {
         }
     }
 
+    private static int s = 1;
```
kingthorin, line 114: I don't mean to be picky but 6mo from now no one is going to remember what `s` is without digging through the code, could you make this `iconSize`? (Or something like that)

Siddharth3330:
> That should be stored in `OptionsParamView` (check how other options are initialised/saved).

I tried doing it but I believe its not done correctly as I still can't see any changes in Icon size.

ricekot:
Rebased and reworked this PR.

psiinon:
Looks good!
The toolbars dont scale with the icons, only the text.
I dont think this is a problem tbh, just mentioning it.
I did notice that the top Sites icon isnt scaled - that could be an existing bug of course, but be good to fix it now 😉 
Screenshot 2023-02-01 at 11 15 28

It also doesnt work in quite the same way as the font controls.
Theres no way to work out what the default size is once you've changed it.
With the fonts if you go to -1 then it defaults, but with the icons the min is 1.
0 or -1 should probably be treated as default?

ricekot:
Tweaked.

ricekot:
Tweaked to make the existing behaviour the default.
scale-icons

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
```diff
@@ -153,6 +155,8 @@ public int getValue() {
             new EnumMap<>(FontUtils.FontType.class);
     private Map<FontUtils.FontType, String> fontTypeLabels =
             new EnumMap<>(FontUtils.FontType.class);
+    private ZapNumberSpinner iconSize = null;
```
kingthorin, line 158: Remove the null initialization?

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Ping @psiinon

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsViewPanel.java
thc202: I didn't mean to dismiss this comment with my approval (I assumed this would be addressed either way), it's just I don't have additional comments.

kingthorin: I didn’t wanna block on such a minor detail. I guess I could tackle it tonight too 😉

ricekot: Thanks for tackling it @kingthorin.

thc202:
Thank you @ricekot for the rework!

psiinon (approved):

Issue #7228, opened 2022-04-21
Scan Rule with empty name in the Scan Progress window and Scan Policy window


globeone:
### Describe the bug

When viewing the Scan Progress window there is a mystery attack (blank title) between CORS Header and Spring Actuator Information Leak.
Also under Miscellaneous in the Scan Policy window 

### Steps to reproduce the behavior

- Start an Active Scan
- Click on the Progress icon 
-- ![image](https://user-images.githubusercontent.com/10052587/164455219-88058065-f8a2-4b80-a347-1931defc7f7e.png)
- scroll to the bottom of the window listing the attacks
- observe an attack missing a title between CORS Header and Spring Actuator Information Leak

OR 

- Open Analyse menu option
- Scan policy manager...
- Default Policy
- Miscellaneous
- Observe that the first attack is missing a title.

### Expected behavior

Expect for all attacks to have a title so one knows what attack vector is used. Also for debugging if something goes wrong with this specific attack.

### Software versions
```
OWASP ZAP
Version: D-2022-04-19

Installed Add-ons: [[id=accessControl, version=8.0.0],
[id=alertFilters, version=14.0.0], [id=amf, version=3.0.0],
[id=ascanrules, version=47.0.0], [id=ascanrulesAlpha,
version=38.0.0], [id=ascanrulesBeta, version=41.0.0],
[id=attacksurfacedetector, version=1.1.4], [id=authstats,
version=2.0.0], [id=automation, version=0.15.0],
[id=browserView, version=5.0.0], [id=bruteforce,
version=12.0.0], [id=callgraph, version=5.0.0],
[id=callhome, version=0.4.0], [id=commonlib,
version=1.10.0], [id=coreLang, version=16.0.0],
[id=custompayloads, version=0.11.0], [id=diff,
version=12.0.0], [id=directorylistv1, version=6.0.0],
[id=directorylistv2_3, version=4.0.0],
[id=directorylistv2_3_lc, version=4.0.0], [id=domxss,
version=13.0.0], [id=encoder, version=0.7.0], [id=exim,
version=0.2.0], [id=fileupload, version=1.1.0],
[id=formhandler, version=5.0.0], [id=fuzz, version=13.7.0],
[id=fuzzdb, version=8.0.0], [id=gettingStarted,
version=14.0.0], [id=graaljs, version=0.3.0], [id=graphql,
version=0.10.0], [id=help, version=15.0.0], [id=hud,
[...]

PR #7404, opened 2022-08-05
fix: Prevent Blank Scan Rule Names

kingthorin:
- ascan > In policy and progress dialogs display class name if rule name is blank.
- pscan > Options panel, display class name if rule name is blank.
- Missing names are logged via AddOnLoader.

Screenshots:
image
image
Dev example: `11728 [ZAP-BootstrapGUI] ERROR org.zaproxy.zap.control.AddOnLoader - Scan rule org.sasanlabs.fileupload.FileUploadScanRule does not have a name.` 

I didn't have a convenient way to test the pscan one.

Fixes zaproxy/zaproxy#7228

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -85,6 +85,7 @@ import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.commons.configuration.Configuration;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.parosproxy.paros.control.Control;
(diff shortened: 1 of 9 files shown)


kingthorin:
I didn't see a more generic/earlier way to do this. Scripts such as the one in the issue will still fail to retrieve a name. Please let me know if I've overlooked something.

kingthorin:
Ping @psiinon @thc202 @ricekot

kingthorin:
Discussed via IRC. Implement in AbstractPlugin (as final) and PluginPassiveScanner as `getDisplayName()`.

kingthorin:
For ascan I ended up doing this as a default method in `Plugin` because the places things are displayed they're Plugin not AbstractPlugin.

kingthorin:
Discussed via IRC: 
- It should log an error in dev and warn in "stable", could say: "Scan rule {} does not have a name"  (full class name) .
- When added/loaded in ZAP (start-up)

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #7409, opened 2022-08-05
Notify DB-open listeners immediately when opening session

ricekot:


zap/src/main/java/org/parosproxy/paros/control/Control.java
```diff
@@ -482,7 +483,6 @@ public void runCommandLineOpenSession(String fileName) throws Exception {
         Session session = Model.getSingleton().getSession();
         Model.getSingleton().openSession(fileName);
         log.info("Session file opened");
-        control.getExtensionLoader().databaseOpen(model.getDb());
```
thc202, line 485: For consistency we should move all `databaseOpen` notifications.

ricekot:
We don't need this change anymore, closing the PR.

PR #7413, opened 2022-08-10
Docker: Fix integrations tests

psiinon:
This change should make it easier to manage a larger set of "valid" output files for each test, and make the tests more readable.

WIP until the spider has been fixed to not find 404 URLs like https://example.com/a as that has messed up the results of these tests ;)


psiinon:
They now work locally 😁 
They are still a bit flaky, but the idea is that we can easily add as many valid files as we need.
At some point it would be really good to work out why the ZAP results have subtle differences, but that could be non trivial...

thc202 (approved):

kingthorin (approved):

PR #7414, opened 2022-08-10
Use rendered content when finding plain URLs

thc202:
Render the content to remove the HTML tags the elements might have,
which could match a URL (e.g. `/a` would be found in `</a>`).

---
Thanks to the integration tests and someone that was looking at them ;)


psiinon (approved):

psiinon:
> Thanks to the integration tests and someone that was looking at them ;)

More by luck than judgement 😛

kingthorin (approved):

kingthorin:
Updated zaproxy/zaproxy#3113 to ensure I catch this at the same time (should be a no brainer, but better to be specific/safe).

PR #7394, opened 2022-07-29
Misc Sites Tree / DDN Variant Updates

ricekot:

PR #7418, opened 2022-08-12
Update Sites Tree Restructuring Logic

ricekot:
Also fix a bug introduced in #7394 where loading 2 sessions back to back resulted in the second session having no sites tree.
It was happening because `hrefMap` was not being cleared upon loading a session.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/model/Context.java
+++ b/zap/src/main/java/org/zaproxy/zap/model/Context.java
@@ -36,7 +36,6 @@ import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.model.SiteMap;
 import org.parosproxy.paros.model.SiteNode;
 import org.parosproxy.paros.network.HttpMessage;
-import org.parosproxy.paros.network.HttpRequestHeader;
 import org.parosproxy.paros.network.HttpStatusCode;
 import org.parosproxy.paros.view.View;
 import org.zaproxy.zap.authentication.AuthenticationMethod;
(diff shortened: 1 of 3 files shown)


zap/src/main/java/org/zaproxy/zap/model/Context.java
```diff
@@ -604,59 +597,30 @@ public void run() {
 
     private void restructureSiteTreeEventHandler() {
         log.debug("Restructure site tree for context: " + this.getName());
+        // long time = System.nanoTime();
         List<SiteNode> nodes = this.getTopNodesInContextFromSiteTree();
         for (SiteNode sn : nodes) {
-            checkNode(sn);
+            updateNode(sn);
         }
+        // log.debug("restructuring took {} nanoseconds", System.nanoTime() - time);
```
ricekot, line 605: While the updated nodes use the correct name from the variant, the sites tree is updated a little more slowly than before. This is because it updates ALL the nodes now. Any suggestions to improve the performance are welcome here...

psiinon: @ricekot when does the slowdown occur? Whenever the Sites tree is updated or ??

ricekot: Whenever a context is modified and `Context#restructureSiteTree` is called.

psiinon: how significant is the slowdown?

PR #7429, opened 2022-08-19
Do not attempt to load extension with errors

thc202:
Add null check as the extension is null if there were errors loading
it.


kingthorin (approved):

ricekot:
Is it worth adding a log message if the loading fails?

thc202:
The errors are logged earlier (in `AddOnLoaderUtils`).

ricekot (approved):
Thanks!

PR #7439, opened 2022-09-01
Change auth header env vars to use ScriptVars

psiinon:
Allow the auth header env vars to be read and updated by scripts and code.


psiinon:
Will rework to use ScriptVars based on feedback..

psiinon:
Changed  to use ScripVars under the covers :)
The tests may still fail - they pass locally but for some reason the last tests got different header counts back in CI.
Can fix the tests if it keeps happening ...

psiinon:
That should fix some of them .. expecting other test to still fail..

psiinon:
Doh! I created a new unit test in the wrong package, and it was the old unit test (with the same name) in the right package that was failing 🤣

psiinon:
Test passed 😁 No longer WIP...

kingthorin (approved):

PR #7453, opened 2022-09-09
Fix StackOverflow in Analyser

thc202:
Keep track of the nodes traversed instead of using the ones analysed as not all of them might be (e.g. nodes not in context).
Use a `Set` to track them, override `equals` and `hashCode` on the node implementations.


psiinon (approved):

kingthorin (approved):
Thank you!

PR #7463, opened 2022-09-16
Add persistent callbacks

psiinon:
Persistent callbacks are like standard callbacks in that they are associated with an APIImplementor and a site.
They fake a very difficult to guess URL on that site and any requests to that URL to passed to the APIImplementor.
Unlike standard callbacks these are persisted in the ZAP config file, so once an APIImplementor has requested one for a site then it will stay until deleted.
These callbacks are managed by the APIImplementor which request them.
They are not expected to be used by many add-ons, just specific special cases.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/API.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/API.java
@@ -314,6 +314,17 @@ public class API {
                 }
             }
             if (callbackImpl == null) {
+                for (Entry<String, String> entry :
+                        this.getOptionsParamApi().getPersistentCallBacks().entrySet()) {
+                    if (url.startsWith(entry.getKey())) {
+                        callbackImpl = this.getImplementors().get(entry.getValue());
+                        if (callbackImpl != null) {
+                            break;
+                        }
+                    }
+                }
+            }
+            if (callbackImpl == null) {
                 logger.warn(
                         "Request to callback URL "
                                 + requestHeader.getURI().toString()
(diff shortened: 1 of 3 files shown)


kingthorin (commented):
Code looks ok, should the public methods in the API class have javadoc? Should the javadoc have @since tags?

psiinon:
Hopefully addressed those comments

kingthorin (approved):

zap/src/main/java/org/zaproxy/zap/extension/api/API.java
```diff
@@ -115,6 +115,8 @@ public enum RequestType {
 
     private Map<String, ApiImplementor> shortcuts = new HashMap<>();
 
+    private Map<String, Long> persistentCallBacks = null;
```
thc202, line 118: Remove the null initialisation, although why is it needed here?

zap/src/main/java/org/zaproxy/zap/extension/api/API.java
```diff
@@ -929,6 +931,48 @@ public void removeCallBackUrls(ApiImplementor impl) {
         this.callBacks.values().removeIf(impl::equals);
     }
 
+    private String getCallbackKey(ApiImplementor impl, String site) {
```
thc202, line 934: - `static`
 - `Callback` → `CallBack` to keep the same case as other methods.

zap/src/main/java/org/zaproxy/zap/extension/api/API.java
```diff
@@ -929,6 +931,48 @@ public void removeCallBackUrls(ApiImplementor impl) {
...
+        return impl.getClass().getCanonicalName() + ":" + site;
```
thc202, line 935: Might be better to use the API prefix rather than the class name, more stable (e.g. if the class is moved/renamed in a refactoring the key needs to be migrated).

zap/src/main/java/org/zaproxy/zap/extension/api/API.java
```diff
@@ -929,6 +931,48 @@ public void removeCallBackUrls(ApiImplementor impl) {
...
+     * @param impl the ApiImplementor requesting the callback URL
+     * @param site the site the callback will be on
+     * @return a callback URL
+     * @since 2.12.0
+     */
+    public String getPersistentCallBackUrl(ApiImplementor impl, String site) {
+        if (persistentCallBacks == null) {
+            persistentCallBacks = this.getOptionsParamApi().getPersistentCallbacks();
+        }
+        String callbackKey = getCallbackKey(impl, site);
```
thc202, line 950: `callBackKey`? Same in other places which use `callback`.

zap/src/main/java/org/zaproxy/zap/extension/api/OptionsParamApi.java
```diff
@@ -70,6 +76,7 @@ public class OptionsParamApi extends AbstractParam {
     private List<DomainMatcher> permittedAddresses = new ArrayList<>(0);
     private List<DomainMatcher> permittedAddressesEnabled = new ArrayList<>(0);
     private int nonceTimeToLiveInSecs = DEFAULT_NONCE_TTL_IN_SECS;
+    private Map<String, Long> persistentCallbacks = new HashMap<>();
```
thc202, line 79: Any reason not to use `String` for the value?

psiinon:
Updated based on comments received, and fixed the minor problem that the persistent callback URLs didnt actually work 😛

psiinon:
Should have fixed the lgtm comment..

PR #6570, opened 2021-04-29
fix: Disable FlatLaf native decorations on windows

kingthorin:
Native Decorations were causing various issues for me:
- ZAP wouldn't start maximized.
- Split pane dividers would drift noticeably between starts of ZAP.
- The close button would sometimes be inaccessible.

See: https://github.com/JFormDesigner/FlatLaf/pull/267 for further details.

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #7464, opened 2022-09-17
Restore component/view on configuration load

thc202:
Show the component/view previously saved when loading the configuration
to properly restore the state when the panels are dynamically loaded.


kingthorin:
I wonder if this will help with the drifting dividers I had observed prior to https://github.com/zaproxy/zaproxy/pull/6570

kingthorin (approved):

thc202:
Isn't that on the main window? This will just restore the components/views selected in the message panels that are dynamically loaded (dialogues added after ZAP is fully started).

kingthorin:
Yup, guess not.

psiinon (approved):

PR #7465, opened 2022-09-19
Dependency Updates

kingthorin:
This started with me just looking at log4j. But then grew into tackling others as identified by `dependencyUpdates` gradle task.

Dependency changes lead to tweaking the way various unit test setup lenient mocks.

Updated build files, LEGALNOTICE.md, and unit tests.

### DIFF
--- a/zap/zap.gradle.kts
+++ b/zap/zap.gradle.kts
@@ -73,7 +73,7 @@ dependencies {
     api("org.apache.commons:commons-text:1.9")
     api("edu.umass.cs.benchlab:harlib:1.1.3")
     api("javax.help:javahelp:2.0.05")
-    val log4jVersion = "2.17.2"
+    val log4jVersion = "2.19.0"
     api("org.apache.logging.log4j:log4j-api:$log4jVersion")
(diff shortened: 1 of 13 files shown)


zap/zap.gradle.kts
```diff
@@ -89,7 +89,7 @@ dependencies {
     api("org.jfree:jfreechart:1.5.3")
     api("org.jgrapht:jgrapht-core:0.9.0")
     api("org.swinglabs.swingx:swingx-all:1.6.5-1")
-    api("org.xerial:sqlite-jdbc:3.36.0.3")
+    api("org.xerial:sqlite-jdbc:3.39.3.0")
```
thc202, line 92: Not really about this PR but I added a task to #3113 to make this an implementation dependency. Core only needed it for the spider and since the spider is now in an add-on there's no reason to keep bundling in core for much longer, so the sooner we hide it the better. Guess we should add it to `commonlib`, `spider` and `ascanrulesBeta` use it.

thc202 (approved):

thc202:
Thank you!

ricekot (approved):
Thanks!

PR #7471, opened 2022-09-21
site node: use contains check for DDN prefix

ricekot:
Also tweak if conditions in setIncludedFromScope
and setExcludedFromScope to short-circuit && operator earlier.


thc202 (approved):

thc202:
Thank you!

PR #7473, opened 2022-09-25
Remove SQLite dependency

thc202:
Core no longer needs it (deprecated and unused code) and the other two
add-ons that need it will be changed to no longer rely on core.
Remove code that was instantiating the JDBC driver just for logging.

Part of https://github.com/zaproxy/zaproxy/issues/3113.

---
See also zaproxy/zap-extensions#4067.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderSVNEntriesParser.java
+++ b/zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderSVNEntriesParser.java
@@ -128,21 +128,6 @@ public class SpiderSVNEntriesParser extends SpiderParser {
                 fos.write(message.getResponseBody().getBytes());
                 fos.close();
 
-                if (getLogger().isDebugEnabled()) {
-                    org.sqlite.JDBC jdbcDriver = new org.sqlite.JDBC();
-                    getLogger()
-                            .debug(
-                                    "Created a temporary SQLite database file '"
-                                            + tempSqliteFile
-                                            + "'");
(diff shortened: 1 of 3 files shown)


kingthorin (approved):

thc202:
Updated to remove code that was instantiating the JDBC driver just for logging.

zap/src/main/java/org/zaproxy/zap/spider/parser/SpiderSVNEntriesParser.java
```diff
@@ -128,21 +128,6 @@ public boolean parseResource(HttpMessage message, Source source, int depth) {
...
-                            .debug(
-                                    "SQLite JDBC Driver is version "
-                                            + jdbcDriver.getMajorVersion()
-                                            + "."
-                                            + jdbcDriver.getMinorVersion());
-                }
-
                 // now load the temporary SQLite file using JDBC, and query the file entries within.
                 Class.forName("org.sqlite.JDBC");
                 String sqliteConnectionUrl = "jdbc:sqlite:" + tempSqliteFile.getAbsolutePath();
```
kingthorin, line 133: What about this continued part? It's not relevant because this is now done in the Spider add-on?

thc202: Right, this code is not in use anymore.

kingthorin (approved):

psiinon (approved):

PR #7478, opened 2022-09-26
Build releases with Java 11

thc202:
Update workflows and build to use Java 11 when building the releases,
version is defined with a env var to also be used by other repos.


psiinon (approved):

thc202:
No longer WIP.

kingthorin (approved):

PR #7491, opened 2022-09-27
Remove more unnecessary dependencies

thc202:
Remove Bouncy Castle and Commons Validator, core no longer needs them, the code is unused and deprecated.
Remove usages from the deprecated classes.

### DIFF
--- a/LEGALNOTICE.md
+++ b/LEGALNOTICE.md
@@ -31,9 +31,6 @@ and subject to their respective licenses.
 
 | Library                             | License                   |
 |-------------------------------------|---------------------------|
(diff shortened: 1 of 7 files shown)


kingthorin (approved):

ricekot (commented):
Can we delete the deprecated classes completely? After this change, not all of their methods are going to work as expected, so not sure if we should keep them around.

thc202:
Some classes are still being referenced by the Network add-on, for legacy support (it will no longer be needed for/after 2.12 though). But it also avoids breaking binary compatibility, which prevents errors in the unlikely event they are being called by something else.

ricekot (approved):
Thanks!

PR #7494, opened 2022-09-28
Correctly set Content-Length in AbstractPlugin

### LINKED ISSUE #7484: Using the Active Scan for GET targets sends a 'Content-Length' header
When using the Active Scan for a GET target Zap sends the header 'Content-Length: 0' in the requests.

This should not happen and can prevent scanning some servers which do reject the 'Content-Length: 0' header in GET requests.

Reproduction:
Install Python
Start Python http server: `python -m http.server 7777`
Start Zap GUI, proxy at 8080, turn off HUD
Make a request using curl: `curl --proxy http://localhost:8080 http://localhost:7777`
See the request in Zap history
Right click the request in Zap history, select "Open/Resend in request editor"
See that no "content-length" header is present
Right click the request again, select "Attack->Active Scan"
Check the requests in the Active Scan tab
See that all the request have the header `Content-Length: 0`

The requests in the attack should not have the "Content-Length" header as some systems reject GET requests with it.

Zap version used: 2.11.1

thc202:
Do not set the Content-Length header when the method does not require one, otherwise set/update it.
Expose the update method to allow scan rules to properly scan the header.

Fix #7484.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
@@ -74,6 +74,7 @@
 // ZAP: 2022/06/05 Remove usage of HttpException.
 // ZAP: 2022/08/03 Keep enabled state when setting default alert threshold (Issue 7400).
 // ZAP: 2022/09/08 Use format specifiers instead of concatenation when logging.
+// ZAP: 2022/09/28 Do not set the Content-Length header when the method does not require one.
 package org.parosproxy.paros.core.scanner;
 
 import java.io.IOException;
@@ -95,6 +96,7 @@ import org.parosproxy.paros.core.scanner.Alert.Source;
 import org.parosproxy.paros.model.HistoryReference;
 import org.parosproxy.paros.network.HttpHeader;
 import org.parosproxy.paros.network.HttpMessage;
+import org.parosproxy.paros.network.HttpRequestHeader;
 import org.parosproxy.paros.network.HttpStatusCode;
 import org.zaproxy.zap.control.AddOn;
 import org.zaproxy.zap.extension.anticsrf.ExtensionAntiCSRF;
@@ -297,7 +299,8 @@ public abstract class AbstractPlugin implements Plugin, Comparable<Object> {
         // always get the fresh copy
         message.getRequestHeader().setHeader(HttpHeader.IF_MODIFIED_SINCE, null);
         message.getRequestHeader().setHeader(HttpHeader.IF_NONE_MATCH, null);
-        message.getRequestHeader().setContentLength(message.getRequestBody().length());
+
+        updateRequestContentLength(message);
 
         if (this.getDelayInMs() > 0) {
             try {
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -323,6 +326,29 @@ protected void sendAndReceive(
...
+     * @since 2.12.0
+     */
+    protected void updateRequestContentLength(HttpMessage message) {
+        int bodyLength = message.getRequestBody().length();
+        String method = message.getRequestHeader().getMethod();
+        if (bodyLength == 0
+                && (HttpRequestHeader.CONNECT.equals(method)
+                        || HttpRequestHeader.DELETE.equals(method)
+                        || HttpRequestHeader.GET.equals(method)
+                        || HttpRequestHeader.HEAD.equals(method))) {
```
kingthorin, line 345: Should include OPTIONS?

kingthorin: I didn’t check the RFCs

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -323,6 +326,29 @@ protected void sendAndReceive(
...
+     * <p>For methods with no nor expected enclosed content the header is removed otherwise in all
+     * other cases the header is updated to match the length of the content.
+     *
+     * @param message the message to update.
+     * @since 2.12.0
+     */
+    protected void updateRequestContentLength(HttpMessage message) {
+        int bodyLength = message.getRequestBody().length();
+        String method = message.getRequestHeader().getMethod();
+        if (bodyLength == 0
```
kingthorin, line 341: If it has a body but is one of these methods it’ll still be updated. Should the unit tests include that?

thc202: That's `shouldAddContentLengthHeaderWhenNeeded`.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
thc202: The OPTIONS can have content, should add TRACE though.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
kingthorin: `shouldAddContentLengthHeaderWhenNeeded `is the case I was thinking. Thanks! I shoulda skipped around instead of just top downing it :grinning:

kingthorin (approved):

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
```diff
@@ -323,6 +326,30 @@ protected void sendAndReceive(
         parent.performScannerHookAfterScan(message, this);
     }
 
+    /**
+     * Updates the Content-Length header of the request.
+     *
+     * <p>For methods with no nor expected enclosed content the header is removed otherwise in all
```
ricekot, line 332:
```suggestion
     * <p>For methods with absent or unanticipated enclosed content, the header is removed otherwise in all
```

zap/src/test/java/org/parosproxy/paros/core/scanner/AbstractPluginUnitTest.java
```diff
@@ -63,6 +66,14 @@
 /** Unit test for {@link AbstractPlugin}. */
 class AbstractPluginUnitTest extends PluginTestUtils {
 
+    private static final List<String> METHODS_NO_ENCLOSED_CONTENT =
+            Arrays.asList(
```
ricekot, line 70: Can we start using Java 9+ features now? If so, this could be replaced with `List.of(...)`.

zap/src/test/java/org/parosproxy/paros/core/scanner/AbstractPluginUnitTest.java
```diff
@@ -63,6 +66,14 @@
 /** Unit test for {@link AbstractPlugin}. */
 class AbstractPluginUnitTest extends PluginTestUtils {
 
+    private static final List<String> METHODS_NO_ENCLOSED_CONTENT =
```
ricekot, line 69: Move to `AbstractPlugin` and make package-private?

thc202: I don't want to use a collection there, less garbage as is.

zap/src/test/java/org/parosproxy/paros/core/scanner/AbstractPluginUnitTest.java
thc202: Not yet.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractPlugin.java
thc202: Done.

thc202:
Changed to check the methods case insensitive.

kingthorin (approved):

psiinon (approved):

PR #7500, opened 2022-10-04
Exclude browser/OS updates by default

psiinon:
These are typically just noise. Things like Google analytics havnt been included as they might leak "interesting" data...


kingthorin (approved):

kingthorin:
Should this include an update method in the Constant class?

psiinon:
Ricardo asked that on IRC as well 😉 
I'm in 2 minds myself - I think it could be a pain to do and the medium term plan is now to move these to the network add-on and implement them as passthroughs (post 2.12)...

kingthorin:
Okay I’m that case let’s not overthink it. We can note it in the release notes, and any release blog that’s done.

thc202 (approved):

thc202:
Thank you!

PR #7511, opened 2022-10-07
Docker: Stable/live to debian, remove zapcli, update webswing

psiinon:

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2022-10-07
+ - Changed stable image to use debian:unstable-slim.
+ - Changed bare image to use eclipse-temurin:11-jre-alpine.
+ - Removed zap-cli from stable.
+ - Updated to use Webswing 22.1.5.
+
 ### 2022-09-28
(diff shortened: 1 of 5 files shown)


kingthorin (approved):

psiinon:
WIP while I double check the bare image..

psiinon:
I've changed bare to use `openjdk:11-jre` - it went down from 1.33GB to 875MB
Should we look at using that for the other images as well??

psiinon:
Arg https://hub.docker.com/_/openjdk - Deprecation notice ☹️  Will have to choose another one...

psiinon:
I've changed bare to use `eclipse-temurin:11-jre-alpine` - thats maintained and brings us down to 714MB
Note that it uses `apk` so if anyone builds on top of it they will probably have to change things.
If we're all ok with this I'll send a msg to the user group and it will get a mention in the release notes...

kingthorin:
The bare image seems to be using both apt-get and apk currently

psiinon:
Thats because the builder is using a different image :)
Can look at changing that as well if people think its important...

kingthorin:
No, I just didn’t catch the detail. Thanks.

thc202 (approved):

PR #7516, opened 2022-10-11
Restore Session title and Persist Session's state

### LINKED ISSUE #7423: `File`>`Persist Session...` should be disabled with open session
### Describe the bug

If one creates a session and selects the option to persist (I can't recall what it's called) then the persist option does nothing remotely useful.

1. Create a default session (that will save its stuff)
2. Use `File`>`Persist Session...`

### Expected behavior

* It should not be possible to get a dialog saying "you can't do this"
* The menu item should not be enabled if it won't function

### Software versions

2.11.1

### Path
image

image

### Disabled menus
image

image

_No response_

### Additional context

_No response_

- [ ] Yes

thc202:
Properly initialise the view in `ExtensionUiUtils` to have the session changed listener added on hook, which sets the main title to the session's name and controls the state of Persist Session menu/button.
Remove statements no longer needed in `MainMenuBar` that updated the button state, now done in a single method.

Fix #7423.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/MainMenuBar.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/MainMenuBar.java
@@ -287,9 +287,6 @@ public class MainMenuBar extends JMenuBar {
                         public void actionPerformed(java.awt.event.ActionEvent e) {
                             try {
                                 getMenuFileControl().newSession(true);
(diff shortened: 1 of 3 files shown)


psiinon (approved):

kingthorin (approved):

kingthorin:
Thank you!

PR #7517, opened 2022-10-12
Reserve ID for SSRF Scan Rule

ricekot:


docs/scanners.md
```diff
@@ -172,6 +172,7 @@ Scan rules:
 40043   Log4Shell (CVE-2021-44228, CVE-2021-45046)
 40044   Exponential Entity Expansion (Billion Laughs Attack)
 40045   Spring4Shell (CVE-2022-22965)
+40046   Server Side Request Forgery
```
ricekot, line 175: Should the ID be in the 2XXXX series instead?

kingthorin: 🤷

kingthorin (approved):

docs/scanners.md
thc202: The ranges/series no longer indicate anything.

kingthorin: Should we remove those notes?

kingthorin:
You're right we had cleaned it up already.

PR #7524, opened 2022-10-17
Address exceptions in AbstractAppParamPlugin

thc202:
Check that the variant and name value pair are non-null before using them when creating a new alert.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #7525, opened 2022-10-17
Do not rely on HttpSender implementation in tests

thc202:
Verify the HTTP messages sent directly instead of relying on the (old) `HttpSender` implementation and a server.


kingthorin (approved):
Looks okay to me. Haven't tested it. But, I don't see anything out of place or questionable.

psiinon (approved):

PR #7526, opened 2022-10-17
Fix PEP 257 issue in zap-baseline.py

tspearconquest:
Hello,

I noticed today that my internal copy of zap-baseline.py was out of date, so I pulled in the file from upstream and tried to commit it, however I got the below error from my internal repo's pre-commit hooks:

```
check docstring is first.................................................Failed
- hook id: check-docstring-first
- exit code: 1

docker/zap-baseline.py:109: Module docstring appears after code (code seen on line 45).
```

Quickly checking through the file, and googling, I realized the issue was that the alignment of `'''` was incorrect at lines 109 and line 149. These don't follow PEP 257, so that explained the failure.

I thought I'd submit this to fix it in the upstream repo. I've marked it as draft. In case you would rather this be solved in some other way, or left alone altogether, kindly let me know.

Signed-off-by: Thomas Spear <tspear@conquestcyber.com>

### DIFF
--- a/docker/zap-baseline.py
+++ b/docker/zap-baseline.py
@@ -106,12 +106,12 @@ def usage():
     print('')
     print('For more details see https://www.zaproxy.org/docs/docker/baseline-scan/')
 
-'''
+    '''
     This script is in the process of being converted to use the Automation Framework.
     If you map a directory to /zap/wrk then the zap.yaml file generated will be copied to that directory.
-    
+
     The following parameters are currently supported:
-    
+
     -c config_file
     -u config_url
     -m mins
@@ -128,13 +128,13 @@ def usage():
     -T mins
     -z zap_options
 
-    The following parameters are partially supported. 
+    The following parameters are partially supported.
     If you specify the '--auto' flag _before_ using them then the Automation Framework will be used:
 
     Currently none.
-    
+
     If any of the next set of parameters are used then the existing code will be used instead:
-    
+
     -D secs           need new delay/sleep job
     -i                need to support config files
     -l level          ditto
@@ -144,9 +144,9 @@ def usage():
     --hook            will need scripting support in the AF
     -g gen_file       may never support
     --autooff         will never support, may remove at some point
-    
-     
-'''
+
+
+    '''
 
 def main(argv):
     global min_level
@@ -344,7 +344,7 @@ def main(argv):
             home_dir = str(Path.home())
             yaml_file = os.path.join(home_dir, 'zap.yaml')
             summary_file = os.path.join(home_dir, 'zap_out.json')
-            
+
             with open(yaml_file, 'w') as yf:
 
                 # Add the top level to the scope for backwards compatibility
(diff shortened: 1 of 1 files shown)


psiinon:
We should include this as a check on all commits, and ideally have a task that automatically fixes these problems otherwise they will just reoccur...

tspearconquest:
I'll open a separate pull request to include some of the public pre-commit hooks we use! :)

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

thc202:
No pre-commits, should be in CI.

tspearconquest:
Can I ask why *no* pre-commits? I hope it's okay to ask. It's not on the FAQ, but "and/or" is, which is funny to me. :)

I'd also like to make a case for the `pre-commit` tool itself. I understand I'm new here, so I hope you'll hear me out, but feel free to just delete this email and ignore me too. No hard feelings either way. :)

First, I think I should clarify that I'm not referring to anything that would be mandatory for all contributors. The `pre-commit` tool allows hooks to be configured and used in a way that's optional for contributors but saves maintainers time.

It would take all 3 of the following conditions to activate the hooks:

1. A contributor would need to choose to install the tool on their system on their own, ie through homebrew.
2. There would need to be a .pre-commit-config.yml file in the root of the zap repo.
3. A contributor would need to run the command `pre-commit install` to activate the hooks on their own system.

If all 3 conditions are met, then the hooks would run before they commit. Otherwise, it's optional and the hooks just simply don't run, so the contributor's commit happens as normal.

I understand the PEP 257 check should be done in the CI, but I think of that as "the check should be done in the CI *as well* as in a pre-commit hook."

This would allow contributors who opt-in to be able to leverage the hooks without compelling them to install anything. Anyone can still commit without the pre-commit tool installed, or without running `pre-commit install` on their clone or their fork because it's simply a .yml file in the root.

I think you, as maintainers, can appreciate that an end of line fixer to remove unnecessary whitespace would be useful. I was really thinking only of that hook and the hook which caught the PEP 257 issue, though there are several others that you or Simon may find useful which are detailed on the pre-commit tool's website. Please, I encourage you to look and consider it.

Thanks for hearing me out.

--Thomas
________________________________

psiinon:
We need the CI checks to make sure they always run. Maintaining the pre-commits as well increases our work with no added benefit 😄

tspearconquest:
> For my 2 cents. no.1 and no.3 won't happen in 80% of cases.

That's the point, contributors don't necessarily
need to run the hooks or have the tool installed. Its geared towards maintainers.

> Having a gradle task that automatically checks/applies the required format (or whatever you want to refer to it as) uses existing tooling and makes things easier.
>
> similar to the existing Java formatting 😀

I totally get this perspective for "code smell" issues such as Java formatting and PEP 257.

I don't see how other things that aren't "code smell" such as fixing whitespace could necessarily be done through Gradle tasks, but then again I'm new to Gradle so I'll go read some documentation. Maybe there's a way. :)

> We need the CI checks to make sure they always run.

I totally understand and respect that. I think Rick has the right idea with having a Gradle task to catch the PEP 257 issue.

> Maintaining the pre-commits as well increases our work

Yes, updating the version periodically would increase the work, I admit. A Gradle task could do that work. ;)

> With no added benefit

I won't ask...

Thanks again for the feedback, I appreciate it.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

psiinon:
We do have gradle tasks for things like fixing formatting in the java code, and always happy to have more gradle tasks to help us. We just dont see the point in the pre-commit hooks ...

thc202:
My comment sounds more blunt than what was meant to, it's not "no" but "we shouldn't" or "there's no need to", we already have a [workflow to check the packaged scans](https://github.com/zaproxy/zaproxy/blob/f0b06c7d3828b951bef143751cfcb9eab7f95c97/.github/workflows/test-packaged-scans.yml), better update that than add yet another way of checking them.

tspearconquest:
This pull request itself is ready. I apologize that I haven't yet opened a separate issue/pr to make a gradle task as I haven't had time to work on that lately due to other priorities. Feel free to close this if you'd prefer to only merge the gradle task, however.

thc202:
Ok, I thought more was coming. I'm good with this being merged as is and the workflow being updated in other PR, we can also take a look at that.

kingthorin (approved):

PR #7533, opened 2022-10-20
Set passive scan data to scan rules

thc202:
Set the passive scan data to the scan rules and provide the message being scanned through it.
Simplifies the setup both in the production code and tests.
Remove helper test class no longer needed.


kingthorin (approved):

kingthorin:
Thanks.

zap/src/main/java/org/zaproxy/zap/extension/pscan/PassiveScanTask.java
```diff
@@ -119,13 +119,11 @@ public void run() {
...
-                            PluginPassiveScanner pps2 = pps.copy();
-                            pps2.init(psThread, msg, passiveScanData);
-                            scanner = pps2;
-                        } else {
-                            scanner.setParent(psThread);
+                            PluginPassiveScanner pps = ((PluginPassiveScanner) scanner).copy();
+                            pps.setHelper(passiveScanData);
+                            scanner = pps;
                         }
+                        scanner.setParent(psThread);
```
ricekot, line 126: Do we always want to do this? It wasn't being set before if scanner was not an instance of `PluginPassiveScanner`.

ricekot: Disregard, just looked at `init`.

ricekot (approved):
Thanks!

psiinon (approved):

Issue #7440, opened 2022-09-01
Manual Request Editor - Add Content-Type when switching to POST


Leyart:
### Is your feature request related to a problem? Please describe.

Opening a GET request with the request editor and switching the method with the dropdown from GET to POST moves the query parameters to the request body but it is not adding the correct content-type header, making the call silently fail.

### Describe the solution you'd like

As in BURP, when switching HTTP methods via Dropdown from GET to POST, the Content-Type: application/x-www-form-url-encoded should be added

### Describe alternatives you've considered

That seems to be the simplest solution to me. We could even specify how to convert the content to be sent in the body, but would make the flow clumsy

### Would you like to help fix this issue?

- [X] Yes

psiinon:
Sounds good to me 😁 @Leyart would you like this to be assigned to you?

Leyart:
Yes please, made me waste quite some time during an assessment because I was naturally expecting this to happen, I am now pissed at myself enough to fix it for good as soon as I have time

PR #7536, opened 2022-10-23
Set/remove the Content-Type when mutating the method

Leyart:
Opening a GET request with the request editor and switching the method with the dropdown from GET to POST moves the query parameters to the request body but it is not adding the correct content-type header, making the call silently fail.
This PR sets correctly the content-type header when switching methods in the requester add-on.

Fix zaproxy/zaproxy#7440

Signed-off-by: Leyart [leyart@gmail.com](mailto:leyart@gmail.com)

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
+++ b/zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
@@ -1016,6 +1016,8 @@ public class HttpMessage implements Message {
                 }
                 // Clear the body
                 body = "";
+                // Remove Content-Type if present
+                getRequestHeader().setHeader(HttpRequestHeader.CONTENT_TYPE, null);
 
             } else if (method.equals(HttpRequestHeader.POST)) {
                 // To be a port, move all URL query params into the body
@@ -1035,6 +1037,10 @@ public class HttpMessage implements Message {
                             sb.append('=');
                         }
                     }
+                    getRequestHeader()
+                            .setHeader(
+                                    HttpRequestHeader.CONTENT_TYPE,
+                                    HttpRequestHeader.FORM_URLENCODED_CONTENT_TYPE);
                     body = sb.toString();
                     uri.setQuery(null);
                 }
--- a/zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
@@ -38,6 +38,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Stream;
+import org.apache.commons.httpclient.URI;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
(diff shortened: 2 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -990,6 +990,11 @@ public void mutateHttpMethod(String method) {
             if (prevMethod.equalsIgnoreCase(method)) {
                 return;
             }
+            // When changing methods, remove the content-type if a method should not have a body
+            // This does not prevent the user for modifying the request afterwards if needed
```
thc202, line 994: Don't add this.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -990,6 +990,11 @@ public void mutateHttpMethod(String method) {
...
+            if (!(method.equals(HttpRequestHeader.POST) || method.equals(HttpRequestHeader.PUT))) {
```
thc202, line 995: IMO we should only change the Content-Type if we are actually changing the parameters around. (This also misses methods where no body is expected.)

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
```diff
@@ -1035,6 +1040,11 @@ public void mutateHttpMethod(String method) {
                             sb.append('=');
                         }
                     }
+                    // Explicitly set the Content-Type for the body of the POST
```
thc202, line 1043: Don't add this.

zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
```diff
@@ -273,6 +273,45 @@ void shouldNotSetContentEncodingsWhenSettingHttpRequestBody() {
         verify(body, times(0)).setContentEncodings(any());
     }
 
+    @Test
+    void shouldSetContentTypeWhenSettingMethodToPOST() throws Exception {
```
thc202, line 277: - `Setting` → `Mutating`
- `ToPOST` → `ToPost`

zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
```diff
@@ -273,6 +273,45 @@ void shouldNotSetContentEncodingsWhenSettingHttpRequestBody() {
...
+        assertThat(message.getRequestHeader().getMethod(), is(HttpRequestHeader.POST));
+        assertThat(
+                message.getRequestHeader().getURI().toString(), is("http://www.example.com:9000/"));
+        assertThat(
+                message.getRequestHeader().getHeader(HttpRequestHeader.CONTENT_TYPE),
+                is(HttpRequestHeader.FORM_URLENCODED_CONTENT_TYPE));
+    }
+
+    @Test
+    void shouldRemoveContentTypeWhenSettingMethodToGET() throws Exception {
```
thc202, line 296: - `Setting` → `Mutating`
- `ToGET` → `FromPost`
- Should be parameterized.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
Leyart: I am not sure if I understood the comment here. 
I have put this check because if we are switching from a POST or PUT back to a GET (or any other method that does not have a body), there should be no Content-Type set, since all the parameters are moved back as query parameters. That's why by setting it to null we are removing it.

thc202: Mutating from PUT to GET does not change anything. The body is not changed unless mutating from/to POST, so we shouldn't be changing unless it's those cases.

zap/src/test/java/org/parosproxy/paros/network/HttpMessageUnitTest.java
thc202: The parameterization should be on the method.

zap/src/main/java/org/parosproxy/paros/network/HttpMessage.java
Leyart: Thanks for the clarification, it makes sense. I will just remove the content-type if the previous method is POST then

Leyart:
Sorry for addressing the comments so late, I was on holidays. The content-type is not removed just if the previous method was a POST and set accordingly if we mutate to a POST.
The tests are now method parametrized, excluding POST that is set explicitly or mutated from explicitly in the tests.

Leyart:
@psiinon @thc202 should be in a good state now, let me know if it needs further adjustments

Leyart:
@kingthorin @thc202 @psiinon Happy new year! Just a ping for this PR that should be ready, would be great to be able to use it in the next release.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #7574, opened 2022-11-04
Add convenience methods for help and options buttons

kingthorin:
As discussed in https://github.com/zaproxy/zap-extensions/pull/4163

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/ViewDelegate.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/ViewDelegate.java
@@ -30,10 +30,15 @@
 // ZAP: 2019/06/01 Normalise line endings.
 // ZAP: 2019/06/05 Normalise format/style.
 // ZAP: 2022/05/29 Allow to obtain the OptionsDialog.
+// ZAP: 2022/11/07 Add convenience methods for help and options buttons.
 package org.parosproxy.paros.extension;
 
 import java.awt.Toolkit;
+import javax.swing.ImageIcon;
+import javax.swing.JButton;
 import javax.swing.KeyStroke;
+import org.parosproxy.paros.Constant;
+import org.parosproxy.paros.control.Control;
 import org.parosproxy.paros.view.MainFrame;
 import org.parosproxy.paros.view.MainPopupMenu;
 import org.parosproxy.paros.view.OptionsDialog;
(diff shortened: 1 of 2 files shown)


psiinon (approved):

zap/src/main/java/org/parosproxy/paros/extension/option/ExtensionOption.java
```diff
@@ -161,4 +166,29 @@ public String getAuthor() {
...
+     * Gets an options button, accepting a tooltip and options panel name (to be displayed when
+     * opening the options dialog).
+     *
+     * <p>Should be called/used only when in view mode.
+     *
+     * @param tooltip the text to be displayed in the button's tooltip.
+     * @param optionsPanelName the name of the panel to be displayed when the Options dialog panel
+     *     is opened.
+     * @return the options button, never {@code null}.
+     * @since TODO Add version
```
thc202, line 180: Can be replaced with the actual version (if we happen to use other version we have to replace anyway, if not less work).

zap/src/main/java/org/parosproxy/paros/extension/option/ExtensionOption.java
```diff
@@ -161,4 +166,29 @@ public String getAuthor() {
...
+     */
+    public JButton getOptionsButton(String tooltip, String optionsPanelName) {
```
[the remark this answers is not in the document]
kingthorin: Okay, I’ll move it.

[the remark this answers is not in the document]
thc202: To be clear the help button could be moved as well.

zap/src/main/java/org/parosproxy/paros/extension/ViewDelegate.java
```diff
@@ -146,4 +153,46 @@ default KeyStroke getMenuShortcutKeyStroke(int keyCode, int modifiers, boolean o
                 Toolkit.getDefaultToolkit().getMenuShortcutKeyMask() | modifiers,
                 onKeyRelease);
     }
+
+    /**
+     * Gets an options button, accepting a tooltip and options panel name (to be displayed when
+     * opening the options dialog).
+     *
+     * <p>Should be called/used only when in view mode.
```
thc202, line 161: I think this could be removed now, it's in a view class.

zap/src/main/java/org/parosproxy/paros/extension/ViewDelegate.java
```diff
@@ -146,4 +153,46 @@ default KeyStroke getMenuShortcutKeyStroke(int keyCode, int modifiers, boolean o
...
+     *
+     * @param tooltip the text to be displayed in the button's tooltip.
+     * @param optionsPanelName the name of the panel to be displayed when the Options dialog panel
+     *     is opened.
+     * @return the options button, never {@code null}.
+     * @since 2.13.0
+     */
+    static JButton getOptionsButton(String tooltip, String optionsPanelName) {
```
thc202, line 169: I'd go with `default` (I know I said static earlier but that was if they were kept in the extensions), to allow the view implementation to override the behaviour.

kingthorin:
Done & done

thc202 (approved):

thc202:
Thank you!

PR #7613, opened 2022-11-22
Lower case HTTP field names

thc202:
Use lower case for compatibility with HTTP/2 which requires the names to be all lower case.
For HTTP/1.x the case is (or should be) irrelevant.


kingthorin (approved):
Thanks!

PR #7619, opened 2022-11-26
Use correct path for proxy.pac API nonce

ricekot:
This fixes the cause of the following warning that is logged when `PAC file` is clicked on the API web UI homepage.

```
WARN  org.zaproxy.zap.extension.api.API - API nonce path was /OTHER/core/other/proxy.pac/ 
but call was for /OTHER/network/other/proxy.pac/ in request from 127.0.0.1
```

image

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/WebUI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/WebUI.java
@@ -34,6 +34,8 @@ public class WebUI {
 
     private API api;
     private boolean isDevTestNonce = false; // Manually change here to test nonces with the web UI
+    private static final String PAC_FILE_API_PATH = "/OTHER/network/other/proxy.pac/";
+    private static final String ROOT_CA_CERT_API_PATH = "/OTHER/network/other/rootCaCert/";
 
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/zaproxy/zap/extension/api/WebUI.java
```diff
@@ -479,7 +479,7 @@ public String handleRequest(URI uri, boolean apiEnabled) {
                                 + API.API_NONCE_PARAM
                                 + "="
                                 + API.getInstance()
-                                        .getLongLivedNonce("/OTHER/core/other/proxy.pac/")));
+                                        .getLongLivedNonce("/OTHER/network/other/proxy.pac/")));
```
ricekot, line 482: Should we link to the shortcut path instead?

thc202: Either way better to extract the path from the resource message and have it in a single place. (Same for the other API endpoint.)

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thank you

PR #7599, opened 2022-11-11
Fix for typo scanner.antiCSFR should be scanner.antiCSRF

sumeetpatil:
Fix #7598 

Signed-off-by: sumeet patil <sumeetp1991@gmail.com>

PR #7623, opened 2022-11-28
Enable all scripts option

vx-sec:


zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
```diff
@@ -1039,6 +1039,8 @@ private int addScriptsFromDir(File dir, ScriptType type, String targetEngineName
                                             type,
                                             false,
                                             f);
+                            if (this.getScriptParam().isEnableScriptsFromDirs())
+                              this.setEnabled(sw,true);
```
kingthorin, line 1043: Should have braces and proper formatting

kingthorin (commented):
Looks like this needs a `./gradlew spotlessApply`

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
vx-sec: I thought one expression conditionals are preferred without braces, will fix now. Where can I see the formatting guidelines for ZAP? The formatting that jdtls applies completely changes the indentation and other aspects of the file, so I don't think i can follow its standard.

psiinon:
The option should also be exposed in the options panel: https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/zaproxy/zap/extension/script/OptionsScriptPanel.java

psiinon:
Oh, you already mentionedthat 😁

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
psiinon: https://www.zaproxy.org/docs/developer/verifying-your-changes/#code-formatting 😁

kingthorin: The braces thing isn’t explicit, you can likely find out laces in the code base where they aren’t used on single statements. However, we’ve been moving to use them more consistently.

For formatting there are gradle tasks to apply and check.

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
```diff
@@ -1039,6 +1039,9 @@ private int addScriptsFromDir(File dir, ScriptType type, String targetEngineName
                                             type,
                                             false,
                                             f);
+                            if (this.getScriptParam().isEnableScriptsFromDirs()) {
+                                this.setEnabled(sw, true);
+                            }
```
thc202, line 1044: As mentioned in the issue, (IMO) this should only apply to directories specified through the options not to all the directories added.

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
vx-sec: Neat. Seems like spotlessCheck doesn't complain anymore.

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
vx-sec: I think i dont understand this right. What other directories will this be called on (besides the one specified through the options)?

vx-sec: Oh, the default one i suppose?

thc202: The add-ons can add their own directories (e.g. Community Scripts).

vx-sec: Fair enough. I'm guessing just an `and` check in that if for whether the current dir is in `getScriptParam().getScriptDirs()` should suffice?

thc202: Cleaner if the enabled state is specified when adding the options directories (see `postInit`), rather than do extra checks here.

vx-sec: If I understand correctly, that would mean modifying `addScriptsFromDir()` to also accept an optional bool argument?

thc202: Correct, note that it needs to keep binary compatibility (add a new method).

vx-sec: Take a look @ https://github.com/zaproxy/zaproxy/pull/7623/commits/eb1ad4dc8e2c0fed234b5f9ade83966c74872690

vx-sec:
I tried to add the checkbox to `OptionsScriptPanel`, but just can't figure out what the proper way to do it is. I should either modify the `AbstractMultipleOptionsBaseTablePanel` to have the checkbox and override it in `OptionScriptPanel`, or just monkeypatch a JCheckbox into `OptionScriptPanel` directly. Either way, I've never done any work on java guis and have no idea how things work together.

vx-sec:
https://github.com/zaproxy/zaproxy/pull/7623/commits/03ec102fe1742e4c3bc205d73908139937693c6a is what i came up with. Don't know how good it is. But it's lacking any localization support.

vx-sec:
I think i figured out localization too.

Well, this was easy. You guys made zap very approachable for new devs. Props to you!

zap/src/main/java/org/zaproxy/zap/extension/script/OptionsScriptPanel.java
```diff
@@ -62,6 +65,7 @@ private void initialize() {
 
         this.add(new ZapHtmlLabel(Constant.messages.getString("options.script.label.dirs")), gbc);
 
+        this.add(enableLoadedScripts, gbc);
```
kingthorin, line 68: Could you add a screenshot?

kingthorin (approved):
Seems reasonable to me. (Note: I haven't pulled the branch and tested.)

zap/src/main/java/org/zaproxy/zap/extension/script/OptionsScriptPanel.java
kingthorin: Thanks

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
```diff
@@ -46,6 +46,7 @@ public class ScriptParam extends AbstractParam {
     private static final String SCRIPT_ENABLED_KEY = "enabled";
     private static final String SCRIPT_DIRS = "dirs";
     private static final String SCRIPT_CONFIRM_REMOVE_DIR = "confRemdir";
+    private static final String SCRIPT_ENABLE_SCRIPTS_FROM_DIRS = "enableScriptsFromDirs";
```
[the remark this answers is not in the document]
vx-sec: I think  SCRIPTS_BASE_KEY is only used for individual script configs. I thought following the same path as SCRIPT_DIRS would be more reasonable, so now SCRIPT_ENABLE_SCRIPTS_FROM_DIRS and SCRIPT_DIRS are on the same level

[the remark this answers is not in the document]
thc202: Not really, the other two keys are wrong as well, they should be all under the base `script` element.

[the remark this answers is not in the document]
vx-sec: In that case, yes. It would make more sense to put them all under scripts element. But I think that  will be a really big breaking change. People have used  `-confg dirs` for as long as it existed.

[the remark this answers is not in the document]
thc202: That's not a problem, we can migrate the old keys as needed. I'm not asking you to fix the other keys though, just add the new one in the right place but if you want to fix them, great :)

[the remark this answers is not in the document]
vx-sec: From what I see, it's only a matter of 
```diff
diff --git a/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java b/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
index a7a8a5484..c798de1d5 100644
--- a/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
@@ -44,9 +44,9 @@ public class ScriptParam extends AbstractParam {
     private static final String SCRIPT_TYPE_KEY = "type";
     private static final String SCRIPT_FILE_KEY = "file";
     private static final String SCRIPT_ENABLED_KEY = "enabled";
-    private static final String SCRIPT_DIRS = "dirs";
-    private static final String SCRIPT_CONFIRM_REMOVE_DIR = "confRemdir";
-    private static final String SCRIPT_ENABLE_SCRIPTS_FROM_DIRS = "enableScriptsFromDirs";
+    private static final String SCRIPT_DIRS = SCRIPTS_BASE_KEY + ".dirs";
+    private static final String SCRIPT_CONFIRM_REMOVE_DIR = SCRIPTS_BASE_KEY + ".confRemdir";
+    private static final String SCRIPT_ENABLE_SCRIPTS_FROM_DIRS = SCRIPTS_BASE_KEY + ".enableScriptsFromDirs";
 
     private static final Logger logger = LogManager.getLogger(ScriptParam.class);
 
```
Tested, seems to work okay.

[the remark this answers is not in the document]
thc202: That's not enough, you would also have to do something like: https://github.com/zaproxy/zaproxy/pull/7599/files#diff-83e062c0df1e7f4b18030a796521e134c4f6387cf8aad53444d05fce7f10533aR314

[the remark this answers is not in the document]
vx-sec: Done.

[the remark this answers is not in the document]
thc202: It still needs the tests.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
```diff
@@ -54,13 +56,15 @@ public class ScriptParam extends AbstractParam {
     private Set<ScriptWrapper> scripts;
     private List<File> scriptDirs;
     private boolean confirmRemoveDir = true;
+    private boolean enableScriptsFromDirs = false;
```
thc202, line 59: Remove the unnecessary initialisation.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
```diff
@@ -54,13 +56,15 @@ public class ScriptParam extends AbstractParam {
...
 
     public ScriptParam() {}
 
     @Override
     protected void parse() {
         defaultScript = getString(PARAM_DEFAULT_SCRIPT, "");
         defaultDir = getString(PARAM_DEFAULT_DIR, "");
+        migrateOldKeys();
```
thc202, line 67: Better if it's the first statement of the method.

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
```diff
@@ -1015,9 +1022,11 @@ public int addScriptsFromDir(File dir) {
      * @param dir the directory from where to add the scripts.
      * @param type the script type, must not be {@code null}.
      * @param targetEngineName the engine that the scripts must be of, {@code null} for all engines.
+     * @param enableScriptsFromDirs whether or not to enable added scripts
```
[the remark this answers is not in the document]
vx-sec: Do you want to change the variable only in this method or in the whole class? I'd argue enableScriptsFromDirs is a tad bit more descriptive of its real purpose.b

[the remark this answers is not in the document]
vx-sec: Fair enough

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptParam.java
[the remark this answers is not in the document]
thc202: That the old keys are properly migrated to the new ones, e.g. `ScannerParamUnitTest#shouldMigrateOldOptions`. Check also with multiple `dirs` (it works fine but tests should still show it).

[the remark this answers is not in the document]
vx-sec:
```java
    @Test
    void shouldMigrateOldOptions() {
        // Given
        File dir1 = new File("/path/to/dir1");
        File dir2 = new File("/path/to/dir2");
        List<File> dirs = new ArrayList<>();
        dirs.add(dir1);
        dirs.add(dir2);
        configuration.setProperty("script.confRemdir", true);
        configuration.setProperty("script.dirs", dirs);
        // When
        param.load(configuration);
        // Then
        assertThat(param.isConfirmRemoveDir(), is(equalTo(true)));
        assertThat(param.getScriptDirs(), is(equalTo(dirs)));
        assertNull(configuration.getProperty("dirs"));
        assertNull(configuration.getProperty("confRemdir"));
    }
}
```
Something like this should work right? I'm really inexperienced with unit testing.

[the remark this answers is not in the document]
vx-sec: cant get `assertThat(param.getScriptDirs(), is(equalTo(dirs)));` to work. `param.getScriptDirs()` seems to return a null list, but it shouldnt.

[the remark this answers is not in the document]
vx-sec: Okay, I got it working, had to pass a string list to `setProperty`, not File list for `dirs`.  Good enough?
``` java
  @Test
  void shouldMigrateOldOptions() {
    // Given
    File dir1 = new File("/path/to/dir1");
    File dir2 = new File("/path/to/dir2");
    List<File> dirs = new ArrayList<>();
    dirs.add(dir1);
    dirs.add(dir2);
    
    configuration.setProperty("confRemdir", true);
    configuration.setProperty("dirs", new String[]{dir1.toString(), dir2.toString()});
    // When
    param.load(configuration);
    // Then
    assertThat(param.isConfirmRemoveDir(), is(equalTo(true)));
    assertThat(param.getScriptDirs(), is(equalTo(dirs)));
    assertNull(configuration.getProperty("dirs"));
    assertNull(configuration.getProperty("confRemdir"));
  }
```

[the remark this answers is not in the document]
vx-sec: Don't know what actual paths to put into the commit to be honest.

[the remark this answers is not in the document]
vx-sec: Bump?

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
```diff
@@ -1039,6 +1046,9 @@ private int addScriptsFromDir(File dir, ScriptType type, String targetEngineName
                                             type,
                                             false,
                                             f);
+                            if (enableScripts) {
+                                this.setEnabled(sw, true);
+                            }
```
thc202, line 1051: This is enabling before actually loading the contents of the script, for `extender` scripts it means they will not be properly installed (i.e. errors or no-op).

Scripts which are in the directories but whose engine is installed later are not enabled, I think they should but if the behaviour is documented I'm fine either way.

zap/src/test/java/org/zaproxy/zap/extension/script/ScriptParamUnitTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @BeforeEach
+    void setUp() {
+        param = new ScriptParam();
+        configuration = new ZapXmlConfiguration();
+        param.load(configuration);
+    }
+
+    @Test
+    void shouldMigrateOldOptions() throws IOException {
```
thc202, line 65:
```suggestion
    void shouldMigrateOldOptions(@TempDir Path tempDir) throws IOException {
```
which does not require manual cleanup later.

thc202:
The [help](https://github.com/zaproxy/zap-core-help/) should be updated to mention the new option.

zap/src/test/java/org/zaproxy/zap/extension/script/ScriptParamUnitTest.java
```diff
@@ -0,0 +1,87 @@
+/*
+ * Zed Attack Proxy (ZAP) and its related class files.
+ *
+ * ZAP is an HTTP/HTTPS proxy for assessing web application security.
+ *
+ * Copyright 2020 The ZAP Development Team
```
thc202, line 6: Update to current year.

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
thc202: Ping.

vx-sec: I got carried away with some other work, I'll try to get back to this in a week's time, sorry for keeping this open.

thc202: No worries and thank you!

0xcrypto:
Ping. Looks like an approval away.

thc202:
There are still pending comments.

psiinon:
I've fixed the conflicts, and checked that an extender script loads and is enabled ok...

psiinon (approved):

zap/src/main/java/org/zaproxy/zap/extension/script/ExtensionScript.java
thc202: For the record, the first part of the comment is no longer a problem (at least with the engines provided by add-ons). The second part of the comment will be addressed with docs.

thc202 (approved):

thc202:
Thank you!

Issue #7589, opened 2022-11-08
Improve add-on update file handling


thc202:
When ZAP updates add-ons it tries to delete/update old files (the ones bundled in the add-on and the add-on itself) but in Windows that might not always work, it says the files are in use or open (other OSes allow to delete the files anyway).

This is most noticeable with WebDrivers add-ons which provide binaries, that might indeed be in use if the user launched a browser from within ZAP. If that happens the binaries might remain with the old version until the user manually deletes them.

ZAP should not assume that can delete the files. Ideally the update itself should not require to delete any files for the new add-on version to work properly. Any required clean ups should probably be done on a following start to ensure that the old files are no longer in use.

thc202:
Removed the Windows part, better if done for all OSes.

thc202:
Ref: https://groups.google.com/g/zaproxy-users/c/uWKRWUZYFJE/m/16z1WoLEAgAJ

thc202:
Closing with the referenced PRs.

PR #7626, opened 2022-11-30
Postpone uninstallation tasks

### LINKED ISSUE #7593: Check first if an add-on can be uninstalled before uninstalling dependents
If an add-on can't be unloaded its dependents should not be uninstalled, otherwise the functionality provided by those add-ons is unavailable while the add-on that was meant to be uninstalled remains available (thus still usable).

From:
https://groups.google.com/g/zaproxy-users/c/CcEkYYOho34/m/k8_QwzuGAwAJ
https://github.com/zaproxy/zaproxy/issues/7555#issuecomment-1309368672

thc202:
Do not attempt to uninstall the add-on that has extensions that can't be unloaded, instead postpone the uninstallation for the next start.
Fix #7593.

Postpone the deletion of the add-on bundled files that were not successfully deleted (e.g. they are in use) during the uninstallation.
Part of #7589.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/control/AddOnInstaller.java
+++ b/zap/src/main/java/org/zaproxy/zap/control/AddOnInstaller.java
@@ -186,9 +186,46 @@ public final class AddOnInstaller {
      * @see Extension
      * @see PassiveScanner
      * @see org.parosproxy.paros.core.scanner.Plugin
+     * @deprecated (2.13.0) Use {@link #uninstall(AddOn, AddOnUninstallationProgressCallback, Set,
+     *     PostponedTasksRunner)} instead.
      */
+    @Deprecated
     public static boolean uninstall(
             AddOn addOn, AddOnUninstallationProgressCallback callback, Set<AddOn> installedAddOns) {
+        return uninstall(addOn, callback, installedAddOns, null);
+    }
+
+    /**
+     * Uninstalls all the (dynamically installable) components ({@code Extension}s, {@code Plugin}s,
+     * {@code PassiveScanner}s and files) of the given {@code addOn}.
+     *
+     * <p>The components are uninstalled in the following order (inverse to installation):
+     *
+     * <ol>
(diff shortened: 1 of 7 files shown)


zap/src/main/java/org/zaproxy/zap/control/AddOnInstaller.java
```diff
@@ -186,9 +186,46 @@ public static boolean uninstall(AddOn addOn, AddOnUninstallationProgressCallback
...
      */
+    @Deprecated
     public static boolean uninstall(
             AddOn addOn, AddOnUninstallationProgressCallback callback, Set<AddOn> installedAddOns) {
+        return uninstall(addOn, callback, installedAddOns, null);
+    }
+
+    /**
+     * Uninstalls all the (dynamically installable) components ({@code Extension}s, {@code Plugin}s,
+     * {@code PassiveScanner}s and files) of the given {@code addOn}.
```
kingthorin, line 200:
```suggestion
     * {@code PassiveScanner}s and files) of the given {@code AddOn}.
```

kingthorin (commented):
Thanks! Just one minor typo, I believe.

zap/src/main/java/org/zaproxy/zap/control/AddOnInstaller.java
thc202: That's the parameter.

kingthorin (approved):

kingthorin:
Thanks

ricekot (commented):
Should we attempt to run postponed tasks when ZAP is closed as well?
Also, what happens if e.g. `-addoninstall addon` is specified and there is also a postponed task to uninstall the add-on?

thc202:
That would lead to the same problem as we have now, trying to unload/remove things that are/might still be in use. 

The postponed tasks are executed before any add-on/extension is loaded, in that case it will be uninstalled and then installed.

psiinon (approved):

PR #7642, opened 2022-12-05
Build multi-platform Docker images

ricekot:
- Changed all images to use debian:bullseye-slim instead of unstable-slim.  
 - Updated live and weekly GitHub workflows to build multi-platform images (linux/amd64, linux/arm64).  
 - Optimized multi-platform builds by using the same builder stage across platforms ([Ref.](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/)).  
 - Use the `docker/dockerfile:1` image to build containers with BuildKit ([Ref.](https://docs.docker.com/build/buildkit/dockerfile-frontend/)).  
 - Replaced usage of `COPY` with `COPY --link` ([Ref.](https://docs.docker.com/engine/reference/builder/#copy---link)).  
 - Replaced `COPY` then `RUN chown` with `COPY --chown` ([Ref.](https://docs.docker.com/engine/reference/builder/#copy)).

### DIFF
--- a/.github/workflows/release-live-docker.yml
+++ b/.github/workflows/release-live-docker.yml
@@ -29,7 +29,7 @@ jobs:
         with:
           context: docker
           file: docker/Dockerfile-live
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
           push: true
           tags: |
             owasp/zap2docker-live:latest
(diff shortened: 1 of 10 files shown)


thc202:
Thank you!

kingthorin (approved):

psiinon (approved):

PR #7645, opened 2022-12-07
Install Network add-on async through the API

thc202:
Install asynchronously to return a proper response, as the API is being served by the add-on which will be uninstalled.


kingthorin:
Worth mentioning something in the change log?

thc202:
This API endpoint is only available in dev mode.

kingthorin (approved):

psiinon (approved):

PR #7668, opened 2022-12-26
Refactor Dockerfiles with best practices

PeterDaveHello:
Implement some best practices like: prevent leaving and remove unnecessary cache, also remove some unnecessary commands.

### DIFF
--- a/docker/Dockerfile-bare
+++ b/docker/Dockerfile-bare
@@ -7,8 +7,7 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	curl \
 	openjdk-11-jdk \
 	xmlstarlet \
-	unzip && \
-	apt-get clean
+	unzip
 
 WORKDIR /zap
 
--- a/docker/Dockerfile-bare
+++ b/docker/Dockerfile-bare
@@ -7,7 +7,8 @@ RUN apt-get update && apt-get install -q -y --fix-missing \
 	curl \
 	openjdk-11-jdk \
 	xmlstarlet \
-	unzip
+	unzip && \
+	rm -rf /var/lib/apt/lists/*
 
 WORKDIR /zap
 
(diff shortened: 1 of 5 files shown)


thc202:
What are the size and time differences?

PeterDaveHello:
Won't be too much actually, so I didn't precisely calculate the result, but as you can see, some of them are alreay done in the Dockerfiles, just not all of it done in the proper way, so a quick fix should be helpful.

thc202:
The commits could be squashed, changes are not that complex or disjoint that need to be in their own commits.

IMO the changelog should be updated, even if it's just to mention the Docker files were tidied up.

PeterDaveHello:
@thc202 just want to share the idea is trying to focus every commit to a single and simple topic, which will make sure the commit is atomic, should make sure the change are easier to be understood and traced in the future, especially when something went wrong, but the same type of changes cross different Dockerfiles, will just be in the same commit. Still got your point and understand your idea though!

BTW, should I record the change in a certain file? Not sure if I missed it in the docs...

kingthorin:
Thanks for clarifying your thoughts on the commits. I don't disagree, however, the changes are all related and the individual commit history isn't really relevant to the greater project in the end. So squashing it is best.

Docker changes are tracked in: https://github.com/zaproxy/zaproxy/blob/main/docker/CHANGELOG.md

thc202:
This needs to be rebased, one of the files was removed (not in use).

docker/Dockerfile-bare
```diff
@@ -36,7 +36,6 @@ COPY --link --chown=1000:1000 policies /home/zap/.ZAP/policies/
 RUN echo "zap2docker-bare" > /zap/container
 
 RUN /usr/sbin/adduser -u 1000 -h /home/zap -s /bin/bash -D zap && \
```
thc202, line 38:
```suggestion
RUN /usr/sbin/adduser -u 1000 -h /home/zap -s /bin/bash -D zap
```

thc202:
Addressed the comments.

kingthorin (approved):

psiinon (approved):

PeterDaveHello:
Thanks for help from @thc202, and sorry for the late reply/update 😅

thc202:
Thank you!

PR #7690, opened 2023-01-12
History filters enhancements

skomak:
Hi,

I set minimum filters window size as by default it was too packed and horizontal scrollers were unnecessarily activated and data was not fitting textarea width.
Also, I added sample predefined regexps.

Previously it looked like this:
obraz

After:
obraz

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
@@ -74,7 +74,10 @@ public class HistoryFilterPlusDialog extends AbstractDialog {
     private JList<String> confidenceList = null;
     private JList<String> tagList = null;
     private JTextArea regexInc = null;
+    private JList<String> regexSamples = null;
     private JTextArea regexExc = null;
+    private JButton btnRegxInc = null;
+    private JButton btnRegxExc = null;
 
     private DefaultListModel<String> tagModel = null;
 
@@ -85,6 +88,7 @@ public class HistoryFilterPlusDialog extends AbstractDialog {
     private JScrollPane confidenceScroller = null;
     private JComboBox<String> notesComboBox = null;
     private JScrollPane urlRegxIncScroller = null;
+    private JScrollPane urlRegxSamples = null;
     private JScrollPane urlRegxExcScroller = null;
 
     /**
@@ -120,12 +124,14 @@ public class HistoryFilterPlusDialog extends AbstractDialog {
         this.setContentPane(getJPanel());
         this.setVisible(false);
         this.setTitle(Constant.messages.getString("history.filter.title"));
+
         if (Model.getSingleton().getOptionsParam().getViewParam().getWmUiHandlingOption() == 0) {
             this.setSize(600, 300);
         }
+
         centreDialog();
         this.getRootPane().setDefaultButton(btnApply);
-        this.setMinimumSize(new Dimension(1000,600));
+        this.setMinimumSize(new Dimension(1000, 600));
         this.pack();
     }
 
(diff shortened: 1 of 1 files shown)


kingthorin:

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
```diff
@@ -119,11 +124,14 @@ private void initialize() {
...
         this.setVisible(false);
         this.setTitle(Constant.messages.getString("history.filter.title"));
+
         if (Model.getSingleton().getOptionsParam().getViewParam().getWmUiHandlingOption() == 0) {
             this.setSize(600, 300);
         }
+
         centreDialog();
         this.getRootPane().setDefaultButton(btnApply);
+        this.setMinimumSize(new Dimension(1000, 600));
```
thc202, line 134: The minimum size should not be set as a rule (e.g. VM with 800x600 resolution would have problems). Maybe you meant the size and even then, better if packed and let the layout manager size the components to their preferred size.

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
```diff
@@ -423,26 +467,28 @@ private JPanel getJPanel2() {
             jPanel2.add(
                     new JLabel(Constant.messages.getString("history.filter.label.urlincregex")),
                     gbc04);
+            jPanel2.add(new JLabel("Sample regexps:"), gbc06);
```
thc202, line 470: The strings should be i18n'ed.

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
```diff
@@ -494,6 +540,84 @@ private JScrollPane getConfidenceScroller() {
...
+
+            regexSamples = new JList<>(URL_REGX_INC_LIST);
+            regexSamples.setSelectionMode(ListSelectionModel.MULTIPLE_INTERVAL_SELECTION);
+            regexSamples.setLayoutOrientation(JList.VERTICAL);
+            regexSamples.setVisibleRowCount(URL_REGX_INC_LIST.length);
+            urlRegxSamples = new JScrollPane(regexSamples);
+        }
+        return urlRegxSamples;
+    }
+
```
thc202, line 558: Not quite sure of usefulness of this without being persisted.

psiinon:
The Sample Regexes take up a lot of space, and are one of the least important parts IMO.
How about changing to something like:
Screenshot 2023-01-13 at 09 56 56
(just the right hand side).
The bottom control is a pulldown - the <kbd>copy</kbd> button copies the selected regex and then the usual OS `Control-V` can be used to paste it..
And yes! Found a good excuse to include an HTML button in a github comment 😛

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
skomak: Default packing looks terrible, you need to expand the window every single time you run ZAP to see all values. It is not convenient. Also, using tools like ZAP with resolution 800x600 is not something you would find convenient. I can redesign window but I believe we need field values visible without need of expanding the window.

kingthorin:
The Notes control seems to be missing in the "After" screenshot?

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
kingthorin: If the packing looks terrible then preferred sizes could/should be specified, or other mechanisms leveraged to make it work nicely (note: I am not and do not claim to be a java UI expert, far from it). Setting the minimum completely takes control away from the user, and potentially makes things unusable if they happen to be on underspeced systems/VMs.

zap/src/main/java/org/zaproxy/zap/extension/history/HistoryFilterPlusDialog.java
```diff
@@ -494,6 +540,84 @@ private JScrollPane getConfidenceScroller() {
         return confidenceScroller;
     }
 
+    private JScrollPane getUrlRegxSamples() {
+        if (urlRegxSamples == null) {
+            // TODO save the list in better place
+            final String[] URL_REGX_INC_LIST = {
+                ".*\\.(js|css|woff2|png|jpeg|jpg|gif|svg)$", ".*\\.(html|asp|jsp|php)$"
```
kingthorin, line 547: Not sure why I didn't notice this earlier. But doing extensions with an ending anchor will mean that URLs with params are not matched by the filter. ex: https://example.com/foo/bar.php?baz=lmnop

kingthorin:
@skomak ping

PR #7701, opened 2023-01-22
Add nearest match handling to Jump To History functionality

kingthorin:
- ExtensionHistory > Extract and use getHistoryIds() method.
- PopupMenuJumpTo > Add handling for when there isn't an exact match for the ID the user enters. In this case it will be the nearest match after (above) the ID requested, or the end of the list if the requested ID is beyond the end of the existing set.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
@@ -103,6 +103,7 @@
 // ZAP: 2023/01/10 Tidy up logger.
 // ZAP: 2023/01/11 Add "jump to" right-click menu item (Issue 7362).
 // ZAP: 2023/01/11 Prevent NPE in "showInHistory" when tab doesn't have focus.
+// ZAP: 2023/01/22 Add utility getHistoryIds() method.
 package org.parosproxy.paros.extension.history;
 
 import java.awt.EventQueue;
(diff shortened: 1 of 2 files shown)


zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -463,13 +464,19 @@ public void run() {
...
     }
 
-    private void searchHistory(HistoryFilter historyFilter) {
+    /**
+     * Returns a sorted list of History IDs for the visible history references.
+     *
+     * @return a list of History IDs for the visible history references.
+     * @since 2.13.0
+     */
+    public List<Integer> getHistoryIds() {
```
thc202, line 473: - Is this public only because of the context menu?
- `getHistoryIdsOfHistType` by contract does not guarantee ordering. The method says the list is sorted but then it's being sorted in the context menu.

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -463,13 +464,19 @@ public void run() {
...
         Session session = getModel().getSession();
+        List<Integer> list;
```
thc202, line 475: This change is not necessary, could just return the result too.

zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuJumpTo.java
```diff
@@ -109,7 +110,19 @@ private void jump() {
...
                 extension.showInHistory(extension.getHistoryReference(historyId));
             } catch (NullPointerException npe) {
-                LOGGER.info("Could not jump to history Id: {}", historyId);
+                List<Integer> list = extension.getHistoryIds();
+                Collections.sort(list);
+                int nearestIdx = Math.abs(Collections.binarySearch(list, historyId)) - 1;
+                HistoryReference nearestRef = null;
+                try {
+                    nearestRef = extension.getHistoryReference(list.get(nearestIdx));
+                } catch (IndexOutOfBoundsException ioobe) {
```
thc202, line 119: Better check than use the exception.

zap/src/main/java/org/zaproxy/zap/extension/history/PopupMenuJumpTo.java
```diff
@@ -109,7 +110,19 @@ private void jump() {
...
+                    nearestRef = extension.getHistoryReference(list.get(list.size() - 1));
+                }
+                int nearestId = nearestRef.getHistoryId();
+                extension.showInHistory(extension.getHistoryReference(nearestId));
```
thc202, line 123: Can't it use the `nearestRef`?

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
kingthorin: Doh, I had actually relocated the sort to here last night but apparently didn’t push it ☹️

kingthorin: Yes it's public only because of the context menu, though it could be needed by neonmarker or other code.

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
kingthorin: Should be inside the try :wink: Also synchronize is no longer necessary here.

kingthorin:
Seems to work fine when the table is filtered. In that it doesn't error, and if the row is visible it does jump to it.

kingthorin:
Hey team, could I get some reviews on this? I think it’s pretty set to go.

zap/src/main/java/org/parosproxy/paros/extension/history/ExtensionHistory.java
```diff
@@ -463,26 +464,37 @@ public void run() {
...
+            List<Integer> list =
+                    getModel()
+                            .getDb()
+                            .getTableHistory()
+                            .getHistoryIdsOfHistType(
+                                    session.getSessionId(),
+                                    HistoryReference.TYPE_PROXIED,
+                                    HistoryReference.TYPE_ZAP_USER,
+                                    HistoryReference.TYPE_PROXY_CONNECT);
+            Collections.sort(list);
```
psiinon, line 486: How will pulling all of the relevant Ids scale is the session is very large?
Should we be sorting in the db query instead of in the code?
These comments may well have applied to the previous PR, sorry ;)

kingthorin: We think this is okay. It's already done elsewhere in code and we do allow for the whole table to be sorted.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #7729, opened 2023-02-09
Support optional to optional extension dependencies

ricekot:
Allow optional extensions to depend on other optional extensions in the same add-on.

Signed-off-by: ricekot <github@ricekot.com>

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/control/AddOn.java
+++ b/zap/src/main/java/org/zaproxy/zap/control/AddOn.java
@@ -1335,7 +1335,7 @@ public class AddOn {
         ExtensionRunRequirements extensionRequirements =
                 new ExtensionRunRequirements(addOn, extension.getClassname());
         requirements.addExtensionRequirements(extensionRequirements);
-        for (AddOnDep dependency : extension.getDependencies()) {
+        for (AddOnDep dependency : extension.getAddOnDependencies()) {
             String addOnId = dependency.getId();
             if (addOnId == null) {
                 continue;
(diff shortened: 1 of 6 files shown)


zap/src/main/java/org/zaproxy/zap/control/AddOnLoader.java
```diff
@@ -126,6 +126,8 @@ public class AddOnLoader extends URLClassLoader {
      */
     private Map<String, AddOnClassLoader> addOnLoaders = new HashMap<>();
 
+    private Map<String, Collection<AddOnClassLoader>> extensionLoaders = new HashMap<>();
```
thc202, line 129: The class loaders are being added but never removed from the map, when the extension/add-on is unloaded.

zap/src/main/java/org/zaproxy/zap/control/AddOnClassLoader.java
```diff
@@ -320,16 +328,12 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
             }
         }
 
-        try {
```
thc202, line 323: I think we should keep this the default behaviour (classes of the add-on should have priority over the ones in the dependencies) and allow to change it when needed.

zap/src/main/java/org/zaproxy/zap/control/AddOn.java
```diff
@@ -1372,6 +1372,18 @@ private static void calculateExtensionRunRequirements(
             calculateRunRequirementsImpl(
                     availableAddOns, extensionRequirements, addOn, addOnDep, checkLibs);
         }
+
+        for (var dependency : extension.getExtensionDependencies()) {
+            String classname = dependency.getClassname();
+            if (classname == null) {
+                continue;
+            }
+            if (addOn.getExtensionsWithDeps().contains(classname)) {
```
thc202, line 1381: It should also check that it can be loaded (i.e. the transitive dependencies are all met).

ricekot:
Closing as per discussions in Slack.

The original intention of this PR was to support the case where the GraphQL add-on wanted to use some functionalities of the quickstart add-on (launch the user's preferred browser with GraphQL IDE URL open) without explicitly depending on it.

There are other ways to achieve this use case, e.g. move any functionality needed by two or more add-ons into commonlib (in this case, launching a browser).

Issue #7767, opened 2023-03-04
Copy Replacer Rules


hmolsen:
### Is your feature request related to a problem? Please describe.

I sometimes have to write multiple replacer rules which I want to use interchangeably. Today I need to add a new rule all the time and copy paste all the parameters from the old rule, which is difficult, because I can't have both modal dialogs open at the same time.

### Describe the solution you'd like

Next to the "Add..." and "Modify..." buttons, I would love to see a "Duplicate" button, duplicatingy the currently selected Replacer rule.

### Describe alternatives you've considered

Today, I use a text file to write down the options of the first replacer rule and then paste it into a new replacer rule

### Would you like to help fix this issue?

- [ ] Yes

psiinon:
This shouldnt be too hard.
The code is all in https://github.com/zaproxy/zap-extensions/tree/main/addOns/replacer/src/main/java/org/zaproxy/zap/extension/replacer
If anyone fancies having a go at this then comment here...

thc202:
Better if this is handled generically for all options though.

aayushhyadav:
Hi, I would like to work on this issue. Could you please tell how to reproduce the issue on UI?

hmolsen:
Alrighty, sure.
You go to the `Tools` menu and choose `Replacer Options`, or you just press <kbd>Ctrl</kbd>+<kbd>R</kbd>
![image](https://user-images.githubusercontent.com/5484682/223629486-94db5ec8-78a0-4208-9e91-09e535e13da5.png)

In this window you can see, e.g., `Remove CSP` as an entry. Say, I like this rule and want to keep it, but I also would like to have a similar rule, which removes X-Unwanted-Header from all Responses.
Today I would Click `Add...` and then I am presented with this a little unintuitive options window:
![image](https://user-images.githubusercontent.com/5484682/223630050-8beb7aa7-f7fb-43e4-89da-37aedb1d8756.png)

So what would make my life way easier would be a button, where I can select the rule that I like:
![image](https://user-images.githubusercontent.com/5484682/223630432-4f538a68-3470-4b4e-802e-054c72842a1e.png)

and click the to-be-developed  `Copy...` button on the right side, which would create a new rule, which could be called "Remove CSP Copy" and have all the settings of the Remove CSP rule.

That would be awesome 🤩

kingthorin:
Thanks @hmolsen i had intended to loop back around and put something like that together, just hadn’t had a chance yet.

[...]

PR #7783, opened 2023-03-15
added copy options rules feature

aayushhyadav:
related to #7767


psiinon:
Will the Duplicate option be valid for all such dialogs?
Might be better to have an allowDuplication option (as per allowModification) with the default as false.
That way we can enable it per dialog and then double check it works.
We dont really have any UI unit tests so we'd need to manually test these anyway...

aayushhyadav:
So will have to update `AbstractMultipleOptionsBaseTablePanel(
            AbstractMultipleOptionsBaseTableModel<E> model, boolean allowModification)` by adding another parameter of allowduplicate right? This currently has 3 usages, so do you mean that I should pass false from all 3 instances?

aayushhyadav:
Any idea how to resolve the build issue? I have tried to clean and build the project but the issue persists.

aayushhyadav:
> Will the Duplicate option be valid for all such dialogs? Might be better to have an allowDuplication option (as per allowModification) with the default as false. That way we can enable it per dialog and then double check it works. We dont really have any UI unit tests so we'd need to manually test these anyway...

I have added the allowDuplication option and passed the same value as that of allowModification at all instances

aayushhyadav:
> You need to keep binary compatibility, e.g. add a new constructor instead of changing that one.

I am trying to add something like `protected AbstractMultipleOptionsBaseTablePanel(boolean allowDuplicate)`. From where should I invoke this constructor? I cannot use the `this(false)` from `public AbstractMultipleOptionsBaseTablePanel(AbstractMultipleOptionsBaseTableModel<E> model)` or `protected AbstractMultipleOptionsBaseTablePanel(AbstractMultipleOptionsBaseTableModel<E> model, boolean allowModification)` as they already use `this` and `super` respectively.

aayushhyadav:
One more query, can't we use the `allowModification` flag for duplicate functionality? So duplicate will only be available when modify is present. Or is it necessary to have them independent of each other?

kingthorin:
They should be independent. There may be situations where modifying one is fine but duplicating isn’t and vice versa.

zap/src/main/java/org/zaproxy/zap/view/AbstractMultipleOptionsBaseTablePanel.java
```diff
@@ -170,17 +173,43 @@ public void mousePressed(MouseEvent me) {
...
+        if (e != null) {
+            getMultipleOptionsModel().addElement(e);
+        }
+    }
+
     protected void selectionChanged(boolean entrySelected) {
         if (modifyButton != null) {
             modifyButton.setEnabled(entrySelected);
         }
+        duplicateButton.setEnabled(entrySelected);
```
thc202, line 212: This will lead to `NullPointerException` when `allowDuplicate` is false.

zap/src/main/java/org/zaproxy/zap/view/AbstractMultipleOptionsBaseTablePanel.java
```diff
@@ -170,17 +173,43 @@ public void mousePressed(MouseEvent me) {
         }
     }
 
+    protected AbstractMultipleOptionsBaseTablePanel(
+            boolean allowDuplicate, AbstractMultipleOptionsBaseTableModel<E> model) {
+        super(model);
+
+        if (allowDuplicate) {
```
thc202, line 180: Doing this here means the duplicate will be the first button, IMO it should be after the add button.

thc202:
https://github.com/zaproxy/zaproxy/pull/7783#issuecomment-1483781358

You can change the implementation of the existing constructors just not their signature.

zap/src/main/java/org/zaproxy/zap/view/AbstractMultipleOptionsBaseTablePanel.java
aayushhyadav: Cool. I'll initialize the add button in this constructor

aayushhyadav:
> [#7783 (comment)](https://github.com/zaproxy/zaproxy/pull/7783#issuecomment-1483781358)
> 
> You can change the implementation of the existing constructors just not their signature.

Is the current approach fine?

zap/src/main/java/org/zaproxy/zap/view/AbstractMultipleOptionsBaseTablePanel.java
thc202: Better if the logic is kept in the same constructor, add a private constructor which is called by both constructors.

thc202:
See https://github.com/zaproxy/zaproxy/pull/7783#discussion_r1148520421

aayushhyadav:
Could anyone please review the changes? Are anymore changes required?

kingthorin:
I'd be happy to finish this, any further feedback?

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [07d4f7f0\-065d\-4148\-971c\-2d68cfa25e88](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=07d4f7f0-065d-4148-971c-2d68cfa25e88&branch=aayushhyadav%2Fcopy-replacer-rules)

tomaszn:
Closing the dialogue window should cancel the Duplicate operation, and the new entry should not appear, right?

kingthorin:
That seems logical to me.

PR #7791, opened 2023-03-22
Prevent Toolkit init when updating without view

thc202:
Do not initialise the icon if there's no view.

Part of #3798.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #7792, opened 2023-03-23
Remove comments in main resource messages

thc202:
Remove comments that are now outdated, no longer apply, or are unnecessary.


ricekot (approved):
Thanks!

kingthorin (approved):

PR #7793, opened 2023-03-23
Set ZAP silent with `ZAP_SILENT` env var

### LINKED ISSUE #7581: I would like to be able to set `ZAP_OFFLINE` to disable all default remote connections
### Is your feature request related to a problem? Please describe.

If you are running in a CI system, you generally want a tightly controlled environment.

Additionally, you want to be able to set items as defaults for any ZAP instance. The simplest way to do this is via environment variables.

### Describe the solution you'd like

I would like to be able to set `ZAP_OFFLINE=true` and have all 'reach out' options disabled.

I believe that this would be equivalent to setting `-silent` and `-notel` at the command line.

### Describe alternatives you've considered

The only centralized alternative that would work is to pull from another repository and/or set an environment variable with the exact command line options.

The drawback to this is that it is just as much work as having users set the option directly and is not able to be driven centrally.

### Screenshots

_No response_

### Additional context

_No response_

- [ ] Yes

thc202:
Change `CommandLine` to also read the env var to set ZAP silent.

Fix #7581.


zap/src/main/java/org/parosproxy/paros/CommandLine.java
```diff
@@ -160,6 +167,17 @@ public CommandLine(String[] args) throws Exception {
         }
     }
 
+    private void readEnv(UnaryOperator<String> env) {
+        if (env.apply("ZAP_SILENT") != null) {
```
ricekot, line 171: Worth extracting a constant?

zap/src/main/java/org/parosproxy/paros/CommandLine.java
```diff
@@ -144,11 +147,16 @@ public class CommandLine {
     private boolean devMode;
 
     public CommandLine(String[] args) throws Exception {
+        this(args, System::getenv);
+    }
+
+    CommandLine(String[] args, UnaryOperator<String> env) throws Exception {
```
ricekot, line 153: This is pretty neat. I wouldn't have thought of using a `UnaryOperator` here.

PR #7801, opened 2023-03-28
Override `host` in `VariantHeader`

thc202:
Change `VariantHeader` to override the `host` header so the injected payloads are sent as expected.

Part of #1318.


psiinon (approved):

kingthorin (approved):
Thanks!

PR #7804, opened 2023-03-30
Ensure passive scan runs even with high traffic

thc202:
Obtain the ID of the last message to scan after any interruption.

The `PassiveScanController` is interrupted when responses are received, with high/constant traffic it can be interrupted often that it would not be able to get the ID of the last message after the first interrupt preventing the new messages from being scanned.
This only happens with new sessions, ones that had no messages persisted.

---
Reported in IRC.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #7805, opened 2023-03-31
Update timing assertion

thc202:
Allow the time to be equal not just greater when asserting the time the passive task took to execute, it might take less than a millisecond, e.g.:
```
Expected: is a value greater than <1680237962023L>
     but: <1680237962023L> was equal to <1680237962023L>
```


ricekot (approved):
Thanks!

psiinon (approved):

PR #7813, opened 2023-04-06
Do not remove add-ons from installation dir

### LINKED ISSUE #7703: Do not remove add-ons from the install dir
### Describe the bug

We document the `-dir` option as a way to start ZAP with a clean config: https://www.zaproxy.org/docs/desktop/cmdline/
However on some platforms add-ons are deleted from the install dir when they are updated.
If the user updates one of the mandatory add-ons then it breaks this option - ZAP will not start.

1. Install ZAP
2. Update ZAP (making sure at least one mandatory option is updated)
3. Run ZAP using the `-d` option

### Expected behavior

ZAP will start with all of the orignal add-ons

### Software versions

2.12.0

### Screenshots

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes

thc202:
Allow different homes to share the same base add-ons including the mandatory ones.
Block the add-ons that are not mandatory when not being updated.

Fix #7703.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/control/AddOnLoader.java
+++ b/zap/src/main/java/org/zaproxy/zap/control/AddOnLoader.java
@@ -23,6 +23,7 @@ import java.awt.EventQueue;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.FilenameFilter;
+import java.io.IOException;
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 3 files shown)


ricekot (approved):
Thank you!

kingthorin (approved):

kingthorin:
Thanks!

PR #7817, opened 2023-04-10
Remove Crawljax log config

thc202:
Let the `spiderAjax` add-on configure the log instead, which knows better what classes need to be adjusted (i.e. they might change with Crawljax updates).

Depends on zaproxy/zap-extensions#4612.


thc202:
Now depending on other PR since the other PR is taking longer to merge and good to get these changes in 2.13.

kingthorin (approved):

psiinon (approved):

PR #7828, opened 2023-04-14
Fix JavaDoc issues

sigee:


thc202:
I'm not sure there's any advantage fixing JavaDoc problems in deprecated classes.

sigee:
> I'm not sure there's any advantage fixing JavaDoc problems in deprecated classes.

You may be right, but from a clean code perspective, If you need code, you should keep them clean. If you don't need the code. remove it. If it is deprecated but needed, keeping it clean is still a good practice in my opinion.

thc202:
In an ideal world, yes, you can have that perspective no matter what.

thc202 (approved):

thc202:
Thank you! (While I still think this is not needed, it's less JavaDoc warnings so…)

kingthorin (approved):

Issue #7421, opened 2022-08-15
Import default context behavior is painful


jsoref:
### Describe the bug

I had a session, and I liked its context, so I saved it. I created a new session and tried to import the default context. I got an error.

### Steps to reproduce the behavior

1. Have an existing session (from a long time ago)
2. File>Open Session...
3. File>Export context
4. Save
5. File>New session
6. File>Import Context...
7. Select the context
8. Click OK

### Expected behavior

If there's a conflict with an existing context name, it could:
* offer to replace the context
* offer to rename the context
* suggest how to delete the existing conflicting context

### Software versions

2.11.1

### Screenshots

![image](https://user-images.githubusercontent.com/2119212/184708501-389a44ca-c54f-4b0a-9783-4fcff7f725c5.png)

Note that it doesn't even tell me the name of the existing context (if a dialog is going to show something like this, it should at least provide that information...)

### Would you like to help fix this issue?

- [ ] Yes

psiinon:
All good suggestions 😄

PR #7836, opened 2023-04-21
import context: Handle duplicates more gracefully

kingthorin:
- MenuFileCntrol.java > When a name is duplicate prompt the user for an alternative. If they provide one use it, if they don't then skip the import.
- Session.java > Allow to import a XML Config.
- Messages.properties > Name/value pairs supporting the new context name dialog.

Fixes zaproxy/zaproxy#7421


kingthorin:
I haven't yet looked at the API functionality, I wanted to get agreement on this concept of handling first.

thc202:
The API does not need to be changed.

zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -595,8 +596,13 @@ public String getDescription() {
                     return;
                 }
                 // Import the context
-                Model.getSingleton().getSession().importContext(file);
+                Context ctx = Model.getSingleton().getSession().importContext(file);
```
thc202, line 599: Extract a method to create the context and then check if the name already exists (and ask the user for the new name) before importing.

kingthorin: Do you mean in `Session` I should create a new method that creates a `Context` without checking the name? (Or setting it into the Session I guess.)

kingthorin:
I believe this is now ready for review.

zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -595,14 +598,35 @@ public String getDescription() {
...
                 // Import the context
-                Model.getSingleton().getSession().importContext(file);
-
-                // Show the dialog
-                View.getSingleton()
-                        .showSessionDialog(
-                                Model.getSingleton().getSession(),
-                                Constant.messages.getString("context.list"),
-                                true);
+                Context ctx = Model.getSingleton().getSession().importContext(file, false);
```
ricekot, line 601: If we validate the name before importing the context, could we avoid creating the new `importContext(file, validateName)` method? It would mean we validate twice but that should be fine?

zap/src/main/java/org/parosproxy/paros/model/Session.java
```diff
@@ -1490,10 +1492,38 @@ public Context importContext(File file)
...
+        return importContext(file, true);
+    }
+
+    /**
+     * Imports a context from the specified (XML) file.
+     *
+     * @param file the (XML) file that contains the context data
+     * @param validateName boolean indicating whether the context name should be validated and
+     *     handled "locally"
+     * @return the imported {@code Context}, already added to the session.
```
kingthorin, line 1504: …. if validateName was true

zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
kingthorin: How would we have the name if we haven’t imported it?

ricekot: Ah, I missed that.
Alternatively, what do you think about accepting a function for validating the name (e.g. using [Consumer](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/function/Consumer.html) as a parameter in `importContext`?

```java
public Context importContext(File file, Consumer<String> validateContextName)
```

and then `MenuFileControl` can pass a lambda with the custom name validation logic.

The body of the existing `importContext(File file)` could be changed to
```java
return importContext(file, this::validateContextName)
```

thc202: The idea was to read the XML file directly (and validate the name) not call the import method.

kingthorin: Okay I could do it that way, it isn't what we discussed a few weeks ago...

thc202: I think it was what was decided in the end.

kingthorin: So am I going to write the new name to the file then import it? Or am I duplicating the import functionality?

kingthorin: How about I extract the middle part of the importContext method into a new one called readContext or something like that, which returns the Context based on having parsed the file content. Then it can be used in the menu and locally in Session?

thc202: https://github.com/zaproxy/zaproxy/pull/7836#discussion_r1200377197

I was thinking adding a `importContext(ZapXmlConfiguration)`, once the UI overrides the name it can import that.

https://github.com/zaproxy/zaproxy/pull/7836#discussion_r1200472424

In that case yo need to pass the context and the file when importing?

kingthorin: In the second one it would take the file and return the context. No validation, not adding/setting in Session.

But I’m good with you first idea around passing the xmlconfig.

kingthorin:
Okay hopefully this is more what people were picturing :grinning:

zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -594,15 +598,39 @@ public String getDescription() {
...
-                View.getSingleton()
-                        .showSessionDialog(
-                                Model.getSingleton().getSession(),
-                                Constant.messages.getString("context.list"),
-                                true);
+                ZapXmlConfiguration config = new ZapXmlConfiguration(file);
+                String ctxName = config.getString(Context.CONTEXT_CONFIG_NAME);
+                Session session = Model.getSingleton().getSession();
+
+                String name = null;
```
thc202, line 605: Does it really need other variable for the name?

zap/src/main/java/org/parosproxy/paros/model/Session.java
```diff
@@ -1495,6 +1496,40 @@ public Context importContext(File file)
         String name = config.getString(Context.CONTEXT_CONFIG_NAME);
         validateContextName(name);
```
thc202, line 1498: This should be done in `importContext(ZapXmlConfiguration)`.

zap/src/main/java/org/parosproxy/paros/model/Session.java
```diff
@@ -1495,6 +1496,40 @@ public Context importContext(File file)
...
 
+        Context c = importContext(config);
+        model.importContext(c, config);
+
+        c.restructureSiteTree();
+
+        addContext(c);
+        saveContext(c);
+
+        return c;
```
thc202, line 1507:
```suggestion
        return importContext(config);
```

zap/src/main/java/org/parosproxy/paros/model/Session.java
```diff
@@ -1557,13 +1592,6 @@ public Context importContext(File file)
         for (Object obj : config.getList(Context.CONTEXT_CONFIG_DATA_DRIVEN_NODES)) {
             c.addDataDrivenNodes(new StructuralNodeModifier(obj.toString()));
         }
-
-        model.importContext(c, config);
-
-        c.restructureSiteTree();
-
-        addContext(c);
-        saveContext(c);
```
thc202, line 1566: This should still be done here.

zap/src/main/java/org/parosproxy/paros/model/Session.java
kingthorin: But then we would be back to passing a Boolean or consumer and how was then any better than the earlier impl with a file?

kingthorin: Okay, right so it just won't matter if it happens again. Okay 👍

kingthorin:
Okay, done again, I hope :wink:

zap/src/main/java/org/parosproxy/paros/control/MenuFileControl.java
```diff
@@ -594,15 +598,35 @@ public String getDescription() {
...
+                                                    "context.import.duplicate.dialog.message",
+                                                    ctxName),
+                                            Constant.messages.getString(
+                                                    "context.import.duplicate.dialog.title"),
+                                            JOptionPane.QUESTION_MESSAGE,
+                                            null,
+                                            null,
+                                            ctxName);
+                }
+                if (ctxName != null && !ctxName.isBlank()) {
```
kingthorin, line 620: As discussed via IRC this should include handling for empty/null names on the off chance that a user edited or crafted the context file.

kingthorin:
Hopefully good now.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #7840, opened 2023-04-24
Reserve ID for alerts raised by the GraphQL Add-on

ricekot:
This will be used with non-scan-rule alerts e.g. an informational alert used to inform the user about introspection being enabled on a GraphQL endpoint.

Also add `[Deprecated]` to IDs of alerts raised by the retired HTTPS info add-on.


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
Thanks!

PR #7843, opened 2023-04-28
Deprecate `Proxy`/`ProxyServer` related methods

thc202:
Deprecate `ExtensionLoader` methods that use the deprecated classes `Proxy` and `ProxyServer`, those classes were superseded by the Network add-on.
Suppress removal warnings in deprecated class.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java
@@ -99,6 +99,7 @@
 // ZAP: 2022/08/17 Install updates before running other cmdline args.
 // ZAP: 2022/11/23 Refresh tabs menu when tabs are removed.
 // ZAP: 2023/01/10 Tidy up logger.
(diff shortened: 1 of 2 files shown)


psiinon (approved):

zap/src/main/java/org/parosproxy/paros/extension/ExtensionLoader.java
```diff
@@ -278,8 +279,9 @@ public int getExtensionCount() {
      * @param proxyServer the proxy server to add, must not be null.
      * @since 2.8.0
      * @see #removeProxyServer(org.parosproxy.paros.core.proxy.ProxyServer)
+     * @deprecated (2.13.0) Use the network add-on to create proxies.
      */
-    @SuppressWarnings("deprecation")
+    @Deprecated(since = "2.13.0", forRemoval = true)
```
kingthorin, line 284: I like this format, we should leverage this more :wink:

kingthorin (approved):

PR #7844, opened 2023-04-28
Retain add-on's mandatory state

thc202:
Ensure mandatory add-ons keep their state when updated, otherwise they could be uninstalled after.


psiinon (approved):

ricekot (approved):
Thanks!

PR #7847, opened 2023-05-02
Add auto-detect checking strategy

psiinon:
Note that auto-detect is _not_ implemented in the core - its implemented in authhelper.

Also change a load of protected methods to public to make it easier for add-ons like authhelper :grin:

The longer term plan is to move a lot of the auth code out of the core .. but that will be non trivial..


kingthorin (approved):
Seems reasonable to me.

thc202 (approved):

thc202:
Thank you!

PR #7854, opened 2023-05-08
[NodeJS API] remove await from nodejs generator

njmulsqb:
Await has been removed from the return function to let the users be able to use promises as well.


njmulsqb:
redundant? I have removed it from the return statement, need to generate the API files again. @thc202

thc202:
I'm talking about the `async`: https://github.com/zaproxy/zaproxy/blob/7259b47769ef4df332fed6681c85828143f8dc6f/zap/src/main/java/org/zaproxy/zap/extension/api/NodeJSAPIGenerator.java#L150

njmulsqb:
Oh `async` is useless now, you're right. Let me remove it

thc202 (approved):

thc202:
Thank you!

PR #7857, opened 2023-05-09
Allow to add/remove authentication methods #2620

forgedhallpass:
Implemented option to dynamically add `AuthenticationMethodType`s from extensions to be used through the REST API.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
@@ -20,9 +20,11 @@
 package org.zaproxy.zap.extension.authentication;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import org.apache.commons.configuration.Configuration;
 import org.apache.commons.configuration.ConfigurationException;
@@ -35,6 +37,7 @@ import org.parosproxy.paros.db.DatabaseException;
 import org.parosproxy.paros.db.RecordContext;
 import org.parosproxy.paros.extension.ExtensionAdaptor;
 import org.parosproxy.paros.extension.ExtensionHook;
+import org.parosproxy.paros.extension.ExtensionHookMenu;
 import org.parosproxy.paros.extension.ExtensionPopupMenuItem;
 import org.parosproxy.paros.model.Session;
 import org.zaproxy.zap.authentication.AuthenticationMethod;
@@ -60,7 +63,7 @@ public class ExtensionAuthentication extends ExtensionAdaptor
     public static final int EXTENSION_ORDER = 52;
 
     /** The NAME of the extension. */
-    public static final String NAME = "ExtensionAuthentication";
+    public static final String NAME = ExtensionAuthentication.class.getSimpleName();
 
     /** The ID that indicates that there's no authentication method. */
     private static final int NO_AUTH_METHOD = -1;
@@ -71,8 +74,13 @@ public class ExtensionAuthentication extends ExtensionAdaptor
     /** The automatically loaded authentication method types. */
     List<AuthenticationMethodType> authenticationMethodTypes = new ArrayList<>();
 
+    AuthenticationAPI api;
+
     /** The context panels map. */
-    private Map<Integer, ContextAuthenticationPanel> contextPanelsMap = new HashMap<>();
+    private final Map<Integer, ContextAuthenticationPanel> contextPanelsMap = new HashMap<>();
+
+    private final List<AuthenticationMethodsChangeListener> authenticationMethodsChangeListeners =
+            new ArrayList<>();
 
     private PopupContextMenuItemFactory popupFlagLoggedInIndicatorMenuFactory;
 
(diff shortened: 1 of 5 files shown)


zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
```diff
@@ -107,18 +115,25 @@ public String getUIName() {
...
-        extensionHook.addContextDataFactory(this);
+        this.extensionHook = extensionHook;
+        this.extensionHook.addContextDataFactory(this);
 
         if (getView() != null) {
-            extensionHook.getHookMenu().addPopupMenuItem(getPopupFlagLoggedInIndicatorMenu());
-            extensionHook.getHookMenu().addPopupMenuItem(getPopupFlagLoggedOutIndicatorMenu());
+            final ExtensionHookMenu hookMenu = this.extensionHook.getHookMenu();
+            hookMenu.addPopupMenuItem(getPopupFlagLoggedInIndicatorMenu());
+            hookMenu.addPopupMenuItem(getPopupFlagLoggedOutIndicatorMenu());
```
kingthorin, line 124: Please revert these (and any other) unrelated changes.

forgedhallpass: It is generally considered better to extract the result of a method into a variable and use that instead of repeating the getter invocation multiple times.

Extracting the result of a method into a variable allows the value to be computed once and reused multiple times, improving performance and readability. It also allows for easier debugging and modification, as any changes to the method only need to be made in one place.

I've followed the Boy Scout rule ("Always leave the code better than you found it"), and only did minor changes close to the logic that was required to implement the fix for the referenced issue.

That being said, if you don't agree with me, I can revert the change...

kingthorin: I appreciate it as a clean-code thing, I was more concerned with limiting what needs to be reviewed. I’m not stuck on it, leave it for @psiinon or @thc202 to tie break 😉

thc202: Yes, please, only do the necessary changes to address the issue. If you want to "clean code" do it in its own PR.

thc202:
I guess this is WIP since only addresses part of the referenced issue?

thc202:
Please, check this comment as well: https://github.com/zaproxy/zaproxy/pull/6218#issuecomment-727838905 (to make sure it addresses those cases).

forgedhallpass:
> I guess this is WIP since only addresses part of the referenced issue?

For this particular PR the scope was only the authentication, as the title suggests. That is why the ticket was only referenced via its ID, but without any of the `fixes`, `closes` etc. keywords.

That being said, if I'll have time I might consider looking into the other part as well, because it should be fairly similar. Will get back to you on this.

zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
```diff
@@ -106,8 +113,9 @@ public String getUIName() {
     @Override
     public void hook(ExtensionHook extensionHook) {
         super.hook(extensionHook);
+        this.extensionHook = extensionHook;
```
thc202, line 116: This will not work, once the hook phase is done any additions to the `extensionHook` have no effect.

thc202:
This is now closing the issue but it still lacks necessary functionality (most of the points mentioned in https://github.com/zaproxy/zaproxy/pull/6218#issuecomment-727838905).

zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
forgedhallpass: I am not entirely sure I understand what do you mean. Adding a new `AbstractAuthenticationMethodOptionsPanel` to the UI from a ZAP extension through a dynamically added `AuthenticationMethodType` was working.

forgedhallpass:
> This is now closing the issue but it still lacks necessary functionality (most of the points mentioned in [#6218 (comment)](https://github.com/zaproxy/zaproxy/pull/6218#issuecomment-727838905)).

I believe the PR covers what the ticket is asking for. I've tested it and it's working. Your points around the contexts, together with the support for reloading add-ons is a separate topic itself.

Please note that first time contributors might not have the full picture of the project. If you create a follow-up ticket and provide clear requirements I might give it a shot.

thc202:
They are not separate topics, the authentication and session management methods are an integral part of the contexts (see `Context` class, fields `authenticationMethod` and `sessionManagementMethod`). The add-ons can be dynamically installed/updated/removed at anytime so the core code must properly support dynamically adding/removing the methods.

I don't expect anyone to know everything, if the issue is not clear enough just ask. I linked the other comment because it mentions what needs to be considered to address the issue. If you prefer that I update the issue with those points I can do that as well.

zap/src/main/java/org/zaproxy/zap/extension/authentication/ExtensionAuthentication.java
thc202: The class `AbstractAuthenticationMethodOptionsPanel` is unrelated to the `ExtensionHook` (that's part of the method itself), change your method to use the `extensionHook`. Start ZAP, install your add-on, and see if it works properly.

forgedhallpass:
@thc202 @kingthorin I don't have time to work on the remainder of the requirements you have asked for. I would recommend merging the PR and creating a follow-up ticket, otherwise the changes and the effort which improves the code is going to be lost.

kingthorin:
Thanks @forgedhallpass: Please leave it open and someone will tackle it, when we get a chance.

PR #7861, opened 2023-05-13
Add key action badges

psiinon:
It turns out you can add badges for github action results 😁
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

This is what they will look like : https://github.com/psiinon/test#zap-key-actions
I've put them on a separate line, but can add them to the existing badges if peopkle think it would look better?
Any other key actions we should include?


thc202:
That looks good to me, we are already including CI and CodeQL.

thc202 (approved):

kingthorin (commented):
LGTM

kingthorin (approved):

PR #7865, opened 2023-05-15
Improve release tasks

thc202:
Move dynamic data (e.g. versions, japicmp exclusions) to external files to allow update them more reliably programmatically.
Update build and tasks accordingly.
Release tasks should no longer fail when preparing the release or next version.


kingthorin (approved):
Thanks that does seem cleaner/easier 😄

Issue #7784, opened 2023-03-15
maxAlertsPerRule for activeScan job


3XC1T3D:
### Is your feature request related to a problem? Please describe.

Because you can not set maxAlertsPerRule in activeScan, the reports are very bloated.

### Describe the solution you'd like

I would like to have a possibilty to configure maxAlertsPerRule also for the activeScan like it is possible for passiveScan

### Describe alternatives you've considered

unfortunately there is no alternative

### Would you like to help fix this issue?

- [ ] Yes

3XC1T3D:
Hi @thc202 ,

due to the merge from https://github.com/zaproxy/zaproxy/pull/7869 this request is complete or not?

Best Regards

thc202:
No, this requires further changes in the Automation ad-on.

3XC1T3D:
Oh thank you for the hint 😅

So i'll wait till the changes are done.

Thank you for your effort 👍

3XC1T3D:
Hi @thc202 ,

we are running the owasp/zap2docker-bare:latest Docker Image but we are still getting the Warning: Unrecognised parameter for job activeScan : maxAlertsPerRule

Could you give me a hint, how i recognize when this option is in the zap2docker-bare Docker Image?

best regards

thc202:
The add-on was not yet released, in the meantime you would have to use `-live` or `-weekly`.

PR #7869, opened 2023-05-17
Allow to limit alerts per rule during active scan

thc202:
Allow to configure the maximum number of alerts that a scan rule can raise, skipping the rule when the number is reached.

Related to #7784.


thc202:
We don't in the passive options.

kingthorin:
Okay.

kingthorin (approved):

psiinon (approved):

PR #7873, opened 2023-05-22
Sort Sites nodes with different case consistently

thc202:
Sort taking into account the case otherwise the nodes do not have a consistent order, they might be before or after each other.
Extract the node comparator to a top level class to be able to import in the tests.

---
This leads to sorting first uppercase and then lowercase, which makes `/PATH` being farther away than `/path` if other paths are in between… but prevents, e.g.: zaproxy/zaproxy-website#1749, zaproxy/zaproxy-website#1780, zaproxy/zaproxy-website#1792


kingthorin (approved):

ricekot (approved):
Thank you!

PR #7880, opened 2023-05-27
Register and reserve types of context data

thc202:
Register types already in use, by alert filters and custom pages.
Reserve a range for the AJAX spider.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/db/RecordContext.java
+++ b/zap/src/main/java/org/parosproxy/paros/db/RecordContext.java
@@ -75,6 +75,12 @@ public class RecordContext {
 
     public static final int TYPE_ACCESS_CONTROL_RULE = 410;
 
+    public static final int TYPE_ALERT_FILTER = 500;
+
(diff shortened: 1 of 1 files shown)


kingthorin (approved):
I don’t disagree but what are these used for?

thc202:
Load and persist context data, e.g.: https://github.com/zaproxy/zaproxy/blob/f8852857ea2a2c2b238ef3866fb1310541e4f59c/zap/src/main/java/org/zaproxy/zap/extension/custompages/ExtensionCustomPages.java#L202-L228

kingthorin:
Has it been defaulting to zero or something?

I don’t recall seeing warnings or errors ☹️ Sorry I missed it during the implementation.

thc202:
They are already using those values, they have their own constants/types though. The `RecordContext` is not needed/used directly so I didn't replace the other constant, but they still should be added there for documentation purposes.

kingthorin:
Okay, thanks!

psiinon (approved):

PR #7884, opened 2023-05-30
Remove build condition no longer needed

thc202:
Remove condition that configured the installer tasks depending on the Java version, no longer required now that ZAP is built with Java 11.


thc202:
Best viewed ignoring whitespace changes.

kingthorin (approved):

psiinon (approved):

PR #7886, opened 2023-05-31
Include whole `org.zaproxy` package in log config

thc202:
For convenience when debugging include the whole package by default as newer add-ons are under `org.zaproxy.addon` not `org.zaproxy.zap`.
Pointed out by @psiinon in IRC channel.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #7887, opened 2023-05-31
Show alert ref in the Alert panel

thc202:
Allow the user to know what ref the alert has, if any.

---
Is "ref" the right term to use in the GUI? I think it's the first time it's being shown. The site calls it "ID" but that's not exactly correct. e.g. https://www.zaproxy.org/docs/alerts/10020-2/


kingthorin:
"Alert Reference" would maximize clarity (IMHO)

kingthorin (approved):
I assume not editable in edit mode?

Empty in manual add?

LGTM (pending agreement on label/name).

thc202:
It's not shown when editing.

thc202:
No longer WIP, updated to use "Alert Reference".

psiinon (approved):

kingthorin (approved):

PR #7897, opened 2023-06-07
Push Docker images to GHCR

ricekot:


kingthorin:
Are there any storage limits we need to worry about?

ricekot:
According to https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages#about-billing-for-github-packages, there aren't any storage limits for public packages:

> GitHub Packages usage is free for public packages. For private packages, each account on GitHub.com receives a certain amount of free storage and data transfer

kingthorin:
I think the changes look fine. My only other question is whether having different names in different repos/sources is going to cause us support headaches. (Though I'm probably over thinking that.)

psiinon:
@kingthorin yeah, we chatted about that in IRC. Somehow the zap2docker names didnt seem to make sense in GHCR. TBH I'd like to change them in Docker too but thats probably not sensible 😉

psiinon (approved):

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #7906, opened 2023-06-13
Update gradle-launch4j plugin

### LINKED ISSUE #5368: ZAP exe might fail to find Java
Using OpenJDK 12.0.1 (64 bit) the (2.7.0) exe fails with “This application requires a java runtime 1.8.0” but the bat file apparently works fine.
Reported via email.

thc202:
Update gradle-launch4j to 3.0.3 which uses a newer version of Launch4j that improves the Java search algorithm.

Fix #5368.

---
Tested with [Windows 11 64bit (WinDev2305Eval)](https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/), `2.13-SNAPSHOT` is able to find Java (Temurin) installed with default settings while 2.12 failed to find it.


kingthorin (approved):
Thanks!

PR #7918, opened 2023-06-26
Use Adoptium for Java download in the executable

thc202:
The previous URL offered Java 8 while 11 is now the minimum required to run ZAP, also that JRE has license constraints.
Link instead to Adoptium OpenJDK builds, it has newer Java versions and it has no license constraints.


psiinon (approved):

kingthorin:
Thanks!

PR #7937, opened 2023-07-06
Deprecate Active Scan option Delay When Scanning

thc202:
The option has been superseded by the rate limit in the Network add-on.


kingthorin (approved):
Thanks

psiinon (approved):

PR #7944, opened 2023-07-11
Include other required add-ons in Core package

thc202:
The Spider, Database, and OAST are all needed for the Active scanner rules and Quick Start add-ons.


thc202:
Removed WIP, core is fine with those.

PR #7953, opened 2023-07-20
Update baseline integration test results

thc202:
Take into account the two passive scan rules that were promoted to beta in the 2.13 release, which are reported as passes.

---
e.g.: https://github.com/zaproxy/zaproxy/actions/runs/5605853668/jobs/10255426101


kingthorin (approved):
Thanks

ricekot (approved):
Thank you!

PR #7965, opened 2023-07-28
Update snap version when preparing release

thc202:
Update the version of the snap and the ZAP package (plain text replacement).
Remove corresponding release task from the release doc.
Update zapbot's commit message to mention the change.

---
I thought initially on parsing the YAML file for accurate update of the version and download link but there are no other version numbers in the file (and we can always correct any issues before merging zapbot's PR).


ricekot (approved):
Thanks!

kingthorin (approved):

PR #7978, opened 2023-08-03
Remove Java 11 requirement for `buildSrc`

thc202:
It's not necessary and avoids the need to have/download other JDK, when using newer Java versions.


kingthorin (approved):
Thank you!

psiinon (approved):

PR #7982, opened 2023-08-04
Update docs and image used in packaged scans

thc202:
Use the weekly from GHCR in packaged scans (when executed directly).
Update the release doc and READMEs to use the generic image names instead of one specifically, they are now published to several places with different image names/tags.
Fix labels of some links.


kingthorin (approved):

ricekot (approved):
Thank you!

PR #7984, opened 2023-08-06
Allow to display script without focusing

aryangupta701:
- overload display script method to support focus as optional which will be helpful for client-side recording
Related to: zaproxy/zap-extensions/pull/4768

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
@@ -129,7 +129,26 @@ public interface ScriptUI {
      */
     void removeScriptTreeTransferHandler(Class<?> klass);
 
+    /**
+     * Displays the provided script with focus.
+     *
+     * @param script The script to be displayed.
+     */
     void displayScript(ScriptWrapper script);
+
+    /**
+     * Displays the provided script with/without focusing on it. By default, this method delegates
+     * to {@link #displayScript(ScriptWrapper)}.
+     *
+     * @param script The script to be displayed.
+     * @param allowFocus {@code true} to allow focusing on the script display, {@code false}
+     *     otherwise.
+     * @since 2.14.0
+     * @see #displayScript(ScriptWrapper)
+     */
+    default void displayScript(ScriptWrapper script, boolean allowFocus) {
+        displayScript(script);
+    }
 
     boolean isScriptDisplayed(ScriptWrapper script);
 
--- a/zap/gradle/japicmp.yaml
+++ b/zap/gradle/japicmp.yaml
@@ -5,4 +5,6 @@
 packageExcludes: []
 fieldExcludes: []
 classExcludes: []
-methodExcludes: []
+methodExcludes:
+  - org.zaproxy.zap.extension.script.ScriptUI#displayScript(org.zaproxy.zap.extension.script.ScriptWrapper, boolean)
+


aryangupta701:
Isn't there any changelog.md file ?

kingthorin:
Nope we do change tracking differently for core. It'll be fine.

aryangupta701:
Okay thanks

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
```diff
@@ -131,6 +131,8 @@ default void removeSelectionListener(TreeSelectionListener tsl) {
 
     void displayScript(ScriptWrapper script);
 
+    void displayScript(ScriptWrapper script, boolean enableFocus);
```
thc202, line 134: Add default implementation which calls existing `displayScript`.

aryangupta701: I was thinking opposite. Shouldn't it be like the existing display script should have a default implementation which will call the newly added display script? Because we need to prune this setfocus boolean to setscript also.

aryangupta701: Shouldn't it be the opposite? I was thinking to add a default implementation for `void displayScript(ScriptWrapper script);`. Am I thinking wrong?

thc202: The existing implementations don't implement the new method yet, any code calling it will lead to `AbstractMethodError`. The default implementation addresses that (and calling the existing method keeps the current behaviour). Also, worth renaming the method to `displayScriptWithouFocus` and remove the `boolean`.

aryangupta701: okay got it. I have updated the code please check

thc202:
You'll have to suppress the binary incompatible change (should be added to `/zap/gradle/japicmp.yaml`).

aryangupta701:
I did something like this: `methodExcludes: [org.zaproxy.zap.extension.script.ScriptUI#displayScriptWithoutFocus()]` is this correct  ?

thc202:
You need to include the method parameter too `org.zaproxy.zap.extension.script.ScriptUI.displayScriptWithouFocus(org.zaproxy.zap.extension.script.ScriptWrapper`.
Consider adding it as block structure as it makes the diffs simpler when adding other entries.

aryangupta701:
Thank you!

thc202 (dismissed):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Should we have some javadoc on that explaining that by default it probably does take focus and that it should be overridden and focus handled by the caller/implementer?

thc202:
I'm not sure about the latter part but adding some JavaDoc to both methods does not hurt.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
thc202: Sorry, lets go back to `void displayScript(ScriptWrapper, boolean)` to keep it closer to what we do in the rest of the codebase.

aryangupta701:
Okay I will add java doc

kingthorin (approved):

thc202:
Please see also https://github.com/zaproxy/zaproxy/pull/7984#discussion_r1286713759

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
```diff
@@ -129,8 +129,24 @@ default void removeSelectionListener(TreeSelectionListener tsl) {
...
     void displayScript(ScriptWrapper script);
 
+    /**
+     * Displays the provided script without focusing on it. By default delegates to {@link
+     * #displayScript(ScriptWrapper)}.
+     *
+     * @param script The script to be displayed.
+     * @see #displayScript(ScriptWrapper)
+     */
+    default void displayScriptWithoutFocus(ScriptWrapper script) {
```
thc202, line 146: Worth adding the `@since` JavaDoc tag.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
aryangupta701: sorry didn't see this

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
```diff
@@ -129,8 +129,27 @@ default void removeSelectionListener(TreeSelectionListener tsl) {
...
     void displayScript(ScriptWrapper script);
 
+    /**
+     * Displays the provided script with/without focusing on it. By default, this method delegates
+     * to {@link #displayScript(ScriptWrapper)}.
+     *
+     * @param script The script to be displayed.
+     * @param allowFocus {@code true} to allow focusing on the script display, {@code false}
+     *     otherwise.
+     * @since version 2.14.0
```
thc202, line 146: The "version" can be removed.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptUI.java
thc202: No worries :)

kingthorin (approved):

thc202 (approved):

aryangupta701:
So will I need to wait till 2.14 release to use it with zap extensions or is there any easier way to do so?

thc202:
You should use reflection in the meantime, it will work with dev and weekly releases.

Thank you again for the changes!

aryangupta701:
Okay

PR #7985, opened 2023-08-07
Fix release of macOS dist in CI

thc202:
Move the creation of the symlink to the DMG task, being in the copy task makes Gradle check it, leading to out of memory errors.
Correct update of the release:
 - Search the release by name if not found by tag, since it might still be in draft state.
 - Use expected newline char in the table header, otherwise the checksums would not be included.

Increase the max memory in the release tasks to ensure there are no memory errors when creating and uploading the dists/installers.


psiinon (approved):

ricekot (approved):
Thanks!

PR #7988, opened 2023-08-07
Use short name for home dir in Windows

thc202:
Use just ZAP for its name and rename the existing/old one.


thc202:
Tested with Windows VM.

kingthorin (approved):

Issue #7821, opened 2023-04-10
Provide API endpoint to get/download the report


thc202:
The current API endpoint generates the report to the local file system and it's not possible (through ZAP) to get/download it.
https://www.zaproxy.org/docs/desktop/addons/report-generation/api/

Ideally there should be an endpoint that allows to get/download the whole report (e.g. zip file containing all the necessary files) rather than just generate it locally.

From OWASP ZAP User Group: https://groups.google.com/g/zaproxy-users/c/S6KMBBI2gJI/m/H7lrcSb7AAAJ

Add-on:
Report Generation

arunsk7891:
Hi Team ,
Thanks for the great work - would be interested to know if there was an update made to get the generated report in pdf format to the client system ? Also I'm a bit confused - so I get a scan ID for the current spider scan i performed. 
I want a report for this latest scan - do I just blindly give the Site address to the report api and it gets the latest report ?  
Requesting to please guide .

kingthorin:
I’m working on the download reports API bit.

As for the usage question that’s my check better tackled in the [ZAP User Group]( https://groups.google.com/group/zaproxy-users).

thc202:
Title updated to match the latest changes.

PR #8002, opened 2023-08-11
API Support file transfer

### LINKED ISSUE #6069: Allow to download & upload context through the API
The ability to download the context file and upload to create and/or replace over the API could allow for stateful automated scans with a stateless ZAP instance for batch scanning.

Implement a `/JSON/context/action/downloadContext/` and `/JSON/context/action/uploadContext/`.

**Describe alternatives you've considered**
The current import and exportContext calls work with the local filesystem which is impractical when working in containers.

### LINKED ISSUE #7607: Allow to download/upload files through the ZAP API
### Is your feature request related to a problem? Please describe.

I wish to use ZAP as a daemon in a container as part of a CI pipeline.

However, to do so, I either need to roll a new ZAP image for every type of target I wish to examine or I need to use ZAP as the base image itself.

With [GitLab services](https://docs.gitlab.com/ee/ci/services/), you cannot modify the running image on load, so you end up having to run ZAP as the "runner" which is not ideal.

### Describe the solution you'd like

I would like any API endpoint that allows for loading from a file on the local filesystem to either:

1. Have an associated `upload` endpoint that will put the file into the correct location
2. Allow me to upload the entire file contents as part of a `POST`
   * I'm not 100% sure this will work for everything but it will certainly work for anything plain-text

### Describe alternatives you've considered

I haven't found any good way around this issue

### Screenshots

_No response_

### Additional context

_No response_

- [ ] Yes

### LINKED ISSUE #7943: Allow API definitions to be uploaded via the ZAP API
### Is your feature request related to a problem? Please describe.

Right now we only support API definitions loaded from a file or from a URL.
Allowing API definitions to be uploaded via the ZAP API would help where the definition is not available via a URL and is not present on the machine running ZAP.

Should be supported for:

- [ ] OpenAPI
- [ ] GraphQL
- [ ] SOAP
- [ ] Postman

### Describe the solution you'd like

As above :)

### Describe alternatives you've considered

Uploading the API definitions via files or URLs.

### Screenshots

_No response_

### Additional context

_No response_

- [ ] Yes

psiinon:
Fixes #7943, fixes #7607, fixes #6069 

Partially addresses #7821 - will need an option to zip HTML files. 

Ready for code review, but still need to finish the manual tests:

- [x] Context upload
- [x] Context download
- [x] Policy upload
- [x] Exim upload
- [x] AF plan upload
- [x] GraphQL upload
- [x] Network upload
- [x] OpenAPI upload
- [x] Report download
- [x] Script upload
- [x] SOAP upload

Any missing?

Will PR associated changes to the core help file..


psiinon:
Will look at the test failures..

psiinon:
It is.
And this should work with all of the add-ons without any changes (apart from zipped reports).

kingthorin:
I haven't gone through the code yet. Don't the extensions still have to implement the handling?

psiinon:
No, thats the fun part :)
If a parameter has "file", "path" or "dir" in its name (ignoring case) then if the value starts with `${XFER}` then the core replaces that with the transfer directory.
So far they have all worked, core and add-on endpoints, as above 😁

psiinon:
Have manually tested all of the above (and tweaked the code to handle "dir";). No longer WIP..

kingthorin:
The user specifies `${XFER}`?

psiinon:
Yes - just writing the help now. It currently looks like this (have removed links):

The core API also supports uploading and downloading files, but this is disabled by default as a security measure.
If an attacker is able to access the ZAP API then with this feature enabled then they would be able to upload their own script to ZAP and then run it. ZAP scripts run with the same permissions as ZAP and so this is full remote code execution access.

To enable file transfer you will need to have an API key set and to enable it via the Options API screen.

With file transfer enabled you will be able to upload files to the file transfer directory using the 'fileUpload' action
and download them from there using the 'fileDownload' 'other' endpoint.
You can specify subdirectories (e.g. "subdir/myfile") but any attempt to upload or download a file to or from another directory (e.g. using "../../") will be rejected.

To download a file generated by another API endpoint you will need to specify a file name/path starting with "${XFER}".
For example you could export a context to "${XFER}/contexts/my.context" and then download it by specifying "contexts/my.context".

To upload a file to another API endpoint you need to do the reverse - first upload it, e.g. to "plans/myplan.yaml", and 
them import it via the path "${XFER}/plans/myplan.yaml".

^ that make sense? No need to do a full review, can do that in the forthcoming help PR...

kingthorin:
Seems very reasonable to me. I'll try to review this PR in the next few days. Thanks for tackling this!

zap/src/main/java/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -876,12 +883,49 @@ public void run() {
...
+                    getChildPath(apiParam.getTransferDir(), getParam(params, PARAM_FILENAME, ""));
+
+            try {
+                Files.createDirectories(filePath.getParent());
+                Files.writeString(filePath, getParam(params, PARAM_CONTENTS, ""));
+                LOGGER.debug("Uploaded file to {}", filePath);
+            } catch (IOException e) {
+                LOGGER.error("Failed to create file {}", filePath, e);
+                throw new ApiException(ApiException.Type.ILLEGAL_PARAMETER, PARAM_FILENAME, e);
+            }
```
kingthorin, line 902: Would it be helpful to return the path that was written to?

kingthorin: Instead of just the ok at 906

zap/src/main/java/org/zaproxy/zap/extension/api/CoreAPI.java
```diff
@@ -1487,6 +1531,35 @@ public HttpMessage handleApiOther(HttpMessage msg, String name, JSONObject param
...
+                    msg.setResponseHeader(
+                            API.getDefaultResponseHeader(null, msg.getResponseBody().length()));
+                } catch (HttpMalformedHeaderException e) {
+                    LOGGER.error("Failed to create response header: {}", e.getMessage(), e);
+                }
+                LOGGER.debug("Downloaded file {}", filePath);
+                return msg;
+
+            } catch (IOException e) {
+                LOGGER.error("Failed to create file {}", filePath, e);
```
kingthorin, line 1560: Wouldn't it be failed to read in this case?

kingthorin (commented):
Should have tests that prove the traversal prevention handling?

zap/src/main/java/org/zaproxy/zap/extension/api/CoreAPI.java
psiinon: Done

psiinon:
Both ops are now "others" and the API now supports "multipart/form-data" and "application/x-www-form-urlencoded" content. Have also added tests for the path traversal detection.

zap/src/main/java/org/zaproxy/zap/extension/api/API.java
```diff
@@ -860,6 +867,23 @@ public static JSONObject getParams(String params) throws ApiException {
...
+                                && value.startsWith(TRANSFER_DIR_TOKEN)) {
+                            String relPath = value.substring(TRANSFER_DIR_TOKEN.length());
+                            if (!relPath.startsWith("/") && !relPath.startsWith("\\")) {
+                                // Cope with the user not specifying a slash
+                                relPath = "/" + relPath;
+                            }
+                            value = options.getTransferDir().toString() + relPath;
+                        }
+                    }
+
```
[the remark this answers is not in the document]
psiinon: done.

thc202 (approved):

thc202:
Thank you!

zap/src/main/java/org/parosproxy/paros/network/HttpHeader.java
```diff
@@ -47,6 +47,7 @@
 // ZAP: 2022/09/12 Allow only major HTTP version.
 // ZAP: 2022/11/17 Add HTTP/2 constant.
 // ZAP: 2022/11/22 Lower case the HTTP field names for compatibility with HTTP/2.
+// ZAP: 2023/08/15 Add FORM_MULTPART_CONTENT_TYPE.
```
kingthorin, line 50: Should the comment and the variable (and it's use I guess) include the "I" in multipart? (I think so)

kingthorin:
Was the plan to handle zip in core with a boolean? (For download?) Is that to be tackled in another PR or did I misunderstand?

zap/src/main/java/org/parosproxy/paros/network/HttpHeader.java
thc202: Good catch.

psiinon: Ha!

psiinon:
No, the plan was to add a zip boolean parameter to the report API, default false. That make sense?

zap/src/main/java/org/parosproxy/paros/network/HttpHeader.java
psiinon: Fixed 😁

kingthorin:
Okay, that's fine, yup it does make sense.

kingthorin (approved):

PR #8003, opened 2023-08-11
Use newer Python ZAP API client

thc202:
Update some entries that were missed previously (using underscores instead of hyphens).


kingthorin (approved):

ricekot (approved):
Thanks!

PR #7659, opened 2022-12-18
Add SARIF report command line argument to baseline scan

oseki-guy:
Signed-off-by: Mister Oseki <113904192+oseki-guy@users.noreply.github.com>

PR #8005, opened 2023-08-13
[WIP] Add sarif report option to zap-baseline

DanBradbury:
Part of the required work to officially support Code Scanning integration with `action-baseline`

**Original Issue:** https://github.com/zaproxy/action-baseline/issues/63

Will require a change to the `action-baseline` as well to `touch` the right files ([see PR here]() for that change)

*NOTE: hacked together to just get it to work so there are issues with `.json` being added to the report name, etc. This PR is just to demonstrate where the changes should be made to add sarif support*

### DIFF
--- a/docker/zap-baseline.py
+++ b/docker/zap-baseline.py
@@ -86,6 +86,7 @@ def usage():
     print('    -w report_md      file to write the full ZAP Wiki (Markdown) report')
     print('    -x report_xml     file to write the full ZAP XML report')
     print('    -J report_json    file to write the full ZAP JSON document')
+    print('    -S report_sarif   file to write the full ZAP sarif document')
     print('    -a                include the alpha passive scan rules as well')
     print('    -d                show debug messages')
     print('    -P                specify listen port')
@@ -164,6 +165,7 @@ def main(argv):
     report_md = ''
     report_xml = ''
     report_json = ''
+    report_sarif = ''
     target = ''
     zap_alpha = False
     info_unspecified = False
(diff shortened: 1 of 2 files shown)


kingthorin:
I'm not sure this is the right away to go about this. We intend to replace the packaged scans with Automation Framework functionality.

---

thc202:
Ref  #7659

semsuddin:
We need this on my project atm. What's the plan? Why can't we merge this it's not a big change and add the capability to generate sarif?

psiinon:
Because its probably not the direction we want to go in.
The packages scans already have too many arguments and have become difficult to maintain.
The automation framework is a better option, and that now does have a GitHub action: https://github.com/marketplace/actions/zap-automation-framework-scan

semsuddin:
> Because its probably not the direction we want to go in. The packages scans already have too many arguments and have become difficult to maintain. The automation framework is a better option, and that now does have a GitHub action: https://github.com/marketplace/actions/zap-automation-framework-scan

I'm really having issues understanding the documentation. On one side SARIF is already there and we should be able to export in that format while on the other end it doesn't work. Right now I'm using the zap docker container and want to generate the SARIF json report via zap API, is there a way?

thc202:
Yes, there is but https://groups.google.com/g/zaproxy-users is a more appropriate place for usage questions.

kingthorin:
> on the other end it doesn't work

In what way?

As for report generation via the API check: 
- https://www.zaproxy.org/docs/api/#reports
- https://www.zaproxy.org/docs/desktop/addons/report-generation/api/

psiinon:
You will find the AF much easier to use than the API. For more help ask on the ZAP User Group.
https://groups.google.com/g/zaproxy-users

psiinon:
Closing because we do not want to kee adding more and more options to the packaged scans.
The automation framework is the way to go, it already supports all of the reports, not just the sarif one.
It can also be run from the command line as well as in docker.

PR #8017, opened 2023-08-22
Do not remove headers when proxying

### LINKED ISSUE #7353: Header `If-None-Match: *` removed for PUT requests
### Describe the bug

This has been raised before: https://github.com/zaproxy/zaproxy/issues/1284

I am experiencing the same issue, however for a PUT request. Mozilla describes two use cases for the header, where the second reads:

> For other methods, and in particular for [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT), If-None-Match used with the * value can be used to save a file not known to exist, guaranteeing that another upload didn't happen before, losing the data of the previous put;

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match

From what I understood, ZAP strips the header since it's related to caching and one want's fresh data. However, PUT requests shouldn't be cacheable anyways.

Proxy a request containing the header through ZAP

### Expected behavior

Header is not stripped

### Software versions

owasp/zap2docker-bare:2.11.1

### Screenshots

_No response_

_No response_

### Additional context

_No response_

- [X] Yes

thc202:
Change the listener to no longer remove headers, if the client tool or user needs a fresh copy of the message they can do so themselves.

Fix #7353.

---
This behaviour was not documented, though it might affect the active scanner if the user scans 304 messages.

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/extension/history/ProxyListenerLog.java
+++ b/zap/src/main/java/org/parosproxy/paros/extension/history/ProxyListenerLog.java
@@ -38,10 +38,10 @@
 // ZAP: 2020/08/04 Changed to use new SessionStructure method
 // ZAP: 2020/11/26 Use Log4j 2 classes for logging.
 // ZAP: 2023/01/10 Tidy up logger.
+// ZAP: 2023/08/22 Do not modify the requests being proxied (Issue 7353).
 package org.parosproxy.paros.extension.history;
 
 import java.awt.EventQueue;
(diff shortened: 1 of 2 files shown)


thc202:
The option will be added to the Network add-on to allow to change (or remove) whenever needed, rather than on core releases.

Leaving WIP until the changes are merged there.

thc202:
No longer WIP.

psiinon (approved):

kingthorin (approved):

PR #8028, opened 2023-08-25
Set the view to `ExtensionAdaptor` sooner

thc202:
Set the view before starting to initialize the extension to ensure the `hasView()` checks work throughout the extension initialization.
Remove now redundant calls to base implementation of `initView` and replace a null check with `hasView()` call.

Related to zaproxy/zap-extensions#4840.


kingthorin (approved):
Thanks

Issue #8046, opened 2023-09-01
OSX version Ventura (13.5.1) not correctly supported by zap.sh script


Lyr:
### Describe the bug

The double click on the "OWASP ZAP.app" icon on Application directory on OSX Ventura (default installation) doesn't run anything and any error appears.

### Steps to reproduce the behavior

1. Download version 2.13.0 of Zap
2. Unzip the downloaded file
3. Paste the "OWASP ZAP.app"  on Application directory
4. Launch ""OWASP ZAP.app" as other application

### Expected behavior

The ZAP application must be launched or at least an error must appear for the final user.

### Software versions

2.13.0

### Errors from the zap.log file

no zap.log have been written at this step.

### Additional context

When the run in a terminal, this error appears :
_ZAP requires a minimum of Java 11 to run, found_

### Would you like to help fix this issue?

- [X] Yes

thc202:
This is unrelated to the macOS version, running Ventura and not seeing that issue.

Lyr:
Yes, you right, it's more the "$OS" = "Darwin", also, it's the java management on OSX that is the root cause, anyway, as first user on Ventura this part doesn't work, so I precise the OSX version cause I haven't other way to test on other version/OS

thc202:
It's not just `"$OS" = "Darwin"` since I'm using different versions of macOS and they work fine (though I'm still not using 13.5.1 which I guess it might have changed something).
In any case the proposed fix is not correct, should not blanket all 13.x versions.

Lyr:
I guess this part of code first have been good reason to be here (for some specific context on osx) so I'm not able to validate this part on other osx version, just the patch works on osx ventura.

thc202:
Would be better to know why it's not finding the bundled Java version which is the root cause of the issue.

Lyr:
> Would be better to know why it's not finding the bundled Java version which is the root cause of the issue.

I remove all the part 'if [ "$OS" = "Darwin" ]' and it looks ok (also for Ventura only). That could mean this old script code is useless now? But since whic osx version?

thc202:
See https://github.com/zaproxy/zaproxy/blob/438f49c3db13eaff90c6c68a6e5c663a4dc4dc7c/zap/src/main/dist/zap.sh#L22

We want to use the bundled version first.

[...]

PR #8047, opened 2023-09-01
[WIP] Support Ventura Java version (Update zap.sh) [#8046]

Lyr:
To fix issue related in #8046, an basic update on zap.sh, to allow to use directly the java version installed in Ventura (13.5.1)

### DIFF
--- a/zap/src/main/dist/zap.sh
+++ b/zap/src/main/dist/zap.sh
@@ -22,7 +22,9 @@ OS=$(uname -s)
 # If we're on OS X, try to use the bundled Java; if it's not there, then the system Java
 # Life would be much easier if OS X had readlink -f
 if [ "$OS" = "Darwin" ]; then
-  if [ -e ../PlugIns/jre*/Contents/Home/bin/java ]; then
+  OSX_VERSION=$(sw_vers -productVersion)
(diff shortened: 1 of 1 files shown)


psiinon:
Many thanks!
I need to update my Mac in order to test this...

zap/src/main/dist/zap.sh
```diff
@@ -22,7 +22,9 @@ OS=$(uname -s)
 # If we're on OS X, try to use the bundled Java; if it's not there, then the system Java
 # Life would be much easier if OS X had readlink -f
 if [ "$OS" = "Darwin" ]; then
-  if [ -e ../PlugIns/jre*/Contents/Home/bin/java ]; then
+  OSX_VERSION=$(sw_vers -productVersion)
+  OSX_MAJOR_VERSION=${OSX_VERSION%%[.|-]*}
+  if [ -e ../PlugIns/jre*/Contents/Home/bin/java && $OSX_MAJOR_VERSION -lt 13]; then
```
thc202, line 27: Based on latest comments from the issue, the correct fix in the script is to check here if we are able to use this Java instead of skipping based on macOS version.

PR #8063, opened 2023-09-08
Add labeler workflow

kingthorin:
I am HORRIBLE at remembering to add the default label. I know this isn't 100% applicable but it is more often than not. So for the few times it needs to be removed to be replaced by enhancement or something I think that's fine.

Ref: https://github.com/actions/labeler


psiinon:
Me too ;) So what does this actually do?

kingthorin:
So any time a PR is opened it'll apply the `Type-Task` label automatically.

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,16 @@
+name: "Pull Request Labeler"
```
thc202, line 1: Maybe change the name of the file too? I'd like to have a e.g. `to-triage` label for issues.

kingthorin: Happy to rename things.

I'm not going to add the other label/confit/workflow until we discuss how that process will work and how/when we're considering triage complete. (That's probably for another PR.)

kingthorin: Tweaked.

thc202: Yeah, not asking for a new workflow here, just giving the reason to use a more specific name.

thc202 (approved):

thc202:
Thank you!

ricekot (approved):
Thanks!

Issue #8062, opened 2023-09-08
Set a sensible default for HSQL defrag


sjames-au:
### Is your feature request related to a problem? Please describe.

While performing scans on a large (to me) GraqhQL interface, I could not complete the initial scan using the GraphQL addon and automation as the database grew to 128Gb and its maximum size. On inspection, it became clear that the params statistics gathering generated DELETE and INSERT statements, leaving a lot of unused space. In this instance, once I issued a SHUTDOWN COMPACT the .data file was reduced from 128Gb to 144Mb.

Creating a session, closing zap, and altering `SET FILES DEFRAG 0` to `30` solved this problem. However, in hindsight, I think this would be better at a much larger figure. HSQL document suggests useful values are between 30 and 60. With the same scan running, the file size is being kept in check, and the graphql scanner can complete.

Reference for `SET FILES DEFRAG`
https://www.hsqldb.org/doc/2.0/guide/management-chapt.html 

### Describe the solution you'd like

Some way to easily set/unset the DEFRAG threshold OR a default where if we are seeing more than 60/70/80% reclaimable so that out of the box, larger scans are likely to work. Of course, if the defragging is a performance problem, users should be able to disable it.

### Describe alternatives you've considered

As described above, You can manually edit the .script file for a session and make a specific session defrag at any percentage point you like. 

### Additional context

The scan filled the disk at around 26,000 requests mark in multiple attempts. The scan is being executed via Automation, there was no feedback in the GUI that the DB is having an issue. 

I have said yes to helping fix this issue. I am happy to try and work on other potential fixes, but I don't know Java.

### Would you like to help fix this issue?

- [X] Yes

psiinon:
Sounds good to me!
We already have an options panel for the db, so the DEFRAG threshold could be set there.
We already support the recoverly log option, so this should be pretty similar.
Relevant code:

- [ParosDatabaseServer.java](https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/db/paros/ParosDatabaseServer.java) - see where `databaseOptions.isRecoveryLogEnabled()` is used
- [DatabaseParam.java](https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/extension/option/DatabaseParam.java) - where the ZAP db options are stored
[...]

PR #8064, opened 2023-09-08
Allow to configure HSQLDB DEFRAG value

sjames-au:
* Added 'defrag' database param
* Proposes a default of 90%
* Added GUI for setting percentage
* read, update and initialize setting
* requires session restart for change to take effect


zap/src/main/java/org/parosproxy/paros/extension/option/DatabaseParam.java
```diff
@@ -154,6 +187,25 @@ public void setCompactDatabase(boolean compactDatabase) {
         getConfig().setProperty(PARAM_COMPACT_DATABASE, compactDatabase);
     }
 
+    /**
+     * Gets the defrag percentage point
+     *
+     * @return the percentage (int) at whiuch point the database should defraf
```
kingthorin, line 193: Drop the u in which 😉

zap/src/main/java/org/parosproxy/paros/extension/option/DatabaseParam.java
```diff
@@ -154,6 +187,25 @@ public void setCompactDatabase(boolean compactDatabase) {
...
+     */
+    public int getDefragPercentage() {
+        return defragPercentage;
+    }
+
+    /**
+     * Sets the defrag percentage point
+     *
+     * @param defragPercentage the request body size
```
kingthorin, line 202: Ummm

zap/src/main/java/org/parosproxy/paros/extension/option/DatabaseParam.java
kingthorin: Oh and fix defrag

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsDatabasePanel.java
```diff
@@ -172,7 +196,15 @@ public OptionsDatabasePanel() {
                 LayoutHelper.getGBC(0, 5, 1, 0.5D, new Insets(2, 2, 2, 2)));
         panel.add(
                 comboNewSessionOption(),
-                LayoutHelper.getGBC(1, 5, 1, 1.0D, new Insets(2, 2, 2, 2)));
```
kingthorin, line 175: This is confusing if there’s no 1,5,1 why jump to 1,6,1?

kingthorin (commented):
This would definitely be easier to review if all the unnecessary formatting changes were suppressed 😉

kingthorin:

zap/src/main/java/org/parosproxy/paros/extension/option/OptionsDatabasePanel.java
sjames-au: I think I fat fingered something along with learning as I go and didnt realise I had done this. With formatting fixed it has made reviewing myself alot easier, thank you so much for having a stab at my low quality PR and for the guidance :)

kingthorin (commented):
Thanks that's much easier to review now. I think it's largely good, I want to pull the branch and see what the UI and tooltip look like. I'll try to do that tomorrow.

sjames-au:
Thank you, your calm approach to my mess helped me give in further. I still have a decision point stuck in my head on the "default" position for this; while the DB software defaults to zero they hint that this should be between 30 and 60, 30 would help keep the file size smallish and zap has a max file size of 128Gb.

I pitched 90% simply because it seemed as close to the current default of never defrag (though zero is still a viable option).
I am leaning towards taking a 30% approach for myself as if I am running a scan, I would rather the potential performance hit of a DB defrag than maxing out the DB size.

Also as a random contributor, while I happy to raise this as a decision point, I am not the right person to make such a choice. If it is purely a me call, I would probably default to 0 just so it is aligned with what has been the zaps history.

kingthorin:
I need to read about this option/feature. If zero is off and also close to 90 then what's 30? If the suggestion is 30 to 60 then 45 seems logical 🤷‍♂️

sjames-au:
from: http://www.hsqldb.org/doc/2.0/guide/management-chapt.html
 SET FILES DEFRAG

set files defrag statement

<set files defrag statement> ::= SET FILES DEFRAG <unsigned integer literal>

Sets the threshold for performing a DEFRAG during a checkpoint. The <unsigned integer literal> is the percentage of abandoned space in the *.data file. When a CHECKPOINT is performed either as a result of the .log file reaching the limit set by SET FILES LOG SIZE m, or by the user issuing a CHECKPOINT command, the amount of space abandoned since the database was opened is checked and if it is larger than the specified percentage, a CHECKPOINT DEFRAG is performed instead of a CHECKPOINT. As the DEFRAG operation uses a lot of memory and takes a long time with large databases, setting the threshold well above zero is suitable for databases that are around than 500 MB or more.

The default is 0, which indicates no DEFRAG. Useful values are between 30 to 60.

Only a user with the DBA role can execute this statement.

This is equivalent to the connection property hsqldb.defrag_limit.

kingthorin:
In that case I think my original suggestion of 45 is probably a good balance between performance and disk usage. 
When the DB is just about half empty tiddy it up 😉

sjames-au:
That sounds completely reasonable to me. :) 

I will get that updated :)

kingthorin:
You could wait till tomorrow, others might chime in early Monday.

sjames-au:
Sure. I have a local workaround, so this isnt blocking me from anything and it will be nice to have helped :) I am already dealing with Mondayitis at my location, the weather is nice though. ;)

I am a little curious about a code quality issue, specifically, I mirrored/copied others work and have ended up with the default being defined in two locations. OptionsDatabasePanel.jave, which sets  default within the spinner and DatabaseParam.java which does the real work. Ideally the default should be defined centrally and just referenced where its needed.

If someone could point me to a preferred pattern for that, I would be happy to incorporate it into this change :)

kingthorin:
Bump 😁

sjames-au:
Seeing as the only view on the default percentage has been at 45% by @kingthorin, I will get refresh this PR with that value as a default this coming weekend :)

thc202:
Would be good to get an answer to this first https://github.com/zaproxy/zaproxy/issues/8062#issuecomment-1712454341

sjames-au:
Closing; I am not in a position to perform the level of testing requested.

kingthorin:
@sjames-au thanks for trying to tackle this. Sorry it didn't pan out.

PR #8068, opened 2023-09-12
Use the current database body size values

thc202:
Use the current options rather than reading the configuration file directly, which would miss changes done during runtime.

---
From ZAP User Group: https://groups.google.com/g/zaproxy-users/c/syQEhnY03k0/m/kEyvaSsEAgAJ


kingthorin (approved):

kingthorin:
Thanks @thc202!

PR #8070, opened 2023-09-14
Prevent concurrent usage of ZAP home

thc202:
Lock a file to prevent different ZAP instances from using the same home at the same time.
The home can't be shared, the resources it has (e.g. config files, databases) are not expected to be accessed/changed concurrently.


zap/src/main/java/org/parosproxy/paros/Constant.java
```diff
@@ -769,6 +780,28 @@ private void initializeFilesAndDirectories(ControlOverrides overrides) {
...
+        try {
+            Path lockFile = Paths.get(zapHome, ".homelock");
+            if (Files.notExists(lockFile)) {
+                Files.createFile(lockFile);
+            }
+            homeLockFileChannel = FileChannel.open(lockFile, StandardOpenOption.WRITE);
+            homeLock = homeLockFileChannel.tryLock();
+            if (homeLock == null) {
+                System.err.println(
+                        "The home is already in use. Ensure no other ZAP instances are running with the same home: "
```
kingthorin, line 793: Should this and the one just below be i18n?

thc202: We do not i18n the start up errors.

kingthorin (approved):

psiinon (approved):

Issue #6292, opened 2020-11-12
Including main technology through the API does not include their specific technologies


VjekoslavKrainovic:
Zap Version 2.9.0
OS: Windows 10

When i try to include all tehnologies of specific field (e.g. languages, dbs...) nothing happens, but when i try specific language it works(e.g. Db.Oracle).
Request url for to get tehnology list: http://localhost:8081/JSON/context/view/technologyList/?apikey=123456
Request url to include tehnology: http://localhost:8081/JSON/context/action/includeContextTechnologies/?apikey=123456&contextName=ss&technologyNames=Language

PR #8072, opened 2023-09-15
fix: Context API including a parent Tech should include children

kingthorin:
- ContextAPI > Adjust code to process child technologies when including and excluding.
- TechSet > Add a method that gets the wanted Tech and its children as a Set of Techs. Add new includeAll/excludeAll methods that accept Tech (as parent).
- TechUnitTest > Added a test for the existing Tech.is(Tech) method.
- TechSetUnitTest > Added tests for the new methods.

Fixes zaproxy/zaproxy#6292

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/ContextAPI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/ContextAPI.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import net.sf.json.JSONArray;
@@ -314,9 +315,7 @@ public class ContextAPI extends ApiImplementor {
                 context = getContext(params);
                 techSet = context.getTechSet();
                 techNames = getParam(params, PARAM_TECH_NAMES, "").split(",");
-                for (String techName : techNames) {
-                    techSet.include(getTech(techName));
-                }
+                handleTechs(techNames, techSet::includeAll);
                 context.save();
                 break;
             case ACTION_INCLUDE_ALL_TECHS:
@@ -329,9 +328,7 @@ public class ContextAPI extends ApiImplementor {
                 context = getContext(params);
                 techSet = context.getTechSet();
                 techNames = getParam(params, PARAM_TECH_NAMES, "").split(",");
-                for (String techName : techNames) {
-                    techSet.exclude(getTech(techName));
-                }
+                handleTechs(techNames, techSet::excludeAll);
                 context.save();
                 break;
             case ACTION_EXCLUDE_ALL_TECHS:
@@ -347,6 +344,13 @@ public class ContextAPI extends ApiImplementor {
         }
 
         return ApiResponseElement.OK;
+    }
+
+    private void handleTechs(String[] techNames, Consumer<Tech> handler) throws ApiException {
+        for (String techName : techNames) {
+            Tech tech = getTech(techName);
+            handler.accept(tech);
+        }
     }
 
     private void addExcludeToContext(Context context, String regex) {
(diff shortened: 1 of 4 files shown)


psiinon:
Does this just affect the API or could there be other side effects?

kingthorin:
Well if you're running GUI mode you see the changes.... but should impact anything else.

thc202:
Should have tests.

zap/src/main/java/org/zaproxy/zap/extension/api/ContextAPI.java
```diff
@@ -349,6 +346,17 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
         return ApiResponseElement.OK;
     }
 
+    private void handleTech(String[] techNames, Consumer<Tech> handler) throws ApiException {
+        for (String techName : techNames) {
+            Tech t = getTech(techName);
+            if (t != null) {
```
thc202, line 352: Should not need this check.

zap/src/main/java/org/zaproxy/zap/model/Tech.java
```diff
@@ -287,4 +288,28 @@ public static Tech get(String name) {
...
+        Set<Tech> set = new HashSet<>();
+        for (Tech tech : getAll()) {
+            if (tech.getParent() == null) {
+                continue;
+            }
+            // 3 layers, prioritized
+            if (tech.getParent().equals(t)
+                    || tech.equals(t)
+                    || (tech.getParent().getParent() != null
+                            && tech.getParent().getParent().equals(t))) {
```
thc202, line 309: This does not cover deeper levels that we might add in the future.

zap/src/main/java/org/zaproxy/zap/model/Tech.java
```diff
@@ -287,4 +288,28 @@ public static Tech get(String name) {
...
     }
+
+    /**
+     * Gets the Techs that match the name or for which the parent matches the name, or an empty Set.
+     *
+     * @param name the name of the Tech
+     * @return the matching Techs
+     * @since 2.14.0
+     */
+    public static Set<Tech> getTechs(Tech t) {
```
thc202, line 299: I'm wondering if this should be behaviour of `TechSet`, e.g. `includeAll` or `includeTree`.

zap/src/main/java/org/zaproxy/zap/extension/api/ContextAPI.java
kingthorin: I’d have to test over again. IIRC it prevents an NPE when the user input fails to match an actual Tech.

zap/src/main/java/org/zaproxy/zap/model/Tech.java
kingthorin: Agreed, hence the comment. I guess I could move that to the method Javadoc?

zap/src/main/java/org/zaproxy/zap/extension/api/ContextAPI.java
thc202: The method `getTech(String)` throws if the tech does not exist.

zap/src/main/java/org/zaproxy/zap/model/Tech.java
thc202: IMO this is a bug, we should not need to change the method for newer techs (which can be added dynamically by any add-on).

kingthorin: Okay, I'll look at handling it differently then.

kingthorin: I came up with a better mechanism. I just need to add tests.

kingthorin:
Re-worked, ready for review.

kingthorin:
I've re-worked this again :wink:. The UnitTests still need work. Please let me know what you think.

kingthorin:
Now with tests.

kingthorin:
Just occurred to me that for the fix this should test not only branch but also twig and leaf in both scenarios. I'll add more. Please let me know if the implementation is okay?

kingthorin:
Now with further tests and based on recent IRC discussion.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

Issue #2189, opened 2016-01-20
Enable/Disable Script Causes Save Prompt on Exit


kingthorin:
To recreate (this should occur in either 2.4.3 or Dev w/ Script Console v15):
0. This assumes you have a script saved.
1. Launch ZAP.
2. Goto the Scripts tab (tree).
3. Enable or Disable a script (don't touch the Script Console tab at all).
4. Exit ZAP.
5. Note that ZAP asks to save changes to the script, though nothing was actually changed. (If you choose to discard the changes the enabled/disabled state is in fact discarded, however checking the actual script file before and after reveals no change in checksum.)

thc202:
Right, that's expected. ZAP will (also) warn that the script has unsaved changes if one of its properties (name, description, enabled state, load on start...) was not (yet) saved, otherwise the changes would be lost without warning.
Maybe the message should be explicit on what was not saved?

kingthorin:
For my 2 cents the properties and state should just be set (saved) when set, just like everything in the Tools:Options elements/attributes.

thc202:
Yeah, makes sense. So, only the contents of the script would not be persisted when changed?

kingthorin:
Yes if you'd actually changed the script not saved and discarded on exit, then your changes would be lost.

PR #8073, opened 2023-09-16
scripts: Persist enable state automatically

ricekot:
The enabled property of the script is persisted to the configuration
automatically when it is toggled, and does not require the user to save
the script manually to persist the change.

This only happens for scripts which have "Load on Start" enabled.

Fixes https://github.com/zaproxy/zaproxy/issues/2189.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptWrapper.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/script/ScriptWrapper.java
@@ -242,11 +242,7 @@ public class ScriptWrapper {
         if (enabled && engine == null) {
             return;
         }
-
-        if (this.enabled != enabled) {
-            this.enabled = enabled;
-            this.changed = true;
-        }
+        this.enabled = enabled;
     }
 
     public String getLastErrorDetails() {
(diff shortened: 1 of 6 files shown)


kingthorin:
Fixes this? https://github.com/zaproxy/zaproxy/issues/2189

kingthorin (commented):
Should update CHANGELOG too

thc202:
The enabled state is persisted that's why it's being marked as changed.

thc202:
Core change, no changelog.

kingthorin:
Oh oops for some reason I had been thinking this was extension.ns, sorry

ricekot:
Updated to save the enabled state of the script whenever it is enabled/disabled, provided it is already present in the config file. If it is not present in the config file, that means that `loadOnStart` was disabled for it when ZAP started. If `loadOnStart` is still disabled for it, we don't need to persist the enabled property (that's the existing behaviour). If `loadOnStart` was enabled, this means that the script was marked as changed and will need to be saved by the user, which would also save the enabled state along with it.

kingthorin (approved):

thc202:
The PR description/commit should be updated to match the latest changes. Should have tests.

zap/src/main/java/org/zaproxy/zap/extension/script/ScriptWrapper.java
```diff
@@ -245,7 +245,6 @@ public void setEnabled(boolean enabled) {
 
         if (this.enabled != enabled) {
```
thc202, line 246: This if is no longer needed.

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

kingthorin:
Thanks!

Issue #3961, opened 2017-10-17
Unable to drag response tab box larger


pr0letariat:
This is a recent issue I've come across.

I'm unable to drag the `Response` tab box across. I can drag it by a tiny amount then it does not drag any more.

![drag-issue](https://user-images.githubusercontent.com/14975253/31691502-37c0319c-b38e-11e7-9042-7308a39072a1.gif)

Nothing in `zap.log`.

Running `Version 2.6.0`

thc202:
That seems to be caused by the Quick Start tab. Which version of Quick Start add-on are you using? Latest?

pr0letariat:
Yes, when I check for updates, it doesn't say it needs updating.

pr0letariat:
Anything I can do to fix this? Would say this is a pretty big issue as prevents me from seeing the full response in a nice, readable format.

thc202:
Right, could you share a screen shot with both (whole) tabs Quick Start and Response? Which other tabs do you have with the Quick Start? (Just Request?)

psiinon:
As a temp fix you could try just closing the Quick Start tab :)

thc202:
Yeah, was going to suggest trying that (which really helps to know which tab is causing the problem).

pr0letariat:
I tried closing the Quick Start tab but that didn't resolve it. I don't have any other tabs except Quick Start, Request and Requester (with Response on the right). Not sure what you mean by both the whole tabs in the screenshot?

thc202:
Did you try closing the Requester tab? (That might have enough UI components to prevent the minimisation.) Regarding the screen shot with the whole tabs, was to see them all in the same image (to check which components were or not visible).

pr0letariat:
Yep, closing Requester solves the issue.

thc202:
Cool, issue updated. (Maybe the Requester tab should have a scroll pane?)

thc202:
Updated the title to reflect the actual problem (which includes other issues that are caused by that).

ArkaprabhaChakraborty:
Also, I would have liked a double bar instead of a single bar with all options... probably it would also help when I'm "compressing" the tab :)

thc202:
(This issue also applies to the Manual Request Editor which uses the same panels.)

ArkaprabhaChakraborty:
Oh, not in the manual editor tho... There the one single bar looks much better... The requestor on the other hand requires a resizing more often so it might help..

[...]

PR #8083, opened 2023-09-27
Wrap message panel's header components

thc202:
Wrap the components to allow to properly resize the panels, as side effect the Send button is no longer positioned at the end.

Improves #3961.

---
Worth trying to see how it behaves… (IMO it's better than not allowing to resize).


kingthorin (approved):
Seems reasonable.

ricekot:
I like it better than before, but I am unable to resize the panel in one go. I have to resize it multiple times until the components at the edge move to the next line one by one.

resize-panel

thc202:
Yeah, it has some quirks… (probably because of the scroll panes, it behaves fine in the Manual Request Editor dialogue).

psiinon (approved):

PR #8101, opened 2023-10-04
Extend ScanEventPublisher to support params

psiinon:


kingthorin (approved):

zap/src/main/java/org/zaproxy/zap/model/ScanEventPublisher.java
```diff
@@ -65,18 +66,37 @@ public String getPublisherName() {
...
+    }
+
+    public void publishScanEvent(
+            EventPublisher publisher,
+            String event,
+            int scanId,
+            Target target,
+            User user,
+            Map<String, String> params) {
+        assert params != null;
```
thc202, line 93: I'd let the NPE happen, which will since assertions are not enabled by default.

zap/src/test/java/org/zaproxy/zap/model/ScanEventPublisherUnitTest.java
```diff
@@ -0,0 +1,203 @@
...
+        pub.publishScanEvent(pub, ScanEventPublisher.SCAN_STARTED_EVENT, 1);
+
+        // Then
+        assertIsBasicEvent();
+    }
+
+    @Test
+    void shouldPublishBasicScanEventWithNullTargetAndUser() {
+        // Given / When
+        pub.publishScanEvent(pub, ScanEventPublisher.SCAN_STARTED_EVENT, 1, (Target) null, null);
```
thc202, line 90: I'd suggest declaring the variables we are testing (under Given) would be more evident what's being tested and would not require casts, same for following tests.

zap/src/test/java/org/zaproxy/zap/model/ScanEventPublisherUnitTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+        // Then
+        assertIsBasicEvent();
+    }
+
+    @Test
+    void shouldPublishBasicScanEventWithNullUserAndUser() {
```
thc202, line 97: `UrlAndUser`

zap/src/test/java/org/zaproxy/zap/model/ScanEventPublisherUnitTest.java
```diff
@@ -0,0 +1,203 @@
...
+    @Test
+    void shouldGetPublisherName() {
+        // Given / When
+        String name = pub.getPublisherName();
+
+        // Then
+        assertThat(name, is("org.zaproxy.zap.model.ScanEventPublisher"));
+    }
+
+    void assertIsBasicEvent() {
```
thc202, line 67: This could allow to pass the event type and target, would reduce duplication in one more test.

psiinon:
Done :)

thc202 (approved):

thc202:
Thank you!

PR #8115, opened 2023-10-11
Use bold font for `TitledBorder` highlight

### LINKED ISSUE #8111: Raw HTML displayed in options panels for search matches
### Describe the bug

After a user has searched within ZAP's options, the title border(s) within some of the Network options panels display raw HTML instead of properly highlighting.

1. Open ZAP.
2. Go into the Option panel.
3. Type "prox" in the search area and hit the magnifying glass button.
4. Goto: Network > Local Servers/Proxies note both the border titles now include raw HTML.

### Expected behavior

The border titles should properly render the text and highlight without raw HTML.

### Software versions

Tested in both current dev and 2.13.0.

Network add-on 0.11.2.

### Screenshots

image

None

### Additional context

N/A

- [ ] Yes

thc202:
Highlight with bold font instead of using the HTML which is not properly rendered anymore.

Fix #8111.

---
Suggested in IRC by @psiinon :)

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/view/panelsearch/HighlighterUtils.java
+++ b/zap/src/main/java/org/zaproxy/zap/view/panelsearch/HighlighterUtils.java
@@ -20,8 +20,10 @@
 package org.zaproxy.zap.view.panelsearch;
 
 import java.awt.Color;
+import java.awt.Font;
 import javax.swing.BorderFactory;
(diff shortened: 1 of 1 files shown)


psiinon (approved):

kingthorin:
Thanks for tackling this so quickly.

PR #8154, opened 2023-10-30
Docker: Add ZAP client profile

psiinon:
This will prevent us (and users) from having to do these sort of horrible things: https://github.com/zapbot/zap-mgmt-scripts/blob/a1623ba93d69e10458f386f2a01a281e36882caa/.github/workflows/zap-vs-crawlmaze.yml#L44-L48

The files are very small and should have no impact unless the client add-on is used with the AJAX spider.


thc202:
Thank you!

kingthorin (approved):

PR #7433, opened 2022-08-23
deprecate: Core HAR API Endpoints

kingthorin:
Related to zaproxy/zaproxy#6579

image
**Note**: the screenshot says "User the...." I did fix that, though didn't retake the screenshot :wink:

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #8157, opened 2023-10-31
search: Deprecate HAR related APIs

kingthorin:
The plan is to migrate these to exim.

Related to zaproxy/zaproxy#7433


kingthorin:
Withdrawn per newer IRC discussions.

PR #8159, opened 2023-11-01
Reserve id for DNS Active Scanner

magmax:
Reserving ID for [DNS Scanner](https://github.com/zaproxy/zap-extensions/pull/5044), according [to comment.](https://github.com/zaproxy/zap-extensions/pull/5044#issuecomment-1784682631).

### DIFF
--- a/docs/scanners.md
+++ b/docs/scanners.md
@@ -221,6 +221,7 @@ Scan rules:
 90037   Command Injection (Timing Based)
 90038   SQL Injection SQLite (Timing Based)
 90039   NoSQL Injection MongoDB (Timing Based)
+90040   DNS add-on SPF rule
 
 100000  Client/Server HTTP Error Response Codes [Script]
 100001  Unexpected Content Types [Script]


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
Should this be specific to the SPF rule?

magmax:
> Should this be specific to the SPF rule?

I am not thinking in checking just that, @kingthorin .

My intention is, at least:
- check SPF rules
- check dmarc rules by checking `_dmarc.<server>` TXT records
- check CAA records
- extract information in the same sense "wappalizer" addon does from TXT records.

An example:
```
$ dig TXT owasp.org +short | sort
"atlassian-domain-verification=BhaFKFKoRcW20xvi6UJ3U0CKocKOCgLH6LSuiBYPQ5A53cSCUN6gcbzcKSOmlVGs"
"google-site-verification=1zT9Of9pBuTj1rgeGCxMbya3iQQMxFE9-DzUBhftUVQ"
"google-site-verification=I9qx_X9EKlR_rfceG25-iXHBXJvLrmeNbkEdy182iI"
"google-site-verification=_slXlbOCopK1Ss9VQEoxdsNxpScVKvXVB_JtPpyL3eQ"
"google-site-verification=ubHJGF1N2ylOhYxQnIzEIIFaqUodqsIdTLXF-rCX9ps"
"MS=ms73859685"
"v=spf1 include:_spf.google.com include:servers.mcsv.net include:amazonses.com ~all"
```

With that information, you will know that:
- the SPF rule is fine, because `amazonses.com` ends in hard fail (what will ignore the lazy fail in the main rule)
- probably uses AWS technologies
- probably uses Google technologies
- probably uses Atlassian technologies

then, checking the `_dmarc.owasp.org` TXT record:
```
$ dig TXT _dmarc.owasp.org +short | sort                                             
"v=DMARC1; p=quarantine; rua=mailto:88b4c392605549efbbfedf60d5b5ba94@dmarc-reports.cloudflare.net,mailto:owasp-dmarc-reports@owasp.com"
```

we can guess:
- it probably uses cloudflare.net technologies
- `owasp-dmarc-reports@owasp.com` is a probable user if the main site has a login.

Beyond that... well, I just want to see what else can be retrieved.

kingthorin:
The question is more around the intended end user experience. If the different checks/rules should be independently enabled/disabled then there should be rule IDs for each, and this should be specific to SPF since that's all that's done currently.

magmax:
> The question is more around the intended end user experience. If the different checks/rules should be independently enabled/disabled then there should be rule IDs for each, and this should be specific to SPF since that's all that's done currently.

I'm not sure. This is my first contribution to the project. That's why I exposed my intention, to ask you for advice, @kingthorin

kingthorin:
Fair enough. From what we know at this point. I'd suggest each should be individual. It's plausable that a user wants to check dmarc but not spf and vice versa, extending to future rules/checks as well.

So I'd suggest for this one tweak the text to something like "DNS add-on SPF rule".

ricekot:
It could also follow the approach of the GraphQL add-on, where the ID is reserved by the add-on and each alert uses a different alertRef.

kingthorin:
@thc202 / @psiinon can one of you tie break this? 😀

thc202:
I'd need more information on how everything is going to be tied together, which alerts are going to be raised, and how the techs are going to be reported. But, I'd go with @ricekot's suggestion for the time being, if we need more scan rule IDs they can be added and the name of the entry adjusted.

magmax:
I can write the "SPF Scanner" and focus on it. Then I can move to the CAA and application discovery features, where we can rethink if they should be joined or kept separated.

Sounds good?

psiinon:
Agree with @thc202 
I think a scan rule should focus on one specific vulnerability. If there are different variations then thats when we have multiple alertRefs. If a rule find very different vulnerabilities then it should be split into 2 or more rules.
Remember that users can disable specific rules but not the variations.

kingthorin:
Thanks

kingthorin:
So merge as-is or further tweak warranted?

thc202:
I'm ok with merging as-is, the name/description is not important at the moment anyway (IMO), the scan rule is not yet published anywhere.

psiinon (approved):

PR #8164, opened 2023-11-02
workflow: Adjust PR Labeler, skip if already labeled

kingthorin:
Only label if no labels to start with.

~~For reference: https://github.com/search?q=%22%21+contains%28github.event.pull_request.labels%22&type=code~~

Ref: https://stackoverflow.com/a/63639935/7718222


thc202 (approved):

thc202:
Thank you!

PR #8171, opened 2023-11-04
Fix rendering of Physical fonts

ricekot:
Closes zaproxy#8018.

Fonts using actual font libraries, e.g. TrueType or PostScript Type 1 fonts were not being rendered correctly.

The original issue is for Windows, but I'm seeing it on Linux too.

Using `StyleContext#getFont` works because it checks if the requested font is a Physical font and handles it correctly.

### References:
- https://docs.oracle.com/javase/tutorial/2d/text/fonts.html
- https://github.com/openjdk/jdk/blob/da75f3c4ad5bdf25167a3ed80e51f567ab3dbd01/src/java.desktop/share/classes/javax/swing/text/StyleContext.java#L267-L269

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/utils/FontUtils.java
+++ b/zap/src/main/java/org/zaproxy/zap/utils/FontUtils.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.EnumMap;
 import javax.swing.JLabel;
 import javax.swing.UIManager;
+import javax.swing.text.StyleContext;
 
 public class FontUtils {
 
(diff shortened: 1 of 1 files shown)


kingthorin:
IIRC there was an issue in the last month or two about this. I'll see if I can dig it up later (on cell phone now).

kingthorin:
Oops you did have it and I missed it 🤦‍♂️

kingthorin (approved):

zap/src/main/java/org/zaproxy/zap/utils/FontUtils.java
```diff
@@ -114,7 +115,8 @@ public static void setDefaultFont(FontType fontType, String name, int size) {
         }
 
         defaultFontSets.put(fontType, (name != null && !name.isEmpty()));
-        setDefaultFont(fontType, new Font(name, Font.PLAIN, size));
+        setDefaultFont(
+                fontType, StyleContext.getDefaultStyleContext().getFont(name, Font.PLAIN, size));
```
thc202, line 119: Extract a method.

thc202 (approved):

Issue #8179, opened 2023-11-09
Drop "to Clipboard" from ZAP copy menu items (etc)


kingthorin:
### Is your feature request related to a problem? Please describe.

Some ZAP menu items are too verbose. "Copy ____ to Clipboard". This would be applicable to add-ons, help, and core.

### Describe the solution you'd like

Drop the "to Clipboard" as it's superfluous.

### Describe alternatives you've considered

None, it's 2023: default copy destination should be obvious.

### Screenshots

.

### Additional context

- https://github.com/search?q=repo%3Azaproxy%2Fzaproxy%20%22to%20clipboard%22&type=code
- https://github.com/search?q=repo%3Azaproxy%2Fzap-extensions+%22to+clipboard%22+language%3Ahtml+OR+path%3Amessages.properties&type=code
- https://github.com/search?q=repo%3Azaproxy%2Fzap-core-help+%22to+clipboard%22+language%3Ahtml+NOT+path%3A**%2Freleases%2F**+path%3Aaddons%2Fhelp%2Fsrc&type=code

### Would you like to help fix this issue?

- [X] Yes

PR #8180, opened 2023-11-09
Drop "to Clipboard" from ZAP copy menu items

kingthorin:
Part of zaproxy/zaproxy#8179

### DIFF
--- a/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
+++ b/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
@@ -556,9 +556,9 @@ ascan.progress.chart.4xx = 4xx
 ascan.progress.chart.5xx = 5xx
 ascan.progress.chart.responses = Responses / second
 ascan.progress.chart.time = Time 
-ascan.progress.copyclipboard.button.label = Copy to Clipboard
-ascan.progress.copyclipboard.button.tooltip = Copies the contents of the table to the clipboard
-ascan.progress.copyclipboard.error = Failed to copy the contents to the clipboard.
+ascan.progress.copyclipboard.button.label = Copy
+ascan.progress.copyclipboard.button.tooltip = Copies the contents of the table
+ascan.progress.copyclipboard.error = Failed to copy the contents.
 ascan.progress.label.completed = Completed
 ascan.progress.label.host = Host:
 ascan.progress.label.pending = Pending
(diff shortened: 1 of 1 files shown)


zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -556,7 +556,7 @@ ascan.progress.chart.4xx = 4xx
 ascan.progress.chart.5xx = 5xx
 ascan.progress.chart.responses = Responses / second
 ascan.progress.chart.time = Time 
-ascan.progress.copyclipboard.button.label = Copy to Clipboard
+ascan.progress.copyclipboard.button.label = Copy
 ascan.progress.copyclipboard.button.tooltip = Copies the contents of the table to the clipboard
 ascan.progress.copyclipboard.error = Failed to copy the contents to the clipboard.
```
thc202, line 561: Should drop from there too?

kingthorin: Sure. I kinda thought these were more purposefully explanatory but it doesn't really add much, I think clipboard is still a safe assumption here.

kingthorin:
Tweaked

ricekot (approved):
Thanks!

psiinon (approved):

thc202:
There are two more occurrences of clipboard that could be dropped (IMO).

kingthorin:
I don't see any occurrences of "to Clipboard" in *.html or Messages.properrties. There's one in *.java but it seems reasonable as part of a logged message for scan progress.

thc202:
e.g. https://github.com/zaproxy/zaproxy/blob/6609be13464a9bd611d0dc2d54893b0310c84cf1/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L1302

kingthorin:
Okay I was being too literal, I was only looking for "to Clipboard"

kingthorin:
Now includes that change.

thc202:
Only one was changed, on purpose?

kingthorin:
Because that's what was discussed. The other was a log message that seemed reasonable to me sir another I hadn't found.

thc202:
The two occurrences I was referring to were in the resources messages, other one being: https://github.com/zaproxy/zaproxy/blob/6609be13464a9bd611d0dc2d54893b0310c84cf1/zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties#L2283

kingthorin:
I probably skipped it because it was paste. Thanks, tweak coming.

kingthorin:
Tweaked.

thc202 (approved):

thc202:
Thank you!

@ricekot or @psiinon do you want to check again?

ricekot:
Still looks good to me, thanks!

PR #8183, opened 2023-11-09
Reserve Client pscan IDs

psiinon:
Its as start ;)

Feel free to suggest better wording ..

### DIFF
--- a/docs/scanners.md
+++ b/docs/scanners.md
@@ -268,6 +268,9 @@ Scan rules:
 110007  Username Disclosure [Script]
 110008  Suspicious XML Comments Disclosure [Script]
 
+120000  Information Disclosure - Information in Browser Storage
+120001  Information Disclosure - Sensitive Information in Browser Storage
+
 322420463 Retire.js (3rd Party)
 
 ```


thc202 (approved):

thc202:
Thank you!

docs/scanners.md
```diff
@@ -268,6 +268,9 @@ Scan rules:
 110007  Username Disclosure [Script]
 110008  Suspicious XML Comments Disclosure [Script]
 
+120000  Information Disclosure - Information in Browser Storage
+120001  Information Disclosure - Sensitive Information in Browser Storage
```
kingthorin, line 272: What's the distinction here? Just reading titles this sounds like a perfect place to use Alert Refs.

psiinon: The first is for _any_ local / session storage event .. and I will be using alertRefs for those. These will be raised at Info level. These are not vulns, they just help you understand the app better
The second is a rip off of InformationDisclosureInUrlScanRule and will be raised at Low.
I think they should be separate so people can turn one or other off - thats much harder if they have the same IDs.

kingthorin (approved):

PR #8184, opened 2023-11-09
Replace usage of Commons Lang v2

thc202:
Replace usage of Commons Lang v2 with JRE equivalent, Commons Lang v3, or Commons Text.
Add Spotless step to disallow use of v2 imports.


kingthorin (approved):

ricekot (approved):
Thanks!

PR #8190, opened 2023-11-14
alert: Add CWE Alert Tag when building and CWE ID has been set

kingthorin:
I also looked at doing this for WASC IDs but they don't have a handy URL, we'd have to add a pre-populated map which isn't impossible, but did seem like too much work for this morning 😀

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
@@ -67,6 +67,7 @@
 // ZAP: 2022/05/26 Add addTag and removeTag methods
 // ZAP: 2023/01/10 Tidy up logger.
 // ZAP: 2023/09/12 Add NUMBER_RISKS convenience constant.
+// ZAP: 2023/11/14 When setting CWE also add a CWE alert tag with an appropriate URL.
 package org.parosproxy.paros.core.scanner;
 
 import java.net.URL;
@@ -192,6 +193,9 @@ public class Alert implements Comparable<Alert> {
     public static final String[] MSG_CONFIDENCE = {
         "False Positive", "Low", "Medium", "High", "Confirmed"
     };
+
+    private static final String CWE_KEY = "CWE-";
+    private static final String CWE_URL_BASE = "https://cwe.mitre.org/data/definitions/";
 
     private int alertId = -1; // ZAP: Changed default alertId
     private int pluginId = -1;
@@ -1242,6 +1246,9 @@ public class Alert implements Comparable<Alert> {
          * <p>The alert URI defaults to the one from the {@code HistoryReference} or {@code
          * HttpMessage} if set.
          *
+         * <p><strong>Note:</strong> If the Alert has a CWE set then an associated Tag will be added
+         * during {@link #build()}
+         *
          * @return the alert with specified data.
          */
         public final Alert build() {
@@ -1277,9 +1284,27 @@ public class Alert implements Comparable<Alert> {
             if (alertRef != null) {
                 alert.setAlertRef(alertRef);
             }
-            alert.setTags(tags);
+            alert.setTags(withCweTag(tags, cweId));
 
             return alert;
+        }
+
+        private static String createCweUrl(int cweId) {
+            if (cweId <= 0) {
+                return "";
+            }
+            return CWE_URL_BASE + cweId + ".html";
+        }
+
+        private static Map<String, String> withCweTag(Map<String, String> existingTags, int cweId) {
+            String cweUrl = createCweUrl(cweId);
+            if (cweUrl.isEmpty()) {
+                return existingTags == null ? null : new HashMap<>(existingTags);
+            }
+            Map<String, String> newTags =
+                    existingTags == null ? new HashMap<>() : new HashMap<>(existingTags);
+            newTags.put(CWE_KEY + cweId, cweUrl);
+            return newTags;
         }
     }
 
--- a/zap/src/test/java/org/parosproxy/paros/core/scanner/AlertUnitTest.java
+++ b/zap/src/test/java/org/parosproxy/paros/core/scanner/AlertUnitTest.java
@@ -27,6 +27,8 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import java.util.HashMap;
 import java.util.Map;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.parosproxy.paros.core.scanner.Alert.Builder;
 
 class AlertUnitTest {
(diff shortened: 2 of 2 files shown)


ricekot (commented):
The function `Alert#setCweId` should be updated too?
https://github.com/zaproxy/zaproxy/blob/24a4eb5afa852605d2bba7b4f9e7bb9c305667ef/zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java#L931

kingthorin:
I dunno we didn't update the old alert functionality when we implemented addTag/removeTag.

I'm happy to do either, but was trying to stay consistent.

thc202:
Whatever we do we should document these side effects to avoid surprises (e.g. I can see this breaking our unit tests as there will be more tags now).

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -193,6 +195,10 @@ public static Source getSource(int id) {
         "False Positive", "Low", "Medium", "High", "Confirmed"
     };
 
+    private static final String CWE_TEMPLATE_KEY = "CWE-{0}";
+    private static final String CWE_TEMPLATE_URL =
+            "https://cwe.mitre.org/data/definitions/{0}.html";
```
thc202, line 200: It would be simpler to just concatenate. Having hardcoded link in core :/ but guess there's no way around (it other than allowing to set from add-ons).

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1142,6 +1148,7 @@ public Builder setInputVector(String inputVector) {
 
         public Builder setCweId(int cweId) {
             this.cweId = cweId;
+            this.addCweTag(cweId);
```
thc202, line 1151: - This should be done when the alert is being built. There should be a test showing that setting multiple IDs does not lead to multiple tags.
- This behaviour should be documented (in `build`?).

kingthorin:
Okay how about this:

1. I'll deprecate the existing setCwe(int) and add setCwe(int, boolean). This can easily be handled with a regex search and replace (with grouping/interpolation) when we update to new core. Where boolean true is to add the tag.
2. In Builder.build() when boolean true then it's added.
3. Add functionality/tests to not set multiple tags.
4. In the Alert functionality I'll add an addCweTag() method? Or also do 1-3 there?

kingthorin:
Crew I need some feedback on this so that it can be moved ahead.

kingthorin:
Anyone out there?

ricekot:
I like the original approach of adding the tags in the builder, if a valid CWE ID is set.

thc202:
I'm also happy with the builder way, my only concern is the URL (though that one should stay stable…).

kingthorin:
Thank you both! I'll get this reworked.

zap/src/test/java/org/parosproxy/paros/core/scanner/AlertUnitTest.java
```diff
@@ -157,4 +159,55 @@ void shouldBuildAlertWithNoTagsWhenOneRemoved() {
...
+    @Test
+    void shouldAddOneCweTagEvenIfCweIdSetTwice() {
+        // Given
+        int cwe = 618;
+        int cwe2 = 619;
+        String cweUrl = "https://cwe.mitre.org/data/definitions/619.html";
+        String cweKey = "CWE-" + cwe2;
+        Builder builder = new Alert.Builder();
+        builder.setCweId(cwe);
+        builder.setCweId(cwe2);
```
thc202, line 203: Call build before this line as well (the tag needs to be added to the alert rather than the builder).

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1242,6 +1246,9 @@ public Builder removeTag(String tag, String value) {
          * <p>The alert URI defaults to the one from the {@code HistoryReference} or {@code
          * HttpMessage} if set.
          *
+         * <p><strong>Note:</strong> If the Alert has a CWE set then an associated Tag will be added
+         * during @link Builder#build()
```
thc202, line 1250: Missing {}.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1277,10 +1284,26 @@ public final Alert build() {
             if (alertRef != null) {
                 alert.setAlertRef(alertRef);
             }
+            addCweTag(cweId);
             alert.setTags(tags);
```
thc202, line 1288: Not part of this PR but we should be setting a copy, otherwise changes done afterwards to the builder might affect an alert already built.

kingthorin: I looked at this quickly, from my understanding it would need a deep copy which there doesn't seem to be a simple solution to. https://www.baeldung.com/java-copy-hashmap

I only checked that one article so I could have just missed the point.

kingthorin:
Tweaked

zap/src/test/java/org/parosproxy/paros/core/scanner/AlertUnitTest.java
thc202: The semantics of the (new) test were changed, the second call to `setCweId` should still be done to the builder.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
thc202: e.g. `new HashMap(tags)` would be enough to copy. Now that I'm reading the code again, it will have to be addressed in this PR anyway, to add the CWE tag.

zap/src/test/java/org/parosproxy/paros/core/scanner/AlertUnitTest.java
kingthorin: Ok, I misinterpreted something said earlier. Will tweak.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
kingthorin: Okay.

kingthorin:
Okay think I got this into a more expected state.

zap/src/main/java/org/parosproxy/paros/core/scanner/Alert.java
```diff
@@ -1277,10 +1284,32 @@ public final Alert build() {
             if (alertRef != null) {
                 alert.setAlertRef(alertRef);
             }
-            alert.setTags(tags);
+            addCweTag(cweId);
+            alert.setTags(new HashMap<>(tags == null ? new HashMap<>() : tags));
```
thc202, line 1288:
```suggestion
            alert.setTags(withCweTag(tags, cweId));
```
and let `withCweTag` add the tag to the new map if/when needed, it should be simpler than add the tag to the `tags` and then having to undo.

kingthorin:
Tweaked again.

kingthorin:
Okay hopefully this is finally good to go.

kingthorin:
I had to re-work it a bit more because I was getting NPE and unmodifiable map exceptions while working on something else this evening. Hopefully it's in a good state now.

thc202 (approved):

thc202:
Thank you!

kingthorin:
For the record this will mean that we need to update "expected mappings" type unit tests when zap-extensions migrates to 2.15.

psiinon (approved):

PR #8198, opened 2023-11-20
URL Path Input Vector - attack end path too

psiinon:
This will increase the number of paths returned, but not by much and the URL path variant is disabled by default.


kingthorin (approved):
Seems okay, not sure what else this might inadvertently impact. (OpenAPI?)

thc202 (approved):

thc202:
Thank you!

PR #8201, opened 2023-11-20
Allow to build weekly with normal output

thc202:
Expose the quiet state through an env var so it can be tweaked anytime.
Update weekly workflow to pass the env var.

### DIFF
--- a/.github/workflows/release-weekly.yml
+++ b/.github/workflows/release-weekly.yml
@@ -23,4 +23,5 @@ jobs:
         ZAP_RELEASE: 1
         ZAP_JAVA_VERSION: 11
         ZAP_WEEKLY_ADDONS_NO_TEST: ${{ vars.ZAP_WEEKLY_ADDONS_NO_TEST }}
+        ZAP_WEEKLY_QUIET: ${{ vars.ZAP_WEEKLY_QUIET }}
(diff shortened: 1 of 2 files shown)


psiinon (approved):

.github/workflows/release-weekly.yml
```diff
@@ -23,4 +23,5 @@ jobs:
         ZAP_RELEASE: 1
         ZAP_JAVA_VERSION: 11
         ZAP_WEEKLY_ADDONS_NO_TEST: ${{ vars.ZAP_WEEKLY_ADDONS_NO_TEST }}
+        ZAP_WEEKLY_QUIET: ${{ vars.ZAP_WEEKLY_QUIET }}
       run: ./gradlew "-Dorg.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m" :zap:createWeeklyRelease
```
kingthorin, line 27: Have any of you checked this lately? To know if 2G is lots or if we're close to that?

thc202: Should be enough for the add-ons included in the weekly (not building all of them like in zap-extensions).

kingthorin (approved):

PR #8203, opened 2023-11-21
Option to encode cookie values

psiinon:
This changes the default behaviour, but I think for the better.
Help PR will follow.

<details><summary>Active Scan Input Vectors Options dialog</summary>
<p>

Screenshot 2023-11-21 at 12 16 09

</p>
</details>


psiinon:
Looking at the failing tests now..

kingthorin:
🤦‍♂️ spotless

psiinon:
ugh, did that before, so that must be a problem with the properties file :/

psiinon:
Help PR: https://github.com/zaproxy/zap-core-help/pull/536

psiinon:
All tests passing now 😉

kingthorin:
Hmm see the help PR before my comments here

psiinon:
Yeah, looks like this is a misunderstanding.
Its a global option, encode cookie values or dont encode them.
I think the previous way of always encoding them was probably wrong.

I'm not sure attacking both encoded and unencoded makes a lot of sence (from a number of requests point of view), but I'm willing to be convinced otherwise.
I would rather have an initialisation step when we inject non alphanumberic values into a cookie and see if they are reflected encoded or not.
But that would be a future change.

zap/src/main/java/org/zaproxy/zap/extension/ascan/VariantFactory.java
```diff
@@ -135,7 +135,9 @@ public List<Variant> createVariants(ScannerParam scanOptions, HttpMessage messag
         }
 
         if ((targets & ScannerParam.TARGET_COOKIE) != 0) {
-            listVariant.add(new VariantCookie());
+            VariantCookie vc = new VariantCookie();
+            vc.setEncodeCookieValues(scanOptions.isEncodeCookieValues());
```
kingthorin, line 139: Would it be "better" to have a constructor that accepts a boolean and then have the existing constructor deprecated and use "false"? Then setup of the Variant isn't a two step process?

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantCookie.java
```diff
@@ -121,7 +123,7 @@ public List<NameValuePair> getParamList() {
     @Override
     public String setParameter(
             HttpMessage msg, NameValuePair originalPair, String name, String value) {
-        return setParameter(msg, originalPair, name, value, false);
+        return setParameter(msg, originalPair, name, value, !encodeCookieValues);
```
kingthorin, line 126: This negative logic bothers me. Above we set the default as false so won't !false be true? (In the GUI won't the checkbox be unchecked if the default is false?)

zap/src/main/java/org/zaproxy/zap/extension/ascan/VariantFactory.java
psiinon: I thought of that, but ended up copying the way VariantJSONQuery did it above. 
Wasnt sure what the implications would be, but a deprecated constructor would solve that..

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantCookie.java
psiinon: Will be changing this as part of a change suggested by @thc202 in irc 😉

kingthorin (approved):
LGTM

zap/src/main/java/org/parosproxy/paros/core/scanner/VariantCookie.java
```diff
@@ -190,4 +198,13 @@ private String setParameter(
 
         return name + "=" + escapedValue;
     }
+
+    /**
+     * Sets whether or not the cookie values should be encoded.
+     *
+     * @since 2.15.0
+     */
+    public void setEncodeCookieValues(boolean encodeCookieValues) {
```
thc202, line 207: Does it worth keeping this?

zap/src/main/java/org/zaproxy/zap/extension/ascan/OptionsVariantPanel.java
```diff
@@ -415,6 +426,7 @@ public void initParam(ScannerParam param) {
                         getChkInjectableHeaders().isEnabled()
                                 && getChkInjectableHeaders().isSelected());
         this.getChkInjectableCookie().setSelected((targets & ScannerParam.TARGET_COOKIE) != 0);
+        this.getChkEncodeCookieValues().setSelected(param.isEncodeCookieValues());
```
thc202, line 429: Should check if `getChkInjectableCookie()` is enabled as well.

psiinon:
Fixed those..

thc202 (approved):

thc202:
Thank you!

PR #8230, opened 2023-12-07
Docker: better parse error

psiinon:
Example output:

* 2023-12-07 17:43:52,079 Failed to load config file /zap/wrk/gen.conf Unexpected number of tokens on line - there should be at least 3, tab separated 10026 WARN	(HTTP Parameter Override)


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

Issue #8231, opened 2023-12-08
Add application/x-x509-ca-cert to avoid Unexpected Content-Type being returned


ssthom:
Similar to https://github.com/zaproxy/zaproxy/issues/8226

`application/x-x509-ca-cert` is the default response from our Spring-Boot application when there is an error like 404 and it was trying to get a cert file that doesn't exist: https://example.com/key.pem

But it is a valid response type when looking at https://www.iana.org/assignments/media-types/application/x-x509-ca-cert and https://datatracker.ietf.org/doc/html/rfc5280 

>4.2.1.1. CA Certificate Response Message Format
If the CA does not have any intermediate CA certificates, the response consists of a single X.509 CA certificate. The response will have a Content-Type of "application/x-x509-ca-cert".
"Content-Type: application/x-x509-ca-cert"

###  Steps to reproduce the behavior
1. Create spring-boot application
2. Scan a GET endpoint with ZAP, which will try with /key.pem
3. The report will have the "Unexpected Content-Type was returned" finding with application/x-x509-ca-cert

### Screenshots
![image](https://github.com/zaproxy/zaproxy/assets/23663785/4e22cf1d-a0f8-46ae-9aa9-ca4b2a0be843)

kingthorin:
Per comment on the associated PR, it doesn't make sense to add this content type as "expected".

---

> I don't think we should add this. A certificate response is unexpected, perhaps you requested one in the first place but the response shouldn't lie about what's actually going on (if a cert was requested and not found then the server isn't sending a cert back).
> 
> 
> 
> If this is causing a FP for you I believe it should be handled via Alert Filters.

PR #8232, opened 2023-12-08
Add application/x-x509-ca-cert to expected API content types

ssthom:
fixes: https://github.com/zaproxy/zaproxy/issues/8231

`application/x-x509-ca-cert` is the default response from our Spring-Boot application when there is an error like 404 and it was trying to get a cert file that doesn't exist: https://example.com/key.pem

But it is a valid response type when looking at https://www.iana.org/assignments/media-types/application/x-x509-ca-cert and https://datatracker.ietf.org/doc/html/rfc5280 

>4.2.1.1. CA Certificate Response Message Format
If the CA does not have any intermediate CA certificates, the response consists of a single X.509 CA certificate. The response will have a Content-Type of "application/x-x509-ca-cert".
"Content-Type: application/x-x509-ca-cert"

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2023-12-08
+- Alert_on_Unexpected_Content_Types.js > Added Content-Type application/x-x509-ca-cert to the list of expected types. 
+
 ### 2023-12-07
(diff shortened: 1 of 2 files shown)


kingthorin:
I don't think we should add this. A certificate response is unexpected, perhaps you requested one in the first place but the response shouldn't lie about what's actually going on.

If this is causing a FP for you I believe it should be handled via Alert Filters.

PR #8233, opened 2023-12-08
Security document update

kuljotbiring:
Added direct links to reporting a bug, scan policies, and terms of use in order to make the document more robust and accessible to users of the product.  Added (lack of) security advisory. Document updated with intent and scope of product use with words of caution to users while using the application.


kingthorin:

SECURITY.md
```diff
@@ -1,3 +1,4 @@
+
```
kingthorin, line 1: This white space change is unnecessary

SECURITY.md
```diff
@@ -12,3 +13,62 @@
 Please report any vulnerabilities via our [Bug Bounty Program](https://bugcrowd.com/owaspzap).
 
 Remote Code Execution for this program will be rewarded at __$1000__
+
+## Reporting a Bug
+
+Please report any bugs via our issue [Bug Issue Tracker](https://github.com/zaproxy/zaproxy/issues/new/choose)
```
kingthorin, line 19: Drop “issue” before the link

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+
+## Security Advisories
+
+Currently, there are not any published security advisories.
```
kingthorin, line 23: I’m don’t think this is true/accurate, I’d have to go digging. Pretty sure we have/had at least one.

kuljotbiring:
I agree with [developercertificate.org](http://developercertificate.org/).

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+
+## Scan Policy
+
+ZAP supports both active and passive scanning rules.
+
+All rules are contained in add-ons so that they can be updated quickly and easily.
+
+Active scanning is an attack on those targets.
```
kingthorin, line 31: What targets? The text was just talking about scan rules

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+You should NOT use it on web applications that you do not own.
+
+It should be noted that active scanning can only find certain types of vulnerabilities.
+Logical vulnerabilities, such as broken access control, will not be found by any active or automated vulnerability scanning.
```
kingthorin, line 35: This isn’t really true, you can use the Access Control add-on to scan for those. It would be more realistic to say something about business logic issues.

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+Currently, there are not any published security advisories.
+
+## Scan Policy
+
+ZAP supports both active and passive scanning rules.
+
+All rules are contained in add-ons so that they can be updated quickly and easily.
+
+Active scanning is an attack on those targets.
+You should NOT use it on web applications that you do not own.
```
kingthorin, line 32: The note below is more accurate, you don’t have to own it to be authorized to asses something

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+
+## Terms of Use
+
+By using Zaproxy, you agree to comply with the terms and conditions below. Terms and conditions may be subject to change or modification at any time. Users are responsible for reviewing the terms regularly to ensure compliance with the latest version.
+
+1. Authorized Usage: You may only use Zaproxy for the purpose of assessing the security of web applications for which you have explicit and authorized access, permission, or ownership. Unauthorized scanning is strictly prohibited and unlawful.
+
+2. Legality and Compliance: You alone are responsible for ensuring that your use of Zaproxy complies with all applicable laws, regulations, and ethical standards. Please follow the applicable legal requirements and industry best practices at all times.
+
+3. Privacy and Data Protection: When using Zaproxy, refrain from engaging in activities that compromise the privacy or security of data, PII and other sensitive data.
```
kingthorin, line 49: PII and sensitive data are data. This should be shortened

kingthorin (commented):
I'm not sure this is necessary or the correct place but here's a first quick review

thc202:
> I'm not sure this is necessary or the correct place but here's a first quick review

I agree.

SECURITY.md
psiinon: We have had some security bugs reported which we've fixed, but we have not published ny security advisories. 
Maybe we should...

kingthorin: I thought one or two had assigned CVEs (via the reporter not us), which I was lumping into the 'advisory' category.

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+
+Manual penetration testing should always be performed in addition to active scanning to find all types of vulnerabilities.
+
+Please see the Terms of Use and Disclaimer for restrictions.
+
+## Terms of Use
+
+By using Zaproxy, you agree to comply with the terms and conditions below. Terms and conditions may be subject to change or modification at any time. Users are responsible for reviewing the terms regularly to ensure compliance with the latest version.
+
+1. Authorized Usage: You may only use Zaproxy for the purpose of assessing the security of web applications for which you have explicit and authorized access, permission, or ownership. Unauthorized scanning is strictly prohibited and unlawful.
```
psiinon, line 45: We never use `Zaproxy` in this context, so this should be `ZAP`.
https://www.zaproxy.org/docs/developer/dev-rules-and-guidelines/#style-guidelines

SECURITY.md
```diff
@@ -12,3 +13,62 @@
...
+
+2. The use of this tool on web applications without proper authorization is illegal and unethical. You should only scan web applications for which you have explicit permission from the appropriate authority. Unauthorized use of this tool can result in legal ramifications.
+
+3. The results generated by this tool should be used for informational and educational purposes only. The authors and contributors are not morally or legally responsible for any actions taken based on the results of scans. 
+
+4. The authors, contributors, and sponsors of this tool do not endorse or promote any malicious or harmful activities. This tool is intended for security students, security professionals, ethical hackers, web developers, administrators to assess the security of their applications and systems.
+
+5. Use of this tool may cause disruption or unintended consequences to web applications and systems. The user is responsible for any damages or issues that may arise during or after the use of this tool.
+
+
```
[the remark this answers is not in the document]
kingthorin: I prefer the existing text.

Issue #8237, opened 2023-12-12
zap-api-scan.py fails with a config file with empty lines


kovacsbela:
### Describe the bug

I have pulled `docker pull owasp/zap2docker-weekly:latest@sha256:888516653c34567971d5d8e4b58863c53ea5d6c6baae0e345d56860e35f753a4` and I am running 
```zap-api-scan.py -j -d -t openapi-2414775296.json -f openapi -c zap.conf -r report.html -w report.md -J report.json ...``` on latest weekly `owasp/zap2docker-weekly                             latest                 1c6bca9169a1   16 hours ago    1.86GB`
where `zap.conf` contains empty lines between some other configs, then it exits after printing:
```
2023-12-12 08:53:19,341 Failed to load config file /zap/wrk/zap.conf Unexpected number of tokens on line - there should be at least 3, tab separated:

```
(empty line in 2nd row)

These empty lines are probably not ignored because of (this)[https://github.com/zaproxy/zaproxy/commit/0e5a9b6cf478bb8ba5372e01b64f96adac53abc2#diff-98e727b210699131d2194fd611f0b3144be743fd9bad19109380af5b97a17305L155-R155] change in the logic(1 length of a `\n` is now not ignored) in zaproxy/zaproxy#8230 ? 
Can you please check if this is an intended modification?

### Steps to reproduce the behavior

Run `zap-api-scan.py ` with a `-c` config file that contains an empty line. Once I removed empty lines, my scan start up fine.

### Expected behavior

Ignore empty lines in config file. Empty line is in this case probably a singe new line char?

### Software versions

I pulled via `docker pull owasp/zap2docker-weekly:latest@sha256:888516653c34567971d5d8e4b58863c53ea5d6c6baae0e345d56860e35f753a4`

### Would you like to help fix this issue?

- [X] Yes

**Comment by psiinon** [comment]
ZAP should ignore either completely empty lines or lines starting with a # comment.
The fix is for you to remove those lines or comment them out.

**Comment by psiinon** [comment]
@kovacsbela Just checked and the baseline scan is fine with completely blank lines, so I'm guessing yours had one whitespace character?

**Comment by kovacsbela** [comment]
My config file looks something like:
```
# comment
40045	IGNORE	(Spring4Shell)

# other comment
90004	IGNORE	(Insufficient Site Isolation Against Spectre Vulnerability)
```
resulting in an error complaining about number of the tokens in the empty line:
```
[...]

PR #8238, opened 2023-12-12
Docker: ignore config lines that contain only whitespaces

kovacsbela:
Closes: zaproxy/zaproxy#8237

This fix in config file parsing adds stripping whitespaces from the lines, as new line char in empty lines leads to a line length of 1, failing with an error about config file formatting a descibed in the issue

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2023-12-12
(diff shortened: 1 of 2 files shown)


psiinon (approved):

thc202 (approved):

kovacsbela:
@thc202 @psiinon do you think this issue is big enough to get out a new weekly latest docker tag including this change?

thc202:
There's a new tag which should include this change.

PR #8257, opened 2023-12-19
updating API-Minimal policy in Docker

PyaeHlianMoe:
reformat the policy file and remove the deprecated rules from the policy.


thc202:
It's unclear what you intend with these changes, the PR description says that you are removing deprecated rules but e.g. 10051 is still in use (note that how the policy is defined removing rules is the same as enabling them). You are also changing rules to be enabled (e.g. 40015) while leaving the level as OFF which is the same as keeping them disabled.

Closing, feel free to post to the [ZAP Developer Group](https://groups.google.com/g/zaproxy-develop) explaining the reason for these changes.

PR #8265, opened 2023-12-22
Deprecate Script Scan Rules and related classes

ricekot:
These will be made available as part of the scripts add-on.


zap/src/main/java/org/zaproxy/zap/control/CoreFunctionality.java
```diff
@@ -142,24 +141,7 @@ private static synchronized void createExtensions() {
      * @see AbstractPlugin
      */
     public static List<AbstractPlugin> getBuiltInActiveScanRules() {
```
thc202, line 143: Remove it?

ricekot: Kept it because it was public, but I don't mind removing it.

thc202: I think we can consider this class internal, it's to bootstrap ZAP. We can also deprecate it first.

thc202 (approved):

thc202:
Thank you!

ricekot:
Set to WIP because this should only be merged after https://github.com/zaproxy/zap-extensions/pull/5175.

kingthorin (approved):

thc202:
@ricekot ok to merge now?

Issue #8280, opened 2024-01-03
Use CSRF Token partial matching in scan rules


e-belair:
### Describe the bug

I have partial matching enabled and I still get warnings about Anti-CSRF token missing.
As you can see in the picture, the setting is enabled. Then in my form, I have the token "_csrf_token" in the element name.
Maybe I didn't understand the goal of this parameter, but I expected my token name to be detected as anti-CSRF token because in my app, all the token names are generated to be unique on the web page.

![image](https://github.com/zaproxy/zaproxy/assets/596562/53d88267-2314-4ca2-acaf-02e694d28184)
![image](https://github.com/zaproxy/zaproxy/assets/596562/a6fc0f48-4877-4671-bbaf-3a726f6b3fd0)

Note : I've created this issue following the recommendation on google group: https://groups.google.com/g/zaproxy-users/c/TuDKUcg9rR0/m/8ca3H6F7BAAJGoogle Group

### Steps to reproduce the behavior

1. Create a form containing a CSRF token input with a name containing _csrf_token such as foobar_csrf_token
2. Ensure partial matching in checked and you have a rule for _csrf_token
3. Scan the page containing the form
4. Enjoy the error

### Expected behavior

I expect the error to not occur. I mean, if my input name is foobar_csrf_token, zap should find it as it contains the partial name _csrf_token.

### Software versions

2.14.0

### Would you like to help fix this issue?

- [ ] Yes

kingthorin:
It seems when partial matching was added to the core it wasn't accounted for in the scan rule(s).

- CsrfCountermeasuresScanRule.java (passive) seems to be using equalsIgnoreCase.
- CsrfTokenScanRule.java (active) doesn't use those settings/names at all.

thc202:
It was not meant for the scan rules but the anti-csrf regeneration.

kingthorin:
Hmmm the group thread really read like the scan rule. I'm happy to leave it open for both though 🤷‍♂️

thc202:
I was referring to the (original) option.

kingthorin:
Oh, I see what you're getting at.

e-belair:
Ha ok, so if I correctly understood, the option is only used to make zap resend the token on the next requests?

thc202:
It was used only in that functionality but I don't see a reason why it can be used more generally (I was just justifying why the scan rules were not using it).

e-belair:
[...]

PR #8283, opened 2024-01-04
Anti-CSRF Handling should always account for partial matching

kingthorin:
- ExtensionAntiCSRF - Ensure public method always accounts for partial matching.
- ExtensionAntiCSRFUnitTest - Add a method to assert the behavior.

Related to:
- zaproxy/zap-extensions#5196
- zaproxy/zaproxy#8280

### DIFF
--- a/zap/src/test/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRFUnitTest.java
+++ b/zap/src/test/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRFUnitTest.java
@@ -44,6 +44,8 @@ import org.apache.commons.httpclient.URI;
 import org.apache.commons.httpclient.URIException;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.parosproxy.paros.network.HtmlParameter;
 import org.parosproxy.paros.network.HttpMessage;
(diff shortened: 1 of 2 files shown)


zap/src/test/java/org/zaproxy/zap/extension/anticsrf/ExtensionAntiCSRFUnitTest.java
```diff
@@ -178,6 +178,30 @@ void shouldGetTokensFromInputFieldsIfIdIsAKnownToken() {
             assertAntiCsrfToken(tokens.get(2), message, KNOWN_TOKEN_3, "value3", 1);
         }
 
+        @Test
+        void shouldGetTokensFromInputFieldsIfKnownTokenWithPartialMatch() {
```
thc202, line 182: This is already covered by `shouldGetTokensFromInputFieldsIfIdAndNameIncludesAKnownTokenAndPartialMatching*`. Though still worth adding one for `isAntiCsrfToken(String)` specifically.

kingthorin: On it, will drop the println too :grinning:

kingthorin:
Also I noticed one of the constants in the test class is misspelled, which I'm happy to correct here or another PR.

`UNKOWN_TOKEN` ➡`UNKNOWN_TOKEN` (missing N in known).

thc202:
Fine here in a separate commit?

kingthorin:
Sure

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

Issue #4513, opened 2018-03-16
Allow logging levels to be changed dynamically


psiinon:
Initial suggestion is an Options panel which shows a list of all of the log4j loggers and allows their levels to be changed.
Looks like the current list of loggers can be found as per https://stackoverflow.com/questions/17511763/what-is-getcurrentloggerss-analog-in-log4j2

psiinon:
Also see https://stackoverflow.com/questions/4598702/dynamically-changing-log4j-log-level

Am0rphous:
Hi, logging can be configured in the log4j2.properties file which is located in the home folder of the user. E.g. `~/.ZAP/`.

It would be highly beneficial to have the functionality within `Options` in the tool, to disable logging completely.

I just experienced `syslog` in Kali being filled up with errors from Zap, resulting in a full disk.

thc202:
Would be good to know about those errors.

Am0rphous:
Okay so i checked and sorted the logs in `~/.ZAP/zap.log*` which repeats the following through many megabytes of logs. Not sure why the I/O error further down. Seems to be because of plugins, hence the need to turn the logging off, as its wasting my storage. Laptop is business provided with good hardware - no issues during one year of using it. The `connection failed` could be because i was hammering the server i was connecting to.
```shell
2026-06-26 17:35:17,938 [ZAP-ActiveScanner-1] WARN  SQLInjectionScanRule - SQL Injection vulnerability check failed for parameter [lucky.php] and payload [lucky.php%')) UNION ALL SELECT NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL,>
org.apache.hc.client5.http.HttpHostConnectException: Connect to http://<REDACTED>:30753 [/<REDACTED>] failed: Connection refused
        at java.base/sun.nio.ch.Net.pollConnect(Native Method)
        at java.base/sun.nio.ch.Net.pollConnectNow(Net.java:642)
        at java.base/sun.nio.ch.NioSocketImpl.timedFinishConnect(NioSocketImpl.java:555)
        at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:599)
        at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:284)
        at java.base/java.net.Socket.connect(Socket.java:668)
[...]

PR #8295, opened 2024-01-10
Allow setting the log level via a CLI argument

ricekot:


kingthorin:
Related to: https://github.com/zaproxy/zaproxy/issues/4513

kingthorin:
I don't see anything wrong with this. But I"m not clear on the implementation, is this using env var values? (ex: Where/how would the user set/override it?).

ricekot:
These can be set by passing flags to the JVM, e.g. 

```
java -Dzap.log.level=debug -jar /zap-2.15.0-SNAPSHOT.jar
```

Maybe we should also update `zap.sh` to accept a flag and append this to the arguments it passes to the JVM.

psiinon:
I think we should have a relatively easy documented way for users to set it, and if that means updating `zap.sh` (and `zap.bat`??) then we should do so.

ricekot:
Updated as discussed in IRC - added a `loglevel` flag that can be used to set the log level.

psiinon:
`-loglevel debug` gives way more than usual, but I guess thats expected?
`-loglevel bad` is ignored - I expected that to fail with some message about an invalid logging level..

Other than that its looking good 😁

ricekot:
> `-loglevel debug` gives way more than usual

We could change it to only set the level for the ZAP classes, and do not change the root log level.

> `-loglevel bad` is ignored - I expected that to fail with some message about an invalid logging level..

That was intentional. I thought a bad log level didn't warrant crashing ZAP, since we have a log level in the log4j2.properties file anyway, but I can tweak this.

psiinon:
+1 for only setting the ZAP logging levels (others may disagree?)
We should at least give a warning/error on a bad level otherwise it might be hard for a user to debug a typo?
The help will say which log levels are supported?

zap/src/main/java/org/parosproxy/paros/CommandLine.java
```diff
@@ -599,6 +609,15 @@ public boolean isNoStdOutLog() {
         return noStdOutLog;
     }
 
+    /**
+     * Returns the specified log level argument if valid, or {@code null}.
```
thc202, line 613: The "if valid" could be removed now.

zap/src/main/resources/org/zaproxy/zap/resources/Messages.properties
```diff
@@ -1107,7 +1107,7 @@ cfu.warn.nolaunch = The latest ZAP release: {0} has been downloaded to\n{1}\nYou
...
...specified directory instead of the default one\n\t-installdir <dir>        Overrides the code that detects where ZAP has been installed with the specified directory\n\t-h                       Shows all of the command line options available, including those added by add-ons\n\t-help                    The same as -h\n\t-newsession <path>       Creates a new session at the given location\n\t-session <path>          Opens the given session after starting ZAP\n\t-lowmem                  Use the database instead of memory as much as possible - this is still experimental\n\t-experimentaldb          Use the experimental generic database code, which is not surprisingly also still experimental\n\t-nostdout                Disables the default logging through standard output\n\t-loglevel <level>        Sets the log level, overriding the values specified in the log4j2.properties file in the home directory\n\t-sbomzip <path>          Creates a zip file containing all of the available SBOMs\n\t-suppinfo                Reports support info to the command line and exits\n\t-silent                  Ensures ZAP does not make any unsolicited requests, including 'check for updates'\nAdd-on options:\n
```
thc202, line 1110: Just noticed the duplicated resource name (`cmdline.help =`).

kingthorin: Good catch!

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #8298, opened 2024-01-11
Prevent NPE if no user creds

psiinon:
Seen when a user ran an AF plan with all auth/session/verification set to auto-detect.
They should have used auth = Browser Based, but we still shouldnt NPE.


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #8313, opened 2024-01-20
Feat: Support URL input for hostname override (issue #8312)

### LINKED ISSUE #8312: zap-api-scan.py allow to use a URL when overriding hostname (`-O`)
### Describe the bug

[Current logic in zap-api-scan.py](https://github.com/zaproxy/zaproxy/blob/b2aed8b5005495af6177a1ad92645d30d126f811/docker/zap-api-scan.py#L421) uses [urllib.parse.urljoin](https://docs.python.org/3/library/urllib.parse.html) to produce a server url when the -O (host override) option is used. This creates an issue when the target is also a URL and its scheme differs from that of the override host.

The issue can easily be demonstrated:

```python
from urllib.parse import urljoin
target_url = 'https://website.com/openapi.json'
host_override = 'http://127.0.0.1:3000'
target = urljoin(target_url, '//' + host_override)
print(target)
``` 

expected result: http://127.0.0.1:3000
actual result: https://http://127.0.0.1:3000

If the host override is given as just '127.0.0.1:3000', that results in a syntactically correct URL of `https://127.0.0.1:3000' that is still not actually correct.

### Expected behavior

I expect valid URL input to stay valid and for the scheme of the override host to be respected. Something like the following would work.

```python
if host_override and urlparse(host_override).scheme:
    target = host_override
elif host_override:
    target = urljoin(target_url, '//' + host_override)
```   

instead of the current:
```python
if host_override:
    target = urljoin(target_url, '//' + host_override)
``` 

### Software versions

I am using the zap nightly docker image which I pulled today.

<details><summary>Image details</summary>
<p>

  "Id" : "sha256:916f71db4468fb060f44918ec194dd8eb58b539b1e3aec4f354118a3e4c50502",
  "RepoTags" : [ "ghcr.io/zaproxy/zaproxy:nightly" ],
  "RepoDigests" : [ "ghcr.io/zaproxy/zaproxy@sha256:c1175078d375070d992f0c248e9075f00c0e6601a545d8a70e0862f8f881301e" 

</p>
</details> 

### Screenshots

image

_No response_

### Additional context

_No response_

- [X] Yes

cerrussell:
Closes #8312 

This PR provides for overriding the hostname with a URL and resolves the issue created when a remote openapi definition is specified whose scheme does match that of the host override.

### DIFF
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -58,7 +58,8 @@ import subprocess
 import sys
 import time
 from datetime import datetime
-from six.moves.urllib.parse import urljoin
+
+from six.moves.urllib.parse import urljoin, urlparse
 from zapv2 import ZAPv2
 from zap_common import *
 
@@ -110,7 +111,7 @@ def usage():
     print('    -S                safe mode this will skip the active scan and perform a baseline scan')
     print('    -T                max time in minutes to wait for ZAP to start and the passive scan to run')
     print('    -U user           username to use for authenticated scans - must be defined in the given context file')
-    print('    -O                the hostname to override in the (remote) OpenAPI spec')
+    print('    -O                the hostname or URL to override in the (remote) OpenAPI spec')
     print('    -z zap_options    ZAP command line options e.g. -z "-config aaa=bbb -config ccc=ddd"')
     print('    --hook            path to python file that define your custom hooks')
     print('    --schema          GraphQL schema location, URL or file, e.g. https://www.example.com/schema.graphqls')
@@ -418,9 +419,14 @@ def main(argv):
                 logging.debug('Import OpenAPI URL ' + target_url)
                 res = zap.openapi.import_url(target, host_override)
                 urls = zap.core.urls()
+
                 if host_override:
-                    target = urljoin(target_url, '//' + host_override)
-                    logging.info('Using host override, new target: {0}'.format(target))
+                    if urlparse(host_override).scheme:
+                        target = host_override
+                    else:
+                        target = urljoin(target_url, '//' + host_override)
+                    logging.info(
+                        'Using override, new target: {0}'.format(target))
             else:
                 logging.debug('Import OpenAPI File ' + target_file)
                 res = zap.openapi.import_file(target_file)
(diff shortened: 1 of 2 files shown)


kingthorin:

cerrussell:
Yeah I know I fixed.

docker/zap-api-scan.py
```diff
@@ -418,9 +419,14 @@ def main(argv):
...
+
                 if host_override:
-                    target = urljoin(target_url, '//' + host_override)
-                    logging.info('Using host override, new target: {0}'.format(target))
+                    if urlparse(host_override).scheme:
+                        target = host_override
+                    else:
+                        target = urljoin(target_url, '//' + host_override)
+                    logging.info(
+                        'Using host override, new target: {0}'.format(target))
```
kingthorin, line 429: In this log entry maybe remove “host”, and just let it be override? 🤷‍♂️

This may not get a second set of eyes until Monday so don’t feel like you have to rush, it’s fine to wait for agreement or disagreement 😉

cerrussell: @kingthorin Gotcha, I'll check again Monday/Tuesday. Thanks!

thc202:
Probably better to update the description of `-O` to no longer be hostname specific.

cerrussell:
@thc202 So change to `the hostname or URL to override in the (remote) OpenAPI spec`? I don't want to give the impression that it has to be a URL, since it doesn't.

thc202:
That looks good.

cerrussell:
Updated.

thc202:
There's only this one pending https://github.com/zaproxy/zaproxy/pull/8313#discussion_r1460120085 

The date of the changelog entry could be updated to the current day.

docker/zap-api-scan.py
kingthorin: I take the thumbs up to mean they agree 😁

cerrussell:
@thc202 @kingthorin Updated.

kingthorin (approved):
LGTM

thc202 (approved):

thc202:
Thank you!

kingthorin:
Do we want the GUI and API to behave similarly?

thc202:
They already do.

kingthorin:
Perfecto

PR #8321, opened 2024-01-25
Update zap.sh script to get memory usage in containers

yannickvr:
Currently when hard limits are configured on memory in a containerized environment, zap.sh gets the host memory to calculate the memory available to Java, causing unexpected OOM errors.

Since all ZAP container images are given the environment var `IS_CONTAINERIZED`, i think this is currently the best method to detect if we are containerized, and thus use a different method of determining the memory available.

This has been tested in Docker Desktop and Amazon ECS Fargate and works properly. No limits or soft limits will return a linux default value ( 9223372036854771712 ), so there's a fallback to the current method.

### DIFF
--- a/zap/src/main/dist/zap.sh
+++ b/zap/src/main/dist/zap.sh
@@ -65,7 +65,24 @@ if [ -f "$JVMPROPS" ]; then
   # Local jvm properties file present
   JMEM=$(head -1 "$JVMPROPS")
 elif [ "$OS" = "Linux" ]; then
-  MEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    HOSTMEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    if [ "$IS_CONTAINERIZED" = "true" ]; then
+        # Check if memory.stat exists so it can be used to determine the memory limit
+        if [ -f /sys/fs/cgroup/memory/memory.stat ]; then
+            # If we are running in a container, we need to use the cgroup memory limit
+            MEMLIMIT=$(grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat | cut -d' ' -f2)
+            # If the cgroup memory limit is not set, use the host memory
+            if [ "$MEMLIMIT" -eq "9223372036854771712" ]; then
+                MEM=$HOSTMEM
+            else
+                MEM=$(expr $MEMLIMIT / 1024 / 1024)
+            fi
+        else
+            MEM=$HOSTMEM
+        fi
+    else
+        MEM=$HOSTMEM
+    fi
 elif [ "$OS" = "Darwin" ]; then
   MEM=$(system_profiler SPMemoryDataType | sed -n -e 's/.*Size: \([0-9]\{1,\}\) GB/\1/p' | awk '{s+=$0} END {print s*1024}')
   if [ "$MEM" = "0" ]; then
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2024-09-13
+- Update the zap.sh script to use the cgroup memory limit when the IS_CONTAINERIZED environment variable is set to "true".
+
 ### 2024-08-30
 - Updated the API-Minimal scan policy.
 
--- a/docker/Dockerfile-bare
+++ b/docker/Dockerfile-bare
@@ -44,5 +44,6 @@ ENV PATH=$JAVA_HOME/bin:/zap/:$PATH
 ENV ZAP_PATH=/zap/zap.sh
 ENV HOME=/home/zap/
 ENV ZAP_PORT=8080
+ENV IS_CONTAINERIZED=true
 
 HEALTHCHECK CMD curl --silent --output /dev/null --fail http://localhost:$ZAP_PORT/ || exit 1


zap/src/main/dist/zap.sh
```diff
@@ -65,7 +65,17 @@ if [ -f "$JVMPROPS" ]; then
   # Local jvm properties file present
   JMEM=$(head -1 "$JVMPROPS")
 elif [ "$OS" = "Linux" ]; then
-  MEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    if [ "$IS_CONTAINERIZED" = "true" ]; then
+    # If we are running in a container, we need to use the cgroup memory limit
+        MEMLIMIT=$(grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat | cut -d' ' -f2)
+        MEM=$(expr $MEMLIMIT  / 1024 / 1024)
```
ricekot, line 71:
```suggestion
        MEM=$(expr $MEMLIMIT / 1024 / 1024)
```

zap/src/main/dist/zap.sh
```diff
@@ -65,7 +65,17 @@ if [ -f "$JVMPROPS" ]; then
   # Local jvm properties file present
   JMEM=$(head -1 "$JVMPROPS")
 elif [ "$OS" = "Linux" ]; then
-  MEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    if [ "$IS_CONTAINERIZED" = "true" ]; then
+    # If we are running in a container, we need to use the cgroup memory limit
```
ricekot, line 69: Indent the comment?

ricekot (commented):
This does not work for the bare image (https://github.com/zaproxy/zaproxy/blob/main/docker/Dockerfile-bare), because the `IS_CONTAINERIZED` environment variable is not set there, but that can be changed easily.

The `9223372036854771712` value would not work for non 64-bit hosts, but I guess that should be okay since we only build our images for 64-bit architectures at the moment (amd64 and aarch64).

zap/src/main/dist/zap.sh
```diff
@@ -65,7 +65,17 @@ if [ -f "$JVMPROPS" ]; then
...
+        MEMLIMIT=$(grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat | cut -d' ' -f2)
+        MEM=$(expr $MEMLIMIT  / 1024 / 1024)
+        # If the cgroup memory limit is not set, use the host memory
+        if [ "$MEMLIMIT" -eq "9223372036854771712" ]; then
+            MEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
```
thc202, line 74: It would also be good if this was not duplicated.

thc202:
@yannickvr are you going to address the comments?

yannickvr:
> @yannickvr are you going to address the comments?

My bad, totally forgot this issue. I'll take it up later this week.

yannickvr:

zap/src/main/dist/zap.sh
```diff
@@ -65,7 +65,18 @@ if [ -f "$JVMPROPS" ]; then
   # Local jvm properties file present
   JMEM=$(head -1 "$JVMPROPS")
 elif [ "$OS" = "Linux" ]; then
-  MEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    HOSTMEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
+    if [ "$IS_CONTAINERIZED" = "true" ]; then
+        # If we are running in a container, we need to use the cgroup memory limit
+        MEMLIMIT=$(grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat | cut -d' ' -f2)
```
thc202, line 71: Is there a typo here? Checking the ZAP Docker images the file is `/sys/fs/cgroup/memory.stat` (i.e. without the memory dir).

yannickvr: @thc202 that's odd, this is a local test:

```
yannickvr@yannickvr:~/zaproxy$ docker run -it softwaresecurityproject/zap-stable /bin/bash
zap@52d82e516a03:/zap$ grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat
hierarchical_memory_limit 9223372036854771712
yannickvr@yannickvr:~/zaproxy$ docker run -m=2048m -it softwaresecurityproject/zap-stable /bin/bash
zap@b56cd3347b54:/zap$ grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat
hierarchical_memory_limit 2147483648
```

yannickvr: I guess this is exposed by the OS or something that starts the container. I think it's right that it's not in the container image; it's only available during runtime.

thc202: Right, I was checking in the running container not the contents of the image, e.g.:
```
➜  ~ docker run -it zaproxy/zap-nightly:latest bash
zap@b5c36f33e445:/zap$ grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat
grep: /sys/fs/cgroup/memory/memory.stat: No such file or directory
```
(Same for `softwaresecurityproject/zap-stable`, also when setting the memory.)

If this depends on the host we should have an additional check that the file really exists (IMO).

yannickvr: Agreed. Just out of interest: what OS are you testing on? If we're adding a file check it makes sense to see if we can find how to find the memory usage in your scenario as well

thc202: Using macOS Sonoma version 14.3.1 (23D60), with Docker version 26.0.0, build 2ae903e.

yannickvr: Alright, I'll work on the file check. I don't have access to an Apple system currently. Could you check if this is available?

```
/sys/fs/cgroup/memory/memory.limit_in_bytes
```

thc202: That one does not exist:
```output
zap@69b5e3df49dc:/zap$ /sys/fs/cgroup/memory/memory.limit_in_bytes
bash: /sys/fs/cgroup/memory/memory.limit_in_bytes: No such file or directory
```

There are these:
```output
zap@69b5e3df49dc:/zap$ ls -la /sys/fs/cgroup/
total 0
drwxr-xr-x  2 root root 0 May  6 16:21 .
drwxr-xr-x 13 root root 0 May  6 16:21 ..
-r--r--r--  1 root root 0 May  6 16:21 cgroup.controllers
-r--r--r--  1 root root 0 May  6 16:21 cgroup.events
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.freeze
--w-------  1 root root 0 May  6 16:21 cgroup.kill
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.max.depth
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.max.descendants
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.procs
-r--r--r--  1 root root 0 May  6 16:21 cgroup.stat
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.subtree_control
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.threads
-rw-r--r--  1 root root 0 May  6 16:21 cgroup.type
-rw-r--r--  1 root root 0 May  6 16:21 cpu.idle
-rw-r--r--  1 root root 0 May  6 16:21 cpu.max
-rw-r--r--  1 root root 0 May  6 16:21 cpu.max.burst
-r--r--r--  1 root root 0 May  6 16:21 cpu.stat
-r--r--r--  1 root root 0 May  6 16:21 cpu.stat.local
-rw-r--r--  1 root root 0 May  6 16:21 cpu.weight
-rw-r--r--  1 root root 0 May  6 16:21 cpu.weight.nice
-rw-r--r--  1 root root 0 May  6 16:21 cpuset.cpus
-r--r--r--  1 root root 0 May  6 16:21 cpuset.cpus.effective
-rw-r--r--  1 root root 0 May  6 16:21 cpuset.cpus.partition
-rw-r--r--  1 root root 0 May  6 16:21 cpuset.mems
-r--r--r--  1 root root 0 May  6 16:21 cpuset.mems.effective
-r--r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.current
-r--r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.events
-r--r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.events.local
-rw-r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.max
-r--r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.numa_stat
-r--r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.rsvd.current
-rw-r--r--  1 root root 0 May  6 16:21 hugetlb.2MB.rsvd.max
-rw-r--r--  1 root root 0 May  6 16:21 io.latency
-rw-r--r--  1 root root 0 May  6 16:21 io.max
-r--r--r--  1 root root 0 May  6 16:21 io.stat
-r--r--r--  1 root root 0 May  6 16:21 memory.current
-r--r--r--  1 root root 0 May  6 16:21 memory.events
-r--r--r--  1 root root 0 May  6 16:21 memory.events.local
-rw-r--r--  1 root root 0 May  6 16:21 memory.high
-rw-r--r--  1 root root 0 May  6 16:21 memory.low
-rw-r--r--  1 root root 0 May  6 16:21 memory.max
-rw-r--r--  1 root root 0 May  6 16:21 memory.min
-rw-r--r--  1 root root 0 May  6 16:21 memory.oom.group
-r--r--r--  1 root root 0 May  6 16:21 memory.peak
--w-------  1 root root 0 May  6 16:21 memory.reclaim
-r--r--r--  1 root root 0 May  6 16:21 memory.stat
-r--r--r--  1 root root 0 May  6 16:21 memory.swap.current
-r--r--r--  1 root root 0 May  6 16:21 memory.swap.events
-rw-r--r--  1 root root 0 May  6 16:21 memory.swap.high
-rw-r--r--  1 root root 0 May  6 16:21 memory.swap.max
-r--r--r--  1 root root 0 May  6 16:21 memory.swap.peak
-r--r--r--  1 root root 0 May  6 16:21 pids.current
-r--r--r--  1 root root 0 May  6 16:21 pids.events
-rw-r--r--  1 root root 0 May  6 16:21 pids.max
-r--r--r--  1 root root 0 May  6 16:21 pids.peak
-r--r--r--  1 root root 0 May  6 16:21 rdma.current
-rw-r--r--  1 root root 0 May  6 16:21 rdma.max
zap@69b5e3df49dc:/zap$ 
```

yannickvr: Thanks, that's completely different from anything linux. As memory limits are typically a production thing i've opted for testing if /sys/fs/cgroup/memory/memory.stat exists, that should suffice for now.

yannickvr:
I've updated this PR with a test to see if `/sys/fs/cgroup/memory/memory.stat` exists and then get the memory from there. This has been tested successfully on recent ubuntu/debian/amazonlinux environments. It does not work on MacOS, but as this fix is primarily for a production issue it should be fine.

zap/src/main/dist/zap.sh
```diff
@@ -67,11 +67,16 @@ if [ -f "$JVMPROPS" ]; then
 elif [ "$OS" = "Linux" ]; then
     HOSTMEM=$(expr $(sed -n 's/MemTotal:[ ]\{1,\}\([0-9]\{1,\}\) kB/\1/p' /proc/meminfo) / 1024)
     if [ "$IS_CONTAINERIZED" = "true" ]; then
-        # If we are running in a container, we need to use the cgroup memory limit
-        MEMLIMIT=$(grep hierarchical_memory_limit /sys/fs/cgroup/memory/memory.stat | cut -d' ' -f2)
-        MEM=$(expr $MEMLIMIT / 1024 / 1024)
-        # If the cgroup memory limit is not set, use the host memory
-        if [ "$MEMLIMIT" -eq "9223372036854771712" ]; then
+        # Check if memory.stat exists to determine if we are running in a container
```
thc202, line 70: This comment is not correct, that's the previous check.

yannickvr: You're right, I would suggest something like "Check if memory.stat exists so it can be used to determine the memory limit" ?

thc202: Looks good.

zap/src/main/dist/zap.sh
thc202: For the record this seems to be caused by different version of cgroup (i.e. v2 instead of v1), fwiw I'm also seeing this version in Xubuntu 24. Thanks to @ricekot for the details.

Ref: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html

ricekot:
@yannickvr would you be able to address the above review comment? I'd also be happy to commit that if it's okay with you. Also, there are conflicts in `docker/Dockerfile-bare` that need to be resolved.

yannickvr:
> @yannickvr would you be able to address the above review comment? I'd also be happy to commit that if it's okay with you. Also, there are conflicts in `docker/Dockerfile-bare` that need to be resolved.

I'm a bit busy with other priorities, so feel free to finish this up!

ricekot:
Addressed review comments and updated the changelog.

thc202 (approved):

thc202:
Thank you @ricekot and @yannickvr!

psiinon (approved):

PR #8346, opened 2024-02-20
Itests: add missing MariaDB

psiinon:
Failing as per https://github.com/zaproxy/zaproxy/actions/runs/7953280442/job/21709056231

At least one of the AF auth tests are still failing, also looking at that, but it will hopefully address in another PR...


thc202 (approved):

thc202:
Thank you!

psiinon:
Thats ok, I didnt think of it either.
But .. it does raise the question as to whether we should load/save "all" tech as blank?
Right now it looks like existing contexts that previously included all tech now wont target MoriaDB :/
OK, in this case they probably will as its the same as MySQL .. but you get the idea?

PR #8362, opened 2024-02-26
Itests: use autodetect instead of poll

psiinon:
The poll option was for 2.12 compatibility.
That doesnt seem to work now, and in any case we dont need to support it.


thc202:
Thank you!

kingthorin (approved):

PR #8389, opened 2024-03-06
Attempt at fixing support for IPv6 reference URIs

FelixMartel:


kingthorin (changes requested):
This is no longer the right place to tackle this. Instead it would be the network add-on in the zap-extensions repo.

thc202:
It would still be here for as long as we are using the URI class.

FelixMartel:
Opened this hastily. Will rework and reopen then.

PR #8390, opened 2024-03-07
Prevent encoding of [] in the host part of IPv6 reference URIs

FelixMartel:
Currently proxying IPv6 reference URIs is not possible. This is because `HttpRequestHeader.parseURI` first attempts to repair URIs containing unencoded characters by encoding them before calling `new URI` with `encoded=true`. This includes the `[]` characters that are part of the authority section and so `http://[::1]/` becomes `http://%5B::1%5C`.
Here I want to avoid having to add parsing of the authority section to HttpRequestHeader while keeping the URI repairing behavior the same (would be different if using `encoded=false` ). First call `new URI` with `encoded=false` if this was an IPv6 reference then apply repairing to the rest of URI. Otherwise repair the whole URI. Finally call `new URI` with `encoded=true`.

### DIFF
--- a/zap/src/main/java/org/apache/commons/httpclient/URI.java
+++ b/zap/src/main/java/org/apache/commons/httpclient/URI.java
@@ -2240,9 +2240,7 @@ public class URI implements Cloneable, Comparable<Object>, Serializable {
                 next++;
             }
             // In IPv6reference, '[', ']' should be excluded
-            _host = (escaped) ? original.substring(from, next).toCharArray() 
-                : encode(original.substring(from, next), allowed_IPv6reference,
-                        charset);
+            _host = original.substring(from, next).toCharArray();
             // Set flag
             _is_IPv6reference = true;
         } else { // only for !_is_IPv6reference
--- a/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
+++ b/zap/src/main/java/org/parosproxy/paros/network/HttpRequestHeader.java
@@ -462,7 +462,22 @@ public class HttpRequestHeader extends HttpHeader {
             mUri = URI.fromAuthority(sUri);
 
         } else {
-            mUri = parseURI(sUri);
+
+            try {
+                mUri = new URI(sUri, false);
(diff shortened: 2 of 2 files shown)


FelixMartel:

thc202:
Better accept the CLA before continuing with more work, without that the changes will not be merged.

thc202:
This will require tests (a lot of them).

FelixMartel:

FelixMartel:
recheck

zap/src/test/java/org/parosproxy/paros/network/HttpRequestHeaderUnitTest.java
```diff
@@ -342,6 +342,24 @@ private static Stream<Arguments> trueTestUrls(String... exts) {
...
+    @MethodSource("ipv6referenceUrlHostPairs")
+    void extractIpv6refereceHost(String url, String host) throws Exception {
+        // Given
+        HttpRequestHeader header = new HttpRequestHeader("GET " + url + " HTTP/1.1\r\n\r\n");
+        // When / Then
+        assertThat(header.getURI().getHost(), is(equalTo(host)));
+    }
+
+    private static Stream<Arguments> ipv6referenceUrlHostPairs() {
+        List<Arguments> urls = new ArrayList<>();
```
kingthorin, line 355: Could use `List.of`

kingthorin: Or even Stream.of I think, should be other examples in the repo

kingthorin:
Seems to have broken binary compatability

kingthorin:
Does <password>:<username>@<host> parse as expected if password has square brackets?

FelixMartel:
Should behave the same as it does now when given asdf:[]@example.com. In practice I don't think clients supply credentials there.

kingthorin:
They "shouldn't", but it's 'allowed' by spec and browsers so it'd be good to add a test just to ensure that nothing breaks in the future, for my two cents.

Issue #8391, opened 2024-03-09
Sign Docker image with Cosign


explsd:
### Is your feature request related to a problem? Please describe.

In order to secure the supply chain, cluster operators are starting to use tools like [Ratify](https://ratify.dev/) to verify images used on their clusters. 

### Describe the solution you'd like

In order to the ZAP container image to be able to pass signature verification, the docker images should be signed. This can simply be done using [Cosign](https://docs.sigstore.dev/signing/quickstart/) and keyless signing during the build of the image in GitHub Actions.

### Describe alternatives you've considered

The image could also be signed using a private key, but it would take additional work to manage that private key.

### Screenshots

![cosign](https://github.com/zaproxy/zaproxy/assets/76909177/5c5ed4dc-4fec-4ca6-903c-49d7e4544356)
![ratify](https://github.com/zaproxy/zaproxy/assets/76909177/15f96f8b-0852-448f-8a46-ba20261bc303)

### Would you like to help fix this issue?

- [X] Yes

PR #8392, opened 2024-03-09
Sign nightly docker image with cosign keyless

explsd:
This PR updates the ZAP nightly image build to include signing of the image with cosign using the identity of the pipeline. The signature is then stored in the zaproxy github registry and can be validated using [Sigstore's infrastructure](https://www.sigstore.dev/how-it-works) since the signing process will add a record if the signing to their transparency log.

#8391

### DIFF
--- a/.github/workflows/release-live-docker.yml
+++ b/.github/workflows/release-live-docker.yml
@@ -9,6 +9,10 @@ jobs:
   publish:
     name: Build and Push Docker Image
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


explsd:

explsd:
recheck

thc202:
We are going to sign with our own key.

psiinon:
Thanks for this PR, but as @thc202 has indicated we are already working on this, and that work covers more than just the docker images.

Issue #8399, opened 2024-03-13
Zap message history tag not cover all json scenarios 


roisec:
### Is your feature request related to a problem? Please describe.

Hi 
I encourage some use cases that http request contains json requests, and there is no tag on the relevant message.
I think the tagging mechanism only works on HTTP responses.

### Describe the solution you'd like

Cover all the http requests with json 

### Describe alternatives you've considered

Filter out by myself .

### Would you like to help fix this issue?

- [ ] Yes

thc202:
The tagging mechanism works on requests and responses but the default tag is only for responses (as the name indicates `response_json`).

roisec:
I saw in history some HTTP requests and responses without the JSON tag, but they contained the header content-type: application/JSON.
Would you happen to know why?

kingthorin:
Tagging is based on passive scan which uses a queue, so they just may not have been processed/tagged yet.

kingthorin:
Correction the patterns are case insensitive.

kingthorin:
Should the tag also cover other JSON'ish types?
https://github.com/zaproxy/zaproxy/blob/02b5c4f031e8cd6a741fa023310b68dd47d4dc7d/docker/scripts/scripts/httpsender/Alert_on_Unexpected_Content_Types.js#L12-L25

kingthorin:
@psiinon @thc202 @ricekot [☝](https://github.com/zaproxy/zaproxy/issues/8399#issuecomment-2013304425) thoughts?

kingthorin:
This was discussed via IRC. The decision was to not modify the existing pattern at this point. Others can be added, initially disabled.

Without further evidence of something actually misbehaving this issue will be closed.

roisec:
Hi 
I have several examples of HTTP responses with content-type: application/json, and there is no JSON tag on the request in ZAP.

kingthorin:
That's not evidence. Unless you can provide the details to back it up there's nothing we can do for you.

roisec:
![image](https://github.com/zaproxy/zaproxy/assets/68992144/b8ed2d2e-ca1d-4599-bcc3-b6db0337ccf7)
In the following screenshots you are able to see some requests tagged by JSON.
When I send the same request again manually for this use case it works fine.
![image](https://github.com/zaproxy/zaproxy/assets/68992144/f0ea4cf8-2dfb-4535-a6ca-bcead3f45920)

kingthorin:
[...]

PR #8412, opened 2024-03-21
regex tag scanner: Add initial test for detection/tagging of json responses

kingthorin:
Related to zaproxy/zaproxy#8399

https://github.com/zaproxy/zaproxy/blob/02b5c4f031e8cd6a741fa023310b68dd47d4dc7d/zap/src/main/resources/org/zaproxy/zap/resources/config.xml#L116-L125

### DIFF
--- a/zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/DefaultRegexAutoTagScannerTest.java
+++ b/zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/DefaultRegexAutoTagScannerTest.java
@@ -19,18 +19,35 @@
  */
 package org.zaproxy.zap.extension.pscan.scanner;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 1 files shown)


zap/src/test/java/org/zaproxy/zap/extension/pscan/scanner/RegexAutoTagScannerUnitTest.java
```diff
@@ -110,4 +135,43 @@ void shouldNotCountWhenDisabledThoughBodyContainsMatch()
...
+                return tagRule;
+            }
+        }
+        return null;
+    }
+
+    @ParameterizedTest(name = "{0}")
+    @ValueSource(
+            strings = {
+                "application/json",
```
kingthorin, line 151:
```suggestion
                "application/JSON",
```

psiinon:
This will only apply to new installs right?
We will need to update existing config files as part of the upgrade?

kingthorin:
If the pattern is changed we would have to discuss doing that. For now this is just to get a base for testing.

thc202 (approved):

thc202:
Thank you!

Issue #8393, opened 2024-03-10
API search messages by filter tag


roisec:
### Is your feature request related to a problem? Please describe.

No, I would like to search by tag.

### Describe the solution you'd like

Add the filter by tag capabilities for 
search/view/urlsByUrlRegex/

### Describe alternatives you've considered

Now, I need to filter it by myself.

### Would you like to help fix this issue?

- [ ] Yes

PR #8422, opened 2024-03-30
search: Support using Tags as a search item

kingthorin:
- ExtensionSearch > Add new Type.
- Messages.properties > Add resources needed to support the GUI or API functionality.
- SearchAPI > Add handling for urls, messages, and har by history Tag.
- SearchPanel > Adjust insets, and add Tag as a combobox entry, plus handling for enablement of the inverse checkbox.
- SearchThread > Add the actual search functionality.

Fixes zaproxy/zaproxy#8393


kingthorin:
Will update https://github.com/zaproxy/zap-core-help/blob/main/addOns/help/src/main/javahelp/contents/ui/tabs/search.html after this is merged.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
```diff
@@ -270,6 +270,22 @@ private void search() {
                             }
                         }
                     }
+                    if (Type.Tag.equals(reqType)) {
+                        // Tags are based on response so set SearchMatch.Location as resp head
```
thc202, line 274: Tags apply to the whole message not just the response.

kingthorin: So it could be set to `null`, I'll have to handle that in the highlighting. But technically a tag could have multiple regexes in different places (req/resp, head/body), and there's no `Alll` location in the enum. (Though I guess I could add one 🤷‍♂️)

kingthorin: I guess we could do a combination of the two, take the tag string iterate through the actual tags check which field has a regex and choose the Location from that. If more than one then pass null.

It's more change than I was thinking but probably not horrible.

kingthorin:
Reworked and fixed highlighting.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
thc202: It was about the comment not being accurate, not the lack or incorrect highlight.

thc202: I don't oppose trying to guess the highlight but we should not ignore tags just because we can't highlight them, that seems wrong and the user will be left wondering why the message is tagged with something and it's still not shown in the search results.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
```diff
@@ -33,6 +33,8 @@
 import org.parosproxy.paros.model.Session;
 import org.parosproxy.paros.network.HttpMalformedHeaderException;
 import org.parosproxy.paros.network.HttpMessage;
+import org.zaproxy.zap.extension.pscan.PassiveScanParam;
+import org.zaproxy.zap.extension.pscan.scanner.RegexAutoTagScanner;
```
thc202, line 37: This should be pluggable, if/when the passive scanner is moved to the add-on it should not require further changes other than call the methods from the add-on.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
kingthorin: Oh, oh well. I removed the comment when I started deriving the location from the tag scanner details.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
kingthorin: I'm not sure how I'd do that because we need to be able to get the details from the "rule". Unless I close this for now and re-implement after the move so that it only uses add-on classes (which is totally fine with me).

It could be done via a custom search like the fuzzer I guess, but I haven't dug into that.

thc202: The search will need changes either way, since tags are part of the message I don't think the passive scanner is a requirement to search for tags (might be for the highlight but that's secondary). You just need to allow to set something that tells the highlight from/for a specific tag.

kingthorin: If we simply don't highlight (which is totally fine from my point of view), then there wouldn't be a need to rely on the 'rules'. Shall I do away with highlighting for now?

thc202: Sounds good to me.

kingthorin:
Simplified, and less "tangled": No longer attempts to highlight the tag matches in the messages.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
```diff
@@ -270,6 +270,16 @@ private void search() {
                             }
                         }
                     }
+                    if (Type.Tag.equals(reqType)) {
+                        for (String tag : message.getHistoryRef().getTags()) {
+                            matcher = pattern.matcher(tag);
+                            while (matcher.matches() && !pcc.allMatchesProcessed()) {
```
thc202, line 276: - The use of `matches` is a deviation from all other search types, I'd use `find` here too.
 - The loop is not necessary (there's only a single tag being checked in each `for` iteration), an if statement is enough.
 - Should break out of the `for` loop when all matches are processed (no point iterating them all if we are not going to add them).

kingthorin:
Done, done, & done.

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
thc202: The last point was not addressed yet.

thc202: It's still looping them if all matches were already processed, hmm, my initial suggestion is also not the best behaviour, adding the check to the if of the Tag type check is though.

thc202: Move `!pcc.allMatchesProcessed()` to right hand of `Type.Tag.equals(reqType)`.

kingthorin:
I think I've addressed everything now. I guess the only thing outstanding is whether I should add a separator so that it's clear it isn't included in "All"

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
```diff
@@ -270,6 +270,16 @@ private void search() {
                             }
                         }
                     }
+                    if (Type.Tag.equals(reqType) && !pcc.allMatchesProcessed()) {
+                        for (String tag : message.getHistoryRef().getTags()) {
+                            matcher = pattern.matcher(tag);
+                            if (matcher.find()) {
+                                notifyMatchFound(
+                                        currentRecordId, matcher.group(), message, null, 0, 0);
```
thc202, line 278: I'd suggest the name the tag instead of the match, if you do a partial search it only shows part of it, sometimes it's hard to tell what the actual tag is.

thc202 (approved):

thc202:
Thank you!

zap/src/main/java/org/zaproxy/zap/extension/search/SearchThread.java
kingthorin: Ok, I'll switch that shortly.

kingthorin:
Swtiched `matcher.group()` ➡ `tag`

psiinon (approved):

Issue #7959, opened 2023-07-25
Move Passive Scanner to an add-on


thc202:
Move the Passive Scanner to an add-on. Main reason to move it is that it allows to release enhancements and bug fixes sooner and more often (no longer requires releasing a whole new version of ZAP).

thc202:
The move of the passive scanner was done, in core remains the passive scanner interface (and base class) but a new one can be introduced in time.

PR #8427, opened 2024-04-01
Let passive scan extension load the scan rules

thc202:
No longer have core classes load the passive scan rules which allow for an easier migration of the passive scanner to an add-on.

Part of zaproxy/zaproxy#7959.


zap/src/main/java/org/zaproxy/zap/control/AddOnInstaller.java
```diff
@@ -96,15 +92,13 @@ private AddOnInstaller() {}
      *   <li>Files;
      *   <li>Extensions;
      *   <li>Active scanners;
-     *   <li>Passive scanners.
```
kingthorin, line 99: Move the period up to the previous line?

thc202: Removed the semicolons instead, not really needed.

psiinon (approved):

kingthorin (approved):

PR #8445, opened 2024-04-12
host override and local openapi spec pathing

frozenSolid:
Hello and thanks for all of your amazing work you do on ZAP! 

This Pull request adds support for 2 edge-cases that I encountered related to specifying a local openapi spec file and passing a host name override.

1. When specifying openapi spec as a local file, the host override parameter `-O` was ignored. 
2. When specifying an absolute file path for the openapi spec file with the `-t` argument, the `zap-api-scan.py` is expecting the file to be written to `/zap/` but would be written to `/zap/local/file/absolutepath` 

You can reproduce the error with 
 `python zap-api-scan.py -d -t /local/file/absolutepath/openapi.yml -J zap-api-report.json -f openapi -O api.example.us`

### DIFF
--- a/docker/zap-api-scan.py
+++ b/docker/zap-api-scan.py
@@ -58,6 +58,7 @@ import subprocess
 import sys
 import time
 from datetime import datetime
+from pathlib import Path
 
 from six.moves.urllib.parse import urljoin, urlparse
 from zapv2 import ZAPv2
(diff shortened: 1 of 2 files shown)


frozenSolid:

frozenSolid:
recheck

thc202:
fwiw, the override is not an edge-case, it was never meant to work with the file (since the user is (or should) be able to change it). In any case, the packaged scans are being replaced with Automation Framework, would be better to use that than continue changing the packaged scans.

Issue #7695, opened 2023-01-17
gRPC Support


kingthorin:
### Is your feature request related to a problem? Please describe.

ZAP doesn't currently have gRPC support out of the box.

- https://grpc.io/
- https://en.wikipedia.org/wiki/GRPC
- https://github.com/grpc/grpc (Does point to a Java library that may be on benefit)
- https://cloud.google.com/blog/products/api-management/understanding-grpc-openapi-and-rest-and-when-to-use-them

### Describe the solution you'd like

gRPC requests should already be shown in the History / Sites tree as binary requests.
This means that someone would need to:

- Implement a body view/editor
- Implement a variant
- Add fuzzing support
- Implement a set of ascan rules which specifically target gRPC requests, e.g.
    - changing types
    - trying to call other classes / methods

### Describe alternatives you've considered

N/A

### Screenshots

N/A

### Additional context

N/A

### Would you like to help fix this issue?

- [ ] Yes

kingthorin:
Marking as GSoC-Candidate for the time being. We'll have to discuss further I guess.

kingthorin:
This probably isn't a perfect solution, but it should help getting the view etc sorted:

- https://www.gosecure.net/blog/2021/04/13/creating-a-custom-view-for-websocket-in-zap/

The 3rd party JWT add-on also has views and fuzzing support which may help as examples:

- https://github.com/SasanLabs/owasp-zap-jwt-addon
- https://www.zaproxy.org/blog/2020-09-03-zap-jwt-scanner/

aryangupta701:
Thank you for the resources. I will begin my research on it soon.

adwait-godbole:
Looking forward to contributing to ZAP this year in GSoC, thank you for the resources.

amitpanwar789:
Hi! I'm very interested in participating in this project for GSoC 2024 and would love to discuss it further. Do you prefer I ask my questions here or reach out via email?

psiinon:
@amitpanwar789 great! Here is best for general questions about the project, but you can also reach out to use via zaproxy-admin@googlegroups.com

kingthorin:
Something like: https://github.com/grpc/grpc-java might help, but that's kinda part of the work/design. I don't think anyone in the core team has done much research on this 🤷‍♂️

ArkaprabhaChakraborty:
How about using postman on an gRPC API to get some good UI ideas?

psiinon:
[...]

PR #8452, opened 2024-04-17
Support decode response body through the `Variant`

amitpanwar789:
## Overview
This add decoding response body support of variants

## Related Issues
https://github.com/zaproxy/zaproxy/issues/7695

@thc202 

## Checklist
- [ ] Update help
- [ ] Update changelog
- [x] Run `./gradlew spotlessApply` for code formatting
- [ ] Write tests
- [ ] Check code coverage
- [x] Sign-off commits
- [ ] Squash commits
- [x] Use a descriptive title

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
@@ -46,6 +46,7 @@
 // ZAP: 2023/01/10 Use logger provided by base class.
 package org.parosproxy.paros.core.scanner;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -135,6 +136,12 @@ public abstract class AbstractAppParamPlugin extends AbstractAppPlugin {
                 }
             }
         }
+    }
+
+    @Override
+    protected void sendAndReceive(HttpMessage message) throws IOException {
+        sendAndReceive(message, true);
+        variant.decodeResponseBody(message);
     }
 
     /**
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
@@ -141,6 +141,10 @@ public abstract class AbstractAppParamPlugin extends AbstractAppPlugin {
     @Override
     protected void sendAndReceive(HttpMessage message) throws IOException {
         sendAndReceive(message, true);
+        decodeResponseBody(message);
+    }
+
+    protected void decodeResponseBody(HttpMessage message) {
         variant.decodeResponseBody(message);
     }
 
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
@@ -46,7 +46,6 @@
 // ZAP: 2023/01/10 Use logger provided by base class.
 package org.parosproxy.paros.core.scanner;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -139,11 +138,6 @@ public abstract class AbstractAppParamPlugin extends AbstractAppPlugin {
     }
 
     @Override
-    protected void sendAndReceive(HttpMessage message) throws IOException {
-        sendAndReceive(message, true);
-        decodeResponseBody(message);
-    }
-
     protected void decodeResponseBody(HttpMessage message) {
         variant.decodeResponseBody(message);
     }
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
@@ -102,4 +102,6 @@ public interface Variant {
     default List<String> getTreePath(HttpMessage msg) throws URIException {
         return null;
     }
+
+    default void decodeResponseBody(HttpMessage msg) {}
 }
--- a/zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
+++ b/zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
@@ -103,5 +103,12 @@ public interface Variant {
         return null;
     }
 
-    default void decodeResponseBody(HttpMessage msg) {}
+    /**
+     * Decodes the response body of the given {@code HttpMessage}. for example, to change a binary
+     * payload to clear text to allow to do text/string comparisons of the content.
+     *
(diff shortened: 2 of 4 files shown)


zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -137,6 +138,12 @@ protected void scan(List<NameValuePair> nameValuePairs) {
         }
     }
 
+    @Override
+    protected void sendAndReceive(HttpMessage message) throws IOException {
+        sendAndReceive(message, true);
+        variant.decodeResponseBody(message);
```
thc202, line 144: Please, expose the protected method that allows to decode, otherwise it does not allow implementations to override this behaviour.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -102,4 +102,6 @@ default String getLeafName(String nodeName, HttpMessage msg) {
     default List<String> getTreePath(HttpMessage msg) throws URIException {
         return null;
     }
+
+    default void decodeResponseBody(HttpMessage msg) {}
```
thc202, line 106: Needs JavaDoc and since tag.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
amitpanwar789: ok i will make it this , maybe this will work 
`protected void decodeResponseBody(HttpMessage message) {}`

amitpanwar789: Could you please clarify how to expose the protected decoding method? I'm unsure about implementation. Thanks!

thc202: Like you suggested in a previous comment, it just needs to call `variant.decodeResponseBody(message);` (and have the `sendAndReceive` call the new method).

amitpanwar789: sorry but still didn't get it, yeah sendAndReceive have to call variant.decodeResponseBody, which method i need to expose and what you mean by expose ?

amitpanwar789: are you suggesting to write a new method like so that, subclasses of AbstractAppParamPlugin can use the decodeResponseBody method. 
```
protected void decodeResponseBody(HttpMessage message) {
    variant.decodeResponseBody(message);
}
```

thc202: Yes.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
thc202: It does, I'd also add, "... given {@code HttpMessage}, for example, to change a binary payload to clear text to allow to do text/string comparisons of the content."

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -141,6 +141,10 @@ protected void scan(List<NameValuePair> nameValuePairs) {
     @Override
     protected void sendAndReceive(HttpMessage message) throws IOException {
         sendAndReceive(message, true);
+        decodeResponseBody(message);
+    }
+
+    protected void decodeResponseBody(HttpMessage message) {
```
thc202, line 147: Add JavaDoc as well.

zap/src/main/java/org/parosproxy/paros/core/scanner/AbstractAppParamPlugin.java
```diff
@@ -141,6 +141,10 @@ protected void scan(List<NameValuePair> nameValuePairs) {
     @Override
     protected void sendAndReceive(HttpMessage message) throws IOException {
```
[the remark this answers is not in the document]
amitpanwar789: this  `sendAndReceive(HttpMessage, boolean, boolean)` is implemented in AbstractPlugin, if we call decodeResponseBody from there, we need to pass variant as parameter also, but we don't have it in AbstractPlugin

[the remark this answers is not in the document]
thc202: The implementation of that method in `AbstractPlugin` would do nothing, that's just there to allow other implementations to do something (like `AbstractAppParamPlugin` with the variant).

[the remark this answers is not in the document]
amitpanwar789: ok, but how can we call `decodeResponseBody` from` sendAndReceive(HttpMessage, boolean, boolean)` before `parent.notifyNewMessage(this, message) `, we have moved `decodeResponseBody` in `AbstractPlugin` and `decodeResponseBody` need variant instance to do` variant.decodeResponseBody(message)`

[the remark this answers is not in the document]
thc202: The only class that needs access to the variant is still just `AbstractAppParamPlugin`, by overriding `decodeResponseBody` it can access it.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -102,4 +102,13 @@ default String getLeafName(String nodeName, HttpMessage msg) {
...
     }
+
+    /**
+     * Decodes the response body of the given {@code HttpMessage}. for example, to change a binary
+     * payload to clear text to allow to do text/string comparisons of the content.
+     *
+     * @param msg the HTTP message whose response body will be decoded
+     * @since 2.15.0
+     */
+    void decodeResponseBody(HttpMessage msg);
```
thc202, line 113: Missing the default and empty body.

zap/src/main/java/org/parosproxy/paros/core/scanner/Variant.java
```diff
@@ -110,5 +110,6 @@ default List<String> getTreePath(HttpMessage msg) throws URIException {
      * @param msg the HTTP message whose response body will be decoded
      * @since 2.15.0
      */
-    void decodeResponseBody(HttpMessage msg);
+    default void decodeResponseBody(HttpMessage msg) {}
+    ;
```
thc202, line 114: Remove the empty statement.

thc202:
Please, suppress the binary incompatibility.

psiinon (approved):

thc202 (approved):

thc202:
Thank you!

PR #8454, opened 2024-04-18
Include pluginId in alert events

psiinon:
This means event consumers will be able to reliably identify alerts. 
Right now they only have access to the alert names, with are i18ned.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertEventPublisher.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertEventPublisher.java
@@ -43,6 +43,7 @@ public class AlertEventPublisher implements EventPublisher {
     public static final String HISTORY_REFERENCE_ID = "historyId";
 
     public static final String NAME = "name";
+    public static final String PLUGIN_ID = "pluginId";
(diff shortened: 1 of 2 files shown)


kingthorin:
Would we want this to be alert ref if it exists?

psiinon:
Oh, good suggestion..

psiinon:
Updated to use alertRef instead.

kingthorin:
I haven't dug through the code but does this always have a value? Does it default to pluginId if a specific ref hasn't been set?

psiinon:
Yes, its defaulted to the pluginId for standard rules.
It can be an empty string for user defined alerts.
I've tested it and both tests (alertRef and the name) seem to work fine..

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

kingthorin:
To me if it's about Alerts then it should be the thing that most specifically identifies the alert, but whatever we can always do more later.

thc202:
It definitely depends on the use case, and here the scan rule ID is enough.

Issue #8460, opened 2024-04-29
Incomplete report for template Risk and Confidence HTML


GunoH:
### Describe the bug

When generating a report using the template _Risk and Confidence HTML_, with all _Sections_ enabled, the generated report does not show any details of the alerts found:

![image](https://github.com/zaproxy/zaproxy/assets/915288/1e9bd69c-0823-4ece-9d44-b4ff2e80c56e)
![image](https://github.com/zaproxy/zaproxy/assets/915288/3f07eb74-db1f-4ae7-9c32-7fefbf3dd2fb)

The generated report files are a lot smaller than they used to be (with earlier versions, when the details were included). 

Using ZAP 2.14.0, Report Generation plugin 0.31.0.

Reverting to plugin versions bundled with 2.14.0 (by removing the ~/.ZAP/plugin; this reverts Report Generation to 0.26.0) resolves the issue, but as soon as I upgrade plugins to the latest versions, the issue is back.

### Steps to reproduce the behavior

1. Update to Report Generation 0.31.0
2. Perform active scan
3. Generate report (Risk and Confidence HTML) with all sections enabled.

### Expected behavior

The generated report contains alert details, such as http requests and responses.

### Software versions

ZAP
Version: 2.14.0

Installed Add-ons: [[id=alertFilters, version=20.0.0],
[id=ascanrules, version=65.0.0], [id=authhelper,
version=0.12.0], [id=automation, version=0.39.0],
[id=bruteforce, version=15.0.0], [id=callhome,
version=0.11.0], [id=commonlib, version=1.24.0],
[id=database, version=0.3.0], [id=diff, version=14.0.0],
[id=directorylistv1, version=7.0.0], [id=domxss,
version=18.0.0], [id=encoder, version=1.4.0], [id=exim,
version=0.8.0], [id=formhandler, version=6.5.0], [id=fuzz,
version=13.12.0], [id=gettingStarted, version=16.0.0],
[id=graaljs, version=0.6.0], [id=graphql, version=0.23.0],
[id=help, version=17.0.0], [id=hud, version=0.18.0],
[id=invoke, version=14.0.0], [id=network, version=0.15.0],
[id=oast, version=0.17.0], [id=onlineMenu, version=12.0.0],
[id=openapi, version=39.0.0], [id=postman, version=0.3.0],
[id=pscanrules, version=57.0.0], [id=quickstart,
version=46.0.0], [id=replacer, version=16.0.0], [id=reports,
[...]

PR #8469, opened 2024-05-01
Fix NPE in `AbstractAppParamPlugin`

thc202:
Check that the `variant` is non-null before attempting to decode the body, as it can be if the extending class sends requests before (indirectly) initializing the `variant`.

From logs in #8460.


psiinon (approved):

ricekot (approved):
Thanks!

PR #8478, opened 2024-05-08
Update Log4j 2

thc202:
Update to latest version.
Add required libraries at compile time as Log4j 2 Core library leaks annotations from other libraries.


kingthorin (approved):
Thanks!

ricekot (approved):
Thank you!

PR #8494, opened 2024-05-17
Docker: Rename test image

psiinon:
this is a significantly different image so I think it should have a different name, not just a different tag.

Does it also need a changelog entry?
I was thinking not...


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #8541, opened 2024-07-02
Remove stats passive scan rule

thc202:
Let the `pscan` add-on have the scan rule.

Part of #7959.

---
Depends on zaproxy/zap-extensions#5555.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
@@ -55,7 +55,6 @@ import org.zaproxy.zap.control.ExtensionFactory;
 import org.zaproxy.zap.extension.AddOnInstallationStatusListener;
 import org.zaproxy.zap.extension.alert.ExtensionAlert;
 import org.zaproxy.zap.extension.pscan.scanner.RegexAutoTagScanner;
-import org.zaproxy.zap.extension.pscan.scanner.StatsPassiveScanner;
 import org.zaproxy.zap.extension.script.ExtensionScript;
 import org.zaproxy.zap.extension.script.ScriptType;
 import org.zaproxy.zap.view.ScanStatus;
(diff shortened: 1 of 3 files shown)


zap/src/main/java/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
```diff
@@ -310,20 +309,6 @@ protected PassiveScannerList getPassiveScannerList() {
 
             // Read from the configs
             scannerList.setAutoTagScanners(getPassiveScanParam().getAutoTagScanners());
-
-            // Load the  'switchable' plugins
-            List<PluginPassiveScanner> listTest =
-                    List.of(new RegexAutoTagScanner(), new StatsPassiveScanner());
```
kingthorin, line 316: Regex is next?

kingthorin: Okay 😃

kingthorin (approved):

zap/src/main/java/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
kingthorin: Yup I see now, it's in the list but then ignored in the loop.

thc202:
No longer WIP.

zap/src/main/java/org/zaproxy/zap/extension/pscan/ExtensionPassiveScan.java
ricekot: Should we deprecate that too?

thc202: Not yet, that will have to be moved along other classes first.

ricekot (approved):
Thanks!

PR #8546, opened 2024-07-05
Prevent adding same add-on's `ResourceBundle`

thc202:
Do not add the `ResourceBundle` found through the extension if it's already being provided by the add-on, which would lead to an error saying it was duplicated.

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/control/ExtensionFactory.java
+++ b/zap/src/main/java/org/zaproxy/zap/control/ExtensionFactory.java
@@ -28,6 +28,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.MissingResourceException;
+import java.util.Objects;
 import java.util.ResourceBundle;
 import java.util.TreeMap;
 import java.util.Vector;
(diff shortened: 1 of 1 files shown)


zap/src/main/java/org/zaproxy/zap/control/ExtensionFactory.java
```diff
@@ -325,6 +326,13 @@ private static void loadMessages(Extension ext) {
         }
     }
 
+    private static boolean isSame(ResourceBundle rb1, ResourceBundle rb2) {
+        if (rb1 == null || rb2 == null) {
+            return false;
+        }
+        return Objects.equals(rb1.getBaseBundleName(), rb2.getBaseBundleName());
```
kingthorin, line 333: Why not string.equals if we are getting names?

kingthorin: I'm not saying it's wrong I'm just curious.

thc202: Technically they might return null.

zap/src/main/java/org/zaproxy/zap/control/ExtensionFactory.java
```diff
@@ -325,6 +326,13 @@ private static void loadMessages(Extension ext) {
         }
     }
 
+    private static boolean isSame(ResourceBundle rb1, ResourceBundle rb2) {
```
ricekot, line 329: Rename to `isSameResourceBundle` for clarity?

ricekot: Or maybe that's clear from the type of the params.

thc202: Renamed, better be clear.

kingthorin (approved):

ricekot (approved):
Thanks!

PR #8547, opened 2024-07-06
Allow moving tabs to new windows

### LINKED ISSUE #2109: Undock Tabs
Hi, 

a feature to undock Tabs for example the Script Console or the History would be nice!
Similar to the behaviour of eclipse.

Best Regards
Martin

ricekot:
Add a context menu item to the tabs in all tabbed panels (`TabbedPanel2`) that moves the tab to a new window. 

Closing the window puts the tab back in its original position.

The list of detached tabs is not currently persisted and tabs must be detached again when ZAP is restarted.

Closes #2109.


zap/src/main/java/org/zaproxy/zap/view/TabbedPanelTab.java
```diff
@@ -229,6 +249,40 @@ public void setEnabled(boolean enabled) {
         }
     }
 
+    private void moveToNewWindow() {
+        int prevTabIndex = parent.indexOfComponent(component);
+        var frame = new DetachedTabFrame(title, component);
```
kingthorin, line 254: Does this still work properly if other tabs are shown/added or hidden/removed (ex: installation of an add-on, changing the pin 📌 options, or actually closing ❌ a tab) while a tab is windowed.

ricekot: Yes, it does. When a tab is windowed, it's removed from the original tabbed pane, so it can only be closed by closing the window, and closing the window brings it back to the tabbed pane. Here's a quick demo :) 

https://github.com/zaproxy/zaproxy/assets/16446369/7dff3c1f-e552-4eb7-a15b-508b751d4142

kingthorin: No I mean back in the mainframe the tabs list can change so the index might end-up wrong?

kingthorin: Ex what if you have request, response, quick start, Neonmarker. Send Neonmarker to a window. Close quick start. Close Neonmarker window (returning it to the tab panel). Etc scenarios like that.

thc202:
There are some issues when detaching and then changing the layout (e.g. tabs go missing in full layout; the window title is not updated; tab still detached and also in the main window).

thc202:
- It's unexpected that when you change the request/response layout they default back to the work panel if they are windows (more noting than asking for changes).
 - If you move all visible tabs of a panel to windows it's left empty, while expected it looks odd, maybe add a message to the panel saying that all visible tabs are windows? (Fine if kept as is.)
 - When you close the window the tab does not get focus, I think it should since you are acting on it.

ricekot:
Addressed comment and updated with the following details :)

1. A warning is displayed if there are any detached request or response tabs and the request/response layout is changed.
   
    <details>
    <summary> Screenshot (click to expand) </summary>
    image
    </details>
    
2. A message is shown if there are no visible tabs. A different message is shown depending on whether any tabs are hidden.
   
    <details>
    <summary> With hidden and detached tabs (click to expand) </summary>
    image
    </details>
    <details>
    <summary> With only detached tabs and no hidden tabs (click to expand) </summary>
    image
    </details>

3. Upon closing a detached tab, it requests to be focused.

thc202:
Re 2., if you change layouts the "no visible tabs" tab behaves as a normal tab.

ricekot:
> Re 2., if you change layouts the "no visible tabs" tab behaves as a normal tab.

Should be fixed now, along with a couple other bugs I found while testing this.

thc202 (approved):

kingthorin (approved):

kingthorin:
Great work!

PR #8565, opened 2024-07-18
Address warnings when building Docker images

thc202:
Address `LegacyKeyValueFormat: "ENV key=value" should be used…`.

### DIFF
--- a/docker/Dockerfile-live
+++ b/docker/Dockerfile-live
@@ -23,7 +23,7 @@ RUN git clone --depth 1 https://github.com/zaproxy/zaproxy.git && \
 WORKDIR /zap
 
 # Setup Webswing
-ENV WEBSWING_VERSION 24.1.1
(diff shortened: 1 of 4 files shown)


psiinon (approved):

docker/Dockerfile-live
```diff
@@ -23,7 +23,7 @@ RUN git clone --depth 1 https://github.com/zaproxy/zaproxy.git && \
 WORKDIR /zap
 
 # Setup Webswing
-ENV WEBSWING_VERSION 24.1.1
+ENV WEBSWING_VERSION=24.1.1
```
kingthorin, line 26: Update to 24.1.4 while changing things?

thc202: Not without checking all is good.

kingthorin (approved):

PR #8566, opened 2024-07-18
Fix auth type isTypeForMethod implementations

psiinon:
Using `instanceof` means that the methods can match auth method types that extend the current class.
This makes it impossible to select the extended auth method.


thc202 (approved):

thc202:
Thank you!

PR #8571, opened 2024-07-25
ITests: continue on error

psiinon:
I'm hoping the action will still fail if one of the jobs fails, but am not sure.
So lets see...


thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #8574, opened 2024-07-25
Itests: fix error check dir path

psiinon:
Hope it will work this time..

Last time it failed with:

docker: Error response from daemon: create .github/workflows/conf: ".github/workflows/conf" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.


kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #8602, opened 2024-08-20
Make some dialog messages copyable.

psiinon:
Related to https://github.com/zaproxy/zap-extensions/pull/5659 - ie makes it easier to copy a URL.
Also makes it easier to copy the full error message for support queries.

Easy way to test - change the path of one of the webdrivers to be invalid and try to launch the relevant browser.
Without this code the message will not be copyable, with it then it will be.

Dialogs using ZapHtmlLabel will still not be copyable as they extend ZapLabel.
But changing that looks like a bigger and potentially more risky change :)

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/View.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/View.java
@@ -95,6 +95,7 @@
 // ZAP: 2022/09/21 Use format specifiers instead of concatenation when logging.
 // ZAP: 2022/09/27 Added dialog methods with ZapHtmlLabel parameter.
 // ZAP: 2023/01/10 Tidy up logger.
+// ZAP: 2024/08/20 Wrap string dialog messages in ZapLabel so that they can be copied.
 package org.parosproxy.paros.view;
 
 import java.awt.Component;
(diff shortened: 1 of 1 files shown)


kingthorin (approved):

zap/src/main/java/org/parosproxy/paros/view/View.java
```diff
@@ -95,6 +95,7 @@
 // ZAP: 2022/09/21 Use format specifiers instead of concatenation when logging.
 // ZAP: 2022/09/27 Added dialog methods with ZapHtmlLabel parameter.
 // ZAP: 2023/01/10 Tidy up logger.
+// ZAP: 2024/08/20 Warp string dialog messages in ZapLabel so that they can be copied.
```
kingthorin, line 98: OOps saw this just as I hit approve.

wrap not warp 🤣

psiinon: Ha! Fixed 😁

kingthorin (approved):

thc202 (approved):

thc202:
Thank you!

PR #8615, opened 2024-08-30
Remove "Nashorn" JavaScript templates

thc202:
These templates are included with GraalVM JavaScript add-on which is now included by default, and should be used instead of Nashorn which will not be available once ZAP starts targeting newer Java version.
Remove test that checked the scripts, no other script remains.


psiinon (approved):

thc202:
Removed the test that was checking the templates, there are no other templates left.

kingthorin (approved):

PR #8623, opened 2024-09-03
Fix typos in comments and log messages

longhutianjie:


zap/src/main/java/org/apache/commons/httpclient/URI.java
```diff
@@ -2838,7 +2838,7 @@ public String getHost() throws URIException {
     // --------------------------------------------------------------- The port
 
     /**
-     * Get the port.  In order to get the specfic default port, the specific
+     * Get the port.  In order to get the specific default port, the specific
```
thc202, line 2841: Lets revert this and leave the forked class with the minimum changes.

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):

PR #8667, opened 2024-10-07
Retain and expose source message ID of alerts

thc202:
Allow the user to know the ID of the source message that caused the alert, as opposed to the ID of the message that has the alert (e.g. evidence) which might not always be the same (e.g. active scans).


psiinon (approved):

kingthorin:
Tests?

kingthorin:
Should it be exposed/settable in AlertBuilder?

thc202:
It already is.

Issue #8375, opened 2024-03-01
Add telemetry to toolbar buttons


njmulsqb:
### Is your feature request related to a problem? Please describe.

N/A

### Describe the solution you'd like

To de-clutter the toolbar ZAP team has decided to add telemetry to the toolbar buttons to know its usage and better decide which buttons receive the most/least usage.

### Describe alternatives you've considered

Google group discussion thread: https://groups.google.com/g/zaproxy-develop/c/pjTLqdEaTlM/m/0gRkYDQTAQAJ

### Would you like to help fix this issue?

- [ ] Yes

psiinon:
To collect stats about something like a menu click you just need to add a line like this in the relevant place:
```
Stats.incCounter("stats.ui.topmenu.open");
```

Note that the stats uploaded are defined in [ExtensionCallHome.StatsPredicate](https://github.com/zaproxy/zap-extensions/blob/main/addOns/callhome/src/main/java/org/zaproxy/addon/callhome/ExtensionCallHome.java)
so that would need to be change to add `|| key.startsWith("stats.ui.")`

PR #8668, opened 2024-10-07
Add main toolbar stats

kingthorin:
- Updated various classes adding stat counters to button/menu actions.
- Added change note to one paros class.

Part of zaproxy/zaproxy#8375

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/MainFrame.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/MainFrame.java
@@ -38,6 +38,7 @@
 // ZAP: 2022/02/09 Remove method call no longer needed.
 // ZAP: 2022/02/26 Remove code deprecated in 2.5.0
 // ZAP: 2022/08/05 Address warns with Java 18 (Issue 7389).
+// ZAP: 2024/10/07 Add stats to main toolbar buttons (Issue 8375).
 package org.parosproxy.paros.view;
 
 import java.awt.CardLayout;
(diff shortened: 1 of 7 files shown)


kingthorin:
QuickStart, reports, and hud will have to be done in other PRs

zap/src/main/java/org/parosproxy/paros/view/MainFrame.java
```diff
@@ -377,6 +379,7 @@ private JButton getShowAllTabsButton() {
                         @Override
                         public void actionPerformed(ActionEvent e) {
                             View.getSingleton().showAllTabs();
+                            Stats.incCounter("stats.ui.toolbar.button.tabs.showall");
```
thc202, line 382: I'd suggest prefix with "maintoolbar", we have other toolbars that could also have stats in the future.

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanelToolbarFactory.java
```diff
@@ -640,6 +640,7 @@ public void actionPerformed(ActionEvent e) {
             setBreakAll(false);
             setBreakRequest(false);
             setBreakResponse(false);
+            Stats.incCounter("stats.ui.toolbar.button.break.continue");
```
thc202, line 643: These actions are also triggered in the Break tab, just saying we'll not know in which toolbar they are actually being used.

kingthorin: I left these as "toolbar" but changed all the others to "maintoolbar".

I tried to find a parent or ancestor method to use in a conditional but didn't have any success/

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

Issue #8234, opened 2023-12-09
False-positive for hidden file CWE 538


epomatti:
### Describe the bug

The spider is reporting `BitKeeper` as a hidden file. It does not exist in the code base and the response is a redirect 301 Moved Permantently.

Confidence is set to low (default).

![image](https://github.com/zaproxy/zaproxy/assets/6530769/57a59acd-3ebd-43f3-ac32-f61efad4182a)

### Steps to reproduce the behavior

1. Run a traditional or AJAX crawler in a web site using default confidence. Return 301.

### Expected behavior

We would expect no alert when there is a redirect as a response.

### Software versions

```
ZAP
Version: 2.14.0

Installed Add-ons: [[id=alertFilters, version=19.0.0],
[id=ascanrules, version=59.0.0], [id=authhelper,
version=0.10.0], [id=automation, version=0.34.0],
[id=bruteforce, version=15.0.0], [id=callhome,
version=0.10.0], [id=commonlib, version=1.20.0],
[id=database, version=0.3.0], [id=diff, version=14.0.0],
[id=directorylistv1, version=7.0.0], [id=domxss,
version=18.0.0], [id=encoder, version=1.4.0], [id=exim,
version=0.8.0], [id=formhandler, version=6.5.0], [id=fuzz,
version=13.12.0], [id=gettingStarted, version=16.0.0],
[id=graaljs, version=0.5.0], [id=graphql, version=0.21.0],
[id=help, version=17.0.0], [id=hud, version=0.18.0],
[id=importurls, version=9.0.0], [id=invoke, version=14.0.0],
[id=network, version=0.13.0], [id=oast, version=0.17.0],
[id=onlineMenu, version=12.0.0], [id=openapi,
version=38.0.0], [id=postman, version=0.2.0],
[id=pscanrules, version=53.0.0], [id=quickstart,
version=43.0.0], [id=replacer, version=16.0.0], [id=reports,
version=0.26.0], [id=requester, version=7.4.0], [id=retest,
version=0.8.0], [id=retire, version=0.28.0], [id=reveal,
version=7.0.0], [id=saverawmessage, version=7.0.0],
[id=savexmlmessage, version=0.3.0], [id=scripts,
version=43.0.0], [id=selenium, version=15.16.0], [id=soap,
version=20.0.0], [id=spider, version=0.7.0], [id=spiderAjax,
version=23.18.0], [id=tips, version=12.0.0],
[...]

PR #8676, opened 2024-10-17
Fix FPs in isSuccess and isPage200

psiinon:
Fix for https://groups.google.com/g/zaproxy-users/c/QCfFo7Nh1sI/m/KABs-yyRAAAJ and https://github.com/zaproxy/zaproxy/issues/8234

The latter was fixed via https://github.com/zaproxy/zap-extensions/pull/5232 but I did not submit the core change at that time 😦  This is that change 😁


kingthorin (approved):

psiinon:
@kingthorin want to check this again?

kingthorin (approved):

kingthorin:
Shall we just do away with the analyser if we aren't going to trust/fix it?

psiinon:
Possibly yes .. but maybe we should investigate it further, and maybe fix it if we can?

thc202:
It's still good at telling (custom) 404s (and is still used for that).

PR #8678, opened 2024-10-18
Std file chooser fixes

psiinon:
Change the standard dialog filechooser fields to be editable (so that the user can paste a value in) and to choose the parent directory if the selected file does not exist - otherwise if defaults to the users default directory.


thc202 (approved):

thc202:
Thank you!

kingthorin (commented):
Extract a method?

psiinon:
Hum .. that might just be possible, and if so well worth doing..

psiinon:
Done

kingthorin (approved):

thc202 (approved):

PR #8685, opened 2024-10-22
Add manual alert add/modify and vuln selection stats

kingthorin:


zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAddDialog.java
```diff
@@ -205,6 +207,7 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
                                     alert.setSource(Alert.Source.MANUAL);
                                     // Raise it
                                     extAlert.alertFound(alert, historyRef);
+                                    Stats.incCounter("stats.ui.alert.dialog.add");
```
thc202, line 210: IMO we should do this in the `ExtensionAlert#alertFound` based on the source, the user can also raise alerts manually through the API (though I guess we could have in both places).

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertViewPanel.java
```diff
@@ -226,6 +227,9 @@ public void actionPerformed(ActionEvent e) {
                                 org.zaproxy.zap.model.Vulnerability v =
                                         getVulnerability((String) alertEditName.getSelectedItem());
                                 if (v != null) {
+                                    Stats.incCounter(
+                                            "stats.ui.alert.panel.vuln.selected." + v.getWascId());
```
thc202, line 231: It would probably be better to check if the name matches a vuln when the alert is about to be raised.

kingthorin: I thought about that but the names are pretty generic. I wouldn't be surprised if a user pasted or typed some of them 🤷‍♂️

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAddDialog.java
kingthorin: I was about to change this, but is it worth knowing the source? (dialog vs api)

kingthorin: /nudge

thc202: Are you suggesting using the same stat in the API? Or not add stats there?

kingthorin: I'm asking if we want to have two separate stats specifically. Or just one (in the suggested location) that covers both scenarios.

thc202: I'm good with two.

kingthorin:
Done

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
```diff
@@ -427,6 +428,7 @@ public ApiResponse handleApiAction(String name, JSONObject params) throws ApiExc
             newAlert.setCweId(cweId);
             newAlert.setWascId(wascId);
             extension.alertFound(newAlert, msg.getHistoryRef());
+            Stats.incCounter("stats.ui.alert.api.add");
```
thc202, line 431: I was going to comment about this having `ui` and checked the stats page to know of a more appropriate key, we already know how many by looking at `stats.api.call`, sorry.

kingthorin: Oh okay, I'll roll this back.

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #8688, opened 2024-10-26
Allow dependabot to update GitHub actions

kingthorin:


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
```
thc202, line 3: Remove the comments please.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
```
thc202, line 6: `"monthly"`

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+    groups:
+      dependencies:
```
thc202, line 8: `gha` or something that's obvious this is updating the GitHub Actions.

kingthorin: Line 3?

kingthorin: Or do you mean for the commit-message or prefix?

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message

thc202: I mean this line, which is the name of the group (and used for commit too).

kingthorin:
Assuming this is what we want elsewhere as well I'll update them after this is merged.

ricekot (approved):

thc202 (approved):

thc202:
Thank you!

PR #8815, opened 2025-01-21
Bugfix/alert view alert counts by risk xml response fix #8814

### LINKED ISSUE #8814: alert/view/alertCountsByRisk fails for response type XML
### Describe the bug:

Navigate to ZAP's API > alert > alertCountsByRisk (url recurse ) > Select XML as the output type, provide your API key, and click the button. 
Observe the resulting error `<Exception code="internal_error" type="exception">Internal Error</Exception>`

Note JSON and HTML responses work fine.

### Steps to reproduce the behavior:

See above.

### Expected behavior:

XML response should be produced as expected.

### Software versions:

Up-to-date DEV or 2.16.0.

### Screenshots:

n/a

### Errors from the zap.log file:

```
909098 [ZAP-IO-Server-1-3] ERROR org.zaproxy.zap.extension.api.API - Failed to convert API response to XML: INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified.
org.w3c.dom.DOMException: INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified.
	at java.xml/com.sun.org.apache.xerces.internal.dom.CoreDocumentImpl.createElement(CoreDocumentImpl.java:627) ~[?:?]
	at org.zaproxy.zap.extension.api.ApiResponseSet.toXML(ApiResponseSet.java:68) ~[main/:?]
	at org.zaproxy.zap.extension.api.API.responseToXml(API.java:841) ~[main/:?]
	at org.zaproxy.zap.extension.api.API.convertViewActionApiResponse(API.java:707) ~[main/:?]
	at org.zaproxy.zap.extension.api.API.handleApiRequest(API.java:559) ~[main/:?]
	at org.zaproxy.addon.network.internal.server.http.handlers.ZapApiHandler.handleApiRequest(ZapApiHandler.java:111) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.handlers.ZapApiHandler.handleRequest(ZapApiHandler.java:85) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.handlers.ZapApiHandler.handleMessage(ZapApiHandler.java:70) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.MainServerHandler.notifyMessageHandlers(MainServerHandler.java:151) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.MainServerHandler.processMessage(MainServerHandler.java:131) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.LocalServerHandler.processMessage(LocalServerHandler.java:67) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.MainServerHandler.process(MainServerHandler.java:94) ~[?:?]
	at org.zaproxy.addon.network.internal.server.http.MainServerHandler.lambda$channelRead0$0(MainServerHandler.java:82) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [network-0.21.0.zap:?]
	at java.base/java.lang.Thread.run(Thread.java:840) [?:?]
```

### Additional context:

https://github.com/zaproxy/zap-api-java/pull/129

- [ ] Yes

Kaifmohd:
### **Title**:
Fix: XML Response Handling in `alertCountsByRisk` API (#8814)

### **Description**:
This pull request resolves the issue encountered when attempting to retrieve an XML response from the `alertCountsByRisk` API endpoint. The bug was caused by special characters, specifically spaces in the key names (`False Positive`), which were causing XML generation to fail.

### **Issue**:
When the API returned XML for the `alertCountsByRisk` endpoint, an `internal_error` was encountered. Upon debugging, it was found that the space in the key name `False Positive` was causing the failure.

### **Solution**:
To fix the issue, I sanitized the key names by replacing spaces with hyphens (`-`) to ensure they were properly formatted for XML. This resolved the issue, and XML responses are now correctly generated.

### **Changes Made**:
- In the `org/zaproxy/zap/extension/api/ApiResponseSet.java` file, I added the following change to sanitize the keys:
  ```java
  String sanitizedKey = val.getKey().replace(" ", "-");
  

https://github.com/user-attachments/assets/2f98aeee-e95e-4a34-9631-3919307aae86

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
@@ -766,6 +766,9 @@ public class AlertAPI extends ApiImplementor {
 
     @SuppressWarnings("unchecked")
     private static class CustomApiResponseSet<T> extends ApiResponseSet<T> {
+
+        private Map<String, T> values = null;
+
         public CustomApiResponseSet(String name, Map<String, T> values) {
             super(name, values);
         }
@@ -773,7 +776,7 @@ public class AlertAPI extends ApiImplementor {
         @Override
         public void toXML(Document doc, Element parent) {
             parent.setAttribute("type", "set");
-            for (Map.Entry<String, T> val : getValues().entrySet()) {
+            for (Map.Entry<String, T> val : values.entrySet()) {
                 Element el = doc.createElement(val.getKey());
                 if ("tags".equals(val.getKey())) {
                     el.setAttribute("type", "list");
@@ -790,7 +793,8 @@ public class AlertAPI extends ApiImplementor {
                         Element elValue = doc.createElement("value");
                         elValue.appendChild(
                                 doc.createTextNode(
-                                        XMLStringUtil.escapeControlChrs(tag.getValue())));
+                                        XMLStringUtil.escapeControlChrs(
+                                                tag.getValue())));
                         elTag.appendChild(elValue);
 
                         el.appendChild(elTag);
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/ApiResponseSet.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/ApiResponseSet.java
@@ -60,12 +60,13 @@ public class ApiResponseSet<T> extends ApiResponse {
         return jo;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
     public void toXML(Document doc, Element parent) {
         parent.setAttribute("type", "set");
-
         for (Entry<String, T> val : values.entrySet()) {
-            Element el = doc.createElement(val.getKey());
+            String sanitizedKey = val.getKey().replace(" ", "-");
+            Element el = doc.createElement(sanitizedKey);
             String textValue = val.getValue() == null ? "" : val.getValue().toString();
             Text text = doc.createTextNode(XMLStringUtil.escapeControlChrs(textValue));
             el.appendChild(text);
--- a/zap/src/main/java/org/zaproxy/zap/extension/api/ApiResponseSet.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/api/ApiResponseSet.java
@@ -60,7 +60,6 @@ public class ApiResponseSet<T> extends ApiResponse {
         return jo;
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public void toXML(Document doc, Element parent) {
         parent.setAttribute("type", "set");


Kaifmohd:

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [71348058\-fe16\-4fb2\-b361\-9df7ddbf99c8](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=71348058-fe16-4fb2-b361-9df7ddbf99c8&branch=Kaifmohd%2Fbugfix%2Falert-view-alertCountsByRisk-xml-response-fix-%238814)

Kaifmohd:

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
kingthorin, line 1: Are these changes necessary to fix the issue?

zap/src/main/java/org/zaproxy/zap/extension/api/ApiResponseSet.java
```diff
@@ -63,9 +63,9 @@ public JSON toJSON() {
     @Override
     public void toXML(Document doc, Element parent) {
         parent.setAttribute("type", "set");
-
         for (Entry<String, T> val : values.entrySet()) {
-            Element el = doc.createElement(val.getKey());
+            String sanitizedKey = val.getKey().replace(" ", "-");
+            Element el = doc.createElement(sanitizedKey);
```
kingthorin, line 68: @thc202 @psiinon is there a way to check and do this only for XML as it might break people's existing use of the JSON responses from the API?

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
Kaifmohd: No while debugging I made those changes I will revert back those changes. Just created a Map<>.. values.

Thanks

Kaifmohd:
Hey @kingthorin @psiinon  I have made the changes as mentioned now these will not affect the older changes. 
Also the code is in function toXML and these will not affect any other functions.

Please let me know if you have any questions.

zap/src/main/java/org/zaproxy/zap/extension/alert/AlertAPI.java
Kaifmohd: I have reverted back all the changes made to AlertAPI.java and fixed only the main issue which was not related to this.

thc202:
IMHO this is a workaround not a fix, the offending endpoints (not just about `alertCountsByRisk`) should be changed to return properly formed data. Otherwise why is it just "sanitizing" spaces and not other illegal characters? And why hyphen and not underscore or even strip them, applying this workaround in other classes we end up with tags called e.g. `Information-Disclosure---Information-in-Browser-localStorage`.

Kaifmohd:
Hi @thc202,

Thank you for the explanation.

Regarding XMLStringUtil.escapeControlChrs(textValue), how should spaces be handled? The following XML structure is not valid according to the XML parser:
<False Positive>0</False Positive>


To address this issue, I replaced spaces with hyphens (-), which resolved the problem and allowed the XML to be parsed correctly.

If trimming spaces is a preferred solution, I can update the XMLStringUtil class to handle spaces as part of the illegal characters list. However, spaces are currently not included in the list of illegal characters.

Do you think this would be the right approach? If so, we wouldn't need the "sanitizedKey" variable to manually replace spaces.

thc202:
The `XMLStringUtil.escapeControlChrs` is used for the values which can contain spaces and pretty much any other character.

As mentioned in the previous comment this should be fixed in the endpoints, they should return a valid XML response.

Kaifmohd:
> The `XMLStringUtil.escapeControlChrs` is used for the values which can contain spaces and pretty much any other character.

Hey @thc202 , 
it is handling other special characters but not for space "\u0020" is missing for XMLStringUtil.escapeControlChrs 
 "illegalChrs" variables list. But even after adding this will fail because it will still contain the character which is not supported by createElement method of xml.
 
    public void toXML(Document doc, Element parent) {
        parent.setAttribute("type", "set");

        for (Entry<String, T> val : values.entrySet()) {
            Element el = doc.createElement(val.getKey()); // Failing at this point also it will fail if you add your escapeControlChrs method here. Becuase escapeControlChars is just replacing the value
            String textValue = val.getValue() == null ? "" : val.getValue().toString();
            Text text = doc.createTextNode(XMLStringUtil.escapeControlChrs(textValue));
            el.appendChild(text);
            parent.appendChild(el);
        }
    }
    
    
    Even after escaping, the result (e.g., `<False\x0020Positive>`) is still not a valid XML tag. As '\' is not allowed for this ` Element el = doc.createElement(val.getKey());`
   Also fixing at the endpoint. Also need to have changes to deal with any special character that is passed. 


Also, all the unit test that is done on the XMLStringUtil class have not considered the space " " scenarios.

I hope I was able to explain the problem to you.

@kingthorin @psiinon

thc202:
The problem is not in `XMLStringUtil` nor `ApiResponseSet` but `AlertAPI` which is creating an invalid XML `ApiResponse`, that's where the fix should be.

Kaifmohd:
Hi @thc202 , 

Thankyou I have gone through AlertAPI. And found the below.

 ```Map<String, Integer> map = new HashMap<>();
            map.put(Alert.MSG_RISK[Alert.RISK_HIGH], riskCounts[Alert.RISK_HIGH]);
            map.put(Alert.MSG_RISK[Alert.RISK_MEDIUM], riskCounts[Alert.RISK_MEDIUM]);
            map.put(Alert.MSG_RISK[Alert.RISK_LOW], riskCounts[Alert.RISK_LOW]);
            map.put(Alert.MSG_RISK[Alert.RISK_INFO], riskCounts[Alert.RISK_INFO]);
            map.put(Alert.MSG_CONFIDENCE[Alert.CONFIDENCE_FALSE_POSITIVE], falsePositiveCount);
            result = new ApiResponseSet<>(name, map);
            
            
      public static final String[] MSG_CONFIDENCE = {
        "False Positive", "Low", "Medium", "High", "Confirmed" /// Do you want me to change here? So that it will pass the correct data 
    };
```

Kaifmohd:
Hey @kingthorin @thc202  , This is the latest one that I was talking about.

kingthorin:
This does not seem to be what @thc202 mentioned

PR #8819, opened 2025-01-22
Fix error when no Java version is found

middagj:
Small fix to prevent errors like: `bash: [: -ge: unary operator expected` I got when there is no Java installed.

I couldn't really find what the target shell is, but even macos bash 3.2 supports `[[ ]]` which is much easier to work with. Considering that you put `bash` in the command line and there is more `[[ ]]` usage, you could also switch to:

```bash
if [[ $JAVA_MAJOR_VERSION -ge 11 ]]; then
```

and standardize on `[[ ]]` instead of `[ ]`. Also the backticks can be replaced by the more modern `$()`. If you want to target any POSIX compliant sh, you should of course replace the `[[ ]]` with `[ ]` and replace `$()` with backticks and also update the shebang to `sh`.


psiinon:
Arg - we should have also updated this as the min Java verion is now 17!

middagj:

zap/src/main/dist/zap.sh
```diff
@@ -47,10 +47,10 @@ JAVA_VERSION=$(java -version 2>&1 | awk -F\" '/version/ { print $2 }')
 JAVA_MAJOR_VERSION=${JAVA_VERSION%%[.|-]*}
 JAVA_MINOR_VERSION=$(echo $JAVA_VERSION | awk -F\. '{ print $2 }')
 
-if [ $JAVA_MAJOR_VERSION -ge 11 ]; then
+if [ ${JAVA_MAJOR_VERSION:-0} -ge 17 ]; then
```
thc202, line 50: We'll be changing the build to set it automatically.

middagj: I will remove the commit

middagj:
Anything I need to add before this PR can be merged?

thc202 (approved):

thc202:
Thank you!

kingthorin (approved):
LGTM, thanks!

PR #8879, opened 2025-03-06
Look and feel: Use native decorations on Windows

kingthorin:
- GuiBootstrap > No longer disable native decorations on windows.
- AbstractFrame > Save the window size regardless of state..

Ref: 
- https://github.com/zaproxy/zaproxy/pull/6570
- https://docs.oracle.com/javase/tutorial/uiswing/components/splitpane.html

### DIFF
--- a/zap/src/main/java/org/parosproxy/paros/view/AbstractFrame.java
+++ b/zap/src/main/java/org/parosproxy/paros/view/AbstractFrame.java
@@ -31,6 +31,7 @@
 // ZAP: 2022/08/05 Address warns with Java 18 (Issue 7389).
 // ZAP: 2022/09/08 Use format specifiers instead of concatenation when logging.
 // ZAP: 2023/01/10 Tidy up logger.
+// ZAP: 2025/03/07 Save window size regardless of state.
 package org.parosproxy.paros.view;
 
 import java.awt.Dimension;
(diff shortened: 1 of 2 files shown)


kingthorin:
Please test this on other platforms.

I'm currently doing a weekly build to test on Kali. But I'd like to know if there are any adverse effects on macOS or whatever other Linux the team has in use.

thc202:
There's a reason why they were being set and removing them does not seem to be the correct fix for the drift issue (as was mentioned in the other PR).

kingthorin:
Since the drifting only seems to happen when maximized I wonder if it has something to do with the preferred size of the components

https://docs.oracle.com/javase/tutorial/uiswing/components/splitpane.html

>To guarantee that both contained components appear, make sure that either the split pane is initially at or above its preferred size, or the minimum sizes of the contained components are greater than zero.

> This should usually happen if the splitpane is given its preferred size, which depends upon the layout manager containing the split pane. Another option is to explicitly set a preferred size on the split pane that is larger than the size of the contained components.

🤷‍♂️ will investigate some more.

kingthorin:
Oh you weren’t meaning specifically those weights being mentioned you mean the save/restoration of the values.

https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/view/WorkbenchPanel.java#L1430

https://github.com/zaproxy/zaproxy/blob/main/zap/src/main/java/org/parosproxy/paros/view/AbstractFrame.java#L222

thc202:
Right, the problem is around there.

kingthorin:
I need to stop looking at it for tonight but I've found that the values are saved as expected (or can be using a SessionChangedListener to be sure). They seem to initially be restored correctly. However when the UI actually displays they've been adjusted. I suspect they're being adjusted because of some internal component preferred Dimension, for example the SiteMapPanel is trying to set it's toolbar to width 800...I have no idea why.

I'll have to dig more, but it's late and I'm tired.

kingthorin:
My wife made fun of my really technical solution to tracking divider locations:
image

kingthorin:
Ready for review & testing on other platforms, I believe.

zap/src/main/java/org/parosproxy/paros/view/AbstractFrame.java
```diff
@@ -194,23 +195,14 @@ private SimpleWindowState restoreWindowState() {
     }
 
     /**
-     * Saves the size of this frame, but only, if window state is 'normal'. If window state is
-     * iconified or maximized, the size is not saved!
+     * Saves the size of this frame, but only, if window state is 'normal'.
```
thc202, line 198: This still needs update.

kingthorin: Done

thc202 (approved):

thc202:
Thank you!

psiinon (approved):

PR #8885, opened 2025-03-14
Allow API access to dynamically added Authn & Session Mgmt Method Types

kingthorin:
Allow authn method type and session mgmt type to be dynamically listed and used via the web API. (Ex: When they're added by the Auth Helper add-on).

Verified that setting them reflects in the context config, and that they are/aren't seen/settable if the "provider" isn't available (installed).

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/sessions/SessionManagementAPI.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/sessions/SessionManagementAPI.java
@@ -62,10 +62,7 @@ public class SessionManagementAPI extends ApiImplementor {
     private static final String PARAM_METHOD_NAME = "methodName";
     private static final String PARAM_METHOD_CONFIG_PARAMS = "methodConfigParams";
 
-    @SuppressWarnings("unused")
     private ExtensionSessionManagement extension;
-
-    private Map<String, ApiDynamicActionImplementor> loadedSessionManagementMethodActions;
 
     public SessionManagementAPI(ExtensionSessionManagement extension) {
         super();
@@ -82,17 +79,6 @@ public class SessionManagementAPI extends ApiImplementor {
                         ACTION_SET_METHOD,
                         new String[] {PARAM_CONTEXT_ID, PARAM_METHOD_NAME},
                         new String[] {PARAM_METHOD_CONFIG_PARAMS}));
-
-        this.loadedSessionManagementMethodActions = new HashMap<>();
-        // Load the session management method actions
-        if (extension != null) {
-            for (SessionManagementMethodType t : extension.getSessionManagementMethodTypes()) {
-                ApiDynamicActionImplementor i = t.getSetMethodForContextApiAction();
-                if (i != null) {
-                    loadedSessionManagementMethodActions.put(i.getName(), i);
-                }
-            }
-        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


kingthorin:
So far I've only looked at getting the correct list for the view (example output below), and handled the case of duplicates we've identified. I have not investigated if there is more to do.

```json
{
  "supportedMethods": [
    "formBasedAuthentication",
    "scriptBasedAuthentication",
    "httpAuthentication",
    "manualAuthentication",
    "browserBasedAuthentication",
    "jsonBasedAuthentication",
    "autoDetectAuthentication"
  ]
}
```

`13967 [AWT-EventQueue-0] WARN  org.zaproxy.zap.extension.authentication.ExtensionAuthentication - Authentication Method Type "Client Script Authentication" with ApiDynamicActionImplementor "scriptBasedAuthentication" seems to be a duplicate, and should implement it's own ApiDynamicActionImplementor.`

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [848c4eb6\-8deb\-48d6\-a885\-ea69286fbc56](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=848c4eb6-8deb-48d6-a885-ea69286fbc56&branch=kingthorin%2Fauth-api-dynamic)

kingthorin:
```json
{
  "supportedMethods": [
    "autoDetectSessionManagement",
    "headerBasedSessionManagement",
    "scriptBasedSessionManagement",
    "cookieBasedSessionManagement",
    "httpAuthSessionManagement"
  ]
}
```

zap/src/main/java/org/zaproxy/zap/extension/sessions/SessionManagementAPI.java
```diff
@@ -65,8 +65,6 @@ public class SessionManagementAPI extends ApiImplementor {
     @SuppressWarnings("unused")
```
thc202, line 65: This should be removed, it's being used now.

kingthorin: Done

psiinon:
If you set headerBasedSessionManagement with no methodConfigParams then it fails with an Internal Error:

```
94284 [ZAP-IO-Server-1-1] ERROR org.zaproxy.zap.extension.api.API - Exception while handling API request:
net.sf.json.JSONException: JSONObject["headers"] not found.
        at net.sf.json.JSONObject.getString(JSONObject.java:2092) ~[json-lib-2.4-jdk15.jar:?]
        at org.zaproxy.addon.authhelper.HeaderBasedSessionManagementMethodType$1.handleAction(HeaderBasedSessionManagementMethodType.java:474) ~[?:?]
        at org.zaproxy.zap.extension.sessions.SessionManagementAPI.handleApiAction(SessionManagementAPI.java:125) ~[main/:?]
        at org.zaproxy.zap.extension.api.API.handleApiRequest(API.java:538) ~[main/:?]
```

Strictly speaking this might be ok, but header based session management doesnt actually require any headers to be specified (it will still work for cookies) so I think it would be better not to error.
Example API valiues in the help would also be good? I'd expect the first question from people trying to use this would be: "what config values to I need to use?"

kingthorin:
Yeah, I had trouble figuring out the format the other day.

kingthorin:
Though I guess that's Auth Helper help not core help?

psiinon:
Yes - good point!

psiinon (approved):

kingthorin:
GUI requires the user to set one:
image

thc202 (approved):

PR #8892, opened 2025-03-19
Add TOTP to credentials

thc202:
Expose TOTP through the credentials to make it easier to use by the add-ons.
Add base implementation and change the credential implementations `UsernamePasswordAuthenticationCredentials` and
`GenericAuthenticationCredentials` to make use of it to allow authentication methods to opt-in for the TOTP usage.


zap/src/main/java/org/zaproxy/zap/extension/api/ApiDynamicActionImplementor.java
```diff
@@ -41,6 +42,21 @@ public ApiDynamicActionImplementor(
...
     }
 
+    /**
+     * Instantiates a new api dynamic action implementor.
+     *
+     * @param name the name
+     * @param mandatoryParamNames the mandatory param names, or <code>null</code> if there are no
+     *     mandatory parameters
+     * @param optionalParamNames the optional param names, or <code>null</code> if there are no
+     *     optional parameters
```
kingthorin, line 52: This says null but then in the next classes use List.of

thc202: One thing does not prevent the other, it also disambiguates.

kingthorin: Ok

psiinon (approved):

kingthorin (approved):

PR #8932, opened 2025-04-15
Fix: Ensure "Always on Top" is Cleared When Leaving Breakpoints to Prevent Stuck Popups  

Vrohs:
This PR addresses an issue where the ZAP main window could remain "Always on Top" after leaving a breakpoint, causing dropdowns and popups to block other applications if ZAP freezes or is not properly reset. The fix ensures that the "Always on Top" property is always cleared when a breakpoint is left, whether by continuing, dropping, or resetting.

### Summary of changes:
- Calls to `breakpointLeft()` have been added to the `cont()`, `drop()`, and `reset()` methods in `BreakPanel.java` to guarantee the main window is never left "Always on Top" after a breakpoint.
- The logic remains self-contained and respects user preferences for the "Always on Top" option.
- No global UI hacks or changes to popups/dropdowns are introduced.

### Rationale:
This change makes ZAP's breakpoint handling more robust and user-friendly, preventing UI elements from interfering with other applications and aligning with user expectations. It also follows the ZAP [contribution guidelines](https://github.com/zaproxy/zaproxy/blob/main/CONTRIBUTING.md) for atomic, descriptive, and well-documented changes.

Please review and let me know if any further changes are needed!

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
@@ -526,6 +526,8 @@ public class BreakPanel extends AbstractPanel implements BreakpointManagementInt
     public void reset() {
         this.msg = null;
         breakToolbarFactory.reset();
+        // Ensure Always on Top is cleared on reset
+        breakpointLeft();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


kingthorin (commented):
This is all/mostly LLM generated code right?

zap/src/main/java/org/zaproxy/zap/extension/brk/BreakPanel.java
```diff
@@ -571,11 +573,15 @@ public void cont() {
...
         breakToolbarFactory.setBreakResponse(false);
+        // Ensure Always on Top is cleared when leaving a breakpoint
+        breakpointLeft();
     }
 
     @Override
     public void drop() {
         breakToolbarFactory.drop();
+        // Ensure Always on Top is cleared when leaving a breakpoint
+        breakpointLeft();
```
thc202, line 584: For the record, the "always on top" state is reset by the calls to `clearAndDisableRequestEventHandler´ and `clearAndDisableResponseEventHandler` which are triggered when the request/response is cleared from the GUI.

PR #9276, opened 2026-03-10
Handle exceptions of individual writers in MultipleWriters

iamsanjaymalakar:
**Problem**
`MultipleWriters.close()` closes the underlying writers sequentially. If one `Writer.close()` throws an `IOException`, the method exits early and the remaining writers are never closed.

**Fix**
Update `close()` to attempt closing all writers, collect the first `IOException`, suppress subsequent close failures, and throw after cleanup is complete.


iamsanjaymalakar:

psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [060a41bf\-065c\-4de9\-aedf\-3d3a1414aa94](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=060a41bf-065c-4de9-aedf-3d3a1414aa94&branch=iamsanjaymalakar%2Fmain)


---
*Examples:*
`@Checkmarx rescan this PR`

thc202:
Can you share the failure?

iamsanjaymalakar:
> Can you share the failure?

The failure mode is purely control-flow/exception-safety: `MultipleWriters.close()` iterates underlying writers and calls `close()` on each. If any one writer’s `close()` throws `IOException`, the method exits immediately and never attempts to close the remaining writers. This is independent of OS timing/non-determinism: it’s just “first exception stops the loop”. In practice the `IOException` could come from a writer backed by a failing/closed stream or a filesystem issue, and leaving the rest unclosed can keep resources/handles open. This change makes `close()` best-effort (continue closing others) while still surfacing the error.

thc202:
So you didn't see any failure?

iamsanjaymalakar:
> So you didn't see any failure?

No.

thc202:
If you want to change the behaviour of `MultipleWriters` do that consistently, all methods (write, flush, close) have the same theoretical problem you are describing. Tests should also be added.

zap/src/main/java/org/zaproxy/zap/extension/script/MultipleWriters.java
```diff
@@ -50,9 +50,16 @@ public void flush() throws IOException {
 
     @Override
     public void close() throws IOException {
+        IOException first = null;
         for (Writer writer : writers) {
-            writer.close();
+            try {
+                writer.close();
+            } catch (IOException e) {
+                if (first == null) first = e;
```
thc202: There's a preference in the codebase to use braces even for single statements.

zap/src/main/java/org/zaproxy/zap/extension/script/MultipleWriters.java
```diff
@@ -36,22 +36,58 @@ public class MultipleWriters extends Writer {
...
+                if (first == null) {
+                    first = e;
+                } else {
+                    first.addSuppressed(e);
+                }
+            }
+        }
+        if (first != null) {
+            throw first;
         }
```
thc202, line 56: Extract a method instead of triplicating the same logic.

thc202:
It would be good to add tests to verify the new behaviour.

zap/src/main/java/org/zaproxy/zap/extension/script/MultipleWriters.java
kingthorin: With like `Consumer.accept()`?

thc202: Yes.

iamsanjaymalakar:
Addressed. I refactored the duplicated logic into a shared helper and applied it consistently to `write`, `flush`, and `close`. I used a small throwing functional interface rather than `Consumer<Writer>`, because `Consumer.accept()` cannot throw checked IOException. I did not add tests since this is a defensive exception-safety change, not a reliably reproducible failure case.

thc202:
It's definitely reproducible, you can add a writer that throws the exception (and writers that don't) and verify that the methods are called even in that case (you can use mockito).

iamsanjaymalakar:
> It's definitely reproducible, you can add a writer that throws the exception (and writers that don't) and verify that the methods are called even in that case (you can use mockito).

I've added the test.

thc202:
Thank you!

PR #9282, opened 2026-03-13
Remove Apache Commons Codec use

psiinon:
As per https://github.com/zaproxy/zap-extensions/pull/7207
It looks like core does not use commons Hex.

### DIFF
--- a/LEGALNOTICE.md
+++ b/LEGALNOTICE.md
@@ -32,7 +32,6 @@ and subject to their respective licenses.
 | Library                             | License                   |
 |-------------------------------------|---------------------------|
 | commons-beanutils-1.11.0.jar        | Apache 2.0                |
-| commons-codec-1.20.0.jar            | Apache 2.0                |
 | commons-collections-3.2.2.jar       | Apache 2.0                |
 | commons-configuration-1.10.jar      | Apache 2.0                |
 | commons-csv-1.14.1.jar              | Apache 2.0                |
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -24,7 +24,6 @@ tasks.withType<JavaCompile>().configureEach {
 dependencies {
     implementation("com.github.javaparser:javaparser-core:3.15.21")
     implementation("com.github.zafarkhaja:java-semver:0.9.0")
-    implementation("commons-codec:commons-codec:1.17.1")
     implementation("commons-configuration:commons-configuration:1.10")
     implementation("commons-collections:commons-collections:3.2.2")
     implementation("commons-io:commons-io:2.13.0")
--- a/buildSrc/src/main/java/org/zaproxy/zap/tasks/CreateGitHubRelease.java
+++ b/buildSrc/src/main/java/org/zaproxy/zap/tasks/CreateGitHubRelease.java
@@ -24,10 +24,12 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.HexFormat;
 import java.util.stream.Collectors;
-import org.apache.commons.codec.digest.DigestUtils;
 import org.gradle.api.Action;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.InvalidUserDataException;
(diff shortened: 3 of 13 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [9bd4b578\-e779\-46c8\-b7af\-89499d513eec](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=9bd4b578-e779-46c8-b7af-89499d513eec&branch=psiinon%2Fbuild%2Fbase64)


---
*Examples:*
`@Checkmarx rescan this PR`

thc202:
Thank you!

thc202:
This was pending review.

psiinon:
Sorry, my bad - missed that 😦 
Let me know if you'd like any changes..

LEGALNOTICE.md
```diff
@@ -32,7 +32,6 @@ and subject to their respective licenses.
 | Library                             | License                   |
 |-------------------------------------|---------------------------|
 | commons-beanutils-1.11.0.jar        | Apache 2.0                |
-| commons-codec-1.20.0.jar            | Apache 2.0                |
```
thc202, line 35: The dependency is still being included (it's a transitive dependency).

buildSrc/build.gradle.kts
```diff
@@ -24,7 +24,6 @@ tasks.withType<JavaCompile>().configureEach {
 dependencies {
     implementation("com.github.javaparser:javaparser-core:3.15.21")
     implementation("com.github.zafarkhaja:java-semver:0.9.0")
-    implementation("commons-codec:commons-codec:1.17.1")
```
thc202, line 27: There was no need to change the build, this does not affect the dependencies of ZAP.

buildSrc/src/main/java/org/zaproxy/zap/tasks/CreateGitHubRelease.java
```diff
@@ -227,14 +228,21 @@ private void appendChecksumsTable(StringBuilder body) throws IOException {
...
-                    .append("](")
-                    .append(baseDownloadLink)
-                    .append(fileName)
-                    .append(") | `")
-                    .append(digestUtils.digestAsHex(file))
-                    .append("` |\n");
+            try {
+                byte[] digest =
+                        MessageDigest.getInstance(algorithm)
+                                .digest(Files.readAllBytes(file.toPath()));
```
thc202, line 234: This should be done with chunks to prevent OOME, reading the whole file into memory is not good given the sizes the dist files have. The try catch only needs to wrap this statement not the whole append. (Same in other places)

zap/src/main/java/org/apache/commons/httpclient/URI.java
```diff
@@ -1703,7 +1702,50 @@ public URI(URI base, URI relative) throws URIException {
      * @return URI character sequence
      * @throws URIException null component or unsupported character encoding
      */
-        
+    private static byte[] encodeUrl(BitSet urlsafe, byte[] bytes) {
```
thc202, line 1705: - Preferable to copy verbatim the current code we are using and acknowledge that with a comment.
- This needs to be moved after the encode method, the JavaDoc is now applying to this method.

zap/zap.gradle.kts
```diff
@@ -65,13 +65,19 @@ tasks.named<JacocoReport>("jacocoTestReport") {
...
             ValidateImports(
                 mapOf(
                     "import org.apache.commons.lang." to
                         "Import/use classes from Commons Lang 3, instead of Lang 2.",
+                    "import org.apache.commons.codec.binary.Base64" to
+                        "Use java.util.Base64 instead.",
+                    "import org.apache.commons.codec.binary.Hex" to
+                        "Use java.util.HexFormat instead.",
+                    "import org.apache.commons.codec.digest.DigestUtils" to
+                        "Use org.zaproxy.zap.utils.DigestUtils instead.",
```
thc202, line 80: This class does not exist. Since we no longer want to use the codec classes we should disallow all imports like done with lang v2 (depending on what we do, if we end up removing it we can remove these rules).

zap/zap.gradle.kts
```diff
@@ -90,7 +96,6 @@ dependencies {
     api("com.fifesoft:rsyntaxtextarea:3.6.0")
     api("com.github.zafarkhaja:java-semver:0.10.2")
     implementation("commons-beanutils:commons-beanutils:1.11.0")
-    api("commons-codec:commons-codec:1.20.0")
```
thc202, line 93: This should be changed to implementation instead, to not expose it as a transitive dependency. Or, we should exclude in the respective direct dependency. In any case it's preferable to do first a full release without exposing the dependency but still including it to give time to third-party add-ons to act accordingly.

Issue #7420, opened 2022-08-15
Upgrading add-ons tossed up many transient dialogs and 3 modal dialogs


jsoref:
### Describe the bug

1. Upgrading should not result in a bunch of modal dialogs
2. Upgrading should not result in a bunch of flashing dialogs

### Steps to reproduce the behavior

1. Help > Check for updates
2. Click `Update all`
3. Click `Yes`
4. Watch as a slew of tiny dialogs/windows appear / disappear in rapid succession
5. Stare in frustration at each successive modal dialog that's telling me something which probably doesn't matter and definitely didn't need to be displayed in a modal dialog

### Expected behavior

Applications shouldn't flash the screen. It doesn't matter if it's a giant dialog or a tiny dialog. It's ok to use a status line to place some installation progress, and it's ok to have a window tied to that status line with detailed information.

If there is important information (such as the removal of a component or migration), it should have been made available in the upgrade dialog (in a summary view) or in a post upgrade window (in a summary view).

### Software versions

2.11.1

### Screenshots

![image](https://user-images.githubusercontent.com/2119212/184700577-e720bda0-fe01-4ebb-bf82-dddb2699341a.png)
![image](https://user-images.githubusercontent.com/2119212/184700588-f7284322-cec7-48ab-8fad-f56332fec09f.png)
![image](https://user-images.githubusercontent.com/2119212/184700602-78e6ae96-f393-4358-b79f-1a23163ac788.png)

### Would you like to help fix this issue?

- [ ] Yes

kingthorin:
If your screen is flashing, that's a video issue not a ZAP issue. I've gone through what you're reporting numerous times on Linux and Windows without any screen flashing.

Also for the majority of users it's doubtful that they'd have all 3 add-ons which are being superseded installed.

I guess they could be made non-modal but then we'd run the risk of them being hidden and the user failing to recall the action that triggered them.

However your concerns are valid, having a way to register post-install messages and display them aggregated might be a good move. (That was kinda discussed in this issue: https://github.com/zaproxy/zaproxy/issues/3588)

jsoref:
The flashing was from other tiny dialogs appearing and disappearing. They were telling me about the progress of upgrading components. I think. But they ran away too fast for me to read/capture.

kingthorin:
Oh okay, that makes more sense. I guess it isn't really flashing, just that you know something was displayed but it clears seemingly before it's useful.

[...]

PR #9363, opened 2026-06-12
Reuse uninstall dialog on batch upgrade

kingthorin:
- Reuse one UninstallationProgressDialogue across per-add-on upgrade uninstalls so Update all does not flash a new modal each time. Keeps modal blocking on the EDT.

This addresses zaproxy/zaproxy#7420

### DIFF
--- a/zap/src/main/java/org/zaproxy/zap/extension/autoupdate/ExtensionAutoUpdate.java
+++ b/zap/src/main/java/org/zaproxy/zap/extension/autoupdate/ExtensionAutoUpdate.java
@@ -125,6 +125,10 @@ public class ExtensionAutoUpdate extends ExtensionAdaptor
     // Files currently being downloaded
     private List<Downloader> downloadFiles = new ArrayList<>();
 
+    private boolean batchUpgradeSessionActive;
+    private UninstallationProgressDialogue sharedUninstallDialogue;
+    private boolean batchUninstallHadFailures;
+
     private static final int ARG_CFU_INSTALL_IDX = 0;
     private static final int ARG_CFU_INSTALL_ALL_IDX = 1;
     private static final int ARG_CFU_UNINSTALL_IDX = 2;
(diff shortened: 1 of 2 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [8b9e58c6\-1710\-47a9\-a9ee\-3d79ee8867bb](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=8b9e58c6-1710-47a9-a9ee-3d79ee8867bb&branch=kingthorin%2Faddon-batch)


---
Communicate with Checkmarx by submitting a PR comment with @Checkmarx followed by one of the supported commands. Learn about the supported commands [here](https://docs.checkmarx.com/en/34965-68683-code-repository-integration-usage---results.html#UUID-40215a0c-63e4-58a7-7a20-eb8732bc3d5e_section-idm235086038509635).

thc202:
How did you verify these changes? And were the behavior changes on purpose?

kingthorin:
Tested live (dev build). Installing and uninstalling language add-ons mainly.
I also grabbed a few older add-ons from tags in zap-extensions, replaced things in .ZAP_D and ran ZAP to get prompted for upgrades.

kingthorin:
Behaviour changes were only intentional in so far as supporting the dialog change.

Will look at it further.

thc202:
https://github.com/zaproxy/zaproxy/pull/9363#issuecomment-4698329622

Verify also the API and check the GUI is consistent in its state.

https://github.com/zaproxy/zaproxy/pull/9363#issuecomment-4698356875

If they are not intentional I think we should undo those changes, maybe better to agree on how the whole process should behave first (e.g. blocking the GUI just because the files are being downloaded is not ideal).

kingthorin:
I did hit an EDT issue at one point, so I'd be unsurprised if cursor and I got "fixing" that wrong I guess.

My intent was the smallest UI/functional change to move from many independent dialogs to one for the activity(ies).

I'm quite happy to address/incorporate any behavior suggestions.

kingthorin:
For the ce8ce8a / f03a5cc pushes.

**Threading/polling** — Unchanged. Still ZAP-DownloadInstaller thread, 100ms download poll, same SwingWorker for uninstall. Only addition: poll hook forwards download % to batch shell.

**Modality** — Partial change for batch only (>1 task). Batch shell is non-modal; sub-steps still block via existing invisible modal (waitForWorkerOnEdt) during batch uninstall. Single add-on ops keep original modal behaviour.

**Sequencing / progress** — Unchanged. Same order: dependency uninstall → async download → per-add-on install (with inline upgrade uninstall). Progress math unchanged; shell displays {completed} of {total} on top of existing per-step events.

**Public APIs / state** — Unchanged surface. No public method signature changes. New types are package-private. Small additive UI sync: notifyInstallationsCompleted() on ManageAddOnsDialog.

--- 

Tested live again. Tested API as well. Did notice a slight issue with Manage Add-ons Dialog state (with these changes in place), was fixed.

kingthorin:
It just occurred to me that although the API install and uninstall did work, it may be using the wrong path, I'm pretty sure I saw the new dialog. I'll check further in a few hours.

kingthorin:
I went a different direction for an ultra compact/minimal change :wink: 84f403b

Impacts vs previous concerns:
- **Threading/polling** &gt; None. Same SwingWorker, same download thread/polling.
- **Modality** &gt;  Changed (GUI only): uninstall progress modal suppressed; EDT still blocks via worker.get(). Status via Manage Add-ons / output tab.
- **Sequencing / progress** &gt; None. Same uninstall order and progress events.
- **Web API** > None. No changes to AutoUpdateAPI endpoints, params, or responses. Install/uninstall/update actions behave same; headless path already bypassed uninstallAddOnsWithView (getView() == null).

thc202:
So we block the GUI without any indication of what's happening?

kingthorin:
It's still in the manage add-ons dialog and output.

kingthorin:
Yes, I agree that's not optimal I guess. The machines/add-ons I tested were fast enough I guess that I never noticed/felt the blocking at all.

I'm open to suggestions.

thc202:
https://github.com/zaproxy/zaproxy/pull/9363#issuecomment-4717796881

That assumes the user kept the dialogue or has the output open while the files were being downloaded.

https://github.com/zaproxy/zaproxy/pull/9363#issuecomment-4717879645

For me add-ons with many files usually show the dialogue (e.g. PTK), but I guess that depends on disk speed…
That's why I was suggesting we discuss how we want it to behave, the problem is that we are updating the add-ons asynchronously as they have finished downloading so we process one show dialogue, process another show dialogue, we can keep the dialogue open if there's still some ready rather than hide it (the initial batch idea but without reworking the installation process since that will break things (uninstall + install is not the same as update)).

kingthorin:
It's only the uninstalls that were blocking before right? 
Do they "need" to block? 
Was the blocking simply to handle add-ons/warning when things don't go right (warning to restart zaps, etc)?

thc202:
Yes, I think they need to block (when it takes longer, we only block/show the dialogue after a threshold). No, that's to prevent the user from using the features that are being unloaded/uninstalled.

PR #9431, opened 2026-08-21
Support sub-rule IDs in packaged scan configs

### LINKED ISSUE #9394: Unable to specify rules with dashes in OUTOFSCOPE/IGNORE rules
Rule 10055 is a blanket CSP rule, but has sub rules, including 10055-7 "script-src unsafe-hashes"

There are many other types of rules that are broken down like this, but you aren't able to specify a rule with a dash in the conf file, because it expects the rule id to be an integer.

This prevents more granular IGNORE/OUTOFSCOPE rules to be defined for false positives. It would be unwise to create a rule at a high level, so you end up with more noise in the resulting scan than necessary.

sainadh777:
## Summary

Packaged scans currently treat scan-rule configuration keys as parent plugin IDs, while ZAP reports sub-rules through dashed `alertRef` values such as `10055-7`. This makes exact `IGNORE` and `OUTOFSCOPE` entries ineffective or causes the baseline Automation Framework path to reject them as non-integer rule IDs.

This change:

- matches exact dashed sub-rule IDs against each alert's `alertRef`
- groups only explicitly configured sub-rules separately, leaving normal parent-rule behavior unchanged
- keeps sibling alerts active when one sub-rule is ignored or excluded
- makes baseline scans use the API implementation when sub-rule configuration is present, because Automation Framework summary/filter rule IDs are integer-only
- adds regression and malformed-ID corner-case coverage
- documents the packaged-scan behavior change

## Impact

Users can now target an individual sub-rule in baseline, full, and API scan configuration without suppressing other alerts from the same parent scanner.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/zap9394-venv.QojQif/bin/python docker/tests/suite.py` — 72 tests, 0 failures, 0 errors, 0 skips
- `python -m py_compile` for all four affected scan scripts and both changed test modules — passed
- `zap-baseline.py -h`, `zap-full-scan.py -h`, and `zap-api-scan.py -h` in the isolated test environment — all exited 0
- `git diff --check` — passed
- `./gradlew --no-daemon :addOns:commonlib:jarZapAddOn :addOns:custompayloads:jarZapAddOn :addOns:pscan:jarZapAddOn` in zap-extensions — BUILD SUCCESSFUL, 122 tasks (70 executed, 2 from cache, 50 up-to-date)
- Fresh-home local ZAP 2.18.0-SNAPSHOT smoke test with a live HTTP endpoint and built passive-scan add-ons:
  - ZAP emitted sibling alert refs `10055-7` and `10055-13`
  - exact `10055-7 IGNORE` produced a separate ignored group while retaining `10055-13`
  - exact `10055-7 OUTOFSCOPE` removed only that alert and retained `10055-13`
  - no relevant ZAP runtime errors were logged

The local Docker daemon was unavailable, so the smoke test used a fresh local ZAP daemon rather than launching the container image. The repository has no configured coverage task for this Python suite.

Fixes #9394

### DIFF
--- a/docker/CHANGELOG.md
+++ b/docker/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to the docker containers will be documented in this file.
 
+### 2026-08-21
+- Added support for scan sub-rule IDs in packaged scan `IGNORE` and `OUTOFSCOPE` configuration (Issue 9394).
+
 ### 2026-07-06
 - Default to the Client Spider when -j (the modern option) specified.
 
--- a/docker/tests/test_zap_baseline_plan.py
+++ b/docker/tests/test_zap_baseline_plan.py
@@ -176,6 +176,40 @@ class TestZapBaselinePlan(unittest.TestCase):
             self.assertTrue(any("-n" in message for message in log_capture.output))
             self.assertFalse(os.path.exists(plan_path))
 
+    def test_plan_only_rejects_sub_rule_config(self):
+        for config_content in [
+            "10055-7\tIGNORE\t(script-src unsafe-hashes)\n",
+            "10055-7\tOUTOFSCOPE\thttps://example\\.com/ignored\n",
+        ]:
+            with self.subTest(config_content=config_content):
+                zap_baseline = self.load_module()
+                args = ["--plan-only", "-t", self.target, "-c", "config.conf"]
+
+                with tempfile.TemporaryDirectory() as home_dir:
+                    plan_path = os.path.join(home_dir, "zap.yaml")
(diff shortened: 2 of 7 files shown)


psiinon:
Logo
**Checkmarx One – Scan Summary & Details** – [cc67f3da\-04dc\-4b6c\-8d38\-c82c64b2039f](https://eu.ast.checkmarx.net/projects/641c4d0f-0dae-4786-973b-ead9338157fd/scans?id=cc67f3da-04dc-4b6c-8d38-c82c64b2039f&branch=sainadh777%2Fsupport-subrule-alert-config-9394)


---
*Examples:*
`@Checkmarx rescan this PR`

